PR #1, opened 2024-06-24
Address Test Failures

Satrat:
Fixing a few errors that were carried over from the old repo:
* defaults in `State` needed to be initialized with a default_factory for python 3.11 compatability
* missing dependencies: sparsezoo (will remove in a follow-up PR)
* remove un-needed dependencies: evaluate, safetensors, sentencepiece 
* fixing a malformed recipe in the quantization tests
* removing the `mypy` quality check for now, it was a new addition and there are many style failures. I think its best to address all of them once the repo is in a more stable state

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,12 +14,12 @@ ifneq ($(findstring pytorch,$(TARGETS)),pytorch)
 endif
 
 # run checks on all files for the repo
+# leaving out mypy src for now
 quality:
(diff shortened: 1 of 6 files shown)


mgoin (approved):
This looks like a reasonable sweep to me. We should also fix the link issues, seems to mostly be missing files or neuralmagic links

PR #3, opened 2024-06-24
Initial README

bfineran:


README.md
```diff
@@ -1,6 +1,77 @@
 # LLM Compressor
+Easy to use model compression library that supports a growing list of formats including
+int and float quantization, activation quantization, and 2:4 sparsity.
 
-A library for compressing large language models utilizing the latest techniques and research in the field for both training aware and post training techniques. 
-The library is designed to be flexible and easy to use on top of PyTorch and HuggingFace Transformers, allowing for quick experimentation.
+## Overview
+A library for compressing large language models utilizing the latest techniques and research for both training aware and post training techniques.
+llm-compressor is designed to be flexible and easy to use on top of PyTorch and HuggingFace Transformers, allowing for quick experimentation.
```
robertgshaw2-neuralmagic, line 7: I think here we should include

List of supported formats:
- Mixed Precision: W4A16 + W8A16
- Integer Quantization: W8A8 (int8)
- Floating Point Quantization: W8A8 (fp8)
- 2:4 Semi-structured Sparsity
- Unstructured Sparsity

List of Supported Techniques:
- PTQ
- GPTQ
- SmoothQuant
- SparseGPT
- QAT (coming soon)

README.md
```diff
@@ -1,6 +1,77 @@
...
+
+```python
+from llmcompressor.transformers import oneshot
+from llmcompressor.modifiers.quantization.gptq import GPTQModifier
+
+# sets parameters for the GPTQ algorithms - target Linear layer weights at 4 bits
+gptq = GPTQModifier(scheme="W4A16", targets="Linear")
+
+oneshot(
+    model="TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T",  # sample model
```
robertgshaw2-neuralmagic, line 43: I would update this to use `TinyLlama/TinyLlama-1.1B-Chat-v1.0`

README.md
```diff
@@ -1,6 +1,77 @@
...
+
+
+## Learn More
+The llm-compressor library provides a rich feature-set for model compression below are examples
+and documentation of a few key flows.  If you have any questions or requests
+open an [issue](https://github.com/vllm-project/llm-compressor/issues) and we will add an example or documentation.
+
+* One-Shot API Documentation (Coming Soon!)
+* Modifiers Documentation (Coming Soon!)
+* [One-Shot Quantization](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization/llama7b_one_shot_quantization.md)
```
robertgshaw2-neuralmagic, line 72: use relative paths

README.md
```diff
@@ -1,6 +1,77 @@
...
+```
+
+### Inference with vLLM
+To run inference with vLLM, first install vLLM from pip `pip install vllm`.
+
+```python
+from vllm import LLM
+
+model = LLM("llama-compressed-quickstart")
+output = model.genetate("I love 4 bit models because")
```
robertgshaw2-neuralmagic, line 61: nit: typo

README.md
robertgshaw2-neuralmagic: I think we should instead pass a dataset that the user created here, like this example

- https://github.com/robertgshaw2-neuralmagic/vllm-examples/blob/main/quantization/w8a8-example.py

The "open_platypus" is a bit too opaque

README.md
```diff
@@ -1,6 +1,77 @@
...
+### Compression
+Compression is easily applied by selecting an algorithm (GPTQ) and calling the `oneshot` API.
+
+```python
+from llmcompressor.transformers import oneshot
+from llmcompressor.modifiers.quantization.gptq import GPTQModifier
+
+# sets parameters for the GPTQ algorithms - target Linear layer weights at 4 bits
+gptq = GPTQModifier(scheme="W4A16", targets="Linear")
+
```
robertgshaw2-neuralmagic, line 41: Can we initialize the model outside of the `one-shot` via `AutoModelForCausalLM` or does this require `SparseAutoModelForCausalLM`?

README.md
```diff
@@ -1,6 +1,77 @@
...
+    recipe=gptq,
+    save_compressed=True,
+    output_dir="llama-compressed-quickstart",
+    overwrite_output_dir=True,
+    max_seq_length=256,
+    num_calibration_samples=256,
+)
+```
+
+### Inference with vLLM
```
Satrat, line 54: Could we also add example code for inference with transformers? Just to make it clear that both are supported with the caveat that transformers runs in fake quant mode

bfineran: let's add this after we can get the HFQuantizer integration landed since that will affect that pathway

README.md
bfineran: after the HFQuantizer is upstreamed the base auto model will work

README.md
bfineran: agreed - wanted to keep the code minimal in the readme, taking a look at what we can do

robertgshaw2-neuralmagic (approved):

PR #4, opened 2024-06-24
Add FP8 Support

Satrat:
Migrating this PR from sparseml: https://github.com/neuralmagic/sparseml/pull/2306 

* add code for inferring compression format from model
* e2e fp8 example
* e2e tests

### DIFF
--- a/examples/quantization/llama7b_fp8_quantization.py
+++ b/examples/quantization/llama7b_fp8_quantization.py
@@ -1,0 +1,38 @@
+import torch
+from datasets import load_dataset
+from transformers import AutoTokenizer
+
+from llmcompressor.modifiers.quantization import GPTQModifier
+from llmcompressor.transformers import SparseAutoModelForCausalLM, oneshot
+
+model_stub = "meta-llama/Meta-Llama-3-8B-Instruct"
+output_dir = "Meta-Llama-3-8B-Instruct-FP8-Compressed"
+num_calibration_samples = 512
+
+tokenizer = AutoTokenizer.from_pretrained(model_stub, use_fast=True)
+tokenizer.pad_token = tokenizer.eos_token
+
+
+def preprocess(batch):
+    text = tokenizer.apply_chat_template(batch["messages"], tokenize=False)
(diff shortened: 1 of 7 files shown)


bfineran (approved):

tests/llmcompressor/transformers/compression/recipes/new_quant_fp8.yaml
```diff
@@ -0,0 +1,19 @@
+quant_stage:
+  quant_modifiers:
+    GPTQModifier:
```
robertgshaw2-neuralmagic, line 3: Can we do this with just quantization modifier?

We shouldn't need GPTQ for this case

examples/quantization/llama7b_fp8_quantization.py
```diff
@@ -0,0 +1,38 @@
...
+def preprocess(batch):
+    text = tokenizer.apply_chat_template(batch["messages"], tokenize=False)
+    tokenized = tokenizer(text, padding=True, truncation=True, max_length=2048)
+    return tokenized
+
+
+ds = load_dataset("mgoin/ultrachat_2k", split="train_sft")
+examples = ds.map(preprocess, remove_columns=ds.column_names)
+
+recipe = GPTQModifier(targets=["Linear"], scheme="FP8", ignore=["lm_head"])
```
robertgshaw2-neuralmagic, line 25: Is there a way to do fp8 without `GPTQModifier`?

robertgshaw2-neuralmagic: Resolved via slack

tests/llmcompressor/transformers/compression/recipes/new_quant_fp8.yaml
robertgshaw2-neuralmagic: Can do as follow up

Satrat: Gotcha, I just updated the PR with this change (both in the example and the integration test). However the nice scheme/target UX is only for GPTQModifier so it made the example a bit messier. I'll move the scheme parsing to QM in a follow up PR

PR #5, opened 2024-06-25
[ Examples ] E2E Examples

robertgshaw2-neuralmagic:
SUMMARY:
* Added examples where user controls dataset preprocessing
* Added examples with leading models  
* W8A8 Channelwise Weights, Dynamic Per Token Example (Llama-3-8B-Instruct) - GPTQ and SmoothQuant
* W4A16 G=128 Weights Example (Llama-3-8B-Instruct) - GPTQ
* READMES for the full user flow

FOLLOW UP PRs:
* Updated W4A16 example to use act_order=True (once supported)
* Migrate W8A8 to code based modifiers


Satrat (approved):

examples/quantization/llama7b_fp8_quantization.py
```diff
@@ -0,0 +1,38 @@
...
+def preprocess(batch):
+    text = tokenizer.apply_chat_template(batch["messages"], tokenize=False)
+    tokenized = tokenizer(text, padding=True, truncation=True, max_length=2048)
+    return tokenized
+
+
+ds = load_dataset("mgoin/ultrachat_2k", split="train_sft")
+examples = ds.map(preprocess, remove_columns=ds.column_names)
+
+recipe = QuantizationModifier(targets="Linear", scheme="FP8")
```
Satrat: Is this missing `ignore=["lm_head"]` or do we quantize the head for fp8?

examples/quantization/w8a8_fp8/README.md
```diff
@@ -0,0 +1,141 @@
...
+# Configure the quantization algorithm to run.
+recipe = QuantizationModifier(targets="Linear", scheme="FP8", ignore=["lm_head"])
+
+# Apply quantization.
+oneshot(
+    model=model, dataset=ds,
+    recipe=recipe,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+)
```
Satrat: Note that this will same an uncompressed copy of the model, if that isn't desired we should set `ouput_dir=SAVE_DIR` here then lines 103-105 will overwrite the uncompressed model

examples/quantization/llama7b_fp8_quantization.py
Satrat: actually we can probably deleted this example all together since its the same as `w8a8_fp8`? Not sure there is anything unique here except the dataset

Satrat (commented):
There are 143 files changed here, could you break out the removal of the copyright into a separate PR?

examples/quantization/w8a8_fp8/README.md
robertgshaw2-neuralmagic: Is there a way to not save via `output_dir`

examples/quantization/llama7b_fp8_quantization.py
robertgshaw2-neuralmagic: yup this will be deleted

robertgshaw2-neuralmagic:
> There are 143 files changed here, could you break out the removal of the copyright into a separate PR?

Im not seeing the 143 files?

examples/quantization/w8a8_fp8/README.md
Satrat: You can set output_dir to None explicitly, by default it saves to "./output"

Satrat:
> > There are 143 files changed here, could you break out the removal of the copyright into a separate PR?
> 
> Im not seeing the 143 files?

Nevermind, when I reviewed it was showing removal of all the copyright headers as changes, just checked again and its fixed

robertgshaw2-neuralmagic:
@bfineran @Satrat 

Examples ready to go. Once we cleanup the defaults for W8A8 INT8, I can update the recipe accordingly

examples/quantization_w4a16/README.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+### 3) Apply Quantization
+
+With the dataset ready, we will now apply quantization.
+
+We first select the quantization algorithm.
+
+In our case, we will apply the default GPTQ recipe for `int4` (which uses static group size 128 scales) to all linear layers.
+> See the `Recipes` documentation for more information on making complex recipes
```
Satrat, line 87: Can we add a link here or does this documentation not exist yet?

examples/quantization_w8a8_fp8/README.md
```diff
@@ -0,0 +1,144 @@
...
+    return tokenizer(sample["text"], padding=False, max_length=MAX_SEQUENCE_LENGTH, truncation=True, add_special_tokens=False)
+ds = ds.map(tokenize, remove_columns=ds.column_names)
+```
+
+### 3) Apply Quantization
+
+With the dataset ready, we will now apply quantization.
+
+We first select the quantization algorithm. In our case, we will apply the default recipe for `fp8` (which uses static-per-tensor weights and static-per-tensor activations) to all linear layers.
+> See the `Recipes` documentation for more information on making complex recipes
```
Satrat, line 85: Same here

examples/quantization_w8a8_int8/README.md
```diff
@@ -0,0 +1,149 @@
...
+### 3) Apply Quantization
+
+With the dataset ready, we will now apply quantization.
+
+We first select the quantization algorithm. For W8A8, we want to:
+* Run SmoothQuant to make the activations easier to quantize
+* Quantize the weights to 8 bits with channelwise scales using GPTQ
+* Quantize the activations with dynamic per token strategy
+
+> See the `Recipes` documentation for more information on recipes
```
Satrat, line 89: and here

examples/quantization_w4a16/README.md
robertgshaw2-neuralmagic: Not exist yet

examples/quantization_w8a8_fp8/README.md
robertgshaw2-neuralmagic: Not exist yet

examples/quantization_w8a8_int8/README.md
robertgshaw2-neuralmagic: Does not exist yet

Satrat (approved):
LGTM, will we need to re-run the evals after the preset schemes land?

robertgshaw2-neuralmagic:
> LGTM, will we need to re-run the evals after the preset schemes land?

The evals take 10s so easy to do. The whole flow end to end is 10min on an H100

mgoin (approved):
LGTM! Just not a fan of the current modifier scheme syntax i.e. `recipe = QuantizationModifier(targets="Linear", scheme="FP8", ignore=["lm_head"])`, I would like to import and use python objects directly for the schemes for easy lookup in source.

PR #8, opened 2024-06-25
Fix Weekly Test Failure

Satrat:
Original failure was from sparseml: https://github.com/neuralmagic/sparseml/actions/runs/9636050064/job/26573555739 

Issue was that when doing sparsity -> quantization, the quantization modifier needs to be included before SparseGPT. After restructuring the recipe the test passes locally for the 7b model. Also updating the device argument to be "auto" for all models


bfineran (approved):

dsikka (changes requested):
Could we make sure `auto` works with the intended runner before we make this change?

Satrat:
> Could we make sure `auto` works with the intended runner before we make this change?

Sure, how would I test that? Right now I'm just setting `export CADENCE="weekly"` then running pytest on a kubernetes pod

dbogunowicz (approved):

PR #11, opened 2024-06-28
Loguru logging standardization for LLM Compressor

markurtz:
Enable logging across LLM compressor and enable environment configuration overrides for library imports

SUMMARY:
- Added LogerConfig and refactored logging implementation to enable environment and configuration overrides for libraries that import LLM compressor.
- Enhanced docstrings and comments for clarity for logging implementations
- Refactored across the repo to remove _LOGGER and logging imports to move into loguru pathways.
- Added unit tests for logger functionality.

TEST PLAN:
Automation tests added, needs further manual testing for verfiication

### DIFF
--- a/src/llmcompressor/transformers/finetune/text_generation.py
+++ b/src/llmcompressor/transformers/finetune/text_generation.py
@@ -54,7 +54,6 @@ def train(**kwargs):
     """
     CLI entrypoint for running training
     """
-    configure_logger()
     model_args, data_args, training_args = parse_args(**kwargs)
     training_args.do_train = True
     main(model_args, data_args, training_args)
@@ -64,7 +63,6 @@ def eval(**kwargs):
     """
     CLI entrypoint for running evaluation
     """
-    configure_logger()
     model_args, data_args, training_args = parse_args(**kwargs)
     training_args.do_eval = True
     main(model_args, data_args, training_args)
@@ -74,7 +72,6 @@ def oneshot(**kwargs):
     """
     CLI entrypoint for running oneshot calibration
     """
-    configure_logger()
     model_args, data_args, training_args = parse_args(**kwargs)
     training_args.do_oneshot = True
     main(model_args, data_args, training_args)
@@ -88,7 +85,6 @@ def apply(**kwargs):
     """
     CLI entrypoint for any of training, eval, predict or oneshot
     """
-    configure_logger()
     report_to = kwargs.get("report_to", None)
     model_args, data_args, training_args = parse_args(**kwargs)
     training_args.run_stages = True
@@ -259,6 +255,8 @@ def main(
     for training and eval
     :param training_args: Arguments pertaining to training loop configuration
     """
+    configure_logger()
+
     # Temporary warning, to be removed
     if model_args.tie_word_embeddings is True:
         logger.warning(
--- a/src/llmcompressor/transformers/finetune/text_generation.py
+++ b/src/llmcompressor/transformers/finetune/text_generation.py
@@ -20,8 +20,6 @@
 import os
 from pathlib import PosixPath
 
-import datasets
-import transformers
 from loguru import logger
 from transformers import (
     AutoConfig,
@@ -269,13 +267,17 @@ def main(
             "weights without errors. Detected tie_word_embeddings=True. "
             "This may cause issues with the one-shot algorithm on save. "
         )
+
     # Setup logger
-    log_level = training_args.get_process_log_level()
-    logger.setLevel(log_level)
-    datasets.utils.logger.set_verbosity(log_level)
-    transformers.utils.logger.set_verbosity(log_level)
-    transformers.utils.logger.enable_default_handler()
-    transformers.utils.logger.enable_explicit_format()
+    # TODO: taking this out for now, these aren't valid calls and I was unsure what
+    # the intended purpose was originally
+    #
+    # log_level = training_args.get_process_log_level()
+    # logger.set_verbosity(log_level)
+    # datasets.utils.logger.set_verbosity(log_level)
+    # transformers.utils.logger.set_verbosity(log_level)
+    # transformers.utils.logger.enable_default_handler()
+    # transformers.utils.logger.enable_explicit_format()
 
     # Setup based on stage types if running stage mode
     if training_args.run_stages and training_args.recipe is not None:
--- a/src/llmcompressor/transformers/finetune/text_generation.py
+++ b/src/llmcompressor/transformers/finetune/text_generation.py
@@ -17,12 +17,12 @@
 # Adapted from https://github.com/huggingface/transformers
 # neuralmagic: no copyright
 
-import logging
 import os
 from pathlib import PosixPath
 
 import datasets
 import transformers
+from loguru import logger
 from transformers import (
     AutoConfig,
     AutoTokenizer,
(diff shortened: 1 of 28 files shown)


src/llmcompressor/transformers/sparsification/sparse_model.py
```diff
@@ -78,23 +75,17 @@ def skip(*args, **kwargs):
         # instantiate compressor from model config
         compressor = ModelCompressor.from_pretrained(pretrained_model_name_or_path)
 
-        # temporarily set the log level to error, to ignore printing out long missing
-        # and unexpected key error messages (these are EXPECTED for quantized models)
-        logger = logging.getLogger("transformers.modeling_utils")
-        restore_log_level = logger.getEffectiveLevel()
-        logger.setLevel(level=logging.ERROR)
```
mgoin, line 85: Are we going to see these key error messages from transformers now? It would be nice to avoid this IMO

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -260,19 +263,19 @@ def main(
...
-    _LOGGER.setLevel(log_level)
-    datasets.utils.logging.set_verbosity(log_level)
-    transformers.utils.logging.set_verbosity(log_level)
-    transformers.utils.logging.enable_default_handler()
-    transformers.utils.logging.enable_explicit_format()
+    logger.setLevel(log_level)
+    datasets.utils.logger.set_verbosity(log_level)
+    transformers.utils.logger.set_verbosity(log_level)
+    transformers.utils.logger.enable_default_handler()
+    transformers.utils.logger.enable_explicit_format()
```
mgoin, line 278: I don't think this is valid

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -50,13 +51,12 @@
...
 from llmcompressor.transformers.utils.helpers import detect_last_checkpoint
 
-_LOGGER: logging.Logger = logging.getLogger(__name__)
-
 
 def train(**kwargs):
     """
     CLI entrypoint for running training
     """
+    configure_logger()
```
mgoin, line 59: Should we put these configure_logger() calls somewhere higher up in the architecture if we are using this globally for llmcompressor so we need to call it in less places?

tests/unit/test_logger.py
```diff
@@ -1,79 +1,103 @@
...
+        log_contents = f.read()
+    assert "Info message" in log_contents
+    assert "Error message" in log_contents
+
+
+def test_environment_variable_override(monkeypatch, capsys, tmp_path):
+    # Test environment variables override settings
+    monkeypatch.setenv("GUIDE_LLM_LOG_LEVEL", "ERROR")
+    monkeypatch.setenv("GUIDE_LLM_LOG_FILE", str(tmp_path / "env_test.log"))
+    monkeypatch.setenv("GUIDE_LLM_LOG_FILE_LEVEL", "DEBUG")
```
mgoin, line 74: These should be for LLM_COMPRESSOR?

tests/unit/test_logger.py
```diff
@@ -1,79 +1,103 @@
...
+    with open(tmp_path / "env_test.log", "r") as f:
+        log_contents = f.read()
+    assert "Info message" in log_contents
+    assert "Error message" in log_contents
+    assert "Debug message" in log_contents
+
+
+def test_environment_variable_disable_logging(monkeypatch, capsys):
+    # Test environment variable to disable logging
+    monkeypatch.setenv("GUIDE_LLM_LOG_DISABLED", "true")
```
mgoin, line 95: Ditto

mgoin (commented):
Is loguru a lightweight dep? 

I ask because vLLM does all their logging through the native python library, so I think it would be nice to match up: https://github.com/vllm-project/vllm/blob/b90d8cd832669b9ad7c48cd9a431e80836778b56/vllm/logger.py

However if there is engineering value in a new logging library I can see the value, especially since it seems to preserve the same interface for creating new logs.

src/llmcompressor/transformers/sparsification/sparse_model.py
Satrat: Yes removing this causes the giant key error message to print out. Agreed we should avoid it, @markurtz I added this back in (just renamed logger to transformers_logger) and confirms its working with the new framework

src/llmcompressor/transformers/finetune/text_generation.py
Satrat: Since we don't have a single global entry points I think we need to call it in each of train/oneshot/apply

Satrat: Moved this to main()

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -260,19 +261,23 @@ def main(
...
+    # Setup logger
+    # TODO: taking this out for now, these aren't valid calls and I was unsure what
+    # the intended purpose was originally
+    #
+    # log_level = training_args.get_process_log_level()
+    # logger.set_verbosity(log_level)
+    # datasets.utils.logger.set_verbosity(log_level)
+    # transformers.utils.logger.set_verbosity(log_level)
+    # transformers.utils.logger.enable_default_handler()
+    # transformers.utils.logger.enable_explicit_format()
```
Satrat, line 278: @markurtz these weren't valid function calls, I couldn't tell what the purpose of this code block was so commenting it out for now, it was breaking a bunch of tests and the logging seems to work fine without it

Satrat:
@markurtz Got the tests in a passing state. Ran some manual tests ofoneshot and reloading, logs look good. Just had a few notes for you above

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -258,21 +255,27 @@ def main(
     for training and eval
     :param training_args: Arguments pertaining to training loop configuration
     """
+    configure_logger()
```
bfineran, line 258: why do we need to configure here when we're running on import of the logger config module?

tests/unit/test_logger.py
```diff
@@ -1,79 +1,105 @@
-import os
```
bfineran, line 1: we've had issues with double logging in the past so may be good to add an assert in against that

src/llmcompressor/transformers/finetune/text_generation.py
markurtz: This is left over from attempting to get it to play correctly as a library. It's no longer needed

markurtz, line 258:
```suggestion
```

src/llmcompressor/transformers/sparsification/sparse_model.py
markurtz: We need to figure out a better way around this, as the solution put in place is extremely hacky and fragile. We'll update this later, though, as we get to auditing the transformers pipeline pathways

src/llmcompressor/transformers/finetune/text_generation.py
markurtz: These are no longer needed and are a remnant from trying to get the logger to play nicely as a library. Unfortunately, due to the inability to consistently detect if we are imported as a library for use in other libraries or as a library for a script, this functionally was removed and configure_logger is now the default for any pathways.

src/llmcompressor/transformers/finetune/text_generation.py
markurtz: Noted elsewhere, but removing this completely and we'll cover it during the audit of the transformer pipeline pathways

tests/unit/test_logger.py
markurtz: I've updated tests to ensure exact counts

Satrat (approved):

bfineran (approved):

PR #13, opened 2024-07-02
SmoothQuant Mapping Defaults

Satrat:
SUMMARY:
Previously users were required to manually enter mappings for `SmoothQuant`, adding in a default that should work for most models.

TEST PLAN:
Added unit test for default, updated one of the integration tests to use the default and confirmed it passes on the nightly tests

### DIFF
--- a/src/llmcompressor/modifiers/smoothquant/base.py
+++ b/src/llmcompressor/modifiers/smoothquant/base.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from typing import Callable, Dict, List, Optional, Tuple
+from pydantic import field_validator
 
 import torch
 from loguru import logger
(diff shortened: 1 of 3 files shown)


src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -13,6 +13,11 @@
 
 MINIMUM_SMOOTHING_SCALE = 1e-5
 
+DEFAULT_SMOOTHQUANT_MAPPINGS = [
```
robertgshaw2-neuralmagic, line 16: QQ: if the model does not have the right structure, will we fail loudly?

Satrat: Yes, if it can't find a particular layer it'll fail like this:

```
Could not find targets ['re:.*does_not_exist'] in module LlamaForCausalLM(
```

It won't explicitly state that the error is due to the SmoothQuant defaults but should be sufficient for debugging

PR #16, opened 2024-07-02
actorder

horheynm:
SUMMARY:
Activation rerodering - groups the weight matrix based on the diagonal of the Hessians.
Grouping similar values so "like-alike" values should provide better scale, zp for the selected groups.

TEST PLAN:
Tested model with actoder = True and False. With actorder should provide a drop of increase in accuracy


bfineran (approved):

bfineran (approved):

bfineran (approved):

mgoin (approved):

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -171,6 +171,8 @@ def compress(
                                 update_layer_weight_quant_params,
                             )
 
+                            observer = getattr(self.layer, "weight_observer", None)
+                            observer.reset()
```
bfineran, line 175: need to only do this if it isn't None

kylesayrs:
Moved to https://github.com/vllm-project/llm-compressor/pull/94

PR #18, opened 2024-07-08
Rs/concepts

robertgshaw2-neuralmagic:
SUMMARY:
* explanation of why quantization is useful

TEST PLAN:
* N/a

### DIFF
--- a/docs/conceptual_guides/inference_acceleration.md
+++ b/docs/conceptual_guides/inference_acceleration.md
@@ -1,0 +1,74 @@
+# Inference Acceleration with Quantization
+
+Broadly speaking, there are two "types" of quantization:
+* Weight-Only Quantization
+* Weight and Activation Quantization
+
+## Weight-Only Quantization
+
+With weight-only quantization, weights are quantized to low precision (typically `int8`, `fp8`, or `int4`) while activations remain at higher precision `fp16` or `bf16`. To perform matrix multiplication, we upconvert each weight to`fp16` before computing `A*B`.
+
+### How Can We Speed Up Weight-Only Quantization?
+
+Roughly speaking, the time required to execute a matrix multiplication on a GPU equals:
+* Latency of moving the weights from main memory (DRAM) to the compute (SRAM)
+* Latency of the tensor-core compute operations
+
+While weight-only quanitzation does not change the latency of the tensor-core operations (since the compute still runs at `bf/fp16`), it can reduce the latency of moving the weights from DRAM to SRAM with "fused" inference kernels that upconvert the weights to `fp16`after moving them into SRAM (thereby reducing the total amount of data movement between DRAM and SRAM). LLM Inference Serving is usually dominated by batch size < 64 "decode" operations, which are "memory bandwidth bound", meaning we can speed up the `Linear` matmuls with weight-only quantization.
+
+### Accelerating Inference Serving in vLLM with `Marlin`
+
+[`Marlin`](https://neuralmagic.com/blog/pushing-the-boundaries-of-mixed-precision-llm-inference-with-marlin/) is an optimized fused inference kernel for weight-only quantization, supporting `int4`, `int8`, and `fp8` weights with `fp16` and `bf16` activations. vLLM uses `Marlin` when executing inference for weight-only quantized models created via `llm-compressor`.
+
+vLLM achieves strong end-to-end speedups from activation quantization on Nvidia A10G GPUs with `Meta-Llama-3-8B-Instruct` running the `sharegpt` online serving benchmark with 1 query per second:
+
+| Weight Precision  | Activation Precision  | Time Per Output Token (ms)    | Speedup vs `fp16` |
+|-                  |-                      |-                              | -                 |
+|`fp16`             | `fp16`                | 42.29                         | 1.0x              |
+|`fp8`              | `fp16`                | 25.24                         | 1.7x              |
+|`int8`             | `fp16`                | 24.23                         | 1.7x              |
+|`int4-g128`        | `fp16`                | 15.0                          | 3.0x              |
+
+> Performance results computed as of `vllm==v0.5.1` via [online serving performance benchmark](../../examples/benchmarking)
+
+### Examples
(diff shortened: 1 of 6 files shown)


docs/conceptual_guides/quantization_schemes.md
```diff
@@ -0,0 +1,88 @@
+# Quantization Schemes
+
+Quantization is a technique to reduce the computational and memory costs of running inference by representing the weights and activations with low-precision data types like 8-bit integer (`int8`) instead of the usual 16-bit floating point (`float16`).
+
+## Theory
+
+Performing quantization to go from `float16` to `int8` (or lower) is tricky. Only 256 values can be represented in `int8`, while `float16` can represent a very wide range of values. The idea is to find the best way to project our range [a, b] of `float32` values to the `int8` space.
```
bfineran, line 7: good space for a diagram (bucketing the weight distribution to 256 buckets)

docs/conceptual_guides/quantization_schemes.md
```diff
@@ -0,0 +1,88 @@
...
+
+```bash
+x = S * (x_q - Z)
+```
+
+where:
+
+- `x_q` is the quantized `int8` value associated to `x`
+- `S` is the scale, and is a positive `float16`
+- `Z` is called the zero-point, it is the `int8` value corresponding to the value 0 in the `float16` realm. If zero-point is ommited, we call this "symmetric" quantization.
```
bfineran, line 19:
```suggestion
- `Z` is called the zero-point, it is the `int8` value corresponding to the value 0 in the `float16` realm. If zero-point is ommited, we call this "symmetric" quantization because the default zero point of 0 is in the true middle of the distribution.
```

bfineran, line 19:
```suggestion
- `Z` is called the zero-point, it is the `int8` value corresponding to the value 0 in the `float16` representation. If zero-point is ommited, we call this "symmetric" quantization.
```

docs/conceptual_guides/quantization_schemes.md
```diff
@@ -0,0 +1,88 @@
...
+Let’s consider a float x in [a, b], then we can write the following quantization scheme:
+
+```bash
+x = S * (x_q - Z)
+```
+
+where:
+
+- `x_q` is the quantized `int8` value associated to `x`
+- `S` is the scale, and is a positive `float16`
```
bfineran, line 18:
```suggestion
- `S` is the scale, and is a positive `float16`.  It is used to "rescale" a distribution from the base range in `float16` to the desired width (ie 256 for `int8`).
```

docs/conceptual_guides/quantization_schemes.md
```diff
@@ -0,0 +1,88 @@
...
+There are several flavors of quantization.
+
+### Static vs Dynamic
+
+The section above described how quantization from `float16` to `int8` works, but did not explain how to compute the scales and zero points.
+
+With weights, since the full range is known ahead of time, we can just compute the scales and zero points statically (sometimes using a more sophisticated algorithm like `GPTQ`).
+
+With activations, however, there are two approaches:
+* **Dynamic quantization**: the range for each activation is computed on the fly at runtime. This gives us the best possible values, but it can be a bit slower than static quantization because of the overhead introduced by computing the range each time. It is also not an option on certain hardware.
```
bfineran, line 45:
```suggestion
* **Dynamic quantization**: the range for each activation is computed on the fly at runtime so that the quantization range matches exactly the current runtime range. This gives us the best possible values, but it can be a bit slower than static quantization because of the overhead introduced by computing the range each time. It is also not an option on certain hardware.
```

docs/conceptual_guides/quantization_schemes.md
```diff
@@ -0,0 +1,88 @@
...
+
+* **Static quantization**: the range for each activation is computed in advance at quantization-time, typically by passing representative "calibration" data through the model and recording the activation values. In practice, we run a number of forward passes on a calibration dataset is done and compute the ranges according to the observed calibration data.
```
bfineran, line 47:
```suggestion
* **Static quantization**: the range for each activation is computed in advance at quantization-time.  This is typically done by passing representative "calibration" data through the model and recording the range of activation values. In practice, we run a number of forward passes on a calibration dataset is done and compute the ranges according to the observed calibration data.
```

docs/conceptual_guides/quantization_schemes.md
```diff
@@ -0,0 +1,88 @@
...
+
+In general, it is best practice to start your experiments with:
```
bfineran, line 49: why?

docs/conceptual_guides/quantization_schemes.md
```diff
@@ -0,0 +1,88 @@
...
+Weight and activation quantization can be performed at different levels of granularity depending on accuracy / latency tradeoffs being targeted.
+
+#### Weight Quantization Granularity
+
+For weight quantization, there are three "levels" (in order of increasing of granularity):
+* **Per-Tensor**: one pair of quantization parameters (`S, Z`) is used per tensor
+* **Per-Channel**: one pair of quantization parameters (`S, Z`) is used per element of one of the dimensions of the tensor. For instance, with a weight matrix of shape `[N,M]`, the scales are a vector of shape [`M`] scales.
+* **Per-Group**: one pait of quantization parameters is (`S, Z`) is used per group of items in a tensor. For instance, with a weight matrix of shape `[N,M]` with `M=4096`, the scales are a matrix of shape `[32, M]` (note: `4096 / 128 = 32`).
+
+Incresing quantization granularity typically helps with accuracy at the expense of less memory reduction and slower inference performance. In general, it is best practice to start your experiments with:
```
bfineran, line 64:
```suggestion
Incresing quantization granularity typically helps with accuracy at the expense of less memory reduction and slower inference performance.  This is because we compute quantization ranges over smaller distributions with the trade off of needing more memory to represent them. In general, it is best practice to start your experiments with:
```

docs/conceptual_guides/quantization_schemes.md
robertgshaw2-neuralmagic: Why is it best practice?

PR #25, opened 2024-07-18
GPTQ - move calibration of quantiztion params to after hessian calibration

bfineran:
fixes a small bug in GPTQ implemetation where scales and zero points were calculated before hessian calibration passes. this likely is the cause of the small drift in results we notice in the implemetation

additionally this change is important for the activation reordering feature as delaying the update until after the hessian is computed allows us to set the scales and zero points easily wrt the permuted weight matrix

### DIFF
--- a/src/llmcompressor/modifiers/quantization/gptq/base.py
+++ b/src/llmcompressor/modifiers/quantization/gptq/base.py
@@ -1,7 +1,12 @@
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 import torch
-from compressed_tensors.quantization import QuantizationScheme
+from compressed_tensors.quantization import (
+    QuantizationScheme,
+    disable_quantization,
+    enable_quantization,
+    freeze_module_quantization,
+)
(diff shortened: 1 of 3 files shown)


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -250,6 +258,11 @@ def apply_compression(
         logger.info(
             f"Running {class_name} calibration with " f"{len(dataloader)} samples..."
         )
+
+        # quantization scales and zp are already initialized but we do not
+        # want to calibrate wrt to these
+        self.model.apply(disable_quantization)
+
```
Satrat, line 265: One thought here: in the case where we have activation quantization and GPTQ on the weights, we will end up running calibration twice and I don't think we need to anymore. Since we aren't using the scale/zp during the Hessian calculation maybe we should try and set them in the same calibration as the hessian.

Not a requirement for now but just something to think about for the future.

horheynm (approved):

PR #33, opened 2024-07-22
[WIP] Llama3 weight analysis and visualization using safetensors

pythonLoader:
SUMMARY:
Weight analysis and visualization for llama3 models loaded from safetensors.

TEST PLAN:
No automatic tests written yet. It's a work in progress. Will add test after review.


markurtz:
Thanks for he contribution @pythonLoader, I'm going to see if there's a bit more we can do to generalize this further over the next few weeks once I can find sometime. 

I'm worried about landing specific scripts under utils or the code base. 

If you'd like to get this in, then we can package it as an example but we'll need some more things around it such as a readme fo rusage.

PR #47, opened 2024-08-01
Layer by Layer Sequential GPTQ Updates

Satrat:
SUMMARY:
Our previous sequential implementation of GPTQ ran calibration forward passes over the whole model when compressing each layer. We now instead calibrate (and compress) one transformer layer at a time. This requires us to cache the intermediate outputs between each layer, which equates to `(hidden_size*calibration_samples*max_calibration_sequence_length)`. These intermediate outputs are stored on CPU, and moved to GPU one by one during calibration so there is no extra GPU memory required.
* Added an `EarlyStopException` for capturing the intermediate output of the model at the start of the decoder layers
* Instead of calling a full forward calibration pass in `GPTQModifier`, we run all the calibration data through one layer at a time (implemented in `LayerCompressor.calibrate_layer()`, caching the intermediate outputs

This update supports model offloading for sequential runs. It also has multi-gpu support: set with either `device_map="auto"` or with `calculate_offload_device_map(num_gpus=....)`. The latter option is recommended as it takes into account the memory required to store the hessians and quantization information when assigning devices and cpu offloading

TEST PLAN:
Updated w8a8 big model example to use the new sequential flow, tested with 8n and 70b.
```
python examples/big_model_offloading/big_model_w8a8_calibrate.py
```

Sequential now takes 21min for 8b W8A8 on an A100 (for reference non-sequential was 19min). vLLM eval results on gsm8k for the 8b example, results are equivalent to the non-sequential run
```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.744|±  |0.0277|
|     |       |strict-match    |     5|exact_match|↑  |0.744|±  |0.0277|
```

**NOTE:** requires https://github.com/neuralmagic/compressed-tensors/pull/120 be merged first (DONE)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -13,11 +13,11 @@ A clear and concise description of what you expected to happen.
 
 **Environment**
 Include all relevant environment information:
-1. OS [e.g. Ubuntu 18.04]:
+1. OS [e.g. Ubuntu 20.04]:
 2. Python version [e.g. 3.7]:
 3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]:
-4. ML framework version(s) [e.g. torch 1.7.1]:
-5. Other Python package versions [e.g. SparseZoo, DeepSparse, numpy, ONNX]:
+4. ML framework version(s) [e.g. torch 2.3.1]:
+5. Other Python package versions [e.g. vLLM, compressed-tensors, numpy, ONNX]:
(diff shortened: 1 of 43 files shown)


bfineran (approved):

abhinavnmagic (approved):
The proposed changes worked well for 70b, 450b - scale models.

PR #49, opened 2024-08-01
Add: Wrapper flow for W8A8 quantization

rahul-tuli:
Adds a wrapper function to apply `W8A8` quantization using `SmoothQuant` and `GPTQ`

Usage:
```python
from llmcompressor.flows import quantize_w8a8


smoothquant_kwargs = {"smoothing_strength": 0.8}
gptq_kwargs = {"targets": "Linear", "scheme": "W8A8", "ignore": ["lm_head"], "sequential_update": False}

quantize_w8a8(
    model_id="Xenova/llama2.c-stories15M",
    dataset="ultrachat-200k",
    output_dir="./test_output",
    splits={"calibration": "train_gen[:5%]"},
    max_seq_length=2048,
    pad_to_max_length=False,
    num_calibration_samples=8,
    smoothquant_kwargs=smoothquant_kwargs,
    gptq_kwargs=gptq_kwargs
)
```

Truncated Output:
```bash
done
2024-08-01T16:17:16.370529+0000 | compress_module | INFO - Compressing model.layers.5.model.layers.5.mlp.down_proj...
2024-08-01T16:17:16.704587+0000 | compress | INFO - time 0.33
2024-08-01T16:17:16.704723+0000 | compress | INFO - error 1.58
done
manager stage: Modifiers initialized
2024-08-01T16:17:16.708711+0000 | initialize | INFO - Compression lifecycle initialized for 1 modifiers
manager stage: Modifiers finalized
2024-08-01T16:17:16.708925+0000 | finalize | INFO - Compression lifecycle finalized for 1 modifiers
2024-08-01T16:17:16.709005+0000 | save_pretrained_wrapper | INFO - Inferring a sparsity configuration requires a global sparsity calculation. This can be costly for large models. To skip the calculation of compression statistics set skip_compression_stats=True
Calculating model sparsity: 100%|█████████████████████████████████████████████████████████████████████████████████████| 141/141 [00:00<00:00, 2253.45it/s]
Calculating quantization compression ratio: 111it [00:00, 2465.75it/s]
Compressing model: 100%|██████████████████████████████████████████████████████████████████████████████████████████████| 141/141 [00:00<00:00, 3498.23it/s]
2024-08-01T16:17:17.142808+0000 | save_model_and_recipe | INFO - Saving output to /root/llm-compressor/test_output
```

### DIFF
--- a/src/llmcompressor/flows.py
+++ b/src/llmcompressor/flows.py
@@ -1,0 +1,60 @@
+from typing import Dict, Optional, Union
+
+
+def quantize_w8a8(
+    model_id: str,
+    dataset: str,
+    output_dir: str,
+    splits: Dict[str, str],
+    max_seq_length: int,
+    pad_to_max_length: bool,
+    num_calibration_samples: int,
+    smoothquant_kwargs: Optional[Dict[str, Union[int, float, str]]] = None,
+    gptq_kwargs: Optional[Dict[str, Union[int, float, str]]] = None,
+) -> None:
+    """
+    Apply W8A8 and SmoothQuant + GPTQ quantization to a model using the oneshot method.
+
+    :param model_id: The HF model id, or local path to the model to be quantized.
+    :param dataset: The dataset to be used for calibration.
+    :param output_dir: The directory where the quantized model will be saved.
+    :param splits: The data splits to be used for calibration
+        ex. {"calibration": "train_gen[:5%]"}
+    :param max_seq_length: The maximum sequence length for the model.
+    :param pad_to_max_length: Whether to pad sequences to the maximum length.
+    :param num_calibration_samples: The number of samples to be used for calibration.
+    :param smoothquant_kwargs: Optional dictionary of keyword arguments for
+        the SmoothQuantModifier.
+    :param gptq_kwargs: Optional dictionary of keyword arguments for the GPTQModifier.
+    """
+    from llmcompressor.modifiers.quantization.gptq import GPTQModifier
+    from llmcompressor.modifiers.smoothquant import SmoothQuantModifier
+    from llmcompressor.transformers import SparseAutoModelForCausalLM, oneshot
+
+    default_gptq_kwargs = dict(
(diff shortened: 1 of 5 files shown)


src/llmcompressor/flows.py
```diff
@@ -0,0 +1,60 @@
...
+def quantize_w8a8(
+    model_id: str,
+    dataset: str,
+    output_dir: str,
+    splits: Dict[str, str],
+    max_seq_length: int,
+    pad_to_max_length: bool,
+    num_calibration_samples: int,
+    smoothquant_kwargs: Optional[Dict[str, Union[int, float, str]]] = None,
+    gptq_kwargs: Optional[Dict[str, Union[int, float, str]]] = None,
```
Satrat, line 13: I don't love this UI of having to specify the args as individual dictionaries. If the user is manually setting parameters I think its easier for them to just write out 

```
    recipe = [
        SmoothQuantModifier(....),
        GPTQModifier(....),
    ]
```
rather than calling this function. I would simplify this to only accept smoothing strength and a preset string (ie "FP8", "W8A8" etc)

Satrat (changes requested):
I feel like adding in this wrapper around oneshot is just going to complicate things, I don't see how it makes the UX any simpler for the user. If anything its klunkier because they have to pass in multiple dictionaries of arguments.

Instead, could we just create some concept of preset recipe schemes? Similar to how we have a few GPTQ schemes pre-defined in compressed-tensors:`PRESET_SCHEMES`, can we just define a global dictionary of recipes that users can pass in as strings?

src/llmcompressor/flows.py
rahul-tuli: What about model-specific arguments like `mappings`?

Satrat: At that point I just don't see the benefit of having this be a separate function, its not really saving any lines of code. I'll be all for it if we had a preset dictionary and only passed in a preset name though

PR #75, opened 2024-08-11
[ Docs ] Update `FP8` example to use dynamic per token

robertgshaw2-neuralmagic:
SUMMARY:
* convert example to use `FP8_DYNAMIC` 

TEST PLAN:
* manually running examples

### DIFF
--- a/examples/quantization_w8a8_fp8/README.md
+++ b/examples/quantization_w8a8_fp8/README.md
@@ -48,7 +48,7 @@ tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 
 For `fp8` quantization, we can recover accuracy with simple PTQ quantization.
 
-We recommend targeting all `Linear` layers using the `FP8_Dynamic` scheme, which uses:
+We recommend targeting all `Linear` layers using the `FP8_DYNAMIC` scheme, which uses:
 - Static, per-channel quantization on the weights
 - Dynamic, per-token quantization on the activations
 
@@ -58,7 +58,7 @@ Since simple PTQ does not require data for weight quantization and the activatio
 from llmcompressor.transformers import oneshot
 from llmcompressor.modifiers.quantization import QuantizationModifier
 
-# Configure the simple PTQ quantization to run with the FP8_DYNAMIC scheme.
+# Configure the simple PTQ quantization
 recipe = QuantizationModifier(
   targets="Linear", scheme="FP8_DYNAMIC", ignore=["lm_head"])
 
--- a/examples/quantization_w8a8_fp8/README.md
+++ b/examples/quantization_w8a8_fp8/README.md
@@ -1,6 +1,6 @@
 # `fp8` Weight and Activation Quantization
 
-`llm-compressor` supports quantizing weights and activations to `fp8` for memory savings and inference acceleration with `vLLM`
+`llm-compressor` supports quantizing weights and activations to `fp8` for memory savings and inference acceleration with `vllm`
 
 > `fp8` compuation is supported on Nvidia GPUs with compute capability > 8.9 (Ada Lovelace, Hopper).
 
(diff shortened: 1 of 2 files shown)


examples/quantization_w8a8_fp8/README.md
```diff
@@ -9,9 +9,7 @@
 To get started, install:
 
 ```bash
-git clone https://github.com/vllm-project/llm-compressor.git
-cd llm-compressor
-pip install -e .
+pip install llmcompressor
```
mgoin, line 12: I think we should pin the version so it is clear when this was made/updated

examples/quantization_w8a8_fp8/README.md
```diff
@@ -22,122 +20,102 @@ The example includes an end-to-end script for applying the quantization algorith
...
+Since simple PTQ does not require data for weight quantization and the activations are quantized dynamically, we do not need any calibration data for this quantization flow.
 
 ```python
 
-recipe = QuantizationModifier(targets="Linear", scheme="FP8", ignore=["lm_head"])
+recipe = QuantizationModifier(targets="Linear", 
+                              scheme="FP8_Dynamic",
```
mgoin, line 64: Is it fine to not use all caps? I thought the scheme was `FP8_DYNAMIC`

examples/quantization_w8a8_fp8/README.md
```diff
@@ -22,122 +20,102 @@ The example includes an end-to-end script for applying the quantization algorith
...
-    dataset=ds,
-    recipe=recipe,
-    max_seq_length=MAX_SEQUENCE_LENGTH,
-    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
-)
+oneshot(model=model, recipe=recipe)
 
 # Save to disk compressed.
-SAVE_DIR = MODEL_ID.split("/")[1] + "-W8A8-FP8"
-model.save_pretrained(SAVE_DIR, save_compressed=True)
```
mgoin, line 105: `save_compressed=True` is the default now?

examples/quantization_w8a8_fp8/README.md
```diff
@@ -22,122 +20,102 @@ The example includes an end-to-end script for applying the quantization algorith
...
+
+#### Install
 
-With the model created, we can now load and run in vLLM (after installing).
+Neural Magic's fork of `lm-evaluation-harness` implements the evaluation strategy used by Meta in the Llama3.1 launch. You can install this branch from source below:
+
+```bash
+pip install vllm
+pip install git+https://github.com/neuralmagic/lm-evaluation-harness.git@a0e54e5f1a0a52abaedced474854ae2ce4e68ded
+```
```
mgoin, line 87: It may be best to use a task that doesn't require a fork of lm-eval to reproduce results. AFAIK it is only ARC-C and GSM8k that require these custom changes. Winogrande is pretty fast, so maybe use that with lm-eval==0.4.3

examples/quantization_w8a8_fp8/README.md
robertgshaw2-neuralmagic: yes

examples/quantization_w8a8_fp8/README.md
robertgshaw2-neuralmagic: it seems to be working with `FP8_Dynamic`, but I can adjust it

examples/quantization_w8a8_fp8/README.md
robertgshaw2-neuralmagic: I like GSM because its easy to understand + its a good proof point for users that its working in a generative task

examples/quantization_w8a8_fp8/README.md
robertgshaw2-neuralmagic: good idea

examples/quantization_w8a8_fp8/README.md
mgoin: Do you need to have the reproduction of paper results then? I think we shouldn't push people towards our fork if possible. I think it also makes the most realistic example to show evals for both the unquantized and quantized checkpoint, so it shouldn't matter to get this specific COT setup.

Satrat (approved):

PR #76, opened 2024-08-11
[ Docs ] Overhaul `accelerate` user guide

robertgshaw2-neuralmagic:
SUMMARY:
* update `accelerate` examples with README
* update `accelerate` cpu offloading example to use fp8 + be consistent with other examples
* update `accelerate` calibration example to highlight multi-gpu setup

TEST PLAN:
* manually running examples


examples/big_model_offloading/big_model_w8a8_calibrate.py
```diff
@@ -1,90 +0,0 @@
-import torch
-from datasets import load_dataset
```
robertgshaw2-neuralmagic, line 2: removed since this user flow will be too slow for use (e.g. issues that teknium ran into)

examples/big_model_offloading/big_model_fp8.py
```diff
@@ -1,59 +0,0 @@
-import torch
-from transformers import AutoTokenizer
```
robertgshaw2-neuralmagic, line 2: renamed to `cpu_offloading_fp8.py` + updated to use better `recipe` arguments

examples/big_models_with_accelerate/README.md
```diff
@@ -0,0 +1,80 @@
...
+`llmcompressor` is designed to respect the `device_map`, so calls to `oneshot` will work properly.
+
+### Practical Advice
+
+When working with `accelerate`, it is important to keep in mind that CPU offloading and naive pipeline-parallelism will slow down forward passes through the model. As a result, we need to take care to ensure that the quantization methods used fit well with the offloading scheme as methods that require many forward passes though the model will be slowed down.
+
+General rules of thumb:
+- CPU offloading should only be with data-free quantization methods (e.g. PTQ with `FP8_DYNAMIC`)
+- Multi-GPU is fast enough to be used with calibration data-based methods with `sequential_update=False`
+- It is possible to use Multi-GPU with `sequential_update=True`, but the runtime will be slower
```
Satrat, line 36: We should note that device_map="auto" does always work with GPTQ, because the automatic memory allocations won't take into account the hessian allocations for GPTQ. To take these into account we need to use the helper function:

```
    model_stub, reserve_for_hessians=True, num_gpus=2, torch_dtype=torch.float16
)
```

And this only works with sequential_update=True. Without this helper fn cpu offloading won't work correctly

examples/big_models_with_accelerate/README.md
```diff
@@ -0,0 +1,80 @@
...
+```bash
+export CUDA_VISIBLE_DEVICES=0
+python cpu_offloading_fp8.py
+```
+
+The resulting model `./Meta-Llama-3-70B-Instruct-FP8-Dynamic` is ready to run with `vllm`!
+
+### Multi-GPU: `INT8` Quantization with `GPTQ`
+
+For quantization methods that require calibration data (e.g. `GPTQ`), CPU offloading is too slow. For these methods, `llmcompressor` can use `accelerate` multi-GPU to quantize models that are larger than a single GPU. For example, when quantizing a model to `int8`, we typically use `GPTQ` to statically quantize the weights, which requires calibration data.
```
Satrat, line 67: Worth mentioning that GPTQ needs to allocate extra memory, ~2x the size of the fp16 model for sequential_update=False

Satrat (changes requested):

examples/big_models_with_accelerate/README.md
robertgshaw2-neuralmagic: thats a nice rule of thumb

btw - are the hessians the same shape as the weight matrix? I thought they were the shape of X?

Satrat: Theyre (column x column) for each linear layer, and must be fp32. So ~2x the model size extra memory is a rough rule of thumb. The exact calculation is here: https://github.com/vllm-project/llm-compressor/blob/d232ea8b3ca30a70890feac1b56f5720c6e6b3c0/src/llmcompressor/transformers/compression/helpers.py#L100

Satrat (approved):

README.md
```diff
@@ -95,6 +98,17 @@ model = LLM("TinyLlama-1.1B-Chat-v1.0-INT8")
 output = model.generate("My name is")
 ```
 
+## Examples
+
```
robertgshaw2-neuralmagic, line 102: this is now duplicated with the above

PR #77, opened 2024-08-11
[ Docs ] Update main readme

robertgshaw2-neuralmagic:
SUMMARY:
* flip order to put examples more front and center
* update install instructions to suggest pypi install
* update example to use w8a8 instead of w4a16
* update example to have pointer to using custom calibration data

TEST PLAN:
* manually run example


Satrat (approved):
looks great!

mgoin (approved):

Issue #81, opened 2024-08-12
Phi-3 had its config.json changed by llm-compressor


Lin-K76:
**Describe the bug**

Using llm-compressor to quantize Phi-3-medium-128k-instruct changes the config.json. Specifically, "rope_scaling" is changed from "type": "su" to "type": "longrope". This is not a problem with Phi-3-mini-128k-instruct for some reason (although the type is also changed), but for Phi-3-medium, this prevents evaluation of the model through lm eval harness. 

To fix this issue, simply change the "type" in config back to "su".

The recipe I use is below:

```
recipe = """
quant_stage:
    quant_modifiers:
        QuantizationModifier:
            ignore: ["lm_head"]
            config_groups:
                group_0:
                    weights:
                        num_bits: 8
                        type: float
                        strategy: channel
                        dynamic: false
                        symmetric: true
                    input_activations:
                        num_bits: 8
                        type: float
                        strategy: token
                        dynamic: true
                        symmetric: true
                    targets: ["Linear"]
"""
```

**Expected behavior**

Successful evaluation through lm eval harness.

**Environment**
Include all relevant environment information:

1. OS [e.g. Ubuntu 18.04]: Ubuntu 22.04.4
2. Python version [e.g. 3.7]: 3.11.9
3. LLM Compressor version or commit hash [e.g. 0.1.0, f7245c8]: v0.1.0
4. ML framework version(s) [e.g. torch 1.7.1]: torch 2.4.0, transformers 4.40.2
6. Other Python package versions [e.g. SparseZoo, DeepSparse, numpy, ONNX]: lm_eval 0.4.3
7. Other relevant environment information [e.g. hardware, CUDA version]: CUDA 12.5

**To Reproduce**
Exact steps to reproduce the behavior:

Install lm eval harness, vllm,  and llm-compressor. Use the recipe above (https://github.com/vllm-project/llm-compressor/blob/main/examples/big_model_offloading/big_model_fp8.py) to quantize Phi-3-medium-128k-instruct (https://huggingface.co/microsoft/Phi-3-medium-128k-instruct). 

**Errors**
If applicable, add a full print-out of any errors or exceptions that are raised or include screenshots to help explain your problem.

**Additional context**
Add any other context about the problem here. Also include any relevant files.

Tagging @robertgshaw2-neuralmagic

robertgshaw2-neuralmagic:
@horheynm can you take a look at this?

[...]

PR #84, opened 2024-08-14
Fix for issue #81

horheynm:
SUMMARY:
Fix for https://github.com/vllm-project/llm-compressor/issues/81

steps:
1. clone repo
2. create and activate virtual eng
3. pip3 install -e .[dev]
4. change `llm-compressor/examples/big_model_offloading/big_model_fp8.py` line 34 from 
`model_stub = "meta-llama/Meta-Llama-3-70B-Instruct"`
to
`model_stub = "microsoft/Phi-3-medium-128k-instruct"`
5. run the code with 
```python3
# determine which layers to offload to cpu based on available resources
    model_stub, reserve_for_hessians=False, num_gpus=1, torch_dtype=torch.float16, trust_remote_code=True
)
```

the output config.json will get rid of line 201 from
`    "type": "longrope"`
to
`    "type": "su"`

TEST PLAN:
Run the code to completion, make sure that the issue does not come up.


src/llmcompressor/transformers/compression/helpers.py
```diff
@@ -191,6 +191,7 @@ def calculate_offload_device_map(
     model_stub: str,
     reserve_for_hessians=False,
     num_gpus: int = 1,
+    trust_remote_code: bool = True,
```
robertgshaw2-neuralmagic, line 194: this should probably be defaulted to false

Lin-K76 (approved):
Tested changes, fixes error regarding the rope type and the configuration python files.

Lin-K76 (approved):
Looks good to me!

Lin-K76 (approved):
Example changes look good to me

bfineran (approved):

Issue #83, opened 2024-08-13
lm_eval compatibility with generated model


horheynm:
**Describe the bug**
After a model is generated running `big_model_fp8.py`, lm_eval dont not work unless the .py files from the original base model is transferred to the generated model folder. Happens for 
https://huggingface.co/microsoft/Phi-3-medium-128k-instruct

```bash
OSError: test_phi_3_medium_128k_instruct_fp8 does not appear to have a file named configuration_phi3.py. Checkout 'https://huggingface.co/test_phi_3_medium_128k_instruct_fp8/tree/None' for available files.
```

**Expected behavior**
run lm_eval without any errors

**Environment**
Include all relevant environment information:
1. OS [e.g. Ubuntu 20.04]:
2. Python version [e.g. 3.7]:
3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]:
4. ML framework version(s) [e.g. torch 2.3.1]:
5. Other Python package versions [e.g. vLLM, compressed-tensors, numpy, ONNX]:
6. Other relevant environment information [e.g. hardware, CUDA version]:

**To Reproduce**
```python3
import torch
from transformers import AutoTokenizer

from llmcompressor.transformers import SparseAutoModelForCausalLM, oneshot
from llmcompressor.transformers.compression.helpers import (  # noqa
    calculate_offload_device_map,
    custom_offload_device_map,
)

# define a llmcompressor recipe for FP8 quantization
# this recipe requires no calibration data since inputs are dynamically quantized
recipe = """
quant_stage:
    quant_modifiers:
        QuantizationModifier:
            ignore: ["lm_head"]
            config_groups:
                group_0:
                    weights:
                        num_bits: 8
                        type: float
                        strategy: channel
                        dynamic: false
                        symmetric: true
                    input_activations:
                        num_bits: 8
                        type: float
                        strategy: token
                        dynamic: true
                        symmetric: true
                    targets: ["Linear"]
"""

# model_stub = "meta-llama/Meta-Llama-3-70B-Instruct"
model_stub = "microsoft/Phi-3-medium-128k-instruct"

# determine which layers to offload to cpu based on available resources
device_map = calculate_offload_device_map(
    model_stub, reserve_for_hessians=False, num_gpus=1, torch_dtype=torch.float16
)

# alternatively, specify the maximum memory to allocate per GPU directly
# device_map = custom_offload_device_map(
[...]

PR #85, opened 2024-08-14
Fix for issue 83

horheynm:
SUMMARY:
Fixing issue 83. https://github.com/vllm-project/llm-compressor/issues/83

Issue is that when a Phi-3-medium-128k-instruct model is generated, and lm_eval is run, lm_eval gives error bc some files that are expected are not included. These are python files that can be found in the model cache folder. lm_eval will fail if not provided. 

Solution was to add the logic in Trainer, a common place for oneshot, etc. - a shared pathway. 
So far only Phi-3-medium-128k-instruct has python files in its cache folder, so if they exist, copy them to the output directory.

TEST PLAN:
Ran 

```
import torch

from llmcompressor.transformers.compression.helpers import (  # noqa
    calculate_offload_device_map,
    custom_offload_device_map,
)

# define a llmcompressor recipe for FP8 quantization
# this recipe requires no calibration data since inputs are dynamically quantized
recipe = """
quant_stage:
    quant_modifiers:
        QuantizationModifier:
            ignore: ["lm_head"]
            config_groups:
                group_0:
                    weights:
                        num_bits: 8
                        type: float
                        strategy: channel
                        dynamic: false
                        symmetric: true
                    input_activations:
                        num_bits: 8
                        type: float
                        strategy: token
                        dynamic: true
                        symmetric: true
                    targets: ["Linear"]
"""

model_stub = "meta-llama/Meta-Llama-3-70B-Instruct"

# determine which layers to offload to cpu based on available resources
    model_stub, reserve_for_hessians=False, num_gpus=1, torch_dtype=torch.float16
)

# alternatively, specify the maximum memory to allocate per GPU directly
# device_map = custom_offload_device_map(
#    model_stub, max_memory_per_gpu="10GB", num_gpus=2, torch_dtype=torch.float16
# )

    model_stub, torch_dtype=torch.float16, device_map=device_map
)

output_dir = "./test_output_llama3b_70b_fp8"


oneshot(
    model=model,
    recipe=recipe,
    output_dir=output_dir,
    save_compressed=True,
    tokenizer=AutoTokenizer.from_pretrained(model_stub),
)

```
with and without the commit. python should be included and flow should be the same.


src/llmcompressor/transformers/finetune/trainer.py
```diff
@@ -6,4 +7,27 @@
 
 
 class Trainer(SessionManagerMixIn, HFTransformersTrainer):
-    pass
+    def __init__(self, **kwargs):
```
robertgshaw2-neuralmagic, line 10: why not in SparseAutoModel?

horheynm: anything downstream of HFTransformersTrainer init should be ok. HFTransformersTrainer init generates the output folder

horheynm: Lemme check what i can do w SparseAutoModel

Satrat (changes requested):
I don't love saving the python files on init rather than save. This seems like in should belong in `src/llmcompressor/transformers/finetune/session_mixin.py:save_model` instead, where we save the model itself and tokenizer. Or I suppose it could also go in `src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py:modify_save_pretrained`

horheynm:
> I don't love saving the python files on init rather than save. This seems like in should belong in `src/llmcompressor/transformers/finetune/session_mixin.py:save_model` instead, where we save the model itself and tokenizer

I dont mind doing that, but with the current flow it doesnt go to save_model

horheynm:
> src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py

Ok it does go to `modify_save_pretrained`, can save it there

src/llmcompressor/transformers/finetune/trainer.py
horheynm: SparseAutoModel gets called before HFTransformersTrainer, so there wont be a place to save the .py files.

Satrat (approved):

PR #87, opened 2024-08-14
MOE example with warning

rahul-tuli:
SUMMARY:
This PR changes the default for token_distribution to `1/num_experts` and raises a warning for if any expert sees lesser percentage of tokens, this also adds an example to quantize an MOE model to FP8 in the examples directory along with a `README.md`

TEST PLAN:
ran the example attached and made sure the right warnings were raised based on threshold and the percentage of tokens seen by each expert

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208216504432981


examples/quantizing_moe_fp8/tinymixtral_4_248_moe.py
```diff
@@ -0,0 +1,29 @@
+from llmcompressor.modifiers.quantization.gptq import GPTQModifier
+from llmcompressor.transformers import oneshot
+
+# model to quantize
+model = "Isotonic/TinyMixtral-4x248M-MoE"
```
Satrat, line 5: Does this model give valid results at all? I'm thinking it might make sense to use a larger model like mistralai/Mixtral-8x7B-Instruct-v0.1 (We would then also want to enable cpu offloading and run GPTQ sequentially)

Satrat (dismissed):

src/llmcompressor/modifiers/utils/pytorch_helpers.py
```diff
@@ -100,3 +105,28 @@ def run_calibration_forward(
...
+    Check if the model is a mixture of experts model
+
+    :param model: the model to check
+    :return: True if the model is a mixture of experts model
+    """
+
+    # Check for MoE components
+    for _, module in model.named_modules():
+        module_name = module.__class__.__name__
+        if "MoE" in module_name or "Expert" in module_name:
```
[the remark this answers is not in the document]
rahul-tuli: I've left this if condition as is because it's checking the class name(s), (which should respect the case); even if this condition fails, the one below it should catch if a module is an moe

rahul-tuli:
Note: The failing tests are unrelated to this PR

examples/quantizing_moe_fp8/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+from llmcompressor.modifiers.quantization.gptq import GPTQModifier
+
+config_groups = {
+                "group_0": {
+                    "targets": ["Linear"],
+                    "input_activations": None,
+                    "output_activations": None,
+                    "weights": {
+                        "num_bits": 8
```
kylesayrs, line 89: Need commas here

kylesayrs (changes requested):

horheynm (dismissed):

examples/quantizing_moe_fp8/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+The provided example script demonstrates an end-to-end process for applying the quantization algorithm:
+
+```bash
+python3 mixtral_moe.py
+```
+
+## Creating a Quantized MoE Model
+
+This example leverages `LLMCompressor` and `Compressed-Tensors` to create an FP8-quantized `Mixtral-8x7B-Instruct-v0.1` model. The model is calibrated and trained using the `open_platypus` dataset.
```
dsikka, line 25: should probably just list the package names? `llm-compressor` and `compressed-tensors`

examples/quantizing_moe_fp8/mixtral_moe.py
```diff
@@ -0,0 +1,59 @@
+from typing import List
+
+from transformers import AutoTokenizer
+
```
dsikka, line 4: Could we add an int4 example? vLLM at the moment only supports int4 models. 

It would be good to add that example and then validate that it runs smoothly by adding it to our e2e tests on vLLM.

examples/quantizing_moe_fp8/README.md
dsikka: let's also make this 4 bits

src/llmcompressor/transformers/finetune/data/data_args.py
```diff
@@ -154,13 +154,13 @@ class DataTrainingArguments(CustomDataTrainingArguments):
         },
     )
     min_tokens_per_module: Optional[float] = field(
-        default=0.2,
+        default=None,
```
dsikka, line 157: Why are we changing the default here?

dsikka (changes requested):
question + it would be good to add an MoE example for int4 + updated e2e tests with whatever is in the example to make sure it runs in vLLM

src/llmcompressor/transformers/finetune/data/data_args.py
rahul-tuli: This is calculated during run time now and defaults to 1/num_experts; since this would be dependent on the model

examples/quantizing_moe_fp8/README.md
dsikka: nvm I misread.

examples/quantizing_moe_fp8/README.md
rahul-tuli: Done!

src/llmcompressor/transformers/finetune/data/data_args.py
```diff
@@ -154,13 +154,13 @@ class DataTrainingArguments(CustomDataTrainingArguments):
...
         metadata={
             "help": (
                 "The minimum percentage of tokens (out of the total number) "
                 "that the module should 'receive' throughout the forward "
                 "pass of the calibration. If a module receives fewer tokens, "
-                "a warning will be logged."
+                "a warning will be logged. Defaults to 1/num_of_experts"
```
dsikka, line 163: isn't this default for all models? not just MoE?

examples/quantizing_moe_fp8/README.md
rahul-tuli: Good catch!

examples/quantizing_moe_fp8/mixtral_moe.py
rahul-tuli: Will add that as a follow up task, could you create a ticket to track?

src/llmcompressor/transformers/finetune/data/data_args.py
rahul-tuli: All models don't have experts so no this applies only to  MOEs

dsikka: sorry I meant you're updating the help for `min_tokens_per_module` in `DataTrainingArguments`  - wouldn't these arguments be used for cases that are not MoEs?

rahul-tuli: Syncing offline

kylesayrs (approved):

horheynm (approved):

PR #90, opened 2024-08-14
Propagate `trust_remote_code` Argument

### LINKED ISSUE #89: Add support to trust_remote_code
**Is your feature request related to a problem? Please describe.**
I had an issue using oneshot() for a model that required trust_remote_code since the instantiation of the tokenizer also requires this argument. I was able to circumvent by passing the tokenizer to oneshot(), but this still is a needed feature.

**Describe the solution you'd like**
Add trust_remote_code as a valid argument and use it when instantiating the model and the tokenizer.

kylesayrs:
SUMMARY:
FIX #89

The `trust_remote_code` argument is propagated to the following class initializers
* tokenizer
* model
* teacher model
>! Note that `trust_remote_code` is already propagated to compressor from model

Model kwargs are added to functions which construct models to allow for the `trust_remote_code` argument to be passed
* `custom_offload_device_map`
* `calculate_offload_device_map`

TEST PLAN:
The tokenizer and model propagations were both tested using `microsoft/Phi-3-small-128k-instruct`, which requires the trust remote code option. One regression test of the one_shot function was performed.

### DIFF
--- a/src/llmcompressor/transformers/compression/helpers.py
+++ b/src/llmcompressor/transformers/compression/helpers.py
@@ -165,7 +165,7 @@ def custom_offload_device_map(
     :param max_memory_per_gpu: Max memory to allocate on each GPU, as either a string
         such as "10GB" or an integer number of bytes
     :param num_gpus: number of gpus to utilize
-    :param \**model_kwargs: keyword arguments to pass to model initializer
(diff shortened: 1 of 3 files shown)


Satrat (approved):
nice work!

rahul-tuli (approved):

mgoin (approved):

PR #100, opened 2024-08-20
Validate Recipe Parsing Output

kylesayrs:
Previously there was an issue with tests which resulted in test yaml paths being read incorrectly. The current behavior is that the path string is parsed as if it was a yaml. This returns a `str` type which is not only wrong content but leads to type conflicts downstream.

Now the user gets a clean error
```
Could not parse recipe from string tests/llmcompressor/transformers/obcq/recipes/quant_and_sparse.yaml. If you meant load from a file, please make sure that the specified file path exists
```


src/llmcompressor/recipe/recipe.py
```diff
@@ -593,13 +593,20 @@ def _load_json_or_yaml_string(content: str) -> Dict[str, Any]:
...
-        return json.loads(content)
+        ret = json.loads(content)
     except json.JSONDecodeError:
         try:
-            return yaml.safe_load(content)
+            ret = yaml.safe_load(content)
         except yaml.YAMLError as err:
             raise ValueError(f"Could not parse recipe from string {content}") from err
 
+    if not isinstance(ret, dict):
```
dsikka, line 603: Would this value error be redundant? In what cases would you not get a dictionary and not have hit the Value Error on line 601?

kylesayrs: @dsikka This happens for any string which does not look like a dictionary, for example
```python3
import yaml
assert isinstance(yaml.safe_load("1"), int)
assert isinstance(yaml.safe_load("asdf"), str)
assert isinstance(yaml.safe_load("my/mistaken/path"), str)
```

dsikka: interesting, so yaml.safe_load() will load the strings without issue?

kylesayrs: Yes, `safe` refers to how pyyaml will throw an error if the string contains dangerous python objects such as functions, libraries, ect. It does not guarantee that the deserialized value is of dictionary type.

PR #103, opened 2024-08-22
remove workflows + update runners

dsikka:
# Summary
- Many of the workflows are no longer relevant/are being moved to another repo. This PR cleans up to remove the workflows
- Also updates the runners to get the workflows to run


.github/workflows/test-check.yaml
```diff
@@ -74,26 +70,16 @@ jobs:
       - name: "⚙️ Install dependencies"
         run: pip3 install .[dev]
       - name: "🔬 Running pytorch tests"
-        run: make test TARGETS=pytorch
+        run: |
+          pytest tests/llmcompressor/pytorch
```
rahul-tuli, line 74: Could we run with -v?

rahul-tuli (approved):

.github/workflows/test-check.yaml
```diff
@@ -74,26 +70,16 @@ jobs:
       - name: "⚙️ Install dependencies"
         run: pip3 install .[dev]
       - name: "🔬 Running pytorch tests"
-        run: make test TARGETS=pytorch
+        run: |
```
kylesayrs, line 73: No need for working-dir?

kylesayrs (approved):

.github/workflows/test-check.yaml
dsikka: the checked out code will already bring you to the root level of the repo

.github/workflows/test-check.yaml
x67samurai: Yeah I usually add -s for verbose error messages if fail

horheynm: I usually add -s so it tells the error message if fail. 
-v applies to both success and fail?
If so it might cause longer run time

dsikka: I think -s will also show prints but -v doesnt always. I'm gonna land this for now and we can adjust if we aren't happy with the level of logging. Does that sound good?

horheynm: Yes

PR #117, opened 2024-08-26
Add vLLM e2e tests

dsikka:
# Summary
- Add e2e test with vLLM 
- Adds test cases for fp8, int8, w4a16, w8a16

# Testing
- Tested on `llm-compressor-testing`: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/10568408144/workflow


tests/e2e/vLLM/test_vllm.py
```diff
@@ -0,0 +1,108 @@
...
+            **self.oneshot_kwargs,
+            output_dir=self.save_dir,
+            clear_sparse_session=True,
+            oneshot_device=self.device,
+        )
+
+        # Run vLLM with saved model
+        print("================= RUNNING vLLM =========================")
+        sampling_params = SamplingParams(temperature=0.80, top_p=0.95)
+        llm = LLM(model=self.save_dir)
```
robertgshaw2-neuralmagic, line 101: having a test for tp>1 is also a good idea if we can

tests/e2e/vLLM/test_vllm.py
```diff
@@ -0,0 +1,108 @@
...
+    dataset_split = None
+
+    def setUp(self):
+        print("========== RUNNING ==============")
+        print(self.scheme)
+
+        self.save_dir = None
+        self.device = "cuda:0"
+        self.oneshot_kwargs = {}
+        self.num_calibration_samples = 512
```
robertgshaw2-neuralmagic, line 40: you could get away with a lot fewer samples here, especially if not testing correctness of vllm

kylesayrs (approved):

tests/e2e/vLLM/test_vllm.py
```diff
@@ -0,0 +1,108 @@
...
+            oneshot_device=self.device,
+        )
+
+        # Run vLLM with saved model
+        print("================= RUNNING vLLM =========================")
+        sampling_params = SamplingParams(temperature=0.80, top_p=0.95)
+        llm = LLM(model=self.save_dir)
+        outputs = llm.generate(self.prompts, sampling_params)
+        print("================= vLLM GENERATION ======================")
+        print(outputs)
```
kylesayrs, line 104: Logic for perplexity tests can be borrowed from 
https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/transformers/compression/test_quantization.py

kylesayrs (approved):

tests/e2e/vLLM/test_vllm.py
robertgshaw2-neuralmagic: I would suggest running gsm8k on 200 samples

tests/e2e/vLLM/test_vllm.py
```diff
@@ -0,0 +1,119 @@
...
+        sampling_params = SamplingParams(temperature=0.80, top_p=0.95)
+        llm = LLM(model=self.save_dir)
+        outputs = llm.generate(self.prompts, sampling_params)
+        print("================= vLLM GENERATION ======================")
+        for output in outputs:
+            assert output
+            prompt = output.prompt
+            generated_text = output.outputs[0].text
+            print("PROMPT", prompt)
+            print("GENERATED TEXT", generated_text)
```
Satrat, line 116: In addition to printing is it possible to store a perplexity metric or similar? I think it would be good to automatically determine if the model is outputting gibberish or not

Satrat (commented):
Looks great and covers all the main cases I can think of! Just had one note on validating output

tests/e2e/vLLM/test_vllm.py
dsikka: Yah the plan is to add a perplexity check but until we finalize runners, I'm using smaller models and perplexity for that likely isn't worth it

tests/e2e/vLLM/test_vllm.py
dsikka: Yah I think that'll be a follow-up test since the structure will change a bit to deal with tp>1 with the same process

I do think that's more of a vLLM test. If anything, we could extend this to publish test models which are then pulled down for all vllm tests.

Satrat (approved):

PR #120, opened 2024-08-27
GPTQ Algorithm Cleanup

kylesayrs:
## Purpose ##
1. Clean up implementation for easier reading (comments, better structure)
2. Allow the algorithm to be skipped if the layer is not being targeted
3. Fix bug where layer is not frozen after QuantizationModifier
4. Prevent weight observer misuse
5. Depreciate weight_fake_quant use case

## Changes ##
* ensure that `freeze_quantization` is True (default), even if QuantizationModifier is wrapped by GPTQModifier
* implement `get_attr_chain` helper function to be used for getting chained attributes
* use `get_attr_chain` to get weight quantization arguments and skip computation if weight does not have valid args
* directly use memoryless observer to avoid misuse with unsupported observers
* perform transpose and float conversion in place to reduce memory use
* break out logging operations to separate function
* remove weight_fake_quant cases

## Testing ##
Regression tested saving, loading, and vllm inferencing with group quantized model


src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -108,17 +123,20 @@ def compress(
...
+            scale = weight_fake_quant.scale
+            zero_point = weight_fake_quant.zero_point
+            dtype = weight_fake_quant.dtype
+            tensor_scheme = weight_fake_quant.qscheme in [
+                torch.per_tensor_affine,
+                torch.per_tensor_symmetric,
+            ]
+        else:  # weight_quant_args is not None
+            observer = MemorylessObserver(weight_quant_args)
+            scale, zero_point = observer(W)
```
Satrat, line 137: Could you add a comment noting we need this scale/zp update for activation reordering?

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -152,61 +170,46 @@ def compress(
...
+                    strategy = weight_quant_args.strategy
+                    if strategy == QuantizationStrategy.TENSOR:
+                        q = fake_quantize(
+                            q,
+                            scale,
+                            zero_point,
+                            self.layer.quantization_scheme.weights,
+                        )
+                    elif strategy == QuantizationStrategy.CHANNEL:
+                        # TODO: for channelwise why isn't this just a 1d tensor?
```
Satrat, line 189: To answer the TODO: it makes it more standardized with the group case

Satrat: Hm I think I actually wrote the original todo lol

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -228,48 +232,55 @@ def compress(
...
-                            f"GPU {i} | usage: {perc:.2f}%"
-                            f" | total memory: {total_memory} GB"
-                        ),
-                    )
-
-            logger.log(
-                "METRIC",
-                f"Compressed layer size: {get_layer_size_bytes(self.layer)} MB",
-            )
+            self.log_metrics(tick, Losses)
```
Satrat, line 235: Nice, I like moving this out to a helper

Satrat (changes requested):
Looks good, thanks for cleaning this up! Had a few minor notes, and could we also add a test to confirm skipping layers works as intended?

src/llmcompressor/utils/helpers.py
```diff
@@ -1008,3 +1009,35 @@ def import_from_path(path: str) -> str:
         return getattr(module, class_name)
     except AttributeError:
         raise AttributeError(f"Cannot find {class_name} in {_path}")
+
+
+def getattr_chain(obj: Any, chain_str: str, *args, **kwargs) -> Any:
```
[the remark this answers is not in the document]
kylesayrs: It's a little pedantic but this signature is required to copy the behavior of `getattr`
> When a default argument is given, it is returned when the attribute doesn't exist; without it, an exception is raised in that case.

If `default` defaults to `None`, then there is no way to tell if the user passed `default=None` as an argument or if `default` was assigned to `None` by the default argument constructor

```
getattr_chain(obj, "dne", None) -> returns None
getattr_chain(obj, "dne") -> raises an Exception
```

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -169,9 +169,7 @@ def on_initialize(self, state: "State", **kwargs) -> bool:
         if not self.initialized_structure_:
             self.on_initialize_structure(state, **kwargs)
         if self.quantization_modifier_:
-            self.quantization_modifier_.initialize(
-                state, freeze_quantization=False, **kwargs
-            )
+            self.quantization_modifier_.initialize(state, **kwargs)
```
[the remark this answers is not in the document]
kylesayrs: From grepping llm-compressor's source, this is the only use of the arg. Let's remove it

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
kylesayrs: In the future we'll only do this in the activation ordering case

kylesayrs: Added, in the future we'll only update for activation ordering

src/llmcompressor/utils/helpers.py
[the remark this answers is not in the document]
kylesayrs: This is useful for cases when a user want to distinguish between "the value at the specified location is None" and "there is no value at the specified location"

kylesayrs:
@Satrat Can you specify what you're looking for in a skip test?

Satrat:
> @Satrat Can you specify what you're looking for in a skip test?

You could just initialize a module with some modules skipped (more than the lm_head) and others quantized, then search the logs for the debug string, or just testing your getattr_chain helper function directly on the model would be fine too

src/llmcompressor/utils/helpers.py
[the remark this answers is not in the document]
Satrat: Gotcha, are you aware of `operator.attrgetter` in the standard python library? Its similar to the helper you've written but doesn't support passing in a default. You might consider making this helper a wrapper of that and just adding in the default functionality rather than implementing the chaining directly. Not a requirement, but if something already exists as a stdlib helper I like to default to that rather than reimplementing

Satrat (approved):

kylesayrs:
Yeah the failing base test is because of a bug from the previous release which I fixed in the main branch
See: https://github.com/neuralmagic/compressed-tensors/blame/4b214e582c8434733efea79239cfadec9358b7fb/src/compressed_tensors/quantization/observers/base.py#L165-L167

kylesayrs:
Using my local machine and the main branch of compressed_tensors, I confirmed that the `tests/llmcompressor/modifiers/` and `tests/llmcompressor/transformers/compression/` are passing

Satrat (approved):

src/llmcompressor/utils/helpers.py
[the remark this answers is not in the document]
kylesayrs: I wasn't aware at the time. I agree using the standard `attrgetter` is better, although I do like being able to pass a default value without resorting to try except. Handling exceptions is fairly costly and annoying to read in python, so I think this function still has a place.

PR #140, opened 2024-09-03
Fix recipe nested in configs

kylesayrs:
## Purpose ##
* Fix failing tests due to recipe file being nested in configs

## Changes ##
* Revert back to original implementation of `parse_params` rather than recursing


tests/testing_utils.py
```diff
@@ -68,54 +68,48 @@ def _validate_test_config(config: dict):
 # Set cadence in the config. The environment must set if nightly, weekly or commit
 # tests are running
 def parse_params(
-    path: str, type: Optional[str] = None
+    configs_directory: Union[list, str], type: Optional[str] = None
```
dsikka, line 71: can you confirm that this runs the e2e tests correctly?

horheynm (approved):

horheynm:
Approved contingent on Dipika's comment about end to end testing

PR #141, opened 2024-09-04
[KV Cache] kv-cache end to end unit tests

horheynm:
SUMMARY:
Tests for 
https://github.com/neuralmagic/compressed-tensors/pull/148

Note: 
Blocked on transformers to update to include
https://github.com/huggingface/transformers/commit/181c962aabb4be59ddba82071d82adbbe3b3922d


dsikka (changes requested):
This is a fantastic test. Since we're using some pretty big models, we should restructure this a bit to leverage the config architecture. 

Basically for every model, you should create a yaml file which defines some required fields (`cadence` , `test_type`) and then any other required fields which are relevant for each test case.  Example file and directory structure: https://github.com/vllm-project/llm-compressor/tree/main/tests/e2e/vLLM/configs/FP8

For the larger models, we can move these test to the `e2e` folder (so they would be identical to the vLLM tests above). 

For the smaller models (so just TinyLlama), we can just have it run on a per-commit basis; similar to https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/transformers/obcq/test_consecutive_runs.py `TestConsecutiveRunsSmall`

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,311 @@
...
+from compressed_tensors.quantization.utils.helpers import iter_named_quantizable_modules
+from datasets import load_dataset
+from safetensors import safe_open
+from transformers import AutoTokenizer
+
+from llmcompressor.transformers import SparseAutoModelForCausalLM, oneshot
+
+MODEL_IDS = [
+    "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+    "microsoft/Phi-3-mini-4k-instruct",
```
dsikka, line 16: How long do we think the Phi Model will take to run?

dsikka (approved):

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,310 @@
...
+        dataset="open_platypus",
+        recipe=recipe,
+        num_calibration_samples=16,
+    )
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        for model_id in MODEL_IDS:
+            oneshot_args["output_dir"] = os.path.join(tmpdir, model_id)
+            used_args["output_dir"] = oneshot_args["output_dir"]
+            yield oneshot(model=model_id, **oneshot_args), used_args
```
kylesayrs, line 57: Why not cache the results of this generator? Then you can test the results of each model with each test, rather than only testing the first model with the first test and the second model with the second test

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,310 @@
...
+
+@pytest.fixture(scope="session")
+def oneshot_fixture():
+    num_bits = 8
+    _type = "float"
+    strategy = "tensor"
+    dynamic = False
+    symmetric = True
+
+    recipe = f"""
```
kylesayrs, line 27: It might be easier to just create a Recipe object, rather than having to serialize it into a string and dict

kylesayrs:
Wouldn't these tests fail because loading quantization configs from `ModelCompressor.from_pretrained` is [not supported](https://github.com/neuralmagic/compressed-tensors/pull/173)?

tests/llmcompressor/test_kv_cache.py
horheynm: oh its does that already

horheynm:
> Wouldn't these tests fail because loading quantization configs from `ModelCompressor.from_pretrained` is [not supported](https://github.com/neuralmagic/compressed-tensors/pull/173)?

does that path use
https://github.com/huggingface/transformers/commit/181c962aabb4be59ddba82071d82adbbe3b3922d
?

kylesayrs:
@horheynm What I'm referring to has nothing to do with your changes (although that's another reason why some of these tests would be expected to fail). I'm referring to how `ModelCompressor` saves as a q config, but only loads c configs. This means that without the above patch, we cannot reload a model with `ModelCompressor` that we have saved.

tests/llmcompressor/test_kv_cache.py
kylesayrs: @horheynm As I read it, the tests which use this feature only run with one model each (for example, `test_kv_cache_config_format` only runs with the `TinyLlama/TinyLlama-1.1B-Chat-v1.0` model. Is that incorrect?

kylesayrs:
Just noticing that some of these tests don't pass on GPUs

```
WARNING  compressed_tensors.quantization.lifecycle.calibration:calibration.py:47 Attempting set module with status None to calibration mode. but status is not initialized - you may be calibrating an uninitialized module which may fail or attempting to re-calibrate a frozen module 

FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_config_format - AttributeError: 'NoneType' object has no attribute 'quantization_config'                                  
FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_model_state_dict_attr - AttributeError: 'NoneType' object has no attribute 'quantization_config'                          
FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_model_populate_kv_scales_only - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!                                                                                                                                                                
FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_with_gptq - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!                                                                                                                                                                                    
ERROR tests/llmcompressor/test_kv_cache.py::test_kv_cache_with_gptq - exceptiongroup.BaseExceptionGroup: errors while tearing down <Session  exitstatus=<ExitCode.OK: 0> testsfailed=4 testscollected=4> (2 sub-exceptions) 
```

1. I know the warnings are being actively tracked, but we should try to figure out if they're related to kv_cache
2. The None compression configs are related to dropping loading support [as previously mentioned](https://github.com/vllm-project/llm-compressor/pull/141#issuecomment-2388993395)
3. We should try to figure out if the cuda devices error is related to kv_cache
4. I expect the teardown failure is just related to the test failing in general

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,310 @@
...
+    scheme = quant_config.kv_cache_scheme
+    assert scheme.num_bits == used_args["num_bits"]
+    assert scheme.type == used_args["_type"]
+    assert scheme.strategy == used_args["strategy"]
+    assert scheme.dynamic == used_args["dynamic"]
+    assert scheme.symmetric == used_args["symmetric"]
+
+
+def test_kv_cache_model_state_dict_attr(oneshot_fixture):
+    model, used_args = oneshot_fixture
```
kylesayrs, line 81: `model` is not used. Infact, `model is None` since `oneshot` does not return anything

tests/llmcompressor/test_kv_cache.py
kylesayrs: I just confirmed through debugging that `test_kv_cache_config_format` only runs with `TinyLlama/TinyLlama-1.1B-Chat-v1.0`. What I'm suggesting is maybe you could cache the oneshot results and then you'd be able to run each of your tests with each of the models

horheynm:
> Just noticing that some of these tests don't pass on GPUs
> 
> ```
> WARNING  compressed_tensors.quantization.lifecycle.calibration:calibration.py:47 Attempting set module with status None to calibration mode. but status is not initialized - you may be calibrating an uninitialized module which may fail or attempting to re-calibrate a frozen module 
> 
> FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_config_format - AttributeError: 'NoneType' object has no attribute 'quantization_config'                                  
> FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_model_state_dict_attr - AttributeError: 'NoneType' object has no attribute 'quantization_config'                          
> FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_model_populate_kv_scales_only - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!                                                                                                                                                                
> FAILED tests/llmcompressor/test_kv_cache.py::test_kv_cache_with_gptq - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!                                                                                                                                                                                    
> ERROR tests/llmcompressor/test_kv_cache.py::test_kv_cache_with_gptq - exceptiongroup.BaseExceptionGroup: errors while tearing down <Session  exitstatus=<ExitCode.OK: 0> testsfailed=4 testscollected=4> (2 sub-exceptions) 
> ```
> 
> 1. I know the warnings are being actively tracked, but we should try to figure out if they're related to kv_cache
> 2. The None compression configs are related to dropping loading support [as previously mentioned](https://github.com/vllm-project/llm-compressor/pull/141#issuecomment-2388993395)
> 3. We should try to figure out if the cuda devices error is related to kv_cache
> 4. I expect the teardown failure is just related to the test failing in general

Kyle, thanks for going over. 

so the first two error is caused by HF not being updated. 
Second can be avoided by setting cuda to just one gpu.

markurtz:
@horheynm @dsikka @kylesayrs do we have an update on whether this is working correctly? Or is this something we should hold off on until we have the GPU runners available to validate with automation?

kylesayrs (changes requested):
Change to AutoModel before merging

horheynm:
> Change to AutoModel before merging

great catch

examples/quantization_kv_cache/llama3_fp8_kv_example.py
```diff
@@ -5,8 +5,10 @@
 from llmcompressor.transformers import oneshot
 
 # Select model and load it.
-MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
-model = AutoModelForCausalLM.from_pretrained(
+# MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
+MODEL_ID="TinyLlama/TinyLlama-1.1B-Chat-v1.0"
+
+model = SparseAutoModelForCausalLM.from_pretrained(
```
dsikka, line 11: update to use automodel and then rerun to verify everything looks good.

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,312 @@
...
+from compressed_tensors.quantization.utils.helpers import iter_named_quantizable_modules
+from datasets import load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor.core import reset_session
+from llmcompressor.transformers import oneshot
+
+MODEL_IDS = [
+    "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+    "microsoft/Phi-3-mini-4k-instruct",
```
dsikka, line 17: move to under `transformers` so that this uses the gpu runners

horheynm: ok

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,312 @@
...
+            model_args[output_path] = used_args
+            oneshot(model=model_id, **oneshot_args)
+            reset_session()
+
+        yield model_args
+
+
+def test_kv_cache_config_format(oneshot_fixture):
+    for output_dir, used_args in oneshot_fixture.items():
+        compressor = ModelCompressor.from_pretrained(output_dir)
```
kylesayrs, line 69: nit: call AutoConfig directly

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,312 @@
...
+        assert scheme.num_bits == used_args["num_bits"]
+        assert scheme.type == used_args["_type"]
+        assert scheme.strategy == used_args["strategy"]
+        assert scheme.dynamic == used_args["dynamic"]
+        assert scheme.symmetric == used_args["symmetric"]
+
+
+def test_kv_cache_model_state_dict_attr(oneshot_fixture):
+    for output_dir, _ in oneshot_fixture.items():
+        model = AutoModelForCausalLM.from_pretrained(output_dir)
```
kylesayrs, line 86: nit: init_empty_weights

tests/llmcompressor/test_kv_cache.py
```diff
@@ -0,0 +1,312 @@
...
+
+    # check the values match
+    scheme = quant_config.kv_cache_scheme
+    assert scheme.num_bits == kv_cache_num_bits
+    assert scheme.type == kv_cache_type
+    assert scheme.strategy == kv_cache_strategy
+    assert scheme.dynamic == kv_cache_dynamic
+    assert scheme.symmetric == kv_cache_symmetric
+
+    model = AutoModelForCausalLM.from_pretrained(output_dir)
```
kylesayrs, line 301: nit: init_empty_weights

kylesayrs (dismissed):

tests/llmcompressor/transformers/kv_cache/configs/gptq_kv_cache/gptq_kv_cache.yaml
```diff
@@ -0,0 +1,2 @@
+cadence: "commit"
+test_type: "regression"
```
dsikka, line 2: why is this empty?

tests/llmcompressor/transformers/kv_cache/configs/kv_cache/phi3.yaml
```diff
@@ -0,0 +1,3 @@
+cadence: "commit"
+test_type: "regression"
+model: "microsoft/Phi-3-mini-4k-instruct"
```
dsikka, line 3: we only run e2e tests on a nightly cadence.
It should ideally be a larger model

dsikka (changes requested):

tests/llmcompressor/transformers/kv_cache/test_kv_cache.py
```diff
@@ -0,0 +1,330 @@
...
+
+KV_CACHE_CONFIGS_DIRECTORY = (
+    "tests/llmcompressor/transformers/kv_cache/configs/kv_cache"
+)
+KV_CACHE_GPTQ_CONFIGS_DIRECTORY = (
+    "tests/llmcompressor/transformers/kv_cache/configs/gptq_kv_cache"
+)
+
+
+@parameterized_class(parse_params(KV_CACHE_CONFIGS_DIRECTORY))
```
dsikka, line 31: Have we confirmed this works with the current e2e set-up?
Does the bash command need to be updated?

tests/llmcompressor/transformers/kv_cache/configs/gptq_kv_cache/gptq_kv_cache.yaml
horheynm: the tests use the same model for kv-cache-gptq so its in the code itself.
Regardless, the model needs to be here?

tests/llmcompressor/transformers/kv_cache/configs/kv_cache/phi3.yaml
horheynm: Nighly ok. 
Didnt we want a smaller model for deepseek?

horheynm:
@dsikka its ready

dsikka (commented):
this LGTM once the empty file is removed.

rahul-tuli (dismissed):
Nice!

tests/llmcompressor/transformers/kv_cache/test_kv_cache.py
```diff
@@ -0,0 +1,252 @@
...
+    ):
+        counts += 1
+        assert "self_attn" in name
+        assert hasattr(submodule, KVCacheScaleType.VALUE.value)
+        assert hasattr(submodule, KVCacheScaleType.KEY.value)
+    assert counts > 0
+
+
+@pytest.fixture(scope="session")
+def kv_cache_fixture():
```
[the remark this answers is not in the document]
horheynm: ok

rahul-tuli (approved):

dsikka (approved):

PR #144, opened 2024-09-05
fp8 config for original, default, default kv"

horheynm:
SUMMARY:

TEST PLAN:


markurtz:
@horheynm is this still in progress or should we close this out?

horheynm:
This was for generating to-be new default config, where None values are removed from the config. We do not need this for anyting

PR #149, opened 2024-09-09
Add tests for examples

dbarbuzzi:
SUMMARY:
This PR adds automated tests for (most of) the content in the `examples` folder:
* `examples/finetuning` is not covered
* `examples/quantizing_moe/deepseek_moe_w4a16.py` is covered but skipped due to exceptionally long run time

The general testing approach covers some of the 'install' snippets as well as the example commands in the READMEs, as well as running the various other example scripts within the covered folders. The commands/scripts are plainly executed and deemed passing as long as the execution completes successfully (i.e., exit code 0) — no further validation occurs.

TEST PLAN:
Automation is configured internally. There is one failure that is reported and unrelated to the automation code itself.


tests/examples/utils.py
```diff
@@ -0,0 +1,137 @@
...
+
+import pytest
+from bs4 import BeautifulSoup, ResultSet, Tag
+from cmarkgfm import github_flavored_markdown_to_html as gfm_to_html
+
+from tests.testing_utils import is_gpu_available, is_torch_available
+
+_T = TypeVar("_T")
+
+requires_gpu = pytest.mark.skipif(not is_gpu_available(), reason="GPU is required")
```
dsikka, line 16: what is this marking?

tests/examples/test_quantization_w8a8_fp8.py
```diff
@@ -0,0 +1,56 @@
...
+
+@pytest.mark.example
+@requires_gpu
+@requires_torch
+class TestQuantizationW8A8_FP8:
+    """
+    Tests for examples in the "quantization_w8a8_fp8" example folder.
+    """
+
+    def test_doc_example_command(self, example_dir: str, tmp_path: Path):
```
dsikka, line 30: we seem to repeat the code for this test case across the different examples - is there anyway we can reuse some of the functionality?

dbarbuzzi: I’ll take a look through today. There is probably a balance to abstract some of it while not obfuscating what the test is doing to shave off a bit of the duplication.

tests/examples/utils.py
dbarbuzzi: It gets imported in the test files and is used throughout as a decorator for all tests that require a GPU.

tests/examples/test_quantizing_moe.py
```diff
@@ -0,0 +1,75 @@
+import shlex
```
dsikka, line 1: Just so that you're aware: https://github.com/vllm-project/llm-compressor/pull/192/files#diff-7311b466a391f25c3d0951bbba59f2b3db9b3391ccc0c5949b22320b1e63a39e

tests/examples/test_quantization_w8a8_fp8.py
dbarbuzzi: I’ve added a pair of helper functions that reduce duplication in most test implementation code.

tests/examples/test_big_models_with_accelerate.py
```diff
@@ -0,0 +1,78 @@
...
+            ),
+            pytest.param(
+                "multi_gpu_int8_sequential_update.py",
+                "",
+                id="multi_gpu_int8_sequential_update",
+                marks=[requires_gpu_count(2), pytest.mark.multi_gpu],
+            ),
+        ],
+    )
+    @requires_gpu
```
andy-neuma, line 60: is the expectation that these will always require multiple GPUs? could we have separate annotations for number of gpus required?

tests/examples/test_compressed_inference.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.mark.example
+@requires_gpu
+class TestCompressedInference:
+    """
+    Tests for examples in the "compressed_inference" example folder.
+    """
+
+    def test_fp8_example_script(self, example_dir: str, tmp_path: Path):
```
andy-neuma, line 20: i think we need to annotate these so they don't run if the GPU doesn't support "fp8".

in general, all the tests need to be annotated, so they only run if the GPU supports the feature(s).

tests/examples/test_quantization_24_sparse_w4a16.py
```diff
@@ -0,0 +1,61 @@
...
+        assert result.returncode == 0, gen_cmd_fail_message(command, result)
+
+    def test_alternative_recipe(self, example_dir: str, tmp_path: Path):
+        """
+        Test for the example command in the README with the alternative recipe file.
+        """
+        shutil.copytree(str(Path.cwd() / example_dir), tmp_path / example_dir)
+
+        # replace recipe file with alternative
+        script_path = tmp_path / example_dir / "llama7b_sparse_w4a16.py"
```
andy-neuma, line 51: nit: it is likely better to consolidate all the raw strings into variables. some of them are repeated and this also guards against typos.

andy-neuma (commented):
looking good

tests/examples/test_big_models_with_accelerate.py
dbarbuzzi: There are multiple marks and decorators being used with great specificity:

**skip decorators** (used to determine if a test should be skipped)
* `requires_gpu` — the test requires at least one GPU, skipped if not met
* `requires_gpu_count(NUM)` — the test requires at least NUM GPUs, skipped if not met
* `requires_gpu_mem(AMOUNT)` — the test requires at least AMOUNT GPU memory (in GB), skipped if not met

**markers** (used to control test selection)
* `example` — indicates the test is for an example (applied to all in this PR)
* `multi_gpu` — indicates the test requires multiple GPUs (this is used to select tests for the `k8s-h100-duo` run as none require more than two/exclude them from the `-solo` run)

tests/examples/test_compressed_inference.py
dbarbuzzi: I can take a look into this, but I don't think that is an issue with the current target runners (A100/H100).

tests/examples/test_quantization_24_sparse_w4a16.py
dbarbuzzi: So, this is one of very few cases where raw strings aren't variables, and only one of them is re-used, meaning that for two of the three raw strings, swapping to variables would not introduce any safety, only noise. I will update the one string that gets repeated to use a variable.

tests/examples/test_compressed_inference.py
dbarbuzzi: @dsikka Are you able to help shed light here on GPU compatibility concerns regarding the "fp8" example(s)?

tests/examples/utils.py
dbarbuzzi: Sort of, yes. This is a pytest-compatible version of the existing decorator which is a very thin wrapper over the core logic that the other one is based on (`is_gpu_available()`), and it’s generally limited to the examples tests. As long as that underlying function exists and returns a boolean, this wrapper will not need any changes.

Unfortunately, to use the other existing “requires gpu” wrapper, I would have to swap to using `unittest` to compose the tests, adding a lot of overhead that pytest provides for free (namely both their built-in fixtures such as `tmp_path` as well as their ability to very easily create simple fixtures and other test decorators).

Alternatively, I did create a separate, more granular pytest skip decorator that checks for a specific number of GPUs and skips if there are fewer available. If you’d prefer, can drop the pytest version of `requires_gpu` and replace all of its usages with that decorator (e.g. `requires_gpu_count(1)`).

Let me know how you’d like me to rectify this.

dbarbuzzi: I have taken the latter approach — my pytest-compatible `requires_gpu` skip decorator has been resolved, and all uses of it have been replaced with `requires_gpu_count(1)`.

tests/examples/test_compressed_inference.py
dbarbuzzi: Per @dsikka, this is not a concern for this project.

dsikka (approved):

rahul-tuli (approved):
Great Job @dbarbuzzi

mgoin (approved):

PR #151, opened 2024-09-09
Fix help dialogue

kylesayrs:
## Purpose ##
* Fix bug where value error would be thrown if user tried to use any entrypoint with the `--help` flag
```
AttributeError: 'tuple' object has no attribute 'strip'
```

## Changes ##
* Replace the tuple which was giving the error with a string

## Testing ##
Running these commands now properly shows the help dialogue
```bash
llmcompressor.transformers.text_generation.oneshot -h
llmcompressor.transformers.text_generation.finetune -h
llmcompressor.transformers.text_generation.oneshot --help
llmcompressor.transformers.text_generation.finetune --help
```


rahul-tuli (approved):
I love it.
Thank you for fixing this. I really appreciate it.

kylesayrs:
@rahul-tuli Thank you for bringing it to attention!

PR #153, opened 2024-09-10
Fix Sparsity Unit Test

Satrat:
SUMMARY:
Fix for failing test detailed in https://app.asana.com/0/1208216504433080/1208216504433083/f. The issue was the quantization modifier was affecting the sparsity, and we weren't checking the mask in the appropriate location. 

TEST PLAN:
Failing nightly test now passes, confirmed locally on janice

### DIFF
--- a/src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
+++ b/src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
@@ -124,7 +124,6 @@ class SparseGptWrapper(ModuleCompressionWrapper):
                 torch.tensor(0, dtype=torch.bool),
             )
             current_sparsity = mask.sum() / W.numel()
-            print(current_sparsity, sparsity)
             if current_sparsity > sparsity:
                 raise ValueError(
                     "The target sparsity is lower than the sparsity "
--- a/src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
+++ b/src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
@@ -1,6 +1,11 @@
 import time
 
+from compressed_tensors.quantization.lifecycle.forward import (
+    maybe_calibrate_or_quantize,
+)
+
 from llmcompressor.modifiers.utils.compression_wrapper import ModuleCompressionWrapper
+from llmcompressor.utils import getattr_chain
 
 try:
     import transformers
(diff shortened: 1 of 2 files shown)


src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
```diff
@@ -83,6 +88,10 @@ def compress(
         final_shape = self.layer.weight.shape
         final_dtype = self.layer.weight.dtype
         W = self.layer.weight.data.clone()
+        args_loc = "quantization_scheme.weights"
+        weight_quant_args = getattr_chain(self.layer, args_loc, None)
+        if weight_quant_args is not None:
+            W = maybe_calibrate_or_quantize(self.layer, W, "weight", weight_quant_args)
```
dsikka, line 94: why do we need this additional call to `maybe_calibrate_or_quantize`?

src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
```diff
@@ -198,6 +209,8 @@ def compress(
         if isinstance(self.layer, transformers.Conv1D):
             W = W.t()
         W = W.reshape(final_shape).to(final_dtype)
+        if weight_quant_args is not None:
+            W = maybe_calibrate_or_quantize(self.layer, W, "weight", weight_quant_args)
```
dsikka, line 213: why do we need this additional call to maybe_calibrate_or_quantize?

src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
```diff
@@ -114,6 +123,14 @@ def compress(
                 torch.tensor(1, dtype=torch.bool),
                 torch.tensor(0, dtype=torch.bool),
             )
+            current_sparsity = mask.sum() / W.numel()
+            print(current_sparsity, sparsity)
```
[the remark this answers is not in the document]
Satrat: oops, done!

tests/llmcompressor/transformers/obcq/recipes/quant_and_sparse.yaml
```diff
@@ -13,12 +13,13 @@ test_stage:
           weights:
             num_bits: 8
           targets: ["Linear"]
+    GPTQModifier:
```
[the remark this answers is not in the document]
Satrat: QuantizationModifier and GPTQModifier should be in order, otherwise we end up in a situation were the "naive" minmax quantization is used for the first sparsity calculation, then the more accurate GPTQ quantization is applied afterwards. We want sparsity and quantization to be fully separate operations, not apply sparsity in the middle of the algorithm

src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
Satrat: the sparseGPT algorithm adjusts the weights and they aren't guaranteed to be in the quantized bins, we need to requantize after updating

PR #155, opened 2024-09-10
fix package name bug for nightly

dhuangnm:
SUMMARY:
fix package name bug for nightly where nightly wheel still uses "llmcompressor" instead of "llmcompressor-nightly".

TEST PLAN:
manually tested for nightly, release and dev builds


dsikka:
Will the name be updated by a bash script from another action?

dsikka (approved):

PR #157, opened 2024-09-10
DeepSeek: Fix Hessian Estimation

Satrat:
SUMMARY:
We were running out of memory using GPTQ on deepseek because our hessian memory estimation code assumes all the decoder layers are the same shape. This is not the case for deepseek. Updating the function to find the largest layer across all the decoders fixed the issue.

TEST PLAN:
```
import torch

# select a Mixture of Experts model for quantization
MODEL_ID = "deepseek-ai/DeepSeek-Coder-V2-Instruct"

# adjust based off number of desired GPUs
# if not enough memory is available, some layers will automatically be offlaoded to cpu
    MODEL_ID,
    reserve_for_hessians=True,
    num_gpus=2,
    torch_dtype=torch.bfloat16,
    trust_remote_code=True,
)

    MODEL_ID, device_map=device_map, torch_dtype=torch.bfloat16, trust_remote_code=True
)

# Select calibration dataset.
DATASET_SPLIT = "train_sft"
NUM_CALIBRATION_SAMPLES = 512
MAX_SEQUENCE_LENGTH = 2048


ds = load_dataset(DATASET_ID, split=DATASET_SPLIT)
ds = ds.shuffle(seed=42).select(range(NUM_CALIBRATION_SAMPLES))


def preprocess(example):
    return {
        "text": tokenizer.apply_chat_template(
            example["messages"],
            tokenize=False,
        )
    }


ds = ds.map(preprocess)


# Tokenize inputs.
def tokenize(sample):
    return tokenizer(
        sample["text"],
        padding=False,
        truncation=True,
        add_special_tokens=False,
    )


ds = ds.map(tokenize, remove_columns=ds.column_names)

# define a llmcompressor recipe for W416 quantization
# since the MoE gate layers are sensitive to quantization, we add them to the ignore
# list so they remain at full precision
recipe = [
    GPTQModifier(
        targets="Linear",
        scheme="W8A8",
        ignore=["lm_head", "re:.*mlp.gate$"],
        sequential_update=True,
    ),
]

oneshot(
    model=model,
    dataset=ds,
    recipe=recipe,
    save_compressed=True,
)
```

### DIFF
--- a/examples/quantization_24_sparse_w4a16/2:4_w4a16_group-128_recipe.yaml
+++ b/examples/quantization_24_sparse_w4a16/2:4_w4a16_group-128_recipe.yaml
@@ -23,12 +23,14 @@ quantization_stage:
   run_type: oneshot
   quantization_modifiers:
     GPTQModifier:
-      sequential_update: false
+      sequential_update: true
+      ignore: ["lm_head"]
       config_groups:
         group_0:
           weights:
             num_bits: 4
             type: "int"
             symmetric: true
-            strategy: "channel"
+            strategy: "group"
+            group_size: 128
           targets: ["Linear"]
--- a/examples/quantization_24_sparse_w4a16/2:4_w4a16_recipe.yaml
+++ b/examples/quantization_24_sparse_w4a16/2:4_w4a16_recipe.yaml
@@ -23,7 +23,7 @@ quantization_stage:
   run_type: oneshot
   quantization_modifiers:
     GPTQModifier:
-      sequential_update: false
+      sequential_update: true
       ignore: ["lm_head"]
       config_groups:
         group_0:
--- a/examples/quantization_24_sparse_w4a16/llama7b_sparse_w4a16.py
+++ b/examples/quantization_24_sparse_w4a16/llama7b_sparse_w4a16.py
@@ -23,7 +23,7 @@ max_seq_length = 512
 num_calibration_samples = 512
 
 # set training parameters for finetuning
-num_train_epochs = 0.5
+num_train_epochs = 0.01
 logging_steps = 500
 save_steps = 5000
 gradient_checkpointing = True  # saves memory during training
--- a/examples/trl_mixin/ex_trl_distillation.py
+++ b/examples/trl_mixin/ex_trl_distillation.py
@@ -58,6 +58,7 @@ training_args = TrainingArguments(
     logging_steps=50,
     gradient_checkpointing=True,
     bf16=True,
+    save_safetensors=False,  # workaround for shared tensors
 )
 trainer = SFTTrainer(
     model=model,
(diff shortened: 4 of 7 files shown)


src/llmcompressor/transformers/compression/helpers.py
```diff
@@ -107,21 +107,26 @@ def hessian_memory_requirements(model: torch.nn.Module) -> int:
...
-                    max_column_size = column_size
-
+    total_hessian_elems = {}
+    max_column_size = {}
+    for no_split_name, no_split_layer in transformer_layers.items():
+        total_hessian_elems[no_split_name] = 0
+        max_column_size[no_split_name] = 0
+        for name, module in no_split_layer.named_modules():
+            if isinstance(module, Linear):
+                print(name)
```
dsikka, line 117: remove

src/llmcompressor/transformers/compression/helpers.py
```diff
@@ -107,21 +107,26 @@ def hessian_memory_requirements(model: torch.nn.Module) -> int:
...
-    for _, module in single_layer.named_modules():
-        if isinstance(module, Linear):
-            for param in module.parameters():
-                column_size = param.shape[1]
-                total_hessian_elems += column_size * column_size
-                if column_size > max_column_size:
-                    # max extra memory for inverse calculation
-                    max_column_size = column_size
-
+    total_hessian_elems = {}
```
dsikka, line 110: just so I understand, instead of looking at just a single layer, we're looking at a list of layers, given by the `get_layers` function?

tests/llmcompressor/transformers/compression/run_compressed_configs/w8a8.yaml
```diff
@@ -1,3 +1,3 @@
 cadence: "commit"
 test_type: "regression"
-model_stub: "nm-testing/tinyllama-w4a16-compressed"
\ No newline at end of file
+model_stub: "nm-testing/tinyllama-w8a8-compressed"
```
dsikka, line 3: was this change related to this PR or just something you found?

dsikka (commented):
LGTM - this was tested on the full deepseekv2 model?

src/llmcompressor/transformers/compression/helpers.py
Satrat: Yep, we need to find the biggest layer and largest column to get the total max

tests/llmcompressor/transformers/compression/run_compressed_configs/w8a8.yaml
Satrat: Just another typo I found

Satrat:
> LGTM - this was tested on the full deepseekv2 model?

I let it run to completion on the lite model, the full model is really slow so I killed it halfway through

PR #160, opened 2024-09-10
Add MoE and Compressed Inference Examples

Satrat:
SUMMARY:
* Adding an additional MoE example for deepseek
* Adding an example for compressed inference using an fp8 model

TEST PLAN:
Confirmed scripts run locally


examples/quantizing_moe/deepseek_moe_w4a16.py
```diff
@@ -0,0 +1,86 @@
...
+SAVE_DIR = MODEL_ID.split("/")[1] + "-W4A16"
+
+oneshot(
+    model=model,
+    dataset=ds,
+    recipe=recipe,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+    save_compressed=True,
+    output_dir=SAVE_DIR,
```
dsikka, line 85: yay!

dsikka (commented):
We'll want to automate these examples like the rest of them and Domenic has been working on that here: https://github.com/vllm-project/llm-compressor/pull/149/files

Could you add a follow-up ticket and add it to the backlog on adding these two tickets to the automated examples as well? You can assign me then

dsikka:
Actually, scratch that, we should pull these in now

examples/quantizing_moe/deepseek_moe_w4a16.py
```diff
@@ -0,0 +1,86 @@
...
+)
```
dsikka, line 86: could we update to add a sample generation here as well?

examples/quantizing_moe/deepseek_moe_w4a16.py
```diff
@@ -0,0 +1,86 @@
...
+
+ds = ds.map(tokenize, remove_columns=ds.column_names)
+
+# define a llmcompressor recipe for W416 quantization
+# since the MoE gate layers are sensitive to quantization, we add them to the ignore
+# list so they remain at full precision
+recipe = [
+    GPTQModifier(
+        targets="Linear",
+        scheme="W4A16",
```
dsikka, line 70: This uses group size 128 which I dont think is actually compatible with the model's architecture

PR #166, opened 2024-09-11
Implement Missing Modifier Methods

kylesayrs:
## Purpose ##
* Fix model loading when recipe contains a modifier such as `OutputDistillationModifier` which previously did not implement critical functions like `on_initialize_structure` (see `reload_stories_distill.py` test script)
* Make implementing modifiers in the future easier than it is now
* Make existing implementations of modifiers easier to read by removing dummy methods

<details><summary>Traceback</summary>

```
  File "/home/ksayers/llm-compressor/reload_stories_distill.py", line 42, in <module>
  File "/home/ksayers/llm-compressor/src/llmcompressor/transformers/sparsification/sparse_model.py", line 152, in from_pretrained
    initialize_recipe(model=model, recipe_path=recipe)
  File "/home/ksayers/llm-compressor/src/llmcompressor/pytorch/model_load/helpers.py", line 77, in initialize_recipe
    pre_initialize_structure(model=model, recipe=recipe_path)
  File "/home/ksayers/llm-compressor/src/llmcompressor/core/session_functions.py", line 70, in pre_initialize_structure
    active_session().pre_initialize_structure(**kwargs)
  File "/home/ksayers/llm-compressor/src/llmcompressor/core/session.py", line 90, in pre_initialize_structure
    mod_data = self._lifecycle.pre_initialize_structure(
  File "/home/ksayers/llm-compressor/src/llmcompressor/core/lifecycle.py", line 94, in pre_initialize_structure
    data = mod.pre_initialize_structure(state=self.state, **extras)
  File "/home/ksayers/llm-compressor/src/llmcompressor/modifiers/stage.py", line 105, in pre_initialize_structure
    modifier.pre_initialize_structure(state, **kwargs)
  File "/home/ksayers/llm-compressor/src/llmcompressor/modifiers/modifier.py", line 87, in pre_initialize_structure
    self.on_initialize_structure(state, **kwargs)
  File "/home/ksayers/llm-compressor/src/llmcompressor/modifiers/modifier.py", line 234, in on_initialize_structure
    raise NotImplementedError()
NotImplementedError
```
</details>

## Changes ##
* Remove requirement for subclass implementations of methods most relevant to training modifiers such as `on_finalize`, `on_start`, `on_update`, `on_end`, and `on_event` methods
* Enforce requirement for subclass implementations of methods relevant to training and one shot modifiers such as `on_initialize`
* Remove requirement for subclass implementations of methods which should be depreciated in the future such as `on_initialize_structure`

## Testing ##
* Modifiers were found using command `grep -P 'class\s+\w+\s*\(.*\bModifier\b.*\)' -r src/` (thank you chatgpt)
* Used below test script to confirm changes.

<details><summary>reload_stories_distill.py</summary>

```python3
import os, shutil
from llmcompressor.core import create_session
from llmcompressor.transformers import (
    SparseAutoModelForCausalLM,
    oneshot, train
)

output_dir = "./distill_out"
if os.path.exists(output_dir):
    shutil.rmtree(output_dir)
dataset = "open_platypus"
concatenate_data = False
splits = "train[:50%]"
max_steps = 50
num_calibration_samples = 64
recipe_str = "tests/llmcompressor/transformers/finetune/test_finetune_recipe.yaml"

# base
    "Xenova/llama2.c-stories15M", device_map="auto"
)
distill_teacher = SparseAutoModelForCausalLM.from_pretrained(
    "Xenova/llama2.c-stories15M", device_map="auto"
)

# distill
with create_session():
    train(
        model=model,
        distill_teacher=distill_teacher,
        dataset=dataset,
        output_dir=output_dir,
        num_calibration_samples=num_calibration_samples,
        recipe=recipe_str,
        concatenate_data=concatenate_data,
        splits=splits,
        max_steps=max_steps,
    )

# load
    output_dir, device_map="auto"
)
```

</details>


kylesayrs:
Confirmed no new test failures

src/llmcompressor/modifiers/modifier.py
```diff
@@ -9,7 +10,7 @@
 __all__ = ["Modifier"]
 
 
-class Modifier(BaseModel, ModifierInterface):
+class Modifier(BaseModel, ModifierInterface, ABC):
```
horheynm, line 13: ModifierInterface is ABC, do we need ABC for Modifier?

src/llmcompressor/modifiers/modifier.py
```diff
@@ -255,7 +258,7 @@ def on_finalize(self, state: State, **kwargs) -> bool:
         :return: True if the modifier was finalized successfully,
             False otherwise
         """
-        raise NotImplementedError()
+        return True
```
horheynm, line 261: I think if it is an abstract method, it is good practice to raise NotImplementedError. 
If not abstract method, then having default logic is ok

src/llmcompressor/modifiers/obcq/base.py
```diff
@@ -6,7 +6,7 @@
 from torch.nn import Module
 from tqdm import tqdm
 
-from llmcompressor.core.state import State
+from llmcompressor.core import State
```
horheynm, line 9: Could you check if the pathway that uses always avoids circular import?
Usually then the library import is less strict circular import can be introduced

kylesayrs: The tests pass and cover importing all modifiers, so I think

src/llmcompressor/modifiers/modifier.py
kylesayrs: Agree

kylesayrs: I actually think that this is a github diff bug. This has already been changed in the branch: https://github.com/vllm-project/llm-compressor/blob/kylesayrs/modifier-missing-funcs/src/llmcompressor/modifiers/modifier.py

src/llmcompressor/modifiers/modifier.py
kylesayrs: Technically no, but abstract classes and methods are basically type hints anyways, so I think it's nice to keep it explicit

src/llmcompressor/modifiers/modifier.py
kylesayrs: Very strange, `git diff` also shows this change. I'll need to investigate this further

kylesayrs: Tried changing base, dummy commits, merging with main. This bug is also seen when running `git diff main kylesayrs/modifier-missing-funcs`. It's a somewhat known bug from what I've heard

horheynm (approved):

src/llmcompressor/modifiers/obcq/base.py
```diff
@@ -83,25 +83,12 @@ class SparseGPTModifier(Modifier):
...
-        return True
-
     def on_initialize(self, state: "State", **kwargs) -> bool:
         """
         Initialize and run the OBCQ algorithm on the current state
 
         :param state: session state storing input model and calibration data
         """
-        if not self.initialized_structure_:
-            self.on_initialize_structure(state, **kwargs)
```
dsikka, line 103: why is this check/condition ok to remove?

kylesayrs: Because `on_initialize_structure` does not do anything. Nearly all `on_initialize_structure` calls outside of GPTQ are no-ops, and GPTQ's `on_initialize_structure` should be depreciated anyways

dsikka (approved):

src/llmcompressor/modifiers/modifier.py
kylesayrs: Looks like when the merge occurred, github ignored the incorrect diff you see here and used the correct changes: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/modifier.py#L249

PR #167, opened 2024-09-11
Make `Modifier` an ABC

kylesayrs:
## Purpose ##
* Avoid coding mistakes such as the one described in #166 

## Changes ##
* Make `Modifier` an ABC

## Testing ##
* TODO


kylesayrs:
Folded into #166

PR #171, opened 2024-09-12
Add DeepSeek V2.5 Example

dsikka:
SUMMARY:
- Add an example for quantizing deepseek v2.5 and running on vLLM

### DIFF
--- a/examples/quantizing_moe/deepseek_moe_w4a16.py
+++ b/examples/quantizing_moe/deepseek_moe_w4a16.py
@@ -1,0 +1,121 @@
+import torch
+from datasets import load_dataset
+from transformers import AutoTokenizer
+
+from llmcompressor.modifiers.quantization import GPTQModifier
+from llmcompressor.transformers import SparseAutoModelForCausalLM, oneshot
+from llmcompressor.transformers.compression.helpers import calculate_offload_device_map
+
+# select a Mixture of Experts model for quantization
+MODEL_ID = "deepseek-ai/DeepSeek-V2.5"
+
+# adjust based off number of desired GPUs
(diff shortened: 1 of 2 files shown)


examples/quantizing_moe/deepseek_moe_w4a16.py
```diff
@@ -0,0 +1,121 @@
...
+# select a Mixture of Experts model for quantization
+MODEL_ID = "deepseek-ai/DeepSeek-V2.5"
+
+# adjust based off number of desired GPUs
+# if not enough memory is available, some layers will automatically be offlaoded to cpu
+device_map = calculate_offload_device_map(
+    MODEL_ID,
+    reserve_for_hessians=True,
+    num_gpus=2,
+    torch_dtype=torch.bfloat16,
```
mgoin, line 18: In the future we should rely on auto dtype, but it seems like deepseek v2.5 is bfloat16 so this is okay

examples/quantizing_moe/deepseek_recipe_w4a16.yaml
```diff
@@ -0,0 +1,9 @@
+quant_stage:
+  quant_modifiers:
+    GPTQModifier:
+      sequential_update: true
+      ignore: [lm_head, "re:.*mlp.gate$"]
+      config_groups:
+        group_0:
+          weights: {num_bits: 4, type: int, symmetric: true, strategy: channel, dynamic: false}
```
mgoin, line 8: We may want to use grouped and actorder for the final example based on how the accuracy comes out, but this is good for now

mgoin (approved):
LGTM only nits for future consideration

PR #177, opened 2024-09-16
GPTQ Quantized-weight Sequential Updating

kylesayrs:
## Background ##
There are many ways in which sequential updating can be implemented

1. (off/False) Since running non-sequentially, ie one forward pass, gives no accuracy benefits and uses the most memory, I would prefer to remove support for this
2. (block_unquantized) Running sequentially, block-by-block but without outputs from quantized weights saves memory, but does not give accuracy benefits. Mark Kurtz referred to this as essentially a bug in the sequential_update implementation, so I'd prefer not to support this
3. (block/layer) Unfortunately the names are badly conflicting here (LC refers to transformer blocks as layers), but this option refers to using quantized-weight outputs for each transformer block (ie true_sequential=False). I think this should be supported
4. (linear/true/True) This option refers to quantizing each linear layer within each block separately and continuing with weight-quantized outputs. I think this should be supported

## Purpose ##
* Implement options 1,3 for better replication of AutoGPTQ

## Changes ##
* always pre_compress during `initialize_compression` (has no effect)
* run additional weight-quantized forward pass to calculate weight-quantized intermediate outputs
* clean up typing in `LayerCompressor`

## Testing ##
* Compressed a model `sequential_update=True` and `sequential_update=False`
* Compared evaluation from `sequential_update=False` and `sequential_update=False` on main and found the model evaluations to be the same

This branch `sequential_update=False`
```
vllm (pretrained=/home/ksayers/llm-compressor/actorder20240917_144813,dtype=auto,add_bos_token=True,max_model_len=4096,tensor_parallel_size=1), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 32
|gsm8k|      3|flexible-extract|     5|exact_match|?  |0.2358|?  |0.0117|
|     |       |strict-match    |     5|exact_match|?  |0.2335|?  |0.0117|
```

main `sequential_update=False`
```
vllm (pretrained=/home/ksayers/llm-compressor/actorder20240917_145431,dtype=auto,add_bos_token=True,max_model_len=4096,tensor_parallel_size=1), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 32
|gsm8k|      3|flexible-extract|     5|exact_match|?  |0.2358|?  |0.0117|
|     |       |strict-match    |     5|exact_match|?  |0.2335|?  |0.0117|
```


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -240,12 +279,10 @@ def initialize_compression(
             args = self._pruning_arguments()
             comp_cls = self._compression_class()
             compressor = LayerCompressor(comp_cls, self.model, layer, idx, name, args)
-            if not self.sequential_update:
-                # add all batch processing hooks before the forward pass
-                compressor.pre_compress()
+            compressor.pre_compress()
```
Satrat, line 282: I don't think we want to change this. `pre_compress` triggers the initialization of the `GPTQWrapper`s, which allocates the memory for the Hessians. We lose the memory benefit of sequential updates by allocating the Hessians up front

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -270,28 +307,41 @@ def apply_compression(
...
         num_layers = len(self.compressible_layers_)
         for idx, layer_compressor in enumerate(self.layer_compressors_):
             logger.info(f"\n===== Compressing layer {idx+1}/{num_layers} " " =====")
 
-            # Prune/quantize using GPTQ
-            if self.sequential_update:
+            if self.sequential_update != UpdateMethod.OFF:
                 # in sequential mode we run the forward pass for each transformer layer
                 # one at a time, caching the intermediate outputs between layers
-                layer_compressor.pre_compress()
```
Satrat, line 288: Just echoing my above comment: this is the call that allocates the Hessian, so to save memory during sequential updates we want to leave this in

dsikka (changes requested):
It's not very clear from this PR in how you're differing between the layer or block cases you're adding in support for

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: Makes sense, I was wondering why we were delaying the precompression

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -26,6 +29,24 @@
...
+    Module: apply compression and proceed with weight-quantized outputs for each
+        module within a transformer block (layer). NOT IMPLEMENTED\n
+    Layer: apply compression and proceed with weight-quantized outputs for each
+        transformer block (layer)\n
+    Off: compute all outputs and hessians in one, weight-unquantized forward pass\n
+    """
+
+    MODULE = "module"
+    LAYER = "layer"
+    OFF = "off"
```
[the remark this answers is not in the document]
kylesayrs: I added the "off" enum to reduce the number of types the member `sequential_update` can take. I can use `None` instead if that's preferred

Satrat:
I think we should want to keep the current non-sequential implementation (option 1), because it is faster by ~20%. Does option 4 provide any accuracy, memory or speed boost compared to option 3? I know AutoGPTQ supports option 4, but unless we have a strong reason I don't think its priority to add it in now. 

As for the change from a boolean to an enum, I don't think its necessary until we decide what to do about option 4, can we leave it as a bool for now? No need to add extra scaffolding for something that isn't on the roadmap

src/llmcompressor/modifiers/utils/layer_compressor.py
```diff
@@ -171,7 +173,6 @@ def compress_module(module):
                 logger.info(f"Compressing {full_name}...")
                 module.compress(**self.args)
                 module.free()
-                print("done")
```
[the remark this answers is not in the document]
Satrat: I think this was just left in from me debugging, all good to delete it :)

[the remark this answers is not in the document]
kylesayrs: I don't think this log adds anything, so I'd rather remove it

kylesayrs:
@dsikka 
> It's not very clear from this PR in how you're differing between the layer or block cases you're adding in support for

Not sure what you're referring to. The PR description specifies implementations for options 1 and 3

dsikka:
> @dsikka
> 
> > It's not very clear from this PR in how you're differing between the layer or block cases you're adding in support for
> 
> Not sure what you're referring to. The PR description specifies implementations for options 1 and 3

We shouldn’t be adding it to the enum until it’s supported

kylesayrs:
Re-performed replication
```
vllm (pretrained=/home/ksayers/llm-compressor/actorder20240919_133905,dtype=auto,add_bos_token=True,max_model_len=4096,tensor_parallel_size=1), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 32
|gsm8k|      3|flexible-extract|     5|exact_match|?  |0.2358|?  |0.0117|
|     |       |strict-match    |     5|exact_match|?  |0.2335|?  |0.0117|
```

Satrat (approved):
LGTM, could you add in eval stats for sequential_update=True to the PR description as well? Just so we can easily track any regression

Satrat (changes requested):
Actually, looks like there is a test failure related to this PR:

```
FAILED tests/llmcompressor/transformers/obcq/test_consecutive_runs.py::TestConsecutiveRunsSmall_0_commit::test_consecutive_runs_small - pydantic_core._pydantic_core.ValidationError: 1 validation error for GPTQModifier
sequential_update
  Input should be a valid boolean [type=bool_type, input_value=None, input_type=NoneType]
  ```

**Comment by kylesayrs** [comment]
@Satrat My mistake, forgot to return the value after validation

**Comment by kylesayrs** [comment]
Looks like the test is passing now

**Review by Satrat — APPROVED** [review]

File: src/llmcompressor/modifiers/quantization/gptq/base.py [hunk 5]
```diff
@@ -277,21 +294,35 @@ def apply_compression(
...
 
-            # Prune/quantize using GPTQ
             if self.sequential_update:
                 # in sequential mode we run the forward pass for each transformer layer
                 # one at a time, caching the intermediate outputs between layers
-                layer_compressor.pre_compress()
                 logger.info(f"Calibrating {layer_compressor.name}...")
-                intermediates = layer_compressor.calibrate_layer(intermediates)
+                layer_compressor.pre_compress()
+                unquantized_outputs = layer_compressor.calibrate_layer(intermediates)
```
dsikka, line 312: should we be caching unquantized_outputs for error calculation you're doing as well?

kylesayrs: @dsikka What does "caching" refers to in this context? The `unquantized_outputs` is used when calculating error

PR #186, opened 2024-09-18
Fix default for sequential updates

dsikka:
# Summary
- Fix default for sequential updates to be make it True
- Remove incorrect comment

### DIFF
--- a/src/llmcompressor/modifiers/quantization/gptq/base.py
+++ b/src/llmcompressor/modifiers/quantization/gptq/base.py
@@ -93,7 +93,7 @@ class GPTQModifier(Modifier):
         and activation 8 bit quantization on the Linear layers.
     """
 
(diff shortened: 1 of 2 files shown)


rahul-tuli (approved):

Satrat (approved):

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -93,7 +93,7 @@ class GPTQModifier(Modifier):
         and activation 8 bit quantization on the Linear layers.
     """
 
-    sequential_update: Optional[bool] = False
+    sequential_update: Optional[bool] = True
```
rahul-tuli, line 96: nit: Should we update the doc string above, that the default for this argument is True?

kylesayrs (approved):

PR #192, opened 2024-09-20
Update MoE examples

mgoin:
Update the FP8 Mixtral example to not use GPTQ and add it to examples list


dsikka (approved):

dsikka:
slightly unrelated once your fp8 moe PR in vllm lands, we should add an e2e test case: https://github.com/vllm-project/llm-compressor/blob/main/tests/e2e/vLLM/test_vllm.py

PR #679, opened 2024-09-26
Increase Sparsity Threshold for compressors

rahul-tuli:
During internal testing, it was found that a sparsity threshold of 0.2 is insufficient. In some cases, applying quantization alone can naturally lead to a 20% sparsity level. To address this, this PR increases the sparsity threshold to 0.3.

Please note that this adjustment specifically pertains to sparse compression.

### DIFF
--- a/src/llmcompressor/transformers/compression/sparsity_config.py
+++ b/src/llmcompressor/transformers/compression/sparsity_config.py
@@ -20,7 +20,7 @@ class SparsityConfigMetadata:
     metadata from the model
     """
 
-    SPARSITY_THRESHOLD: float = 0.3
+    SPARSITY_THRESHOLD: float = 0.4
 
     @staticmethod
     def infer_global_sparsity(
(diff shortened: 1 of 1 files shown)


kylesayrs (approved):

kylesayrs (dismissed):
Generally, shouldn't we be exposing this threshold as an argument?

src/llmcompressor/transformers/compression/sparsity_config.py
```diff
@@ -20,7 +20,7 @@ class SparsityConfigMetadata:
     metadata from the model
     """
 
-    SPARSITY_THRESHOLD: float = 0.2
+    SPARSITY_THRESHOLD: float = 0.3
```
mgoin, line 23: My vote is 40% :)
```suggestion
    SPARSITY_THRESHOLD: float = 0.5
```

rahul-tuli: Your vote is accepted!

mgoin (approved):

horheynm (approved):

PR #695, opened 2024-09-27
Model Initialization Context

kylesayrs:
## Purpose ##
* Restore `torch.nn.init` functions after initialization is done
* Easy api for initializing models


kylesayrs:
Closing for after model loading refactor

PR #769, opened 2024-10-02
Fix 2/4 GPTQ Model Tests

dsikka:
SUMMARY:
-The `quantization_config` that is attached to the model is attached to a slightly different location using the loaded `quantization_config` compared to `compression_config`
- This updates the test based on this change + adds a comment explaining the difference


rahul-tuli (approved):

kylesayrs (approved):
Woop

PR #774, opened 2024-10-03
Update workflows/actions

dbarbuzzi:
SUMMARY:
These changes perform some cleanup in the existing workflows in advance of looking at adding GPU support:

* The `actions/checkout`, `actions/setup-python`, and `actions/github-script` usages have been updated to the latest version
* The use of the deprecated `set-output` has been replaced with the current standard
* Some redundancies have been cleaned up

The first two points will reduce a lot of noise in the runs as there were messages for each usage of the outdated actions due to their age/using outdated Node.js (internally to the actions).

The final point is a general cleanup to keep things simple.

TEST PLAN:
The runs for this PR should have the expected outcomes, but without the various warnings/etc. about the outdated actions.

### DIFF
--- a/.github/workflows/set-comment.yaml
+++ b/.github/workflows/set-comment.yaml
@@ -1,5 +1,5 @@
 name: PR Reminder Comment Bot
-on: 
+on:
   pull_request:
     branches:
       - main
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Remind to add ready label
-        uses: actions/github-script@v6
+        uses: actions/github-script@v7
         with:
           script: |
             github.rest.issues.createComment({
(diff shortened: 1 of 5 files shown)


.github/workflows/set-comment.yaml
```diff
@@ -20,4 +20,4 @@ jobs:
               body: '👋 Hi! Thank you for contributing to llm-compressor. Please add the ready label when the PR is ready for review.'
             })
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
kylesayrs, line 23: Is this visible to people who shouldn't have access to it?

kylesayrs (approved):

rahul-tuli (approved):

mgoin (approved):

PR #821, opened 2024-10-06
Set Sparse compression to save_compressed

rahul-tuli:
PR Description: Before this PR, sparse compression was not being utilized since it was explicitly set to False in the codebase. This PR enables sparse compression, allowing us to take advantage of its benefits for reduced memory on disk

### DIFF
--- a/src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
+++ b/src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
@@ -95,7 +95,7 @@ def modify_save_pretrained(model: PreTrainedModel):
                     "skip_compression_stats=True"
                 )
                 sparsity_config = SparsityConfigMetadata.from_pretrained(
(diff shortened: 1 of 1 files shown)


src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -95,7 +95,7 @@ def save_pretrained_wrapper(
                     "skip_compression_stats=True"
                 )
                 sparsity_config = SparsityConfigMetadata.from_pretrained(
-                    model, state_dict=state_dict, compress=False
+                    model, state_dict=state_dict, compress=save_compressed
                 )
```
mgoin, line 99: Do you think we should change `save_compressed: bool = True` to False in order to preserve previous behavior?

mgoin (changes requested):

mgoin (approved):
Per offline discussion, we will allow this automatic behavior

PR #822, opened 2024-10-07
Enable Sparse compression

rahul-tuli:
This PR makes compressors composable. After this change, both sparse and quantization compression can be applied to a sparse quantized model.

### Notable Changes:
- Reworked the logic for inferring sparse compressor format:
   - Previously, if a model was quantized, the sparse compressor was always set to the Identity (dense) compressor. Now, an appropriate sparse compressor is selected (currently only `sparse_bitmask` or `sparse_24`).
   - A special condition for Marlin-style compression was added (Sparse compressor should be set to Dense if `marlin_24` compressor is used).
   - Updated the SPARSITY_THRESHOLD to 50% as discussed offline

### Dependencies:
- [ ] https://github.com/neuralmagic/compressed-tensors/pull/197
- [ ] [neuralmagic/compressed-tensors#182](https://github.com/neuralmagic/compressed-tensors/pull/182)
- [x] [vllm-project/llm-compressor#821](https://github.com/vllm-project/llm-compressor/pull/821)

<details>
<summary>Test Script</summary>

```python

MODEL_ID = "nm-testing/llama2.c-stories110M-pruned50-compressed-tensors"

# Load model
)

recipe = """
quant_stage:
    quant_modifiers:
        QuantizationModifier:
            ignore: ["lm_head"]
            config_groups:
                group_0:
                    weights:
                        num_bits: 4
                        type: int
                        strategy: tensor
                        dynamic: false
                        symmetric: true
                    input_activations:
                        num_bits: 4
                        type: int
                        strategy: tensor
                        dynamic: true
                        symmetric: true
                    targets: ["Linear"]
    pruning_modifiers:
        ConstantPruningModifier:
            targets: [
                're:.*q_proj.weight',
                're:.*k_proj.weight', 
                're:.*v_proj.weight',
                're:.*o_proj.weight',
                're:.*gate_proj.weight',
                're:.*up_proj.weight',
                're:.*down_proj.weight',
            ]
            start: 0
"""

# Apply quantization.
oneshot(model=model, recipe=recipe)

output = model.generate(input_ids, max_new_tokens=20)
print("==========================================")

# Save to disk in compressed-tensors format.
SAVE_DIR = MODEL_ID.split("/")[1] + "-FP8"

```

</details>


src/llmcompressor/transformers/compression/quantization_format.py
```diff
@@ -36,9 +35,7 @@ def infer_quantization_format(
 
     if save_compressed:
         weight_args, input_args = _get_unique_quant_args(model)
-        is_24_structure = (
-            sparsity_config and sparsity_config.sparsity_structure == "2:4"
-        )
+        is_24_structure = sparsity_structure is not None and sparsity_structure == "2:4"
```
dsikka, line 38: nit: make 2:4 enum

src/llmcompressor/transformers/compression/sparsity_config.py
```diff
@@ -95,10 +96,11 @@ def from_pretrained(
         sparsity_structure = SparsityConfigMetadata.infer_sparsity_structure(
             model=model
         )
-        if is_model_quantized(model):
-            # compressing a sparse quantized model is not supported yet
+        if is_marlin:
+            # sparse compressor should be dense for marlin
+            # compression
```
dsikka, line 101: why

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -77,15 +81,22 @@ def save_pretrained_wrapper(
             if state_dict is None:
                 state_dict = get_state_dict_offloaded_model(model)
 
+            sparsity_stucture = SparsityConfigMetadata.infer_sparsity_structure(model)
```
kylesayrs, line 84: nit: move this into `infer_quantization_format` and copy onto line 99. This way the `infer_quantization_format` has less dependencies

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -77,15 +81,22 @@ def save_pretrained_wrapper(
...
+            quantization_format = infer_quantization_format(
+                model=model,
+                quantization_format=quantization_format,
+                save_compressed=save_compressed,
+                sparsity_structure=sparsity_stucture,
+            )
+            is_marlin = quantization_format == CompressionFormat.marlin_24.value
+
             if sparsity_config is not None:
                 sparsity_config.global_sparsity = (
```
kylesayrs, line 94: tbh we should just get rid of this branch and instead add `SparsityConfigMetadata.from_format`, but that's out of scope

kylesayrs: Since really the only point in being able to pass sparsity config is to be able to override the format afaict

kylesayrs (approved):
These changes highlight how tied sparsity and quantization are (quantization format now depends on sparsity structure, and sparsity config now depends on quantization format).

I think this is the proper way to do it (first infer format, then use format to generate configs), but I hope in the future that we can structure the classes in a way such that the flow is clearer and users have an easier time understanding and using the tools

src/llmcompressor/transformers/compression/quantization_format.py
rahul-tuli: Done!

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
rahul-tuli: Keeping it as is for now!

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
rahul-tuli: I agree; we'll make those updates soon!

dsikka (commented):
We should add a couple of e2e tests to run sample generation in llm-compressor.
We can then eventually expand on them to also run in vllm (once we've integrated).

src/llmcompressor/transformers/compression/sparsity_config.py
```diff
@@ -95,11 +151,15 @@ def from_pretrained(
...
             format = CompressionFormat.dense.value
-        elif compress:
-            format = CompressionFormat.sparse_bitmask.value
+        if compress:
+            if sparsity_structure == SparsityStructure.TWO_FOUR.value:
+                format = CompressionFormat.sparse_24.value
+            else:
+                format = CompressionFormat.sparse_bitmask.value
```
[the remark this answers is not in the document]
rahul-tuli: It will be dense if conditions are not met;

src/llmcompressor/transformers/compression/sparsity_config.py
```diff
@@ -67,13 +67,14 @@ def infer_sparsity_structure(model: Optional[Module] = None) -> str:
...
 
-        return sparsity_structure or "unstructured"
+        return SparsityStructure(sparsity_structure).value
 
     @staticmethod
     def from_pretrained(
         model: Module,
         state_dict: Optional[Dict[str, Tensor]] = None,
         compress: bool = False,
+        is_marlin: bool = False,
```
kylesayrs, line 77: To make this more generic, why not pass in the quantization config? We will for sure have different compression formats which affect sparsity in the future

kylesayrs (approved):
Overall structure is good (having quantization config be dependent on sparsity structure is better than it being dependent on a sparsity config)

I definitely recommend genericising the is_marlin a little bit by passing the compression format directly so it's easier to support more formats in the future

horheynm (approved):

dsikka (changes requested):
Just summarizing offline discussions:

Although marlin_24 should no longer be the default, we still need to make sure we have a pathway to enable it, such as through the addition of a "format" argument in the recipe.

src/llmcompressor/transformers/compression/sparsity_config.py
dsikka: Yeah we should make this generic as marlin will likely not be the only case

PR #829, opened 2024-10-08
bad filename with colon, stops windows users from cloning

shaneholloman:
bad filename with colon, stops windows users from cloning


kylesayrs:
Hi @shaneholloman, thank you for your contribution! Would we not also want to change the `2:4_w4a16_group-128_recipe.yaml` file?

From my grepping, I only see these two files as being files with colons in the name
```
```

Also note that we'll want to change all references to these file names in the following files

`grep -rl '2:4_w4a16' src examples tests docs utils`
```
examples/quantization_24_sparse_w4a16/README.md
examples/quantization_24_sparse_w4a16/llama7b_sparse_w4a16.py
tests/examples/test_quantization_24_sparse_w4a16.py
```

shaneholloman:
yes, any file whith colon will stop windows users from propoer git functionality

kylesayrs:
#846

PR #830, opened 2024-10-08
bad filename with colon, stops windows users from cloning

shaneholloman:
bad filename with colon, stops windows users from cloning

SUMMARY:
bad filename with colon, stops windows users from cloning

TEST PLAN:
bad filename with colon, stops windows users from cloning


kylesayrs:
Hi @shaneholloman, thank you for your contribution! Would we not also want to change the `2:4_w4a16_group-128_recipe.yaml` file?

From my grepping, I only see these two files as being files with colons in the name
```
```

kylesayrs:
Duplicates #829

PR #832, opened 2024-10-09
Remove SparseAutoModelForCausalLM

horheynm:
SUMMARY:
Goal is to get rid of `SparseAutoModelForCausalLM` - users can still use but a deprecation warning pops up
Class removal causes logic change in the saving. 
Two main pathway of saving - fsdp and save_pretrained - both are wrapped on save_pretrained.

Other changes:
* oneshot no longer saves the models, hence example code has output_dir argument removed.
* Deprecation of SparseAutoModel, bc it uses SparseAutoModelForCausalLM to load models

TEST PLAN:
* Example code dry run for small, short-run time models and making sure the saved files are same as main
* Example runs for the below cases.

baseline case:
```python3

# Load model.
)

    targets="Linear", scheme="FP8_DYNAMIC", ignore=["lm_head"]
)

# Apply quantization.
# NOTE: This is where we wrap save_pretrained
oneshot(model=model, recipe=recipe)

# Save to disk in compressed-tensors format.
SAVE_DIR = MODEL_ID.split("/")[1] + "-FP8-Dynamic"

```

vision:
```python3
rom transformers import (
    AutoModelForCausalLM,
    AutoProcessor,
    MllamaForConditionalGeneration,
    AutoTokenizer,
)

# Load model.
MODEL_ID = "meta-llama/Llama-3.2-11B-Vision-Instruct"
model = MllamaForConditionalGeneration.from_pretrained(
)
# model = AutoModelForCausalLM.from_pretrained(
#     MODEL_ID, device_map="auto", torch_dtype="auto"
# )

# Configure the quantization algorithm and scheme.
# In this case, we:
#   * quantize the weights to fp8 with per channel via ptq
#   * quantize the activations to fp8 with dynamic per token
    targets="Linear",
    # scheme="FP8_DYNAMIC",
    scheme="FP8",
    ignore=["re:.*lm_head", "re:multi_modal_projector.*", "re:vision_model.*"],
)

# Apply quantization and save to disk in compressed-tensors format.
SAVE_DIR = MODEL_ID.split("/")[1] + "-FP8-model_save-MllamaForConditionalGeneration"
# SAVE_DIR = MODEL_ID.split("/")[1] + "-FP8-Dynamic-foo"

oneshot(
    model=model,
    recipe=recipe,
    output_dir=SAVE_DIR,
    dataset="open_platypus",
    num_calibration_samples=2,
)
processor.save_pretrained(SAVE_DIR)
```

bert:
```python

# Load model.
MODEL_ID = "google-bert/bert-base-uncased"

    MODEL_ID, 
    # device_map="auto", 
    torch_dtype="auto"
)

    targets="Linear", scheme="FP8_DYNAMIC", ignore=["lm_head"]
)

# Apply quantization.
# NOTE: This is where we wrap save_pretrained
oneshot(model=model, recipe=recipe)

# Save to disk in compressed-tensors format.
SAVE_DIR = MODEL_ID.split("/")[1] + "-FP8-Dynamic"
```


src/llmcompressor/pytorch/model_load/helpers.py
```diff
@@ -106,6 +107,12 @@ def save_model_and_recipe(
     :param save_safetensors: whether to save as safetensors or pickle (bin)
     :param save_compressed: whether to compress sparse weights on disk
     """
+    if is_model_quantized(model):
+        from llmcompressor.transformers.sparsification.compressed_tensors_utils import (
+            modify_save_pretrained,
+        )
+
+        modify_save_pretrained(model)
```
dsikka, line 115: We should probably also take this opportunity to clean-up some of what happens in the `save_pretrained_wrapper` (e.g. seem to always calculate sparsity statistics by default)

Also, oneshot gives the option to provide an output dir, which I think also saves the model:
https://github.com/vllm-project/llm-compressor/blob/b3c474f89621bed5fc2d779226e50cb1b34dd50b/tests/llmcompressor/transformers/obcq/test_consecutive_runs.py#L30 but not 100% sure.

kylesayrs, line 115: Why not just call the modified `save_pretrained` function directly? Doing it this way has the unintended and undocumented side effect that now the model has a modified save_pretrained function

horheynm: youll call vanilla save_pretrained if the model has not gone thru compression.
If compressed, then one of the thing we need to do is to add the quant-config, which vanilla code doesnt do

kylesayrs: > Why not just call the *modified* `save_pretrained` function directly

kylesayrs: Wrapping is not required and leads to an extra side effect.

```python3
def save_compressed(model, *args, **kwargs):
    save_fn = save_pretrained_compressed(model.save_pretrained)
    save_fn(*args, **kwargs)
```

```python3
- modify_save_pretrained(model)
- model.save_pretrained(...)
# now model has a wrapped save function as a side effect

+ save_compressed(model)
```

src/llmcompressor/transformers/sparsification/sparse_model.py
```diff
@@ -62,6 +62,13 @@ def from_pretrained(
             pretrained_model_name_or_path directory and applied if found
         :return the created model for causal language modeling
         """
+        logger.info(
```
dsikka, line 65: In this file, seems like you're still using SparseAutoModel in `text_generation_from_pretrained` pathway?

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -64,75 +73,108 @@ def save_pretrained_wrapper(
...
-            # support float8 types. Fix is posted upstream
-            # https://github.com/huggingface/transformers/pull/30488
-            transformers.modeling_utils.dtype_byte_size = new_dtype_byte_size
+                quantization_format = infer_quantization_format(
+                    model=model,
+                    quantization_format=quantization_format,
+                    save_compressed=save_compressed,
+                    sparsity_config=sparsity_config,
+                )
+                compressor = ModelCompressor.from_pretrained_model(
```
dsikka, line 137: We only want this method to be responsible for saving the files to disk post compression. The ModelCompressor lifecyle should be outside of save_pretrained.

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -170,34 +168,8 @@ def one_shot(self, stage: Optional[str] = None):
...
-        else:
-            save_model_and_recipe(
-                model=self.trainer.model,
-                save_path=self._output_dir,
-                tokenizer=self.tokenizer,
-                save_safetensors=self._training_args.save_safetensors,
-                save_compressed=self._training_args.save_compressed,
-            )
+        # wrap model.save_pretrained(...)
+        modify_save_pretrained(self.trainer, self.tokenizer)
```
dsikka, line 172: why is it ok for us to remove the fsdp model checks?

src/llmcompressor/transformers/sparsification/sparse_model.py
dsikka, line 65: I would throw an error if this is tried, not just an info.

dsikka (changes requested):
Can you provide examples on different loading scenarios (previously quantized models/legacy compression_config models, one shot pathways for causal models,) to understand how the  interface should work?

In general I think we still have to make updates in the following areas:
1. Save pretrained should only be responsible for saving things to disk, the ModelCompressor pathway should be outside of this
2. Seems like we're still using the SparseAutoModel pathway in a couple of places?
3. Is there anymore clean-up we can do in loading/saving?
4. Confirm the loading we're doing works sufficiently for the work @kylesayrs is doing with adding modifier support for multi-modal cases

src/llmcompressor/transformers/sparsification/sparse_model.py
```diff
@@ -62,6 +62,13 @@ def from_pretrained(
...
+            (
+                "`wrap_hf_model_class` is now deprecated. ",
```
dsikka, line 67: Can you also add an updated lifecycle? For save/load. See compressed-tensors lifecycle examples:
https://github.com/neuralmagic/compressed-tensors/blob/506cd36829f6fd386a0a2c75ed8a51cc6dcab619/src/compressed_tensors/compressors/base.py#L30

dsikka (commented):
We should update e2e/example testing and other ci testing as well

horheynm:
With respect to the functionality, the PR is ready.

Next to dos:
* Detach the compressor save_pretrained.

Strategy:
in main (shared flow for oneshot, etc), at the end, call the compressor, attach the compressed state_dict as the attribute as the model.

after oneshot is called, call save_pretrained(SAVE_DIR), and extract the statedict from the model. 

Testing:
* e2e
* vllm
* offloading
* fsdp

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
horheynm: Next to dos, updated below

examples/big_models_with_accelerate/cpu_offloading_fp8.py
```diff
@@ -24,3 +24,4 @@
     tokenizer=AutoTokenizer.from_pretrained(MODEL_ID),
     output_dir=OUTPUT_DIR,
 )
+model.save_pretrained(OUTPUT_DIR)
```
kylesayrs, line 27: Why save twice?

horheynm: Where do you see it being saved twice?

src/llmcompressor/pytorch/model_load/helpers.py
```diff
@@ -212,16 +212,31 @@ def load_safetensors_state_dict(file_path: str) -> Dict[str, torch.Tensor]:
...
         import shutil
 
-        cache_dir = config._name_or_path
-        for file in os.listdir(cache_dir):
-            full_file_name = os.path.join(cache_dir, file)
+        from huggingface_hub import hf_hub_download
+        from transformers import TRANSFORMERS_CACHE
+        from transformers.utils import http_user_agent
+
+        cache_path = config._name_or_path
```
rahul-tuli, line 226: will this work for cases when we have a locally downloaded model?

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -64,75 +73,108 @@ def save_pretrained_wrapper(
             saving a model in dense format
             :param kwargs: additional kwargs to pass on to model.save_pretrained
             """
+            if is_fsdp_model(trainer.model):
```
rahul-tuli, line 76: This function does too many things, would it be possible to break it out into smaller methods?

src/llmcompressor/pytorch/model_load/helpers.py
horheynm: yes

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
dsikka: +1 if we could break it out to help with readability

dsikka (commented):
@horheynm There are new consistent test failures across the different test cases

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
horheynm: okk

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -170,35 +165,6 @@ def one_shot(self, stage: Optional[str] = None):
...
-
-        else:
-            save_model_and_recipe(
-                model=self.trainer.model,
-                save_path=self._output_dir,
-                tokenizer=self.tokenizer,
-                save_safetensors=self._training_args.save_safetensors,
-                save_compressed=self._training_args.save_compressed,
-            )
-
```
dsikka, line 201: Seems like we're still producing empty `output` directories

horheynm: what code are you running?

dsikka: examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py

src/llmcompressor/transformers/sparsification/sparse_model.py
```diff
@@ -1,174 +1,29 @@
...
-        pretrained_model_name_or_path = download_model_directory(
-            pretrained_model_name_or_path, **kwargs
-        )
-
-        # instantiate compressor from model config
-        compressor = ModelCompressor.from_pretrained(
-            pretrained_model_name_or_path, **kwargs
+class SparseAutoModelForCausalLM:
+    def from_pretrained(*args, **kwargs):
+        logger.warning(
```
kylesayrs, line 22: Since this class is lot actually usable, this should be an error rather than a warning

kylesayrs (changes requested):

examples/big_models_with_accelerate/cpu_offloading_fp8.py
kylesayrs: Did more reading, the model is not saved twice but you pass the output_dir to oneshot and to save_pretrained which seems misleading

kylesayrs: This is true of most of the examples. We shouldn't pass the outputdir to oneshot if it's not going to be used (we should even trigger a warning/error for this)

tests/llmcompressor/transformers/finetune/test_oneshot_and_finetune.py
```diff
@@ -22,6 +22,7 @@ def _test_oneshot_and_finetune(self):
         splits = {"train": "train[:50%]", "calibration": "train[50%:60%]"}
         if self.dataset == "ultrachat-200k":
             splits = {"train": "train_gen[:50%]", "calibration": "train_gen[50%:60%]"}
+        breakpoint()
```
kylesayrs, line 25: Remove breakpoint

src/llmcompressor/transformers/sparsification/sparse_model.py
```diff
@@ -1,174 +1,29 @@
...
-    )
-
-    return wrapped_model_class
-
-
-SparseAutoModelForCausalLM = wrap_hf_model_class(AutoModelForCausalLM)
+        return AutoModelForCausalLM.from_pretrained(*args, **kwargs)
 
 
 class SparseAutoModel:
```
kylesayrs, line 29: Remove here and in main

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -170,35 +165,6 @@ def one_shot(self, stage: Optional[str] = None):
 
         self.trainer.one_shot(calibration_data=calib_data, stage=stage)
```
kylesayrs, line 167: I recommend not removing the saving capability of one_shot

kylesayrs (changes requested):
1. depreciate sparseautomodel
2. add back saving to oneshot pathway, or at least remove argument and remove from examples
3. remove breakpoint

examples/big_models_with_accelerate/cpu_offloading_fp8.py
dsikka: yeah the saving twice is confusing and part of this PR is to remove it in from oneshot. I think we should also remove the argument from oneshot.

dsikka (changes requested):
Can you please update the PR description?

Adding in a complete summary of the functionality that was changed, why, and testing that was done to verify behaviour after the changes were applied. The current description is unclear and makes the code hard to review.

It also seems like the test cases are still failing with a new test failure.

rahul-tuli (approved):
The code looks good! Did we run the examples again after latest changes?

src/llmcompressor/transformers/sparsification/sparse_model.py
```diff
@@ -1,174 +1,29 @@
...
-        )
-
-        pretrained_model_name_or_path = download_model_directory(
-            pretrained_model_name_or_path, **kwargs
-        )
-
-        # instantiate compressor from model config
-        compressor = ModelCompressor.from_pretrained(
-            pretrained_model_name_or_path, **kwargs
+class SparseAutoModelForCausalLM:
```
kylesayrs, line 20: As discussed previously, please make this either a subclass with a depreciation warning, or a class which raises an error

src/llmcompressor/transformers/sparsification/sparse_model.py
kylesayrs: Warnings should be used in cases where it's okay to continue. Unless you subclass AutoModel, this is not a case where the user can continue.

kylesayrs (changes requested):
1. subclass automodel in sparseautomodel or raise depreciation error
2. add back saving to oneshot pathway, or at least remove argument and remove from examples

src/llmcompressor/transformers/sparsification/sparse_model.py
rahul-tuli: I agree, let's raise an error instead

horheynm: you can use this code, we can completely delete the pathway later

horheynm:
> 1. subclass automodel in sparseautomodel or raise depreciation error
> 2. add back saving to oneshot pathway, or at least remove argument from model_args and remove from examples

1. we dont need this
2. There are entry points in the pipeline that uses output_dir to training_args, so removing that would need more validation. That would be next to dos.

kylesayrs:
Should we not also depreciate `SparseAutoModel`?

horheynm:
I have broken this down into example changes and component changes, will close the PR

PR #839, opened 2024-10-11
Only untie word embeddings

kylesayrs:
## Purpose ##
* Some models have tied weights which are not the language model word embeddings. For example, `microsoft/Phi-3.5-vision-instruct` has `{'model.embed_tokens.weight', 'model.vision_embed_tokens.wte.weight'}` tied. For these models, loading with the `tie_word_embeddings=False` option should not untie these embeddings. 

## Changes ##
* Modify `patch_tied_tensors_bug` to only untie word embeddings

## Testing ##
* `test_model_reload` and `test_model_shared_tensors` tests pass


mgoin (approved):
tenor

PR #841, opened 2024-10-11
Fix GPTQ Docstring, better argument grouping

kylesayrs:


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -68,8 +68,8 @@ class GPTQModifier(Modifier):
 
     :param sequential_update: Whether or not to update weights sequentially by layer,
         True saves on GPU memory, default is True
-    :param targets: list of layer names to compress during GPTQ, or '__ALL__'
```
rahul-tuli, line 71: Why remove targets from the docstring? We are still accepting it to build the QuantizationModifier

rahul-tuli (changes requested):

dsikka (commented):
Is the targets param not used anywhere? seems like the last PR kept it, on line 101

markurtz:
Why are renaming / standardizing on sequential targets? The original goal, across all modifiers, was to standardize on targets as the base name to represent the scope of which the modifier would target -- I'd prefer not to deviate from that whenever we can.

kylesayrs:
@markurtz The `GPTQModifier` has two targets arguments, `targets` which selects which modules to compress, and `sequential_targets`, which selects which modules are considered layers (transformer blocks) for the purposes of `true_sequential=False`. Both are required, since the inferred defaults are not always correct and should be capable of being overridden.

PR #845, opened 2024-10-13
Fix forward function pass call

dsikka:
Summary
- Fix forward function call after this change: https://github.com/neuralmagic/compressed-tensors/pull/188


kylesayrs:
Outside of the scope of this PR, but we should consider implementing this with [pytorch hooks](https://pytorch.org/docs/stable/generated/torch.nn.modules.module.register_module_forward_hook.html)

dsikka:
> Outside of the scope of this PR, but we should consider implementing this with [pytorch hooks](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_hook) rather than a wrapped forward

Yup. That's what the follow-up PR that Im working on does to some degree. Uses hooks for calibration.  This is meant as a stepping stone to make that PR easier.

PR #846, opened 2024-10-13
[Bugfix] Rename files to remove colons

kylesayrs:
## Purpose ##
* Remove colons from filenames in order to better support cloning on Windows OS
** Reported by #829

## Changes ##
* Renamed files with colons
```
```

* Modified files which reference file names with colons
```
examples/quantization_24_sparse_w4a16/README.md
examples/quantization_24_sparse_w4a16/llama7b_sparse_w4a16.py
tests/examples/test_quantization_24_sparse_w4a16.py
```

## Testing ##
* Grepped codebase for mentions of colons
```bash
#!/bin/bash

check_files() {
  local dir="$1"
  find "$dir" -type f -name "*:*" -print
}

dir="${1:-.}"
check_files "$dir"
```


markurtz (changes requested):
@kylesayrs can we replace these with 2.4 instead of 24? Agree with the goal of enabling better windows support, but rewriting all of them to 24 makes it harder to read that that should be "two four" and not "twenty four"

kylesayrs:
@markurtz How do we feel about 2-4_w4a16_group-128_recipe.yaml?

markurtz:
> @markurtz How do we feel about 2-4_w4a16_group-128_recipe.yaml?

Given we use the dash and underscore in other places to differentiate between sections within, I'm not sure we should overload that.

Can you check with @anmarques for our latest state on naming conventions for models? It'd be great to have this unified with that and we had to go through this there as well I believe due to HF not supporting colons in names

anmarques:
> > @markurtz How do we feel about 2-4_w4a16_group-128_recipe.yaml?
> 
> Given we use the dash and underscore in other places to differentiate between sections within, I'm not sure we should overload that.
> 
> Can you check with @anmarques for our latest state on naming conventions for models? It'd be great to have this unified with that and we had to go through this there as well I believe due to HF not supporting colons in names

The 

> > @markurtz How do we feel about 2-4_w4a16_group-128_recipe.yaml?
> 
> Given we use the dash and underscore in other places to differentiate between sections within, I'm not sure we should overload that.
> 
> Can you check with @anmarques for our latest state on naming conventions for models? It'd be great to have this unified with that and we had to go through this there as well I believe due to HF not supporting colons in names

The current standard for model names is using dashes to split compression techniques (pruned vs quantized) with dots to designate specifiers for each technique. For instance, `-pruned.2of4-quantized.w4a16`. We do not add specifiers if there considered standard, such as group size 128 for w4a16.

mgoin (approved):
This is going to be annoying for a bit since "24" has been what was used since sparse_24_marlin was created, including all in vLLM ([example](https://github.com/vllm-project/vllm/blob/e7116c017c86cb547f4d1888edaf13a9be2a4562/csrc/quantization/marlin/sparse/marlin_24_cuda_kernel.cu#L47), [search](https://github.com/search?q=repo%3Avllm-project%2Fvllm%20%2224_%22&type=code))

It's fair to choose 2of4 since that is what we name models on HF. I can look at renaming in vLLM to make that change so we have more consensus

markurtz (approved):

PR #847, opened 2024-10-14
Use float32 for Hessian dtype

kylesayrs:
## Purpose ##
* Fix failing tests `RuntimeError: "cholesky_cusolver" not implemented for 'BFloat16'`

## Changes ##
* Explicitly set hessian dtype to float32 (an assumption made by other functions such as cpu_offloading)
** This change guards against changes to the pytorch default dtype

## Testing ##
* TODO

### DIFF
--- a/src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
+++ b/src/llmcompressor/modifiers/obcq/utils/sgpt_wrapper.py
@@ -40,7 +40,10 @@ class SparseGptWrapper(ModuleCompressionWrapper):
 
         # for Hessian calculation
         self.register_buffer(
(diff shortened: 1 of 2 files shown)


mgoin (approved):
LGTM but could this cause more OOM issues?

rahul-tuli (dismissed):

kylesayrs:
@mgoin float32 precision is definitely required for the hessian. Float32 is the precision assumed by [custom_offload_device_map](https://github.com/vllm-project/llm-compressor/blob/b3c6d90b8240d31ad7170d423e6a88883f309734/src/llmcompressor/transformers/compression/helpers.py#L139). We point memory-sensitive users towards using this function, so they shouldn't encounter any issues as long as they use it.

mgoin (approved):

PR #849, opened 2024-10-15
[ DOC ] Remove version restrictions in W8A8 exmaple

miaojinc:
SUMMARY:
The latest compressored-tensor 0.8.0 removed some API, https://github.com/neuralmagic/compressed-tensors/pull/156/files If installed the older llmcompressor from pip, it would throw the error like:
```
ImportError: cannot import name 'update_layer_weight_quant_params' from 'compressed_tensors.quantization'
```

TEST PLAN:
already tested locally


kylesayrs (approved):
Thank you for your contribution!

PR #855, opened 2024-10-19
[WIP] Modify Params Context

kylesayrs:
## Purpose ##
* Provide a context within which parameters belonging to FSDP or accelerate-offloaded models can be safely modfied
* Note that the parameters must be modified in-place through in-place operations like `copy_` rather than assignment or assigning to the tensor data directly


kylesayrs:
Closing, can be reopened when better FSDP support is the priority

PR #863, opened 2024-10-23
[GPTQ] Iterative Parameter Updating

kylesayrs:
## Purpose ##
* Add iterative parameter updating in line with the described implementation of GPTQ
* Prepare for observers refactor by reducing GPTQ's dependence on observers

## Changes ##
* rename `weight_quant_args` to `quant_args` for brevity
* replace `_update_quantization_parameters` with `compute_scale_zero_point` and explicitly pass and update quant params to reduce observer dependence
* update quantization parameters to reflect changes resulting from previous blocks

## Testing ##
* Replicated winogrande evaluations to within margin of error

| Actorder | Replication Branch (with dtype changes) | Feature branch | AutoGPTQ |
| ------- | ------------------ | --------------- | ----- |
| group | 0.7924 | 0.7743 | 0.7766 |
| weight | 0.7735 | 0.7766 | 0.7758 |
| none | 0.7727 | 0.7656 | 0.7719 |


rahul-tuli (dismissed):

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -99,20 +96,27 @@ def compress(
...
-        actorder = weight_quant_args.actorder
+        strategy = quant_args.strategy
+        actorder = quant_args.actorder
         final_shape = self.layer.weight.shape
         final_dtype = self.layer.weight.dtype
         W = self.layer.weight.data.clone()
 
+        # create observer for calculating quantization parameters
+        observer = Observer.load_from_registry(
+            "minmax",
```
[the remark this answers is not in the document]
kylesayrs: Yep, the algorithm doesn't make sense with any other kind of observer or averaging constant, so this guards against user misuse

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -127,26 +130,28 @@ def compress(
...
                 torch.arange(self.columns, device=W.device, dtype=torch.int)
-                // weight_quant_args.group_size
+                // quant_args.group_size
             )
 
             if actorder == ActivationOrdering.GROUP:
                 # permute by activation order first, then update groups
                 W, self.H, perm = self._apply_activation_ordering(W, self.H)
-                self._update_quantization_parameters(weight_quant_args, W)
+                scale, zero_point = observer(W, g_idx=None)
```
horheynm, line 139: Do we always need to pass in g_idx?

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -212,16 +217,28 @@ def compress(
...
                 elif strategy == QuantizationStrategy.GROUP:
                     # get the group index for the current column
                     column_idx = i1 + i
                     group_index = g_idx[column_idx]
 
+                    # update quantization parameters to reflect changes
+                    # resulting from previous blocks
+                    if (
+                        actorder != ActivationOrdering.WEIGHT
+                        and column_idx % quant_args.group_size == 0
```
horheynm, line 231: 👍

horheynm (approved):

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
kylesayrs: The default value is None, but I keep it explicit here to emphasize that the scales and zero points are being computed as if there was no g_idx (even though it has been permuted by the `_apply_activation_ordering` function)

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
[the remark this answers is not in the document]
dsikka: I think just log info that if the user has specified a different observer, they're aware we're using something different 
I thought I remembered a conversation around MSE being the preferred observer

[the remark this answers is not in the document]
kylesayrs: Yep makes sense!

src/llmcompressor/modifiers/quantization/gptq/utils/gptq_wrapper.py
```diff
@@ -100,20 +96,27 @@ def compress(
...
-        actorder = weight_quant_args.actorder
+        strategy = quant_args.strategy
+        actorder = quant_args.actorder
         final_shape = self.layer.weight.shape
         final_dtype = self.layer.weight.dtype
         W = self.layer.weight.data.clone()
 
+        # create observer for calculating quantization parameters
+        observer = Observer.load_from_registry(
+            quant_args.observer,
```
dsikka, line 115: rebase bug? Seems like we're no longer hardocding as we were before and reading from args?

rahul-tuli: +1

rahul-tuli (commented):
Looks good once @dsikka 's comment is resolved!

dsikka (approved):
This LGTM but would be nice to add a test to reflect the added functionality

rahul-tuli (approved):

PR #869, opened 2024-10-27
Fix HFTrainer overloads

kylesayrs:
## Purpose ##
* Fix failing tests in `tests/llmcompressor/transformers/finetune/test_finetune_no_recipe_custom_dataset.py`

## Background ##
* The HFTrainer function contracts were changed in the most recent transformers release 4.46.0 ([offending commit](https://github.com/huggingface/transformers/commit/6ba31a8a94bf7cfeaf59ffc3bc9e0b0cd3e25788#diff-ed55888e6665791fe92cc8fc0c499da54f4ace6738551cd9a2591881cda076deR3603))

## Changes ##
* Change mixin overloads to use the same function contract as the original functions
* This change is backwards compatible with earlier versions of transformers since the new arguments are kwargs

## Testing ##
* `tests/llmcompressor/transformers/finetune/test_finetune_no_recipe_custom_dataset.py` now passes

### DIFF
--- a/src/llmcompressor/transformers/finetune/session_mixin.py
+++ b/src/llmcompressor/transformers/finetune/session_mixin.py
@@ -297,7 +297,7 @@ class SessionManagerMixIn:
             model,
             inputs,
             return_outputs=return_outputs,
-            num_items_in_batch=num_items_in_batch
+            num_items_in_batch=num_items_in_batch,
         )
 
         # take the mean across multiple GPUs
--- a/src/llmcompressor/transformers/finetune/session_mixin.py
+++ b/src/llmcompressor/transformers/finetune/session_mixin.py
@@ -293,7 +293,12 @@ class SessionManagerMixIn:
 
         # TODO: do we need these model signature columns?
         inputs = {k: inputs[k] for k in inputs if k in self._signature_columns}
-        loss = super().compute_loss(model, inputs, return_outputs, num_items_in_batch)
+        loss = super().compute_loss(
+            model,
+            inputs,
+            return_outputs=return_outputs,
+            num_items_in_batch=num_items_in_batch
+        )
 
         # take the mean across multiple GPUs
         # this is done outside the compute_loss function in the parent, replicating it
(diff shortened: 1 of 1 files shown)


src/llmcompressor/transformers/finetune/session_mixin.py
```diff
@@ -286,7 +293,7 @@ def compute_loss(
 
         # TODO: do we need these model signature columns?
         inputs = {k: inputs[k] for k in inputs if k in self._signature_columns}
-        loss = super().compute_loss(model, inputs, return_outputs=return_outputs)
+        loss = super().compute_loss(model, inputs, return_outputs, num_items_in_batch)
```
mgoin, line 296: It would be nice to keep named args here

mgoin (dismissed):

src/llmcompressor/transformers/finetune/session_mixin.py
kylesayrs: I’d prefer to keep the call more concise in the case that the reader knows exactly what the function call is, unless there are any other good reasons

rahul-tuli (approved):

src/llmcompressor/transformers/finetune/session_mixin.py
dsikka: I think we should use arg names to help with readability and in the event that one of them is removed, the error will indicate that it is no longer relevant

dsikka (dismissed):

rahul-tuli (dismissed):

rahul-tuli:
Should we add *args, **kwargs and pass them to the super functions, to protect against future changes like this?

src/llmcompressor/transformers/finetune/session_mixin.py
```diff
@@ -265,12 +268,16 @@ def training_step(
         self._check_super_defined("training_step")
 
         callbacks.batch_start(batch_data=inputs)
-        model_outputs = super().training_step(model, inputs)
+        model_outputs = super().training_step(model, inputs, num_items_in_batch)
```
mgoin, line 271: nit: arg names here too :)

mgoin (dismissed):

kylesayrs:
@rahul-tuli Since both functions require some information from the inputs, I think I'd prefer to expose all the arguments. There are tradeoffs either way w.r.t. erroring when arguments change

dsikka (approved):

Issue #109, opened 2024-08-23
Struggling to quantize Llama-3.1-70b -  OOM/linalg.Cholesky issue 


lulmer:
**Describe the bug**
Hello the vLLM team, thank you for your outstanding work. I think llm-compressor is really filling a need : a one simple unified quant franework for vLLM. 

So the bug I am encountering is the following. 
When trying to quantize llama3.1-70b on the int4 format (W4A16) I always get CUDA OOM right after the calibration epoch is finished. I also tried setting `sequential_update=True` in `oneshot` function and I got a more informative log (see bellow).

**Expected behavior**
The model to be successfully quantised in int4 (W4A16), but I also suspect the int8 (W8A8) script to suffer from the same problem. 

**Environment**
Include all relevant environment information:
1. OS [e.g. Ubuntu 20.04]: Debian 12 (Bookworm)
2. Python version [e.g. 3.7]: 3.11.9
3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]: 0.1.0
4. ML framework version(s) [e.g. torch 2.3.1]: 2.3.1
5. Other Python package versions [e.g. vLLM, compressed-tensors, numpy, ONNX]: vLLM 0.5.3.post1, vllm-flash-attn 2.5.9.post1, numpy 1.26.4
6. Other relevant environment information [e.g. hardware, CUDA version]: 6 x H100 , CUDA 12.2, Driver 535

**To Reproduce**
Exact steps to reproduce the behavior:
The script I am using is the following : 
```python
from llmcompressor.transformers import SparseAutoModelForCausalLM
from transformers import AutoTokenizer
from datasets import load_dataset
from llmcompressor.transformers import oneshot
from llmcompressor.modifiers.quantization import GPTQModifier
from llmcompressor.modifiers.smoothquant import SmoothQuantModifier

# Model is downloaded using huggingface_cli download 
pretrained_model_dir = "/data/checkpoints/Meta-Llama-3.1-70B-Instruct/"
quantized_model_dir = f"/data/checkpoints/Meta-Llama-3.1-70B-Instruct-INT4-W4A16-G128"
NUM_CALIBRATION_SAMPLES = 512
MAX_SEQUENCE_LENGTH = 2048

# Best Practices
# Start with 512 samples for calibration data (increase if accuracy drops)
# Use a sequence length of 2048 as a starting point
# Employ the chat template or instruction template that the model was trained with
[...]

Issue #823, opened 2024-10-07
I encountered the same issue: torch._C._LinAlgError: linalg.cholesky


okwinds:
I encountered the same issue: 

https://github.com/vllm-project/llm-compressor/issues/109 https://github.com/vllm-project/llm-compressor/issues/142

`torch._C._LinAlgError: linalg.cholesky: The factorization could not be completed because the input is not positive-definite (the leading minor of order 17915 is not positive-definite).
`
1. Use enough calibration data (≥ 5k samples; model: Qwen2.5-7B-Instruct)
2. Shuffling data so the samples are in a different order
3. I adjusted the `dampening_frac`, which allowed me to complete the model quantization, but running the VLLM inference caused the WSL Ubuntu 22.04 to crash.(No exception information was captured)

In the end, my solution was to switch back to the older version 0.1.0, which resolved the issue, although the quantization process is quite slow.

datasets:
belle_resampled_78_k_cn-train
ultrachat_200k
open-platypus
AI-MO_NuminaMath-CoT

> Unfortunately, without being able to reproduce the setup, I can't provide more help beyond these few suggestions to try:
> 
> 1. Use enough calibration data (≥ 512 samples; if possible, please try 2k, 3k, or 4k as well).
> 2. Once you have enough calibration data, try shuffling it so the samples are in a different order.
> 3. If steps 1 and 2 don't help, try gradually increasing `dampening_frac`. Be aware that this should be the last option, as increasing `dampening_frac` makes your GPTQ algorithms more similar to round-to-nearest quantization, which negatively impacts accuracy.

_Originally posted by @okwinds in https://github.com/vllm-project/llm-compressor/issues/142#issuecomment-2395811942_

markurtz:
Hi @okwinds, can you provide the exact sample dataset so we can attempt to reproduce with the Qwen model? The dampening fraction is the correct pathway to trace down for issues like these. Did you test if the model was runnable as is through HuggingFace before vLLM and if it was creating sensible answers? Because it sounds like quantization went through correctly and his may have been a different crash that happened in vLLM and it was unrelated to the cholesky decomposition.

PR #875, opened 2024-10-29
Add advice about dealing with non-invertable hessians

kylesayrs:
## Purpose ##
* Add advice about dealing with non-invertable hessians when using GPTQModifier
* Fixes https://github.com/vllm-project/llm-compressor/issues/109, https://github.com/vllm-project/llm-compressor/issues/142
* Related to https://github.com/vllm-project/llm-compressor/issues/823

## Changes ##
* Catch error and give advice
  * increasing dampening frac
  * increase number of calibration samples
  * shuffle samples


dsikka (approved):
Thanks. This has been on my todo for a while now.

eldarkurtic (approved):

PR #876, opened 2024-10-30
[DO NOT MERGE] Increase Sparsity Threshold to 50% for sparse compression

rahul-tuli:
#### Background
We have observed instances where accidental sparsity induced by W4-style quantization exceeds 40%. These modules are being identified during the population of `sparsity_config`.

#### Changes
As discussed in our offline conversations, this PR increases the sparsity threshold to 50%.

#### Summary
- Increased the sparsity threshold from 40% to 50% to better handle cases of accidental sparsity induced by W4-style quantization.

### DIFF
--- a/src/llmcompressor/transformers/compression/sparsity_config.py
+++ b/src/llmcompressor/transformers/compression/sparsity_config.py
@@ -20,7 +20,7 @@ class SparsityConfigMetadata:
     metadata from the model
     """
 
-    SPARSITY_THRESHOLD: float = 0.4
+    SPARSITY_THRESHOLD: float = 0.5
 
     @staticmethod
     def infer_global_sparsity(


mgoin (commented):
Putting it right at 0.5 could cause issues due to numerics. Is this threshold a requirement for 2:4 detection as well?

kylesayrs (approved):

src/llmcompressor/transformers/compression/sparsity_config.py
```diff
@@ -20,7 +20,7 @@ class SparsityConfigMetadata:
     metadata from the model
     """
 
-    SPARSITY_THRESHOLD: float = 0.4
+    SPARSITY_THRESHOLD: float = 0.5
```
mgoin, line 23: Change to 0.48 or 0.49 to have a bit of leeway?

rahul-tuli:
Merge pending verification from research team!

dsikka:
Shared the resulting "accidental" sparsity from W4A16 with @anmarques - doesnt seem like anything we should be concerned with

PR #882, opened 2024-10-31
BugFix: Fix Sparsity Reload Testing

dsikka:
# SUMMARY:
- Fix previously failing reload test case 
- This was failing as the lifecycle had changed but the test case was never updated
- As HFQuantizer currently does not support decompressing the entire model on load (only layer by layer) on the forward pass, temporarily commented out that pathway until it is supported and currently using the  model compressor to decompress the model on disk
- HFQuantizer also in general does not support decompressing the embedding layer, even on a layer by layer case.

### DIFF
--- a/tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
+++ b/tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
@@ -174,6 +174,7 @@ def test_quant_model_reload(format, dtype, tmp_path):
         precision=dtype,
     )
 
+    # Fetch the oneshot model
     model = get_session_model()
     og_state_dict = model.state_dict()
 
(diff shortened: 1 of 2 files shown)


tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -156,60 +158,71 @@ def test_quant_model_reload(format, dtype, tmp_path):
     dataset = "open_platypus"
     concatenate_data = False
     num_calibration_samples = 64
-    output_dir = tmp_path / "oneshot_out"
     splits = {"calibration": "train[:10%]"}
+    empty_model = AutoModelForCausalLM.from_pretrained(model_path, torch_dtype=dtype)
```
rahul-tuli, line 162: Eventually, It would be nice if decompression worked with an actual empty model with tensors on a `meta` device, but this is good for the time being!

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -156,60 +158,71 @@ def test_quant_model_reload(format, dtype, tmp_path):
...
+        clear_sparse_session=False,
         precision=dtype,
     )
 
-    model = SparseAutoModelForCausalLM.from_pretrained(
-        tmp_path / "oneshot_out", torch_dtype=dtype
-    )
+    # Fetch the oneshot model
+    model = get_session_model()
+    og_state_dict = model.state_dict()
```
rahul-tuli, line 179: Best variable name in our entire codebase!

rahul-tuli (approved):

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
kylesayrs: Agreed, although you'll want to use `get_state_dict_offloaded_model` to account for the offloaded case

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
kylesayrs: I agree, it'd be nicer if `model_compressor.decompress(path)` returned the decompressed AutoModel instance

kylesayrs (approved):

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -156,60 +158,71 @@ def test_quant_model_reload(format, dtype, tmp_path):
...
+    """
+    # Verify the abs difference between the decompressed model
+    # and the original model
+    reconstructed_state_dict = empty_model.state_dict()
     assert len(og_state_dict) == len(reconstructed_state_dict)
     for key in og_state_dict.keys():
-        dense_tensor = og_state_dict[key]
-        reconstructed_tensor = reconstructed_state_dict[key]
+        dense_tensor = og_state_dict[key].to(device)
+        reconstructed_tensor = reconstructed_state_dict[key].to(device)
```
kylesayrs, line 218: Just for my own education, shouldn't both of these models (and their state dicts) be on the cpu?

dsikka: they can be on either the gpu or cpu, as long as it's the same device. cpu offloading wouldnt really be a case here considering the size of the model.

PR #896, opened 2024-11-04
Fix observing offloaded weight

kylesayrs:
## Purpose ##
* Fix bug where offloaded weights and activations were not be onloaded before being observed

## Changes ##
* Modify `call_observer` function contract to load weights within onloading context if observer basename is weight

## Testing ##
* Ran `examples/quantization_w4a16/llama3_example.py` to completion when previously the example failed


dsikka (approved):
Thanks!

horheynm (approved):

PR #897, opened 2024-11-04
Cap accelerate version to avoid bug

kylesayrs:
## Purpose ##
* A bug introduced in accelerate v1.1.0 breaks loading state_dicts of offloaded models. Until https://github.com/huggingface/accelerate/pull/3217 lands, accelerate's version should be capped to avoid this error


dsikka (approved):
cool thanks. should automatically cap accelerate from compressed-tensors.

mgoin (approved):

PR #898, opened 2024-11-05
fix consecutive oneshot

horheynm:
SUMMARY:
Fix test that runs consecutive oneshots

TEST PLAN:
Pass test

Next todos:
Need a clean way to initialize recipe -- previously done by SparseAutoModelForCausalLM -> need on the second oneshot, so might be able to integrate on HFQuantizer side 
`model=first_tiny_model` should be taken care by using HFQuantizer, not by loading in from the session


tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -48,10 +51,14 @@ def _test_consecutive_runs(
         stages = [stage.group for stage in session_recipe.stages]
         self.assertEqual(len(stages), 1)
         session.reset()
-
+        
+        recipe = resolve_recipe(recipe=self.first_recipe, model_path=self.output_first)
```
kylesayrs, line 55: Very nice, since we are no longer resolving the recipe within SparseAutoModel, we should probably put this in the main function since it's relevant for the finetuning pathway as well?

horheynm: Yeah thats what im thinking about - the code here runs after the first one shot is complete - so the simple solution would be to add this at the end -- will split it to another PR

kylesayrs: Okay

kylesayrs (approved):

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -49,9 +53,13 @@ def _test_consecutive_runs(
         self.assertEqual(len(stages), 1)
         session.reset()
 
+        recipe = resolve_recipe(recipe=self.first_recipe, model_path=self.output_first)
+        if recipe:
+            initialize_recipe(model=first_tiny_model, recipe_path=recipe)
+
```
dsikka, line 59: What is the difference between the `session_recipe` and the recipe we're pulling from the output dir here?

dsikka (commented):
LGTM. just one question

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
horheynm: The recipes are the same. 
The reason we have the logic here is bc we need to let the session know that these recipes has been passed before. 

If no recipes passed, then
https://github.com/vllm-project/llm-compressor/pull/898/files#diff-c9ae8b3ad24d13abeea5b649a5fd6d0b0925f5c9cc40220cbfbe21ae81242f8dR88
will fail

If session is not reset, it will trigger an error 
https://github.com/vllm-project/llm-compressor/blob/b61d4e57954f850a92d4d0332c82cef871abdb03/src/llmcompressor/core/lifecycle.py#L153

Looks like multiple optimization pathway (optimization  = (oneshot (quant, sparse), finetune, etc.) will be in our roadmap then this finalize error becomes a bottleneck

Ill look more into this

dsikka:
I think this wont work until we allow decompression of the entire model

PR #914, opened 2024-11-13
VLM Support via GPTQ Hooks and Data Pipelines

### LINKED ISSUE #91: Layers not skipped with  ignore=[ "re:.*"]
**Describe the bug**
Cosmetic issue. 

Running the code std-out's
```
===== Compressing layer 23/40  =====
2024-08-15T15:22:59.526464+0000 | compress_module | INFO - Compressing model.layers.22.model.layers.22.self_attn.o_proj...
2024-08-15T15:23:00.110515+0000 | compress | INFO - time 0.51
2024-08-15T15:23:00.110713+0000 | compress | INFO - error 0.00
```

**Expected behavior**
exit the compress() function early - but GPTQ will still run. we do do need all the layers in the pipeline for the data to flow properly.

**Environment**
1. OS [e.g. Ubuntu 20.04]:
2. Python version [e.g. 3.7]:
3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]:
4. ML framework version(s) [e.g. torch 2.3.1]:
5. Other Python package versions [e.g. vLLM, compressed-tensors, numpy, ONNX]:
6. Other relevant environment information [e.g. hardware, CUDA version]:

**To Reproduce**
```
recipe = [
    GPTQModifier(targets="Linear", scheme="W8A8", ignore=["lm_head", "re:.*"]),
]
```
using examples/big_models_with_accelerate/multi_gpu_int8.py. 

**Errors**
If applicable, add a full print-out of any errors or exceptions that are raised or include screenshots to help explain your problem.

**Additional context**

### LINKED ISSUE #961: Gptj use gptq quant bug
**Describe the bug**
Exception has occurred: TypeError
GPTJBlock.forward() got multiple values for argument 'hidden_states'
  File "/home/zhoushengguang/workspace/llm-compressor/examples/big_models_with_accelerate/mult_gpus_int8_device_map.py", line 76, in <module>
    oneshot(
TypeError: GPTJBlock.forward() got multiple values for argument 'hidden_states'

**Expected behavior**
Exception has occurred: TypeError
GPTJBlock.forward() got multiple values for argument 'hidden_states'
  File "/home/zhoushengguang/workspace/llm-compressor/examples/big_models_with_accelerate/mult_gpus_int8_device_map.py", line 76, in <module>
    oneshot(
TypeError: GPTJBlock.forward() got multiple values for argument 'hidden_states'

**Environment**
1. OS [e.g. Ubuntu 20.04]:
2. Python version [e.g. 3.7]:
3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]:
4. ML framework version(s) [e.g. torch 2.3.1]:
5. Other Python package versions [e.g. vLLM, compressed-tensors, numpy, ONNX]:
6. Other relevant environment information [e.g. hardware, CUDA version]:

This is my code:
`import torch

from llmcompressor.modifiers.smoothquant import SmoothQuantModifier

model_id="gptj/checkpoint-final"

    model_id, num_gpus=2, reserve_for_hessians=True, torch_dtype=torch.bfloat16
)

model = `AutoModelForCausalLM.`from_pretrained(
    model_id, device_map=device_map, torch_dtype=torch.bfloat16
)
NUM_CALIBRATION_SAMPLES = 512
MAX_SEQUENCE_LENGTH = 2048

recipe = [
    GPTQModifier(
        targets="Linear",
        scheme="W8A8",
        ignore=["lm_head"],
    ),
]
SAVE_DIR="GptJ-6B-GPTQ/"

oneshot(
    model=model,
    dataset="open_platypus",
    recipe=recipe,
    save_compressed=True,
    output_dir=SAVE_DIR,
)
`

### LINKED ISSUE #990: When llava-v1.6 models w8a8_int8 quantization can be supported?
any plan for llava-v1.6 models be supported by w8a8_int8 ?

kylesayrs:
## Purpose ##
* Enable oneshot quantization of vision-language models

VLM Banner
[Llama_3 2-Vision Graphviz](https://github.com/user-attachments/assets/6b371ccc-f9f6-4bf2-b4cd-24ed75a3cad0)

## Related Issues ##
* Fixes #91
* Fixes #961
* Fixes #990

## Prerequisites ##
* https://github.com/neuralmagic/compressed-tensors/pull/193
* #917
* #943
  * #955
    * #950
* #998
* #1014

### VLM Support ###
* Add multimodal examples in `examples/multimodal_vision`
* Modify `custom_offload_device_map` to support models which are not `XForCausalLM`
* Add custom data collators for VLM models in `src/llmcompressor/transformers/utils/data_collator.py`

### GPTQModifier ###
* Implement hooks-based compression in `GPTQModifier`
  * This replaces layer-compressor, which made many assumptions about model architecture
  * This also enables finer-grained sequential compression such as [true_sequential](https://huggingface.co/docs/transformers/main_classes/quantization#transformers.GPTQConfig.true_sequential)
  * Functions previously implemented in `gptq_wrapper.py` are now implemented in `gptq_quantize.py`
* Implement `offload_hessians` parameter in `GPTQModifier`
* Implement data-pipelines-based calibration in `GPTQModifier`
  * First an attempt will be made to trace the model and run the `sequential` pipeline
  * If that fails, assumptions will be made about the model architecture and an attempt will be made to run the `layer_sequential` pipeline
    * This ensures backwards compatibility with any previously supported models
  * If that fails, then the basic pipeline will be used, which is guaranteed to run but may require using `offlo ad_hessians`
* Change hessian instability from a `ValueError` to a `_LinAlgError` so it can be ignored by the gptq pipeline fallback mechanism
* Add support for conv2d as indicated by [AutoGPTQ](https://github.com/AutoGPTQ/AutoGPTQ/blob/6689349625de973b9ee3016c28c11f32acf7f02c/auto_gptq/quantization/gptq.py#L45-L54)

### Data Pipelines ###
* Implement the basic skeletons of data pipelines, which are subject to change when data pipelines are pulled out of modifiers
* Basic Pipeline
  * Performs standard forward passes through the model with provided dataloader
  * Used as fallback, as well as in the future for basic calibration passes
* Layer Sequential Pipeline
  * Refactor of `LayerCompressor` as a straight-forward data pipeline
  * Uses `IntermediatesCache` to handle activation offloading
* Sequential Pipeline
  * Utilizes graph tracing implemented by `torch.fx` to trace the graph in order to determine where sequential targets (layers) exist in the graph and what their inputs and outputs are
  * Implements BFS algorithm to assign nodes to partitions
    * An ideal implementation consolidates partition indices to assign each node to the latest possible partition, delaying execution. The current implementation addresses the most common case (node.op == get_attr)
  * Each partition (`Subgraph`) is compiled as an executable python function with the proper inputs and outputs
  * Uses `IntermediatesCache` to handle activation offloading
* Implement `IntermediatesCache` which automagically handles the offloading and onloading of activations from batches
  * This class is capable of offloading many non-standard activation types such as `Tuple`s and dataclasses such as `BaseModelOutputWithPast`
  * For convenience, the class also handles masking padding
  * The class is tested in `tests/llmcompressor/pipelines/test_cache.py`

### Tracing ###
* In order to support sequential quantization of the large variety of different multimodal model architectures, some model definitions have to be altered to support tracing
  * If the calibration dataset is text only, most LLMs and VLMs are traceable without additional work. Multimodal calibration datasets are more likely to require additional work to make tracable
  * For many VLMs (but not all), the vision tower is not traceable without significant work. However, this only affects sequential error propagation and (minimal?) increased memory usage, which leaves the door open for future support for quantizing modules in the vision tower
* Add traceable model definitions for llava, mistral, mllama, and glm
* All copyright licenses allow for alteration and redistribution, the line `# vllm-project: no copyright` was added in similar style to [text_generation.py](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/transformers/finetune/text_generation.py#L18)

## Future Work/ Follow ups ##
* #1027
* #1032
* #1039
* #1030
* Create better data collators capable of handling larger batch sizes in order to support VLM fine tuning
* Better support prompt masking for multimodal processors in order to support VLM fine tuning
* Add [true_sequential](https://huggingface.co/docs/transformers/main_classes/quantization#transformers.GPTQConfig.true_sequential) option
* lm_head and embedding quantization
* Potential uses targeting hooks on operations rather than just modules, as would be the case with post-rope transformation hooks 

## Winogrande Evaluations ##

Model | Dataset | Scheme | Runtime | Winogrande |
-- | -- | -- | -- | --
Llama-3-8B | ultrachat | W4A16 | 43m, 2xA4000 | 0.7545 
Llama-3-70B | ultrachat | W4A16 | 303m, 1xH100 | 0.8216 
Mixtral-8x7B | ultrachat | W4A16 | 317m, 1xA100 | 0.8200 
openbmb/MiniCPM3-4B | ultrachat | W4A16 | 63m, 1xA100 | 0.6701 
Qwen2-VL-2B-Instruct | ultrachat | W8A8 | 12m, 2xA4000 | 0.6188 
Qwen2-VL-2B-Instruct | flickr | W8A8 | 24m, 2xA4000 | 0.6093 
Llama-3.2-11B-Vision-Instruct | flickr | W8A8 | 75m, 1xA100 | 0.7837 
Pixtral-12B-2409 | flickr | W8A8 | 52m, 1xA100 | 0.7924 
llava-1.5-7b-hf | flickr | W8A8 | 15m, 1xH100 | 0.7214 
Phi-3-vision-128k-instruct | flickr | W4A16 | 51m, 1xA100 | 0.7151 

`lm_eval --model vllm --model_args pretrained="path/to/model",dtype=auto,max_model_len=4096,tensor_parallel_size=1,gpu_memory_utilization=0.8,enforce_eager=True,add_bos_token=True --tasks winogrande --num_fewshot 5 --batch_size 32`
`lm_eval --model vllm --model_args pretrained="path/to/model",dtype=bfloat16,max_model_len=4096,tensor_parallel_size=1,gpu_memory_utilization=0.8,enforce_eager=True,add_bos_token=True,max_num_seqs=1 --tasks winogrande --num_fewshot 5 --batch_size 1`

## MMMU Evaluations ##
Credit to @shubhra 

Model | Dataset | Scheme | MMMU
-- | -- | -- | --
Llama-3.2-11B-Vision | N/A | Dense | 0.4144
Llama-3.2-11B-Vision | N/A | FP8-dynamic | 0.4300
Llama-3.2-11B-Vision | flickr | W4A16 | 0.4377
Llama-3.2-11B-Vision | flickr | W4A16-group | 0.4211

Model | Dataset | Scheme | MMMU
-- | -- | -- | --
Llama-3.2-90B-Vision | N/A | Dense | 0.5388
Llama-3.2-90B-Vision | N/A | FP8-dynamic | 0.5278
Llama-3.2-90B-Vision | flickr | W4A16 | 0.5111
Llama-3.2-90B-Vision | flickr | W4A16-group | 0.5477

Model | Dataset | Scheme | MMMU
-- | -- | -- | --
Pixtral-12B-2409 | N/A | Dense | 0.5022
Pixtral-12B-2409 | N/A | FP8-dynamic | 0.5322
Pixtral-12B-2409 | flickr | W4A16 | 0.4500
Pixtral-12B-2409 | flickr | W4A16-group | 0.4689

## Testing ##
* [Nightly](https://github.com/neuralmagic/llm-compressor-testing/actions/runs/12640439996)


kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/12359686271

kylesayrs:
No new nightly failures: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/12398026591/job/34609853074

dsikka:
Is this ready for review?
If yes, can we update the PR description, show testing results, etc?

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/12536961704

kylesayrs:
@dsikka The PR description has been updated and the PR is ready for review

dsikka (commented):
Generally speaking, we should aim to keep PRs under <2k lines of code. 

For example, some of these changes could be introduced as follow-ups in smaller PRs, such as updates to examples or landing the first two pipelines and then following up with the tracing changes.

Some of the changes are also not necessarily vision model specific, such as removal of the GPTQ layer wrapper, which we could have potentially landed sooner to start.

kylesayrs:
@dsikka I'll note that 6K of those lines are copies of model definitions with slight modifications to enable tracing. For those files, we really only need to review the modifications. I'll make another commit to make the locations of those modifications clearer

src/llmcompressor/pipelines/layer_sequential/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    signature = inspect.signature(first_layer.forward)
+
+    with calibration_forward_context(model), early_stop_hook(first_layer):
+        desc = "Preparing intermediates cache"
+        for batch_index, batch in enumerate(tqdm.tqdm(dataloader, desc=desc)):
+            batch = apply_pad_mask_to_batch(batch) if mask_padding else batch
+            batch = tensors_to_device(batch, model_device)
+
+            try:
+                model(**batch)
```
dsikka, line 50: If we're using multiple batches, i.e batch_size < dataset_size, the early stop would execute and return an intermediate output after the first batch from a layer which has not yet been quantized, correct? As _update_size by default is the entire dataset?

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -0,0 +1,71 @@
...
+        # prepare intermediates cache
+        intermediates = capture_first_layer_intermediates(model, layers, dataloader)
+
+        num_layers = len(layers)
+        for layer_index, layer in enumerate(layers):
+            # prepare tqdm description texts
+            calib_desc = f"({layer_index + 1}/{num_layers}): Calibrating"
+            prop_desc = f"({layer_index + 1}/{num_layers}): Propagating"
+
+            if propagate_error:
```
dsikka, line 53: I think for this particular pipeline, we should probably keep propagate_error=True as the only supported case/consider other cases as we expand data pipelines beyond GPTQ

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+from transformers.utils.fx import HFTracer
+
+from llmcompressor.modifiers.utils.hooks import HooksMixin
+from llmcompressor.utils.helpers import calibration_forward_context
+
+__all__ = ["trace_subgraphs", "Subgraph"]
+
+
+@dataclass
+class Subgraph:
```
dsikka, line 22: docstring

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+    input_names: Set[str]
+    consumed_names: Set[str]
+
+    def compile_forward(self):
+        code = self.graph.python_code("self")
+        exec(code.src, code.globals)
+        return code.globals.get("forward")
+
+
+def trace_subgraphs(
```
dsikka, line 33: docstring

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+
+    # perform subgraph partition
+    partitions = topological_partition(graph, sequential_targets)
+    subgraphs = partition_graph(model, partitions)
+    trace_consumed_names(subgraphs)
+
+    return subgraphs
+
+
+def get_tracer(
```
dsikka, line 77: docstring

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,272 @@
...
+                    user_partitions.append(index)
+                    break
+        partition_index = min(user_partitions)
+        partitions[partition_index].insert(0, node)
+
+    assert set().union(*partitions) == set(graph.graph.nodes)
+    return partitions
+
+
+def partition_graph(model: Module, partitions: List[List[Node]]) -> List[Subgraph]:
```
dsikka, line 187: dosctring

src/llmcompressor/transformers/tracing/glm/LICENSE
```diff
@@ -0,0 +1,84 @@
+The glm-4-9b License
+
+1. 定义
```
dsikka, line 3: this is needed for chatglm?

src/llmcompressor/pipelines/cache.py
```diff
@@ -0,0 +1,137 @@
...
+import tqdm
+
+
+@dataclass
+class IntermediateValue:
+    value: Union[torch.Tensor, "IntermediateValue", Any]
+    device: Union[torch.device, None]
+
+
+class IntermediatesCache:
```
dsikka, line 15: docstring

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -175,150 +177,183 @@ def on_initialize_structure(self, state: State, **kwargs):
...
+        # infer update size
+        if self._update_size is None:
+            self._update_size = len(state.data.calib)
+
+        # infer pipeline
+        model_name = state.model.__class__.__name__
+        input_names = state.data.calib.dataset.column_names
+        unfixable_errors = (torch.OutOfMemoryError, torch._C._LinAlgError)
+        try:
+            run_sequential(
```
dsikka, line 222: I think we need a more descriptive name to help better differentiate with the layer_sequential pipeline

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -0,0 +1,77 @@
...
+        num_subgraphs = len(subgraphs)
+        for subgraph_index, subgraph in enumerate(subgraphs):
+            # prepare tqdm description texts
+            calib_desc = f"({subgraph_index + 1}/{num_subgraphs}): Calibrating"
+            prop_desc = f"({subgraph_index + 1}/{num_subgraphs}): Propagating"
+
+            # compile subgraph forward function
+            forward_function = subgraph.compile_forward()
+
+            if propagate_error:
```
dsikka, line 60: same as other pipeline, let's remove case for `propagate_error=False`

dsikka (changes requested):
Need to go through the tracing pipeline in more detail. 
It would also be helpful to understand the required changes to the models to make them compatible with tracing. Your diagram provides some insight but not 100% clear to me atm.

kylesayrs:
I've removed unnecessary model definitions and reduced the total changes by 3.5K lines

examples/multimodal_vision/llava_example.py
```diff
@@ -0,0 +1,54 @@
+from transformers import AutoProcessor
+
+from llmcompressor.modifiers.quantization import GPTQModifier
+from llmcompressor.transformers import oneshot
+from llmcompressor.transformers.tracing import TracableLlavaForConditionalGeneration
+from llmcompressor.transformers.utils.data_collator import llava_data_collator
+
+# Load model.
+model_id = "llava-hf/llava-1.5-7b-hf"
+model = TracableLlavaForConditionalGeneration.from_pretrained(
```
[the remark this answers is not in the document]
kylesayrs: #1032

src/llmcompressor/transformers/tracing/glm/LICENSE
kylesayrs: Yes

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: I'm all ears for name suggestions. "sequential" is the term used by LC and AutoGPTQ to refer to the process of breaking a model into parts for the purpose of propagating error / reducing the number of active hessians.

"Layer Sequential Pipeline" and "Sequential Pipeline" were the most cogent names I could think of to describe this process at the time.

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs: Good idea

src/llmcompressor/pipelines/layer_sequential/pipeline.py
kylesayrs: Good call

src/llmcompressor/pipelines/layer_sequential/helpers.py
kylesayrs: If I'm reading your explanation correctly, then in most cases no.

If a modifier targets modules which execute before the first "layer" (for example, torch.nn.Embedding), then yes, that module would pass on outputs which do not propagate that module's compression error.

However, in most cases, we make the assumption that this is not the case.

We could fix this by running two forward passes (one calibration, one propagation) similar to the other layers, but this comes at considerable performance cost and is likely overkill for a non-default pipeline anyways.

kylesayrs: Thanks for the catch

src/llmcompressor/pipelines/layer_sequential/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    mask_padding: bool = True,
+) -> IntermediatesCache:
+    model_device = get_execution_device(model)
+    intermediates = IntermediatesCache.empty(len(dataloader), torch.device("cpu"))
+    first_layer = layers[0]
+    signature = inspect.signature(first_layer.forward)
+
+    with calibration_forward_context(model), early_stop_hook(first_layer):
+        desc = "Preparing intermediates cache"
+        for batch_index, batch in enumerate(tqdm.tqdm(dataloader, desc=desc)):
```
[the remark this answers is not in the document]
kylesayrs: By default, [calibration data is batch_size=1](https://github.com/vllm-project/llm-compressor/blob/1b8c7bf1fa3999529fc36e701573effc5ffa6a66/src/llmcompressor/transformers/finetune/data/data_helpers.py#L54).

However, one of the effects of the data pipeline refactor is to allow users to choose batch_size in order to trade memory and runtime, similar to a normal training loop.

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -175,150 +177,183 @@ def on_initialize_structure(self, state: State, **kwargs):
...
+                self._hessians[module],
+                self._num_samples[module],
             )
-            self.layer_compressors_[0].clear_early_stop()
 
-            num_layers = len(self.compressible_layers_)
-            for idx, layer_compressor in enumerate(self.layer_compressors_):
-                logger.info(f"\n===== Compressing layer {idx+1}/{num_layers} " " =====")
+        # After enough samples are accumulated, perform quantization
+        if self._num_samples[module] >= self._update_size:
```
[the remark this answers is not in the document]
kylesayrs: Example (update_size=3, dataset_size = 9)

1. make_empty_hessian + accumulate_hessian
2. accumulate_hessian
3. accumulate_hessian + quantize_weight
4. make_empty_hessian + accumulate_hessian
5. accumulate_hessian
6. accumulate_hessian + quantize_weight
7. make_empty_hessian + accumulate_hessian
8. accumulate_hessian
9. accumulate_hessian + quantize_weight

kylesayrs:
@dsikka Some common examples of untracable code and how to fix them are being detailed in https://github.com/vllm-project/llm-compressor/pull/1030

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def run_pipeline(model: torch.nn.Module, dataloader: torch.utils.data.DataLoader):
+    model_device = get_execution_device(model)
+
+    with calibration_forward_context(model):
+        for batch in tqdm.tqdm(dataloader, desc="Calibrating"):
+            batch = apply_pad_mask_to_batch(batch)
+            batch = tensors_to_device(batch, model_device)
+            model(**batch)
```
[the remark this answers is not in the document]
kylesayrs: I'll add to the docstring

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -0,0 +1,71 @@
...
+        then caught and used to capture the input arguments to the first layer
+    3. The inputs to the first layer are used to calibrate the first layer, and the
+        output of the previous layer is used as inputs to calibrate the next layer
+
+    This pipeline requires that the model have distinct layers defined in its
+    architecture and that the outputs of the previous layer are exactly the inputs
+    to the next layer. This is violated by encoder-decoder architectures among others.
+
+    If your model architecture violates these assumptions, consider using the sequential
+    pipeline (see llmcompressor.pipelines.sequential)
```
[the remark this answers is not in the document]
kylesayrs: Good idea

src/llmcompressor/modifiers/quantization/gptq/base.py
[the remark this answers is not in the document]
dsikka: Ah makes sense. Sorry, missed the reset code at the bottom of the hook.

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -0,0 +1,72 @@
...
+    llmcompressor.transformers.tracing)
+    """
+    # trace subgraphs
+    sample_input = next(iter(dataloader))
+    subgraphs = trace_subgraphs(model, sample_input, sequential_targets, ignore)
+
+    with calibration_forward_context(model):
+        # prepare intermediates cache
+        model_device = get_execution_device(model)
+        intermediates = IntermediatesCache.from_dataloader(dataloader, model_device)
```
dsikka, line 47: Why do we not need to capture the activations before the first subgraph, as we have to do with raising the early exception for the layer_sequential pipeline? Seems like the IntermediateCache is filled with values from the dataloader, not any previous activations.

src/llmcompressor/modifiers/quantization/gptq/base.py
dsikka: Could we do "Layer Sequential" and "Subgraph Sequential" ?  Sequential being indicative of the data/error propagation while using "layer" and "subgraph" to differentiate between data structures?

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -0,0 +1,68 @@
...
+    If your model architecture violates these assumptions, consider using the sequential
+    pipeline (see llmcompressor.pipelines.sequential). Architectures which are known to
+    fail these assumptions include GPT-J and most vision language models
+    """
+    # find layers
+    layers = match_modules(model, sequential_targets)
+
+    with calibration_forward_context(model):
+        # prepare intermediates cache
+        intermediates = capture_first_layer_intermediates(model, layers, dataloader)
```
dsikka, line 45: nit for clarity:
```python
intermediates: IntermediatesCache = capture...
```

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -0,0 +1,68 @@
...
+            # and is only used for capturing outputs from the newly compressed modules
+            with HooksMixin.disable_hooks():
+                for batch_index in tqdm.tqdm(range(len(dataloader)), desc=prop_desc):
+                    inputs = intermediates.fetch(batch_index)
+                    output = layer(**inputs)
+
+                    if layer_index < num_layers - 1:
+                        output = to_next_layer_kwargs(output, layers[layer_index + 1])
+                        intermediates.delete(batch_index)
+                        intermediates.update(batch_index, output)
```
dsikka, line 68: Because this is primarily not going to be used by vision models, is there a need to offload/onload batches?

Do we have any data on the additional time it takes for running through the pipeline, with/without the offload?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -0,0 +1,72 @@
...
+
+            # this pass does not trigger modifier hooks
+            # and is only used for capturing outputs from the newly compressed modules
+            with HooksMixin.disable_hooks():
+                for batch_index in tqdm.tqdm(range(len(dataloader)), desc=prop_desc):
+                    inputs = intermediates.fetch(batch_index, subgraph.input_names)
+                    output = forward_function(model, **inputs)
+
+                    if subgraph_index < num_subgraphs - 1:
+                        intermediates.update(batch_index, output)
```
dsikka, line 71: I would expect with vision models we have to offload batches. Just wondering if we have any data on change in performance when we offload to cpu? Or when combined with parts of the model being offloaded as well?

examples/multimodal_vision/llava_example.py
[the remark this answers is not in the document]
dsikka: For examples, could we add a case where the model is traceable without modifications?
Not urgent. Just as a follow-up if I didn't miss it in this PR

src/llmcompressor/pipelines/layer_sequential/pipeline.py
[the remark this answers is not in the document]
kylesayrs: Agree

src/llmcompressor/pipelines/layer_sequential/pipeline.py
kylesayrs: Activation offloading as nothing to do with vision models. Activation offloading was done by the layer compressor.

Without activation offloading, the algorithm will have the same memory requirements as if we were running a forward pass of the model with `batch_size = ds_size` (which is to say, a lot)

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs: See https://github.com/vllm-project/llm-compressor/pull/914/files#r1904553406

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs: The layer sequential pipeline processes everything before the first layer and captures the activations.

The sequential pipeline assigns everything before the first layer to the first subgraph. Said another way, all the processing that is done by `capture_first_layer_intermediates` is done in the sequential pipeline by processing the first subgraph.

```
subgraph[0] ~= capture_first_layer_intermediates
subgraph[1] ~= layer[0]
```

src/llmcompressor/pytorch/utils/helpers.py
```diff
@@ -320,9 +319,7 @@ def tensors_to_device(
     if isinstance(tensors, Iterable):
         return [tensors_to_device(tens, device) for tens in tensors]
 
-    raise ValueError(
-        "unrecognized type for tensors given of {}".format(tensors.__class__.__name__)
-    )
+    return tensors
```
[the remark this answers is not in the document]
kylesayrs: Thanks for the catch, I believe we can remove this change (although it wouldn't hurt to keep it either).

During development I initially used this function offload/onload intermediates, but I implemented more robustly with the `IntermediatesCache`.

I think it makes sense to keep both functions. I'll remove this change and perform some regression tests to make sure everything still works.

src/llmcompressor/pipelines/layer_sequential/pipeline.py
kylesayrs: > Activation offloading was done by the layer compressor

On main, we use activation offloading.
https://github.com/vllm-project/llm-compressor/blob/47bd37eee115687042c02d48784afc24c7e3b5ec/src/llmcompressor/modifiers/utils/pytorch_helpers.py#L29
https://github.com/vllm-project/llm-compressor/blob/47bd37eee115687042c02d48784afc24c7e3b5ec/src/llmcompressor/modifiers/utils/layer_compressor.py#L136

In this branch, we also use activation offloading. Therefore there is no change in performance due to there being no change in this regard.

kylesayrs: I did a runtime regression a week ago and found that on the llama-8B model there was only about a 9% difference in runtime between main and this branch.

You can also see from the runtimes documented under the Winogrande Evaluations that the runtimes are reasonable.

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: I don't really like exposing the idea of subgraphs to the user, as most users won't have the context to really know what that means or why they should choose it.

In the future, the sequential pipeline will be the default for any recipes which include gptq, wanda, or sparsegpt. I'd prefer to keep this name short.

The layer sequential pipeline is only useful as a fallback for a small subset of models which follow layer-wise assumptions (next.input == prev.output) and are not traceable (due to weird or bad programming). The only model I've found like this is minicpm.

src/llmcompressor/pipelines/cache.py
```diff
@@ -0,0 +1,191 @@
...
+        Initialize a cache with data from the provided dataloader
+
+        :param dataloader: dataloader which generates values to be cached
+        :param model_device: device which values will be onloaded to when fetched
+        :param mask_padding: zero out padding tokens if True. This affects modifiers
+            such as GPTQ and SparseGPT
+        :param offload_device: device to offload values to
+        """
+        batch_intermediates = [
+            {
```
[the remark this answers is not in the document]
kylesayrs: If you're referring to list compression, I agree. This function has about the same runtime with or without list comprehension, so let's remove it.

llama inputs
```
Preparing intermediates cache: 100%|███| 512/512 [00:00<00:00, 848.23it/s]  # with comp
Preparing intermediates cache: 100%|███| 512/512 [00:00<00:00, 860.89it/s]  # no comp
```

mllama inputs
```
Preparing intermediates cache: 100%|███| 20/20 [00:46<00:00,  2.31s/it]  # with comp
Preparing intermediates cache: 100%|███| 20/20 [00:46<00:00,  2.31s/it]  # no comp
```

examples/multimodal_vision/llava_example.py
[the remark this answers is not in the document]
dsikka: thanks

src/llmcompressor/pipelines/layer_sequential/pipeline.py
kylesayrs: Yeah happy to add any comments to make this clearer

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/12640439996

src/llmcompressor/transformers/tracing/glm/LICENSE
```diff
@@ -0,0 +1,84 @@
+The glm-4-9b License
```
rahul-tuli, line 1: Could you check if we need to add this to the manifest file? or will it get picked up in our wheels automatically?

rahul-tuli (approved):
Really like the IntermediatesCache implementation, good job!

src/llmcompressor/transformers/tracing/glm/LICENSE
kylesayrs: I would not have caught that, thank you!

It is not included by default. After I added it to the manifest.in, you can verify that it is picked up as so:
```bash
python setup.py sdist
tar -tzf dist/llmcompressor-dev-0.3.1.dev0.tar.gz | grep LICENSE
```

src/llmcompressor/modifiers/quantization/gptq/base.py
dsikka: hm let me think of other descriptors 
I think we just want each of the pipelines beyond the basic pipeline to be a little more verbose in its name

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,365 @@
...
+
+    return subgraphs
+
+
+def get_tracer(
+    model: Module, sequential_targets: Set[Module], ignore: Set[Module]
+) -> HFTracer:
+    """
+    Get a tracer specialized for the given model. The resulting tracer will not trace
+    inside of sequential targets, ignored targets, or offloaded modules.
```
dsikka, line 108: Trying to understand this comment. If the resulting tracer does not trace offloaded modules, how does this work for cases when we have parts of the model offloaded?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,365 @@
...
+    :param ignore: list of patterns matching modules to ignore during tracing
+    :return: a list of Subgraphs in order of execution
+    """
+    # find modules
+    sequential_targets = match_modules(model, sequential_targets)
+    ignore = match_modules(model, ignore)
+
+    # initialize arguments
+    tracer = get_tracer(model, sequential_targets, ignore)
+    concrete_args = populate_concrete_args(model, sample_input)
```
dsikka, line 71: populate_concrete_args is just creating a dictionary of defaults args?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,365 @@
...
+
+    :param model: model being traced
+    :param sequential_targets: modules which are sequential targets
+    :param ignore: modules which are ignored
+    """
+    offloaded_modules = set(m for m in model.modules() if has_offloaded_params(m))
+    skip_trace_modules = sequential_targets | offloaded_modules | ignore
+
+    class SequentialTracer(HFTracer):
+        def create_arg(self, a: Any) -> Argument:
```
dsikka, line 121: can you explain why this is needed?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,365 @@
...
+    Partition the graph into partitions such that each `target` belongs to exactly one
+    partition and executing each partition depends only on intermediate values produced
+    by executing the partitions before it.
+
+    :param graph: graph being partitioned
+    :param targets: target modules which will be assigned to disjoint partitions
+    :return: list of partitions, where each partition is a list of nodes belonging to
+        that partition
+    """
+    assert check_assumption(graph.graph)
```
dsikka, line 197: nit: update for a more descriptive name e.g check_graph_not_malformed or something

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,365 @@
...
+
+        if len(node.users) != len(set(node.users)) or len(node.all_input_nodes) != len(
+            set(node.all_input_nodes)
+        ):
+            return False
+
+    return True
+
+
+def match_modules(model: Module, target_names: List[str]) -> Set[Module]:
```
dsikka, line 353: we seem to have multiple match_module methods.
https://github.com/vllm-project/llm-compressor/blob/153a4fa3c4d4831e01219b8d7a901366555bd960/src/llmcompressor/pipelines/layer_sequential/helpers.py#L21
We should either update their names, consolidate, make private or, nest if only used by one other method

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+def trace_consumed_names(subgraphs: List[Subgraph]):
+    """
+    Populate the `consumed_names` attribute of each Subgraph according to when inputs
+    are last used in order to vacate the `intermediates` cache and save memory
+
+    :param subgraphs: list of subgraphs with empty `consumed_names` attributes
+    """
+    # populate consumed_names according to when inputs are last used
```
dsikka, line 316: what does this mean by "last used"?

src/llmcompressor/transformers/tracing/mllama.py
```diff
@@ -0,0 +1,162 @@
...
+from torch.fx import wrap
+from transformers.models.mllama.modeling_mllama import (
+    MLLAMA_START_DOCSTRING,
+    MllamaForConditionalGeneration,
+)
+
+logger = logging.get_logger(__name__)
+
+
+# TRACING: This function is not traceable
```
dsikka, line 40: I think it would be helpful to have quick notes to help understand what you had to update to make this model traceable

examples/multimodal_vision/pixtral_example.py
```diff
@@ -0,0 +1,54 @@
...
+    recipe=recipe,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+    trust_remote_code_model=True,
+    data_collator=pixtral_data_collator,
+)
+
+# Confirm generations of the quantized model look sane.
+print("========== SAMPLE GENERATION ==============")
+input_ids = processor(text="Hello my name is", return_tensors="pt").input_ids.to("cuda")
```
dsikka, line 46: Sample generation should probably be image related?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -0,0 +1,365 @@
...
+    ):
+        graph = GraphModule(
+            model,
+            tracer.trace(
+                model,
+                dummy_inputs=sample_input,
+                concrete_args=concrete_args,
+                complete_concrete_args_with_inputs_not_in_dummy_inputs=False,
+                # bug in trace throws an error for variadic
+                # args and kwargs in function signature
```
mgoin, line 86: Is this just explaining that you couldn't pass `*args, **kwargs` here?

src/llmcompressor/pipelines/sequential/helpers.py
mgoin: What does the tracer actually trace?

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: We do expose the idea of "tracing". Just something to think about, I'll sleep on it too

src/llmcompressor/transformers/tracing/mistral.py
```diff
@@ -0,0 +1,252 @@
...
+class MistralForCausalLM(MistralForCausalLM):
+    def __init__(self, config):
+        super(MistralPreTrainedModel, self).__init__(config)
+        # TRACING: Must use MistralModel
+        self.model = MistralModel(config)
+        self.vocab_size = config.vocab_size
+        self.lm_head = nn.Linear(config.hidden_size, config.vocab_size, bias=False)
+
+        # Initialize weights and apply final processing
+        self.post_init()
```
mgoin, line 208: This looks to be the same as the modeling definition https://github.com/huggingface/transformers/blob/12ba96aa3cb3e4ed2a3ffb77b59f53f8ce9ac1fa/src/transformers/models/mistral/modeling_mistral.py#L752-L763
What is the purpose of this comment and code?

mgoin (commented):
I think these Traceable model definitions have very opaque changes compared to the reference model definitions. This architecture seems like an intensive blocker to add support for a new model, as it requires a lot of knowledge of tracing limitations. However I understand the need - I'll look in more detail tomorrow

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: Yes, concrete args are essentially static flags and the set of all args that are not dynamically being traced.

Normally transformer's `trace` implementation automatically creates concrete args, but as noted in this function's description, `populate_concrete_args` also handles a bugged case where some models need to be provided default *args and **kwargs

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: > Tracing within sequential targets and ignored targets is unnecessary, and tracing within offloaded modules may result in meta tensors being added to the model graph

When a module is "not traced", this means that the internals of module are not traced, but the module still appears in the graph as a `call_module` node.

kylesayrs: For example, in this model graph, the internals of `language_model_model_layers_39` are not traced (we don't see individual nodes for the attention mechanism, ffn, ect.) but all the operations after the module are traced with high granularity (we see individual `pow`, `.to` and `.dtype` functions).

Screenshot 2025-01-06 at 23 22 15

kylesayrs: As a side note, even if a module contains untraceable code internally, if the internals of the module skip tracing via `ignore` or `sequential_targets` or `has_offloaded_params`, then the model graph as a whole will still be traceable, but just with less granularity.

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: > special extension allows models which depend on config values to be traced

I overload this function to insert my own definition for creating an argument which is of type `PretrainedConfig`. Many models use values from the config during execution, but `torch.fx` is only capable of "baking" a limited set of class types into the model graph.

This code says, whenever the graph would try to reference an instance of a `PretrainedConfig` (for example, to get an attribute like `config.max_sequence_length`), instead just create a `PretrainedConfig` on the fly and initialize it with all of the args from the original config.

kylesayrs: There are a few things in this file that we should consider upstreaming HF, this might be one of them

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: Some input names are used by multiple subgraphs in the model (for example, the cross attention output is used by every text decoder layer in mllama) while other input names are only used once (for example, the hidden state of a text decoder layer).

All outputs are stored in the IntermediatesCache. However, we only want to keep outputs which will be later used as inputs to later subgraphs and vacate outputs which are never used again (this is really only to reduce cpu memory usage). Therefore, for each name, we need to find the index of the subgraph which is the last user of that name. After that, we can vacate that output from the cache.

Side note, outputs which do not lead to inputs are automatically pruned by the instantiation of `GraphModule`, and this is validated by `check_assumption`).

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: I agree, I'll add a note to consolidate them later

src/llmcompressor/transformers/tracing/mllama.py
kylesayrs: > what you had to update to make this model traceable

"what" I had to update to make this model traceable is exactly the code in this file. "why" and "where" it was done is indicated by the `# TRACING` comments

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: This is explaining why I need to write my own `populate_concrete_args` function rather than rely on [the one provided by transformers](https://github.com/huggingface/transformers/blob/main/src/transformers/utils/fx.py#L1268-L1280)

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: @mgoin The tracer traces all of the objects and operations that are need perform a forward pass of the model.

src/llmcompressor/transformers/tracing/mistral.py
kylesayrs: That comments should say
```
# TRACING: Must use MistralModel with wrapped _prepare_4d_causal_attention_mask_with_cache_position function
```

Note that we define a version of the `MistralModel` which wraps the problematic function, and it is this definition that is used by `MistralForCausalLM`.

https://github.com/vllm-project/llm-compressor/pull/914/files/153a4fa3c4d4831e01219b8d7a901366555bd960#diff-75d8b6c11cbd2f8c8efbb720ba0566c4c895c4bbadeee774382f7fe7a4f6a3baR105-R106

kylesayrs:
@mgoin I think the [Tracing Guide](https://github.com/vllm-project/llm-compressor/pull/1030) will clarify how and why to make changes to your model to make it traceable and why tracing is the best and least invasive solution currently available.

Also note that
1. Unlike vllm, custom model definitions are not needed for every model. For the vast majority of text models, custom definitions are not required. Most vision models _when calibrated with text datasets_ also do not require custom tracing. Custom definitions are mostly required for vision models when calibrated with vision datasets, and even then some models like [phi3_vision](https://github.com/vllm-project/llm-compressor/pull/1032) do not require any changes.
2. Even if a text model is not traceable, gptq falls back to the layer_sequential pipeline, which is equivalent to what is currently on main. Therefore these changes only extend what is possible with llm-compressor now.

examples/multimodal_vision/pixtral_example.py
kylesayrs: Will create follow up PR so we can land this asap

dsikka (approved):
Great work!

rahul-tuli (approved):

mgoin (approved):
Thanks for responding to comments, let's get to the followup items after this

PR #917, opened 2024-11-14
Implement HooksMixin

kylesayrs:
## Purpose ##
* Precursor to #914
* Create a shared API for adding hooks to modules
* Allow code which handles data pipelines to selectively disable hooks for certain passes. This will be needed in cases with custom datapipelines (GPTQ/Wanda/SparseGPTQ) and when multiple modifiers are active at the same time.
  * This is needed for GPTQ-style sequential algorithms which require one pass with hooks in order to accumulate the hessians and compress, and then a second pass without hooks in order to compute compressed (weight-quantized) outputs
  * This is also a tool for research users to be able to control when hooks are enabled from within the data pipelines

```python3
for layer in model_layers:
    # accumulate hessians
    unquantized_outputs = layer(*args, **kwargs)

    # get sequential outputs
    with HooksMixin.disable_hooks():
        quantized_outputs = layer(*args, **kwargs)
    
    print(f"Mean error from quantization: {get_loss(unquantized_outputs, quantized_outputs)}")
```

## Changes ##
* Implement `HooksMixin`
  * The `_HOOKS_DISABLED` attribute is a global variable attached to the class which is used to disable hooks globally
  * The `_hooks` attribute is a local variable attached to each modifier which lists all of the hooks created by that modifier
* Integrate with `QuantizationModifier`, refactor calibration functions to reference the same function rather than generating hook functions
* Integrate with `SmoothQuantModifier`
* Integrate with `WandaPruningModifier` and `SparseGPTModifier`
* Integrate with `MagnitudePruningModifier` and `ConstantPruningModifier` via `LayerParamMasking`
* Purposefully did not integrate with `LayerCompressor` since this will be handled by future data pipelines and doing so would all the `BaseModel` inheritance to the `LayerCompressor` class, which add unnecessary complexity to this PR

## Testing ##
* Added tests in `tests/llmcompressor/modifiers/utils/test_hooks.py`


kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/11881780980/job/33106510578

dsikka (commented):
We briefly looked at the implications of using hooks with FSDP - are we taking care of that already or through this PR?

kylesayrs:
@dsikka I consider that to be out of scope for this PR. I consider FSDP to be unsupported as of now, although this PR makes it easier to support FSDP in the future.

Modifying a module's parameter requires being in special FSDP contexts. 
```python3
@torch.no_grad()
def pre_hook(module, _args):
  # modifying both training and handle training states is required
  with model._use_training_state(TrainingState.IDLE, HandleTrainingState.IDLE):
    with FullyShardedDataParallel.summon_full_params(model):
      # modify module weight. Doing so outside of the contexts will raise a non-contiguous tensor error
      module.weight *= 0
```

We can bake these contexts into the `HooksMixin.register_hook` function, although there's implementation details associated with that I'd like to leave that for a separate task/PR.

src/llmcompressor/modifiers/utils/hooks.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+
+        @wraps(hook)
+        def wrapped_hook(*args, **kwargs):
+            if HooksMixin._HOOKS_DISABLED:
+                return
+
+            return hook(*args, **kwargs)
+
+        handle = getattr(target, f"register_{hook_type}_hook")(wrapped_hook, **kwargs)
```
dsikka, line 71: break into two lines. hard to read

tests/llmcompressor/modifiers/utils/test_hooks.py
```diff
@@ -0,0 +1,81 @@
+import torch
+
+from llmcompressor.modifiers.utils.hooks import HooksMixin
+
+
+class DummyModel(torch.nn.Module):
```
dsikka, line 6: quick docstring would be nice

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -292,27 +284,19 @@ def register_calibration_hooks(self, module: Module):
...
             # hooks for attn modules if running kv_cache quant
             if is_attention_module_:
-                pre_hook_handle = module.register_forward_pre_hook(
-                    calibrate_kv_cache_input_hook(), with_kwargs=True
-                )
-                post_hook_handle = module.register_forward_hook(
-                    calibrate_kv_cache_output_hook()
-                )
+                pre_hook = calibrate_kv_cache_input_hook
+                self.register_hook(module, pre_hook, "forward_pre", with_kwargs=True)
```
dsikka, line 293: Do we need the `pre_hook` variable - we can pass in `calibrate_kv_cache_input_hook` directly?

dsikka (commented):
Overall looks good in cleaning up/unifying hooks

Current testing should test the changes with the `QuantizationModifier`- do we think this is the case for the other modifiers being tested?

The other thought I had was about a less common but potentially useful use case where a modifier may have hooks for different cases and may want to target turning off a specific subset as opposed to all of them - do we think the hooks mixin class can be extended easily to handle that?

src/llmcompressor/modifiers/quantization/quantization/base.py
kylesayrs: That's good as well

kylesayrs:
@dsikka
> Current testing should test the changes with the QuantizationModifier- do we think this is the case for the other modifiers being tested?

I've tested with the e2e tests, although I can perform more rigorous testing if we think that's necessary.

> The other thought I had was about a less common but potentially useful use case where a modifier may have hooks for different cases and may want to target turning off a specific subset as opposed to all of them - do we think the hooks mixin class can be extended easily to handle that?

Yes! There are good arguments to be made for enabling this kind of functionality within the GPTQ algorithm, and unifying hooks makes implementing this functionality much easier.

horheynm (approved):

src/llmcompressor/modifiers/utils/hooks.py
```diff
@@ -0,0 +1,83 @@
...
+    ) -> RemovableHandle:
+        """
+        Registers a hook on a specified module/parameter with the option to disable it
+        with HooksMixin.disable_hooks()
+
+        :param target: the module or parameter on which the hook should be registered
+        :param hook: the hook to register
+        :param hook_type: the type of hook to register corresponding to the
+            `register_{hook_type}_hook` attribute on torch.nn.Module.
+            Ex. "forward", "forward_pre", "full_backward", "state_dict_post", ""
```
dsikka, line 60: Should we make these enums?

src/llmcompressor/modifiers/utils/hooks.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+
+        @wraps(hook)
+        def wrapped_hook(*args, **kwargs):
+            if HooksMixin._HOOKS_DISABLED:
+                return
+
+            return hook(*args, **kwargs)
+
+        register_function = getattr(target, f"register_{hook_type}_hook")
```
dsikka, line 71: do we use the hook_type anywhere else?
Could be used to turn off certain types of hooks for a modifier?

dsikka (commented):
I'd suggest checking out the nightly test cases and making sure we're not running any issues there. LGTM.

dsikka:
oh ignore my nightly comment.

dsikka (approved):

src/llmcompressor/modifiers/utils/hooks.py
kylesayrs: I see no reason to do so, as the error returned from an invalid `hook_type` should be clear enough. Adding an enum just adds another structure to maintain and doesn't add much clarity/performance benefit imho

src/llmcompressor/modifiers/utils/hooks.py
kylesayrs: `hook_type` just exists to create a clear/easy-to-use api
We could turn off certain types of hooks, but I see more of a use case for disabling hooks by modifier or extending this api with hook_metadata and disabling by metadata matching.

dsikka: Wdym by hook metadata?

kylesayrs: We could use metadata schemes in order to identify which hooks to disable in a very extensible way

PR #922, opened 2024-11-17
[1/2] Expand e2e testing to prepare for lm-eval

dsikka:
# Summary
- Update e2e testing to prepare for lm_eval testing. Specifically, refactor the test to run oneshot through a utils function which can also be leveraged by lm_eval
- Each test case now runs in its own process to better manage memory. This is now being managed by a bash script `run_tests.sh`
- Update recipes to be better 
- Save recipes to the HFHub
- Better handle session lifecycle
- Update saving to the hub to save the entire directory

# Running test cases
```bash
bash tests/e2e/vLLM/run_tests.sh
```


tests/e2e/e2e_utils.py
```diff
@@ -0,0 +1,67 @@
...
+            oneshot_kwargs["recipe"] = GPTQModifier(
+                targets="Linear", scheme=scheme, ignore=["lm_head"]
+            )
+        else:
+            oneshot_kwargs["recipe"] = QuantizationModifier(
+                targets="Linear", scheme=scheme, ignore=["lm_head"]
+            )
+
+    # Apply quantization.
+    print("ONESHOT KWARGS", oneshot_kwargs)
```
kylesayrs, line 54: Oneshot already logs arguments?

dsikka: I find those overwhelming and hard to parse through. This is for so to understand test configuration thats running.

horheynm: Could we use logger.debug instead of print?

kylesayrs (approved):

horheynm (approved):

PR #924, opened 2024-11-19
Use llama-1b for faster and more accessible examples

kylesayrs:
## Purpose ##
* Make examples more accessible to users without large GPU resources. Since these are just examples, they should be accessible as possible to encourage easy use and adoption of llm-compressor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Quantization is applied by selecting an algorithm and calling the `oneshot` API.
 from llmcompressor.modifiers.quantization import GPTQModifier
 from llmcompressor.modifiers.smoothquant import SmoothQuantModifier
 from llmcompressor.transformers import oneshot
+from transformers import AutoModelForCausalLM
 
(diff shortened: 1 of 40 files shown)


dsikka (changes requested):
I think we actually want to keep the larger models as most of our questions on vllm are around compressing larger models to run them there. So I dont think we should be making this change as these are an easy reference.

These examples are also used directly in our testing which help identify cases that would otherwise be ignored when dealing with larger memory

horheynm:
I think if the model structure is the same we can use a smaller model. 
for example if the difference from the large and small model is N attention heads only then it should be ok, bc the vllm code execution path pathway will be the same.

If the larger model has different architecture, then we should keep the smaller model, the execution path will be different

kylesayrs:
I agree, it makes sense to keep the 8b models. We already have an example of quantizing a small tiny llama model in the readme.

PR #927, opened 2024-11-19
Fix for e2e tests

horheynm:
SUMMARY:
Issue was caused by trying to save the recipe on modify_save_pretrained. This runs after the session is reset, so no recipe to save. 

With SparseAutoModelForCausalLM before the commit, in the same test, recipe is not saved. It is saved in the save_model_and_recipe, which runs after oneshot. We dont do it here anymore because 1.  if output path is not provided it doesnt save; 2. if we add again, there will be double save

[Notes]
Another way to fix this is to not to reset the session. This allows us to load and save the recipe. pre-commited SparseAutoModelForCausalLM state does not save the recipe.

TEST PLAN:
Checked that the output from oneshot is the same and pass the e2e tests.

### DIFF
--- a/src/llmcompressor/core/session.py
+++ b/src/llmcompressor/core/session.py
@@ -271,7 +271,7 @@ class CompressionSession:
         if recipe is not None and hasattr(recipe, "yaml"):
             return recipe.yaml()
 
-        logger.warning("Recipe not found in session - may been reset")
+        logger.warning("Recipe not found in session - it may have been reset")
 
     def _log_model_info(self):
         # Log model level logs if cadence reached
--- a/src/llmcompressor/core/session.py
+++ b/src/llmcompressor/core/session.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, List, Optional, Union
 
+from loguru import logger
+
 from llmcompressor.core.events import EventType
 from llmcompressor.core.helpers import log_model_info, should_log_model_info
 from llmcompressor.core.lifecycle import CompressionLifecycle
@@ -260,12 +262,16 @@ class CompressionSession:
         self.lifecycle.initialized_ = False
         self.lifecycle.finalized = False
 
-    def get_serialized_recipe(self) -> str:
+    def get_serialized_recipe(self) -> Optional[str]:
         """
         :return: serialized string of the current compiled recipe
         """
         recipe = self.lifecycle.recipe_container.compiled_recipe
-        return recipe.yaml()
+
+        if recipe is not None and hasattr(recipe, "yaml"):
+            return recipe.yaml()
+
+        logger.warning("Recipe not found in session - may been reset")
 
     def _log_model_info(self):
         # Log model level logs if cadence reached
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Quantization is applied by selecting an algorithm and calling the `oneshot` API.
 from llmcompressor.modifiers.quantization import GPTQModifier
 from llmcompressor.modifiers.smoothquant import SmoothQuantModifier
 from llmcompressor.transformers import oneshot
+from transformers import AutoModelForCausalLM
 
 # Select quantization algorithm. In this case, we:
 #   * apply SmoothQuant to make the activations easier to quantize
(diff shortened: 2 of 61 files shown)


dsikka (commented):
Why is the session being reset in these cases and not our example cases? I'm just surprised we didn't see it then

src/llmcompressor/core/session.py
```diff
@@ -260,12 +262,16 @@ def reset_stage(self):
...
         """
         :return: serialized string of the current compiled recipe
         """
         recipe = self.lifecycle.recipe_container.compiled_recipe
-        return recipe.yaml()
+
+        if recipe is not None and hasattr(recipe, "yaml"):
+            return recipe.yaml()
+
+        logger.warning("Recipe not found in session - may been reset")
```
dsikka, line 274: grammar

dsikka: I also think we should be saving the recipe for this case.

horheynm: It doesnt save there bc there is not recipe bc session was reset

horheynm: So the recipe is not saved using SparseAutoModelForCausalLM anyways, Yes i do think we should be saving the recipes but saving/ not saving do not affect the outcome of this test

dsikka: Do you mind updating it in this PR to save it then?

horheynm: Good question, 
the save happens after those lines at the context manager. 
```python3
recipe_path = os.path.join(save_directory, "recipe.yaml")
            session = active_session()
            recipe_yaml_str = session.get_serialized_recipe()
            with open(recipe_path, "w") as fp:
                fp.write(recipe_yaml_str)
```
The problem we have now is that we get an error at `recipe_yaml_str = session.get_serialized_recipe()`. 
Reason is because it assumes that there is an object, which can be serialized into yaml. The object is None in this case because the session is reset. There is nothing there -> None.

dsikka: my question is why it's being reset now and not before to trigger this error

horheynm: Its on the description

```
With SparseAutoModelForCausalLM before the commit, in the same test, recipe is not saved. It is saved in the save_model_and_recipe, which runs after oneshot. We dont do it here anymore because 1. if output path is not provided it doesnt save; 2. if we add again, there will be double save

```

tldr: it follows a differnt save pathway

horheynm: The session has always been reset, it just doesnt save the recipe 
https://github.com/vllm-project/llm-compressor/pull/927/files#diff-a863d6754eda15d0560a67cbe262489d9784b7c0ebf8b1f43cf6d17389b85fc9R135

dsikka (changes requested):
Could we update such that we're saving the recipe now please?
It's informative to have that for the e2e cases.

horheynm:
@dsikka 
The recipe now should be saved.
FYI for this e2e tests, saving recipes do not change the outcome so it wont be informative.

dsikka:
> @dsikka The recipe now should be saved. FYI for this e2e tests, saving recipes do not change the outcome so it wont be informative.

We push the outputs and refer to the frequently. Please look at the complete test

horheynm:
> > @dsikka The recipe now should be saved. FYI for this e2e tests, saving recipes do not change the outcome so it wont be informative.
> 
> We push the outputs and refer to them frequently. Please look at the complete test

If we were referring to it frequently, we should have known recipe wasnt being pushed previously. 
I can add recipe creation check on the test in the follow up pr.

dsikka (approved):

rahul-tuli (approved):

horheynm:
dont merge yet please, not resetting session, so lifecycle is finializing more than once, which raises error.

kylesayrs (approved):

horheynm:
Follow up pr to check the contexts of the files, and to make sure recipes exist in the session is here 

https://github.com/vllm-project/llm-compressor/pull/929

robertgshaw2-neuralmagic (approved):

mgoin (approved):

PR #931, opened 2024-11-25
Kylesayrs/gptq hessian offload

kylesayrs:
Implement GPTQ algorithm using hessian accumulation done on the GPU and offloaded to the CPU


kylesayrs:
Merged into https://github.com/vllm-project/llm-compressor/pull/914

kylesayrs:
Merged into #914

PR #932, opened 2024-11-25
Update pytest command to capture results to file

dbarbuzzi:
This PR updates the recently added `run_tests.sh` script to include flags in the `pytest` command so test results will be captured. It also fixes the improper use of `-s` as well as makes variable usage consistently safer by wrapping in quotes to prevent unexpected globbing/etc. (now fully passes `shellcheck`).


dhuangnm (approved):

dsikka (commented):
Will this output all the info logs?

dbarbuzzi:
> Will this output all the info logs?

Yes, the `tee-sys` value for `--capture` allows pytest to capture all of the output while also mirroring it live to console. If you want live output, this is the value that should be used instead of `-s` in things like CI/CD.

PR #938, opened 2024-11-26
Update text_generation.py

imrankh46:
i avoid this warning.
packages/llmcompressor/transformers/finetune/session_mixin.py:95: FutureWarning: `tokenizer` is deprecated and will be removed in version 5.0.0 for `Trainer.__init__`. Use `processing_class` instead.

SUMMARY:
"replace tokenizer by processing_class in trainer"


kylesayrs:
Thank you very much for the contribution @imrankh46 !
This should be followed up with a more complete genericization of tokenizer -> processor, but this is a good start to get rid of the pesky warning.

Could you please take a look at signing the DCO, otherwise looks good to me!

kylesayrs:
Handled by #955

PR #945, opened 2024-11-29
Add LM Eval Testing

dsikka:
# Summary:
- Add LM Eval Teseting to run on a `weekly` cadence
- Model is quantized after which lm_eval runs and the metrics produced are asserted against metrics defined in the config for the particular test case
- Updates the bash script to take in configs folder and test script, allowing it to work for both the vLLM e2e tests and LM Eval tests

# Testing
- Testing has to be updated to use the following commands:
```bash
bash tests/e2e/vLLM/run_tests.sh -c tests/e2e/vLLM/lm_eval_configs -t tests/e2e/vLLM/test_lmeval.py 
bash tests/e2e/vLLM/run_tests.sh -c tests/e2e/vLLM/configs -t tests/e2e/vLLM/test_vllm.py 
```

- LM Eval testing requires `CADENCE=weekly`


setup.py
```diff
@@ -71,6 +71,7 @@
             "pytest-mock>=3.6.0",
             "pytest-rerunfailures>=13.0",
             "parameterized",
+            "lm_eval",
```
robertgshaw2-neuralmagic, line 74: I would pin a specific version

tests/e2e/vLLM/test_lmeval.py
```diff
@@ -0,0 +1,131 @@
...
+        recipe_yaml_str = session.get_serialized_recipe()
+        with open(recipe_path, "w") as fp:
+            fp.write(recipe_yaml_str)
+        session.reset()
+
+        logger.info("================= Running LM Eval ======================")
+
+        model_args = f"pretrained={self.save_dir}"
+        results = lm_eval.simple_evaluate(
+            model="hf",
```
robertgshaw2-neuralmagic, line 109: this will run the model using transformers, not vllm.

tests/e2e/vLLM/test_lmeval.py
```diff
@@ -0,0 +1,131 @@
...
+            model_args=model_args,
+            tasks=[self.task],
+            num_fewshot=self.num_fewshot,
+            limit=self.limit,
+            device="cuda:0",
+            batch_size=100,
```
robertgshaw2-neuralmagic, line 115: with vllm, use `batch_size="auto"`

tests/e2e/vLLM/test_lmeval.py
dsikka: yes. that is the goal. vllm lm_eval is already covered through vllm testing

tests/e2e/vLLM/test_lmeval.py
dsikka: meant for hf

tests/e2e/vLLM/run_tests.sh
```diff
@@ -2,10 +2,23 @@
...
+while getopts "c:t:" OPT; do
+  case ${OPT} in
+    c ) 
+        CONFIG="$OPTARG"
+        ;;
+    t )
+        TEST="$OPTARG"
+        ;;
+    \? )
+        usage
```
dbarbuzzi, line 14: The `usage` function is not defined anywhere, so this attempt to call it will result in an error that will mask the real error.

dhuangnm: Same as what Domenic mentioned above. I'll update the test workflow to include the lm eval tests once this PR is landed.

tests/e2e/vLLM/test_lmeval.py
robertgshaw2-neuralmagic: my b, I was confused because this is in the `tests/e2e/vLLM` directory

dsikka: good point

tests/e2e/vLLM/run_tests.sh
dsikka: Will update

dsikka:
- To add:
- w4 + act order
- int8 + smoothquant

tests/e2e/vLLM/test_lmeval.py
```diff
@@ -0,0 +1,131 @@
...
+TEST_DATA_FILE = os.environ.get("TEST_DATA_FILE", None)
+
+
+# Will run each test case in its own process through run_tests.sh
+# emulating vLLM CI testing
+@requires_gpu_count(1)
+@pytest.mark.skipif(
+    not lm_eval_installed, reason="lm eval is not installed, skipping test"
+)
+class TestLMEval:
```
horheynm, line 31: Do you not need to inherit unittest here?

horheynm:
one ocmment otherwise lgtm

tests/e2e/vLLM/test_lmeval.py
dsikka: No, using pyest

rahul-tuli: We don't need to since pytest can pick up both!

rahul-tuli (approved):
LGTM pending Domenic's comment about "usage" function in the bash script

tests/e2e/vLLM/run_tests.sh
dhuangnm: The code seems not updated yet to include the usage function definition?

kylesayrs (dismissed):

PR #947, opened 2024-12-02
Record config file name as test suite property

dbarbuzzi:
SUMMARY:
This PR updates the e2e tests to record the config file used as a property on the testsuite which is written to the JUnit XML report file, which gets used downstream to uniquely identify the tests so they can be tracked over time (currently for elapsed time).

I also added an extra skip-guard to the `test_vllm` test around the required environment variable and updated the `pytest` command usage to include “reasons” for all tests that don’t pass. This means that skip reasons/short error messages/etc. will all be logged at the end of the test results (it does not affect anything else).

TEST PLAN:
- [x] Tested locally by running `CADENCE=nightly bash tests/e2e/vLLM/run_tests.sh` to verify all things pass and that the test reports contain the expected new data.


tests/e2e/vLLM/test_vllm.py
```diff
@@ -20,12 +21,19 @@
...
+@pytest.fixture
+def record_config_file(record_testsuite_property: Callable[[str, object], None]):
+    test_data_file_name = TEST_DATA_FILE.split("configs/")[-1]
+    record_testsuite_property("TEST_DATA_FILE_NAME", test_data_file_name)
 
 
 # Will run each test case in its own process through run_tests.sh
 # emulating vLLM CI testing
 @requires_gpu_count(1)
+@pytest.mark.skipif(not TEST_DATA_FILE, reason="must set TEST_DATA_FILE env var")
```
dsikka, line 36: I don't think this should be a reason we skip a test case. we should probably just throw a warning or error out

dbarbuzzi: Generally speaking, it’s the same general idea as the vllm-based skip – both check if an environmental factor that is required for the test is present.

That said, I’ve removed the new skip entirely.

rahul-tuli (approved):
LGTM!

horheynm (approved):

PR #955, opened 2024-12-05
Replace tokenizer with processor

kylesayrs:
## Purpose ##
* Prepare to support processors and vision datasets
* It's important to rename and retype variable to better reflect its more widened definition

## Prerequisites ##
* #950

## Postrequisites ##
* #943

## Changes ##
* Rename and retype instances of tokenizer to processor
* Add `processor` pathway argument to which `tokenizer` is internally reassigned to
* Add typing definitions in `src/llmcompressor/typing.py`
* Special handling of tokenizer in `src/llmcompressor/transformers/finetune/data/base.py`, `src/llmcompressor/transformers/finetune/data/ultrachat_200k.py`, `src/llmcompressor/transformers/finetune/session_mixin.py`

## Testing ##
* No new functionality is added, CI tests should pass


src/llmcompressor/typing.py
```diff
@@ -0,0 +1,17 @@
+from typing import Union
```
rahul-tuli, line 1: Nice!

rahul-tuli (approved):
LGTM! Thanks for this.

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -56,19 +56,19 @@ def __init__(
...
+    def populate_datasets(self, processor: Processor, add_labels: bool = True):
         """
         Loads datasets for each flow based on data_args, stores a Dataset for each
         enabled flow in self.datasets
 
         :param tokenizer: tokenizer to use for dataset tokenization
         """
         if self._data_args.dataset is None:
-            self.tokenizer = self._model_args.tokenizer
+            self.processor = self._model_args.processor
```
dsikka, line 71: Seems like we're keeping the tokenizer in the model_args as well? What if both are specified? Or only tokenizer?

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -226,19 +235,21 @@ def initialize_model_from_path(
...
 
-def initialize_tokenizer_from_path(model_args, model, teacher):
-    tokenizer_src = model_args.tokenizer
-    tokenizer_src = tokenizer_src or get_shared_tokenizer_src(model, teacher)
-    tokenizer = AutoTokenizer.from_pretrained(
-        tokenizer_src,
+def initialize_processor_from_path(
+    model_args: ModelArguments, model: PreTrainedModel, teacher: PreTrainedModel
+) -> Processor:
+    processor_src = model_args.processor
```
dsikka, line 241: same, what if a tokenizer is provided?

dsikka (changes requested):
I think we need to make it very clear:

1) What a processor is vs a tokenizer
2) If either/or can be provided and in what cases

src/llmcompressor/transformers/finetune/runner.py
kylesayrs: See the newly added [model args handling logic](https://github.com/vllm-project/llm-compressor/pull/955/files#diff-4b4df7cb1f6042ff11af61f87b5d6dffd8e07a7ef182a8aa659566a5ee77fc04R139-R145)

src/llmcompressor/transformers/finetune/text_generation.py
kylesayrs: See the newly added [model args handling logic](https://github.com/vllm-project/llm-compressor/pull/955/files#diff-4b4df7cb1f6042ff11af61f87b5d6dffd8e07a7ef182a8aa659566a5ee77fc04R139-R145)

kylesayrs:
@dsikka The current strategy is to treat all possible tokenizers as a subset of all possible processors, as type-defed here
```python3
Processor = Union[
    PreTrainedTokenizer, BaseImageProcessor, FeatureExtractionMixin, ProcessorMixin
]
```

We should continue to support the `tokenizer` model arg, but internally reassign it to the `processor` variable name for code simplicity.

dsikka:
> @dsikka The current strategy is to treat all possible tokenizers as a subset of all possible processors, as type-defed here
> 
> ```python
> Processor = Union[
>     PreTrainedTokenizer, BaseImageProcessor, FeatureExtractionMixin, ProcessorMixin
> ]
> ```
> 
> We should continue to support the `tokenizer` model arg, but internally reassign it to the `processor` variable name for code simplicity.
> 
> ```python
> # silently assign tokenizer to processor
> if model_args.tokenizer:
>     if model_args.processor:
>         raise ValueError("Cannot use both a tokenizer and processor")
>     model_args.processor = model_args.tokenizer
> model_args.tokenizer = None
> ```

I think this is fine. My two comments about clarity were specific to being clear towards users - either in the model_args or through text_generation.py script

kylesayrs:
@dsikka
1. There is help text attached to the newly added `processor` arg which users can read
2. We throw an error if both are passed

I think this should be clear enough messaging without being annoying/verbose

dsikka:
> @dsikka
> 
> 1. There is help text attached to the newly added `processor` arg which users can read
> 2. We throw an error if both are passed
> 
> I think this should be clear enough messaging without being annoying/verbose

Oh sorry, missed the help text.
Sounds good

dsikka (approved):

PR #960, opened 2024-12-05
Fail OneShot then Finetune test artificially

dsikka:
# Summary:
- With thew new transformers release, we can now run training and serialize quantized models (yay!) through AutoModel
- However, this artificially passes our OneShot then Finetune test because as we have to be able to decompress the model before finetune
- Fail until this is enabled in transformers


horheynm:
I don't think we should purposely fail tests - 
Instead we should add tests here to check if the model was decompressed

dsikka:
> I don't think we should purposely fail tests - Instead we should add tests here to check if the model was decompressed

I more so wanted to throw an exception so that we dont forget to come back to this

rahul-tuli (approved):
Yay!

PR #970, opened 2024-12-11
[Test Fix] Fix/update test_run_compressed

horheynm:
~~Contingent on merge of https://github.com/huggingface/transformers/pull/34719~~
^ has been merged not yet released

SUMMARY:
Update run_compressed tests from decompression tests to run_comrpressed tests -> test if run_compressed True/False models generate the same output

Add decompress tests that copies attrs from the source dir path's model to the target model. 

TEST PLAN:
ran the test using transformers main
must pass tests/llmcompressor/transformers/compression/test_decompress.py 
and tests/llmcompressor/transformers/compression/test_run_compressed.py


dsikka:
The generations compared are decompressed and decompressed on the forward pass. So they should be close within a tolerance

dsikka (changes requested):
I think you've changed the goal of the test which is compressed linear vs decompressing the whole model. We still want run_compressed to be  True for one of the models

horheynm:
> I think you've changed the goal of the test which is compressed linear vs decompressing the whole model. We still want run_compressed to be True for one of the models

Honestly, idk what the original test was doing. If you want to add compressed linear, decompressed model and its model.generate thats cool. 

original test is comparing a quantized model and a non-touched base model, that will always fail

dsikka:
> > I think you've changed the goal of the test which is compressed linear vs decompressing the whole model. We still want run_compressed to be True for one of the models
> 
> Honestly, idk what the original test was doing. If you want to add compressed linear, decompressed model and its model.generate thats cool.
> 
> original test is comparing a quantized model and a non-touched base model, that will always fail

1. We load a compressed model: 

https://github.com/vllm-project/llm-compressor/blob/606aab2b57eb0931f93976f6011f2b2869ebdc60/tests/llmcompressor/transformers/compression/test_run_compressed.py#L29

2. We create an empty model to hold the decompressed weights: 

https://github.com/vllm-project/llm-compressor/blob/606aab2b57eb0931f93976f6011f2b2869ebdc60/tests/llmcompressor/transformers/compression/test_run_compressed.py#L37

3) We then use lines 42-50 to decompress the whole model, using the empty model as the skeleton. This is loading the same checkpoint as what is used by the compressed model.
https://github.com/vllm-project/llm-compressor/blob/606aab2b57eb0931f93976f6011f2b2869ebdc60/tests/llmcompressor/transformers/compression/test_run_compressed.py#L42

We then run generations and compare the two cases. You're comparing decompressed in both cases, one is just decompressing on the forward pass.  It's the same checkpoint in both cases. 

We just want to update such that the uncompressed model can be decompressed using your new transformer changes.

Thi

horheynm:
Ok i see, we should update the test name, test_run_compressed is confusing

dsikka (changes requested):
`test_run_compressed` is not testing the `run_compressed` arg which is the point of the test.

the input to both loads should be the same compressed model stub, with one of them being given the config to decompress the entire model as input

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,115 @@
...
+            self.dense_model.model.layers[0].self_attn.q_proj, "weight_scale"
+        )
+
+        self.decompressed_model = None
+        config = AutoConfig.from_pretrained(self.compressed_model_stub)
+
+        compression_config = getattr(config, QUANTIZATION_CONFIG_NAME, None)
+        self.compressor = ModelCompressor.from_compression_config(compression_config)
+        self.compressor.quantization_config.quantization_status = (
+            QuantizationStatus.FROZEN
```
dsikka, line 64: instead of doing this manual decompression, we should be using the AutoModel pathway to decompress?

dsikka (changes requested):

tests/llmcompressor/transformers/compression/test_decompress.py
horheynm: Yeah thats more appropriate

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@requires_gpu
+@parameterized_class(parse_params(CONFIG_DIR))
+class TestQuantizationMatches(unittest.TestCase):
+    """
+    Test decompression - given a skeleton model and path to the optimized model,
+    write the optimized model's safetensors to the skeleton model and decompress
+    Ex. write weight_scale to skeleton model and then fp4 -> fp16
+
```
dsikka, line 26: this docstring isn't very clear as to what is being tested.

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+
+        # use the model_path to load the decompressed weights into dense_model
+        dense_model = copy.deepcopy(self.dense_model)
+
+        # overwrite the weights of the dense model
+        self.compressor.decompress(
+            model_path=self.compressed_model_stub,
+            model=self.dense_model,
+        )
```
dsikka, line 85: we shouldn't need the code from lines 70-85. Scope of the test should be focused on the run_compressed vs decompressing everything upfront

dsikka (changes requested):

tests/llmcompressor/transformers/compression/test_decompress.py
horheynm: Yeah so thats happening. Not sure if you had a chance to look below but the test check thats happening is below here. 27-28. I can put these on top for better visibility

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # self.dense_model should be decompressed
+        assert dense_model is not self.dense_model
+
+        self.decompressed_model_manual = self.dense_model
```
dsikka, line 86: why not just use self.dense_model in the test?

dsikka (approved):
these look great.
great work!

tests/llmcompressor/transformers/compression/test_decompress.py
horheynm: Oh I can but its decompressed so i just want a name change. 
If we hit memory limitation I can use dense_model

horheynm: ^ Actually models are pass by ref so should be ok

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+
+        # Manually decompress this model
+        self.dense_model = AutoModelForCausalLM.from_pretrained(
+            self.skeleton_model_stub,
+            torch_dtype=self.decompressed_model_hf_quantizer.dtype,
+            device_map=self.decompressed_model_hf_quantizer.device,
+        )
+
+        assert not hasattr(
```
kylesayrs, line 62: Should we not also check `decompressed_model_hf_quantizer`?

mgoin (approved):

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.decompressed_model_manual = self.dense_model
+
+        assert hasattr(
+            self.decompressed_model_manual.model.layers[0].self_attn.q_proj,
+            "weight_scale",
+        )
+
+    def test_hf_quantizer_decompress_match_manual_decompress(self):
+        decompressed_model_manual = self.decompressed_model_manual.device
```
kylesayrs, line 94: nit: `decompressed_model_manual` -> `manual_device`

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,128 @@
...
+            decompressed_model_hf_quantizer
+        )
+
+        for input in self.SAMPLE_INPUTS:
+            inputs = self.tokenizer(input, return_tensors="pt", padding=True).to(
+                self.decompressed_model_manual.device
+            )
+            inputs = inputs.to(self.decompressed_model_manual.device)
+
+            decompressed_model_manual_output = self.tokenizer.batch_decode(
```
kylesayrs, line 110: nit: we don't actually have to decode the outputs. We can just compare the outputs directly

tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -0,0 +1,128 @@
...
+                decompressed_model_hf_quantizer_out == decompressed_model_manual_output
+            )
+
+    @classmethod
+    def tearDownClass(self):
+        shutil.rmtree(self.test_dir)
+        del self.dense_model
+        del self.decompressed_model_hf_quantizer
+        del self.decompressed_model_manual
+        torch.cuda.empty_cache()
```
kylesayrs, line 128: Please remove

tests/llmcompressor/transformers/compression/test_run_compressed.py
```diff
@@ -3,77 +3,187 @@
...
-        cls.compressor.quantization_config.quantization_status = (
-            QuantizationStatus.FROZEN
-        )
-        cls.compressor.decompress(
-            model_path=cls.model_stub, model=cls.uncompressed_model
+            cls.uncompressed_model_stub,
+            torch_dtype=cls.decompressed_model.dtype,
+            device_map=cls.decompressed_model.device,
         )
+        breakpoint()
```
kylesayrs, line 58: Please remove

tests/llmcompressor/transformers/compression/test_run_compressed.py
```diff
@@ -3,77 +3,187 @@
...
+        self.uncompressed_model = self.uncompressed_model.to(uncompressed_device)
+
         inputs = self.tokenizer(SAMPLE_INPUT, return_tensors="pt", padding=True).to(
-            self.compressed_model.device
+            decompressed_device
         )
-        compressed_output = self.tokenizer.batch_decode(
-            self.compressed_model.generate(**inputs, max_length=50)
+
+        decompressed_output = self.tokenizer.batch_decode(
```
kylesayrs, line 80: Again, do not need to decode, only compare outputs

tests/llmcompressor/transformers/compression/test_run_compressed.py
```diff
@@ -3,77 +3,187 @@
...
         SAMPLE_INPUT = [
             "I love 4-bit quantization because",
             "What is the capital of France?",
             "def fibonacci(n):",
         ]
 
+        decompressed_device = self.decompressed_model.device
+        uncompressed_device = self.uncompressed_model.device
+
+        # overwrite weights in cpu to cuda
```
kylesayrs, line 72: Why is this needed? It seems like you're already controlling which device the models live on when you load them.

kylesayrs (changes requested):

tests/llmcompressor/transformers/compression/test_run_compressed.py
dsikka: this is fine to keep.

horheynm: Kyle you are correct. I want to keep it to check if the generations make sense. If it takes up too much run time or memory happy to remove

tests/llmcompressor/transformers/compression/test_decompress.py
horheynm: Sure, dense model wont have it so a check for that

tests/llmcompressor/transformers/compression/test_run_compressed.py
horheynm: At this point in the code, although set to cuda:N, I was getting an error of some layers not on the same device. I added that to override that.

PR #971, opened 2024-12-11
[Test Fix] Fix Consecutive oneshot

horheynm:
~~Contingent on merge of https://github.com/huggingface/transformers/pull/34719~~
~~ ^ has been merged not yet released ~~
^ has been released

Blocked on 
https://github.com/neuralmagic/compressed-tensors/pull/237

SUMMARY:
* In multiple optimization tests, automatically decompress model if provided as optimized model
* Fix recipe stage length
* Revive old code
* When running multiple optimizations (ex. oneshot then finetune, oneshot and oneshot), the recipes needs to be added to the session using `initialize_recipe`. Example here https://github.com/vllm-project/llm-compressor/pull/971/files#diff-c9ae8b3ad24d13abeea5b649a5fd6d0b0925f5c9cc40220cbfbe21ae81242f8dR63-R65

TEST PLAN:
ran the test using transformers main
Must pass tests/llmcompressor/transformers/obcq/test_consecutive_runs.py


src/llmcompressor/transformers/utils/helpers.py
```diff
@@ -54,3 +59,223 @@ def detect_last_checkpoint(
             )
 
     return last_checkpoint
+
+
+def is_model_quantized_from_path(path: str):
```
dsikka, line 64: this seems redundant with the method we have in text_generation?

dsikka (changes requested):

src/llmcompressor/transformers/utils/helpers.py
```diff
@@ -54,3 +59,223 @@ def detect_last_checkpoint(
...
+
+def recipe_from_huggingface_model_id(
+    model_path: str, RECIPE_FILE_NAME: str = RECIPE_FILE_NAME
+) -> Tuple[Optional[str], bool]:
+    """
+    Attempts to download the recipe from the huggingface model id.
+
+    :param model_path: Assumed to be the huggingface model id.
+        If it is not, this function will return None.
+    :param RECIPE_FILE_NAME: The name of the recipe file to download.
```
rahul-tuli, line 176: Why is this in UPPERCASE?

src/llmcompressor/transformers/utils/helpers.py
```diff
@@ -54,3 +59,223 @@ def detect_last_checkpoint(
...
+    return False
+
+
+def resolve_recipe(
+    model_path: Union[str, Path],
+    recipe: Union[str, Path, None] = None,
+) -> Union[str, None]:
+    """
+    Resolve the recipe to apply to the model.
+    :param recipe: the recipe to apply to the model.
```
rahul-tuli, line 81: nit: new line b/w description and params

rahul-tuli (commented):
Looks fine, I feel we can make a few functions cleaner by using pathlib. But I'll leave that upto you

src/llmcompressor/transformers/utils/helpers.py
horheynm: sorry which method? 
is_model_quantized_from_path is defined here, text_generation.py uses it from here

src/llmcompressor/transformers/utils/helpers.py
horheynm: ok

src/llmcompressor/transformers/utils/helpers.py
```diff
@@ -54,3 +59,224 @@ def detect_last_checkpoint(
             )
 
     return last_checkpoint
+
+
+def is_model_quantized_from_path(path: str):
+    """
+    Determine if model is quantized based on the config
```
rahul-tuli, line 66: docstring with params and return type?

rahul-tuli (approved):

src/llmcompressor/transformers/utils/helpers.py
horheynm: Ok

rahul-tuli (dismissed):
LGTM!

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -224,6 +228,13 @@ def initialize_model_from_path(
         "trust_remote_code": model_args.trust_remote_code_model,
     }
     # this calls from_pretrained under the hood so should be FSDP safe
+
+    # optimized models must be decompressed to carry out oneshot/train/etc
+    if is_model_quantized_from_path(model_path):
```
kylesayrs, line 233: Is this helper `is_model_quantized_from_path` really necessary? What happens if you attempt to load a non-quantized model with `quantization_config=CompressedTensorsConfig(run_compressed=False)`?

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -5,14 +5,19 @@
...
 
+from llmcompressor.transformers.utils.helpers import resolve_recipe
 from tests.testing_utils import parse_params, requires_gpu
 
 CONFIGS_DIRECTORY = "tests/llmcompressor/transformers/obcq/obcq_configs/consec_runs"
 GPU_CONFIGS_DIRECTORY = (
     "tests/llmcompressor/transformers/obcq/obcq_configs/consec_runs/gpu"
 )
 
+quantization_config = CompressedTensorsConfig(run_compressed=False)
```
kylesayrs, line 19: nit: consider moving into class, similar to other tests

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -36,19 +41,29 @@ def _test_consecutive_runs(
             oneshot_device=self.device,
             clear_sparse_session=False,
         )
-        first_tiny_model = get_session_model()
+
+        first_model = AutoModelForCausalLM.from_pretrained(
```
kylesayrs, line 45: I think the purpose of this test to test consecutive runs *without* model reloading.

kylesayrs (changes requested):
I believe the purpose of this test to test consecutive runs *without* model reloading.

There are quite a few helper functions which I'm not entirely convinced are needed as of now. I think it might be worth revisiting whether they're actually needed for these tests

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
dsikka: The test currently loads the first oneshot model from disk when running the second oneshot run so the focus of this particular case is consecutive one shot runs after first loading from disk

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: Good question, we do have is_model_quantzied in CT, but inputs a module. Here you check the config without loading. 
If you load a non-quantized model with the config, then It would error out -- but here on load it only adds the quant_config if its quantized. Its not a case we should worry about.

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
horheynm: We can do without reloading, but the original test on sparseml (if i remember correctly) was to load a saved model. As Dipika said above, we couldnt do that, bc HFQuantizer takes care of loading logic. 

The use case was that we optimize, save. Then we load from disk and optimize.

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
horheynm: ok

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
kylesayrs: I misread the code

kylesayrs (dismissed):

src/llmcompressor/transformers/utils/helpers.py
```diff
@@ -54,3 +59,101 @@ def detect_last_checkpoint(
...
+def is_model_quantized_from_path(path: str) -> bool:
+    """
+    Determine if model from path is quantized based
+    on the config
+
+    :param path: path to the model or HF stub
+    :return: True if config contains quantization_config from the given path
+
+    """
+    config = AutoConfig.from_pretrained(path)
```
dsikka, line 73: nit: check if also compressed-tensors as all quant models will have this quantization_config

src/llmcompressor/transformers/utils/helpers.py
```diff
@@ -54,3 +59,101 @@ def detect_last_checkpoint(
...
+    Attempts to download the recipe from the Hugging Face model ID.
+
+    :param model_path: Assumed to be the Hugging Face model ID.
+    :param recipe_file_name: The name of the recipe file to download.
+     Defaults to RECIPE_FILE_NAME.
+    :return: A tuple:
+        - The path to the recipe file if found, None otherwise.
+        - True if model_path is a valid Hugging Face model ID, False otherwise.
+    """
+    model_id_url = os.path.join(HUGGINGFACE_CO_URL_HOME, model_path)
```
dsikka, line 127: can we target just downloading the recipe?

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -36,19 +41,29 @@ def _test_consecutive_runs(
...
-        assert qat_active(first_tiny_model)
+        assert qat_active(first_model)
 
         session = active_session()
         session_recipe = session.lifecycle.recipe_container.compiled_recipe
         stages = [stage.group for stage in session_recipe.stages]
         self.assertEqual(len(stages), 1)
         session.reset()
 
+        recipe = infer_recipe_from_model_path(model_path=self.output_first)
```
dsikka, line 63: Can you explain why we need this first recipe?
Seems like all `initialize_recipe` does is create the session and the pre_initialize_ method uses the recipe path, but is that doing anything here?

dsikka (changes requested):

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -36,19 +41,29 @@ def _test_consecutive_runs(
...
+        if recipe:
+            initialize_recipe(model=first_model, recipe_path=recipe)
+
         # reload saved model and up sparsity to 0.7
         oneshot(
```
dsikka, line 68: couldn't we just wrap  this with `with create_session()?`

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
horheynm: Yeah so we reset the session - if no reset the modifiers are finalized.
After reset, the recipe container is reset. We have a check at line 78 on this diff that checks the length of the applied recipes. 
Its not doing anthing else

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
horheynm: Yeah we can

kylesayrs (dismissed):

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -15,69 +19,88 @@
...
+            self.assertEqual(len(stages), 1)
+            # session.reset()
+
+        with create_session():
+            recipe = infer_recipe_from_model_path(model_path=self.output_first)
+            if recipe:
+                initialize_recipe(model=first_model, recipe_path=recipe)
+
+            # reload saved model and up sparsity to 0.7
+            oneshot(
```
dsikka, line 71: I am confused as to what we're doing here: 

1) We're creating a new session in `initialize_recipe` - my comment about using `with create_session()` was potentially about replacing `initialize_recipe`, not using both. Generally speaking, you dont need to wrap all the code, just the oneshot call

2) Why do we have `clear_sparse_session` set to False, while also creating a new session? Seems contradictory

src/llmcompressor/transformers/utils/helpers.py
horheynm: yes it does

src/llmcompressor/transformers/utils/helpers.py
horheynm: done!

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
dsikka: Two sessions are fine but they just be wrapping the oneshot call, not everything esle. I dont get the point of adding the `initialize_recipe` call though, if we're starting a new recipe/model for the second oneshot call.

horheynm: I see, I understand the comment. 
We run oneshot twice, so on save the change should be reflected in the recipe. 
After the first oneshot we get in `oneshot_output/test_2/recipe.yaml`
```yaml
test_stage_0:
  obcq_modifiers:
    SmoothQuantModifier:
      smoothing_strength: 0.5
      mappings:
      - - ['re:.*q_proj', 're:.*k_proj', 're:.*v_proj']
        - re:.*input_layernorm
      - - ['re:.*gate_proj', 're:.*up_proj']
        - re:.*post_attention_layernorm
    QuantizationModifier:
      ignore: [lm_head]
      config_groups:
        group_0:
          weights: {num_bits: 8}
          targets: [Linear]
    GPTQModifier: {block_size: 128}
    SparseGPTModifier:
      sparsity: 0.5
      block_size: 128
      sequential_update: false
      percdamp: 0.01
      mask_structure: 0:0
      targets: [model.layers.0]

```

and after with the current changes on the PR, we get 
```yaml
test_stage_0:
  obcq_modifiers:
    SmoothQuantModifier:
      smoothing_strength: 0.5
      mappings:
      - - ['re:.*q_proj', 're:.*k_proj', 're:.*v_proj']
        - re:.*input_layernorm
      - - ['re:.*gate_proj', 're:.*up_proj']
        - re:.*post_attention_layernorm
    QuantizationModifier:
      ignore: [lm_head]
      config_groups:
        group_0:
          weights: {num_bits: 8}
          targets: [Linear]
    GPTQModifier: {block_size: 128}
    SparseGPTModifier:
      sparsity: 0.5
      block_size: 128
      sequential_update: false
      percdamp: 0.01
      mask_structure: 0:0
      targets: [model.layers.0]
test_stage_1:
  obcq_modifiers:
    SparseGPTModifier:
      sparsity: 0.7
      block_size: 128
      sequential_update: true
      percdamp: 0.01
      mask_structure: 0:0
      targets: [model.layers.0]
      preserve_sparsity_mask: true

```

The goal of `initialize_recipe` is to add all the recipes that has been applied to the model on save. Without it, it will only have `test_stage_1`, not `test_stage_0`

dsikka: I see. Makes sense. Thanks for the explanation. Can you actually add a quick note to the test to with the recipe example?

And then update to remove the `with create_session()`

horheynm: Yup!

PR #978, opened 2024-12-15
Fix SmoothQuant offload bug

dsikka:
SUMMARY:
- Offload/onload properly during SmoothQuant calibration

We may actually not need this change?
The issue is because of the following line:
```python
scales = activation_scales.pow(self.smoothing_strength) / weight_scales.pow(
            1 - self.smoothing_strength
        )
```
i.e ```y/x```

However, we use `y.div_(x)` when actually applying the smoothing scales, which does it in place and does not cause an issue.

Not 100% sure why one is ok and one is not


src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -329,4 +337,9 @@ def _calculate_smoothing_scales(
             1 - self.smoothing_strength
         )
         scales = torch.where(weight_scales > 0.0, scales, activation_scales)
+
+        for layer in balance_layers:
```
kylesayrs, line 341: Why can't the layer be re-offloaded on line 330? As currently written, there is a time where `balance_layers` are onloaded, which may lead to oom issues. The layer only needs to be onloaded while `layer.weight` is being used to calculate `weight_scales`

kylesayrs (changes requested):

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -102,6 +104,7 @@ class SmoothQuantModifier(Modifier):
     resolved_mappings_: Optional[List] = None
     scales_: Optional[Dict] = None
 
+    @log_time
```
kylesayrs, line 107: todo

src/llmcompressor/modifiers/smoothquant/base.py
dsikka: It can. An open question I have is if we need to offload. See PR description

kylesayrs:
@dsikka Apparently pytorch does not throw errors for inplace operations on meta-device tensors. 

```python3
a = torch.rand(10, device="meta")
# tensor(..., device='meta', size=(10,))

a.div_(6)
# no error
# tensor(..., device='meta', size=(10,))
```

Since the `_apply_smoothing` doesn't take this into account, it's likely that this function silently fails on offloaded weights

src/llmcompressor/modifiers/smoothquant/base.py
kylesayrs: https://github.com/vllm-project/llm-compressor/pull/978#issuecomment-2546666709

dsikka:
> @dsikka Apparently pytorch does not throw errors for inplace operations on meta-device tensors.
> 
> ```python
> a = torch.rand(10, device="meta")
> # tensor(..., device='meta', size=(10,))
> 
> a.div_(6)
> # no error
> # tensor(..., device='meta', size=(10,))
> ```
> 
> Since the `_apply_smoothing` doesn't take this into account, it's likely that this function silently fails on offloaded weights

Yeah same behaviour I saw.

rahul-tuli (approved):
LGTM

PR #998, opened 2024-12-19
GPTQ Vision Utils

kylesayrs:
## Purpose ##
* Add utils used by #914

## Changes ##
* Add `DisableQuantization`
* Add `calibration_forward_context`
* Add `CompressionLogger`

## Testing ##
* Added tests in `tests/llmcompressor/utils/test_helpers.py`

### DIFF
--- a/src/llmcompressor/utils/helpers.py
+++ b/src/llmcompressor/utils/helpers.py
@@ -4,6 +4,7 @@ Common functions for interfacing with python primitives and directories/files.
 """
 
 import ast
+import contextlib
 import errno
 import fnmatch
 import glob
@@ -23,6 +24,7 @@ from urllib.parse import urlparse
 
 import numpy
 import torch
+from compressed_tensors.quantization import disable_quantization, enable_quantization
 from loguru import logger
 
 __all__ = [
@@ -61,6 +63,8 @@ __all__ = [
     "import_from_path",
     "getattr_chain",
     "DisableKVCache",
+    "DisableQuantization",
+    "calibration_forward_context",
 ]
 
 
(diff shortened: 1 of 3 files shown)


src/llmcompressor/utils/metric_logging.py
```diff
@@ -51,3 +53,50 @@ def get_layer_size_mb(module: Module) -> float:
     total_size_mb = total_size / (1e6)  # Convert bytes to MB
 
     return total_size_mb
+
+
+class CompressionLogger:
```
dsikka, line 58: Why is this required given the current logging that already exists in gptq?

kylesayrs: Breaking this out means more real estate is dedicated to describing the algorithm, not logging

src/llmcompressor/utils/helpers.py
```diff
@@ -1080,3 +1084,32 @@ def __enter__(self):
...
+
+    - Remove gradient calculations
+    - Disable the KV cache
+    - Disable quantization from QuantizationModifier
+    """
+    model.eval()
+
+    with (
+        torch.no_grad(),
+        DisableKVCache(model),
```
dsikka, line 1112: do we actually disable the kv cache, anywhere outside of gptq?

kylesayrs: No, but technically we should be disabling the kv cache for all calibration forward passes (pipelines), as calibration only consists of prefill and populating the kvcache only costs performance and memory.

dsikka: Yeah I was gonna suggest adding this to SparseGPTQ but can do later

dsikka (approved):

rahul-tuli (approved):

PR #1005, opened 2024-12-20
Fix `test_model_reload` test

kylesayrs:
## Purpose ##
* Fix failing `test_model_reload` resulting from slightly different behavior of `update_offload_parameter`

## Changes ##
* With the newest changes to `update_offload_parameter`, the new parameter data is copied into the existing tensor rather than replacing it. This breaks its use case for `patch_tied_tensors_bug`, which uses the function to create new tensors which are untied.

## Testing ##
* Previously failing tests in `tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py` now pass


rahul-tuli (approved):

dsikka (approved):
awesome

PR #1012, opened 2024-12-23
[Bugfix] Pass `trust_remote_code_model=True` for deepseek examples

dsikka:
SUMMARY:
- Pass `trust_remote_code_model=True` for deepseek models
- Needed due to the slight differences in how tokenizer and processor pull down their relevant configs

### DIFF
--- a/examples/quantizing_moe/deepseek_moe_w4a16.py
+++ b/examples/quantizing_moe/deepseek_moe_w4a16.py
@@ -75,6 +75,7 @@ oneshot(
     max_seq_length=MAX_SEQUENCE_LENGTH,
     num_calibration_samples=NUM_CALIBRATION_SAMPLES,
     save_compressed=True,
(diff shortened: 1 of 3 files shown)


rahul-tuli (approved):

horheynm:
why do we need this now? what change caused this?

kylesayrs (approved):

kylesayrs:
@horheynm Both `AutoConfig` and tokenizers share the same cache for `AutoConfig`, meaning that if you pass `trust_remote_code` to the processor, you do not have the pass it into oneshot (specifically, the instantiation of `AutoConfig` in oneshot).

However, the `AutoConfig` cache is not shared between `AutoConfig` and processors. Previously we were getting away with not passing `trust_remote_code_model` because it was sharing from the tokenizer cache, but now we do since processor and AutoConfig no longer shared from the same cache.

PR #1013, opened 2024-12-23
Update ReadMe and test for cpu_offloading

dsikka:
Summary
- ReadMe incorrectly states the requirement of 8 gpus, it should be 2. This has resulted in users being confused.
- The test case has a memory limit, which is not needed


dbarbuzzi (dismissed):

kylesayrs (dismissed):
Interesting that this was missed in the readme and cuda environment var

PR #1023, opened 2025-01-01
Extend `disable_hooks` to keep subsets

kylesayrs:
## Purpose ##
* Allow subsets of handles to remain active, as is needed in the case of wanda

```python3
def _get_activations(self, model, dataloader, nsamples=128):
  def save_acts(module, input, name):
      ...
  
  hooks = set(
      self.register_hook(mod, partial(save_acts, name=name), "forward_pre")
      for name, mod in self.model.named_modules()
      if isinstance(mod, torch.nn.Linear) and "lm_head" not in name
  )
  # in the future, if the user puts wanda after another modifier,
  # initialize will run after other modifiers have added hooks

  # want to disable hooks from other modifiers, but keep the ones we just added
  with HooksMixin.disable_hooks(keep=hooks):
      run_basic(model, dataloader)
  self.remove_hooks(hooks)
```

## Prerequisites ##
* https://github.com/vllm-project/llm-compressor/pull/1021

## Postrequisites ##
* Layer compressor deprecation

## Changes ##
* Add a `_HOOKS_KEEP_ENABLED` class variable

## Tests ##
* Added `test_disable_hooks_keep` in tests


rahul-tuli (dismissed):
The new changes LGTM! Great work @kylesayrs

src/llmcompressor/modifiers/utils/hooks.py
```diff
@@ -29,18 +29,29 @@ class HooksMixin(BaseModel):
         - modifier.remove_hooks()
     """
 
-    _HOOKS_DISABLED: ClassVar[bool] = False  # attached to global HooksMixin
-    _hooks: Set[RemovableHandle] = set()  # attached to local subclasses
+    # attached to global HooksMixin class
```
dsikka, line 32: Can you update the docstring to show how to enable or disable a subset of the hooks?

src/llmcompressor/modifiers/utils/hooks.py
```diff
@@ -29,18 +29,29 @@ class HooksMixin(BaseModel):
...
+    def disable_hooks(cls, keep: Set[RemovableHandle] = set()):
+        """
+        Disable all hooks across all modifiers. Composing multiple contexts is
+        equivalent to the union of `keep` arguments
+
+        :param keep: optional set of handles to keep enabled
+        """
         try:
             cls._HOOKS_DISABLED = True
+            cls._HOOKS_KEEP_ENABLED |= keep
```
dsikka, line 50: use a better variable name

dsikka (changes requested):

src/llmcompressor/modifiers/utils/hooks.py
kylesayrs: Any suggestions? I think this name perfectly describes the set as "the set of hooks which we will keep enabled"

src/llmcompressor/modifiers/utils/hooks.py
kylesayrs: Done

src/llmcompressor/modifiers/utils/hooks.py
dsikka: On a second reading, that makes sense.
HOOKS_TO_KEEP_ENABLED or HOOKS_KEPT_ENABLED

kylesayrs: I like the current name

src/llmcompressor/modifiers/utils/hooks.py
```diff
@@ -23,24 +23,42 @@ class HooksMixin(BaseModel):
...
+    _HOOKS_KEEP_ENABLED: ClassVar[Set[RemovableHandle]] = set()
+
+    # attached to local subclasses
+    _hooks: Set[RemovableHandle] = set()
 
     @classmethod
     @contextlib.contextmanager
-    def disable_hooks(cls):
-        """Disable all hooks across all modifiers"""
+    def disable_hooks(cls, keep: Set[RemovableHandle] = set()):
```
brian-dellabetta, line 48: you're not mutating `keep` below, so this won't be an issue here, but python has very weird behavior when you use default container values as inputs to functions:

```python
def f(m = []):
    m.append(4)
    return m

print(f()) #prints "[4]"
print(f()) #prints "[4,4]"!
```

might be better to have `keep: Set[RemoveableHandle] | None = None` instead, and just do the `if keep is None` check.

src/llmcompressor/modifiers/utils/hooks.py
brian-dellabetta: I agree it is a little non-descript as is, `HOOKS_TO_KEEP_ENABLED` seems like a minimally invasive update. `HOOKS_KEEP_ENABLED` sounds like a bool type.

None of this will persist in a config/save file right? In that case we don't really have to worry about whether that is the best long-term way to retain state.

dsikka (dismissed):

src/llmcompressor/modifiers/utils/hooks.py
```diff
@@ -23,24 +23,42 @@ class HooksMixin(BaseModel):
...
+    >>> with HooksMixin.disable_hooks(keep=hooks):
+            model.forward(...)
+    >>> modifier.remove_hooks(hooks)
     """
 
-    _HOOKS_DISABLED: ClassVar[bool] = False  # attached to global HooksMixin
-    _hooks: Set[RemovableHandle] = set()  # attached to local subclasses
+    # attached to global HooksMixin class
+    _HOOKS_DISABLED: ClassVar[bool] = False
+    _HOOKS_KEEP_ENABLED: ClassVar[Set[RemovableHandle]] = set()
```
brian-dellabetta, line 41: I think this is also sensitive to my comment below regarding `keep: Set[RemoveableHandle] = set()`. If multiple `HooksMixin` classes are instantiated, they may share this same set. It's super weird and really silly python behavior

rahul-tuli (dismissed):
LGTM pending @brian-dellabetta 's comment on defaulting to None

src/llmcompressor/modifiers/utils/hooks.py
kylesayrs: Totally agree. In the future we should add this to the linter options, no mutable defaults

src/llmcompressor/modifiers/utils/hooks.py
kylesayrs: As mentioned in the comments, this variable is meant to be attached to the global class (Pydantic controls whether it's a class var or instance var using the `ClassVar` type hint)

This is actually intended behavior, as we want the keep set to apply across modifiers

rahul-tuli (approved):

src/llmcompressor/modifiers/utils/hooks.py
brian-dellabetta: resolving after slack huddle, awesome!

brian-dellabetta (approved):
🚀

PR #1041, opened 2025-01-07
[Oneshot] Oneshot Refactor 

horheynm:
SUMMARY:
Edit oneshot pathway to use only the necessary code. 

Problem:
Oneshot uses a pathway that is shared in other entrypoints, eg. train. 
Oneshot doesnt train, so do not need use code that trains, as an example of redundant code. Remove other parts of the code and refactor 

Design
Screenshot 2025-01-16 at 2 36 33 PM

Changes:
* Oneshot class responsible for carrying out the oneshot pipeline - broken up into preprocess (init model, tokenizer), carrying out logic (run oneshot) and post-process (save, clear session).
* Move parse_args into one folder, transformers/utils/arg_parser
* Get rid of def apply in `src/llmcompressor/core/session.py` and `src/llmcompressor/core/session_functions.py`, which is only used in the oneshot flow
* Get rid of def one_shot in `src/llmcompressor/transformers/finetune/session_mixin.py`
* Add model dependent args from training_args to model_args (eg. save_compressed)
* Get rid of unused model_args (eg. use_fast_tokenizer)

TEST PLAN:
Pass existing passing tests
Pass all finetune tests merging transformer main for HFQuantizer support 
Verified new pathway has the same scales for the int8 w8a8 dynamic per token values for Meta-Llama-3-8B-Instruct


examples/quantization_w8a8_int8/llama3_example.py
```diff
@@ -80,6 +80,6 @@ def tokenize(sample):
 print("==========================================\n\n")
 
 # Save to disk compressed.
-SAVE_DIR = MODEL_ID.split("/")[1] + "-W8A8-Dynamic-Per-Token"
+SAVE_DIR = MODEL_ID.split("/")[1] + "-W8A8-Dynamic-Per-Token-oneshotclass"
```
dsikka, line 83: revert

examples/quantization_w8a8_fp8/llama3_example.py
```diff
@@ -3,7 +3,8 @@
 from llmcompressor.modifiers.quantization import QuantizationModifier
 from llmcompressor.transformers import oneshot
 
-MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
+# MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
+MODEL_ID = "TinyLlama/TinyLlama-1.1B-Chat-v1.0"
```
dsikka, line 7: revert

src/llmcompressor/transformers/finetune/session_mixin.py
```diff
@@ -632,3 +643,54 @@ def _calculate_checkpoint_info(self, kwargs) -> Tuple[Optional[str], float]:
             ).epoch
 
         return checkpoint, epoch
+
+
+class OneshotSessionManagerMixIn:
```
dsikka, line 648: I dont think this class is needed. As the goal is to remove abstractions, we can initialize the session directly, this class isn't providing any additional.

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -76,12 +84,17 @@ def eval(**kwargs):
...
 def oneshot(**kwargs):
+    from llmcompressor.transformers.calibration.oneshot import Oneshot
+
     """
     CLI entrypoint for running oneshot calibration
     """
-    model_args, data_args, training_args = parse_args(**kwargs)
-    training_args.do_oneshot = True
-    main(model_args, data_args, training_args)
+    oneshot_calibrator = Oneshot(**kwargs)
```
dsikka, line 92: Why are we calling this a calibrator?

src/llmcompressor/transformers/finetune/model_args.py
```diff
@@ -83,3 +83,91 @@ class ModelArguments:
             "repositories you trust and in which you have read the code"
         },
     )
+
+
+@dataclass
+class OneshotModelArguments:
```
dsikka, line 89: Why do we need to introduce additional model args as opposed to using the model args that we already have defined?

Seems redundant

src/llmcompressor/transformers/finetune/trainer.py
```diff
@@ -1,9 +1,16 @@
...
 
-__all__ = ["Trainer"]
+__all__ = ["Trainer", "Calibrator"]
 
 
 class Trainer(SessionManagerMixIn, HFTransformersTrainer):
     pass
+
+
+class Calibrator(OneshotSessionManagerMixIn):
```
dsikka, line 15: why is this needed?

src/llmcompressor/transformers/calibration/oneshot.py
```diff
@@ -0,0 +1,244 @@
...
+        model_path,
+        **model_kwargs,
+    )
+    if "sequence_length" in model_kwargs:
+        model.seqlen = model_kwargs["sequence_length"]
+
+    return model
+
+
+def initialize_processor_from_path(
```
dsikka, line 215: redundant?

src/llmcompressor/transformers/calibration/oneshot.py
```diff
@@ -0,0 +1,244 @@
...
+        if model_args.processor:
+            raise ValueError("Cannot use both a tokenizer and processor")
+
+        logger.debug("Overwriting processor with tokenizer")
+        model_args.processor = model_args.tokenizer
+
+    return model_args, data_args, recipe_args
+
+
+def initialize_oneshot_model(
```
dsikka, line 159: redundant?

dsikka (changes requested):
It seems like we're introducing more abstractions, not fewer, when separating out the oneshot entrpoint and I'm not sure these are necessary. 

we also seem to be repeating a lot of code everywhere.

dsikka:
> Screenshot 2025-01-07 at 6 49 43 PM
> SUMMARY: Edit oneshot pathway to use only the necessary code.
> 
> Problem: Oneshot uses a pathway that is shared in other entrypoints, eg. train. Oneshot doesnt train, so do not need use code that trains, as an example of redundant code. Remove other parts of the code and refactor
> 
> Changes:
> 
> * Oneshot run will not depend on the session. Only use CompressionLifecycle that has modifer logic support.
> * Return model as a return type to oneshot to obtain model without relying on session.
> 
> TEST PLAN: Pass existing passing tests Verified new pathway has the same scales for the int8 w8a8 dynamic per token values for Meta-Llama-3-8B-Instruct

Please update the PR desription showing what the updated entrypoint looks like as a result of this PR.

src/llmcompressor/transformers/finetune/model_args.py
horheynm: Good question. We got rid of training args. Additional args here are from training args, ex. save_compressed.

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: You are right, its running ptq and compressing. Ill call it compressor/oneshot_compressor

src/llmcompressor/transformers/calibration/oneshot.py
horheynm: yes thank you. Duplicate code will be removed / combined with existing code

horheynm:
@dsikka code is ready

tests/llmcompressor/transformers/gptq/test_oneshot.py
```diff
@@ -75,7 +75,6 @@ def test_oneshot_application(self):
             model=self.model,
             dataset=self.dataset,
             output_dir=self.output,
-            overwrite_output_dir=True,
```
horheynm, line 78: Only used in HF training args
https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L819

kylesayrs:
This PR changes 45 files. It'd be really nice if you could break out
* Renaming `DataTrainingArguments` to `DatasetArguments`
* Renaming `get_shared_processor_src` to `get_processor_from_model`
* Renaming `DVCDatasetTrainingArguments` to `DVCDatasetArguments`
* Renaming `CustomDataTrainingArguments` to `CustomDatasetArguments`
* Restructuring arguments
  * Moving some arguments to different classes, why is `use_fast_tokenizer` removed?
* Replacing the oneshot and apply pathway with the Oneshot class

horheynm:
> This PR changes 45 files. It'd be really nice if you could break out
> 
> * Renaming `DataTrainingArguments` to `DatasetArguments`
> * Renaming `get_shared_processor_src` to `get_processor_from_model`
> * Renaming `DVCDatasetTrainingArguments` to `DVCDatasetArguments`
> * Renaming `CustomDataTrainingArguments` to `CustomDatasetArguments`
> * Restructuring arguments
>   
>   * Add model dependent args from training_args to model_args (eg. save_compressed)
>   * Move parse_args into one folder, transformers/utils/arg_parser
> * Replacing the oneshot and apply pathway with the Oneshot class
> 
> You can probably group the renaming changes into one PR

Ok sure thing

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -251,21 +260,30 @@ def run_sequential_stages(self, checkpoint: Optional[str] = None):
...
             if run_type is StageRunType.ONESHOT:
-                self.one_shot(stage=stage_name)
+                from llmcompressor.transformers.calibration import Oneshot
+
+                model = get_session_model()
+                self._model_args.model = model
+
+                oneshot = Oneshot(
+                    output_dir=self._training_args.output_dir,
+                    **get_dataclass_as_dict(self._model_args, ModelArguments),
```
kylesayrs, line 270: This is a really weird pattern, where we parse the data args, then expand them to kwargs, then parse them again within Oneshot.

This will be simpler in the future if we decide to make Oneshot (or LLMCompressor) the main entrypoint for user args, but in the meantime we can add a Oneshot.from_args class method that just assigns the args directly?

horheynm: ok `Oneshot.from_args` is a good idea

horheynm:
@kylesayrs 
Thanks for reviews, this is the pre-broken down, so don't worry about this pr

src/llmcompressor/core/session.py
```diff
@@ -200,19 +200,6 @@ def finalize(self, **kwargs) -> ModifiedState:
             modifier_data=mod_data,
         )
```
dsikka, line 202: yay

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -65,27 +70,33 @@ def train(**kwargs):
...
-    model_args, data_args, training_args = parse_args(**kwargs)
+    model_args, data_args, recipe_args, training_args, _ = parse_args(
+        include_training_args=True, **kwargs
+    )
     training_args.do_eval = True
-    main(model_args, data_args, training_args)
+    main(model_args, data_args, recipe_args, training_args)
 
 
 def oneshot(**kwargs):
```
dsikka, line 91: We should separate this into its own file so it's not baked under this file

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -288,7 +351,8 @@ def initialize_processor_from_path(
 
 def main(
     model_args: ModelArguments,
-    data_args: DataTrainingArguments,
+    data_args: DatasetArguments,
+    recipe_args: RecipeArguments,
```
dsikka, line 355: Why does oneshot live in the training arguments?
I dont think we want it coming from main at all

src/llmcompressor/transformers/calibration/oneshot.py
```diff
@@ -0,0 +1,263 @@
...
+from llmcompressor.transformers.sparsification.compressed_tensors_utils import (
+    modify_save_pretrained,
+    patch_tied_tensors_bug,
+)
+from llmcompressor.transformers.utils.arg_parser import DEFAULT_OUTPUT_DIR
+
+__all__ = ["Oneshot"]
+
+
+class Oneshot:
```
dsikka, line 25: This should be at the top level

PR #1049, opened 2025-01-08
GPTQ compress using callback not hook

kylesayrs:
## Purpose ##
* Support calibrating MoE models for GPTQ


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -206,26 +205,24 @@ def on_initialize(self, state: "State", **kwargs) -> bool:
         if not self.quantize:
             raise ValueError("To use the GPTQModifier, quantization must be enabled.")
 
+        # prepare module names
+        self._module_names = {m: name for name, m in state.model.named_modules()}
```
horheynm, line 209: Nit:
`self._module_names = {module: name for name, module in state.model.named_modules()}`

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -240,6 +237,7 @@ def on_initialize(self, state: "State", **kwargs) -> bool:
                 state.data.calib,
                 self.sequential_targets,
                 self.ignore,
+                self,
```
horheynm, line 240: Above the try can you add a description of whats going with try catch and going to do 3 different types of algorithm selection

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: I shortened the name so it would fit on one line and be easier to read 🙃

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: > `# infer pipeline`

I feel that the fallback logic here is self-documenting and also not relevant to this PR

mgoin (approved):
LGTM, although having all of these separate module dicts feels a bit off. In the future we could have a class for holding all the state for a given module together

PR #1061, opened 2025-01-13
[E2E Testing] Fix HF upload

horheynm:
SUMMARY:
Fix for e2e tests for models that do not exist on HF. 
Error comes from repo not existing on HF, hence trying to push to a non-existing repo. 

Error from e2e.
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/12729284591/job/35480943434#step:15:6775

TEST PLAN:
Verified that e2e works with one existing config and all of new kv-cache configs locally, pushing to nm-testing. 
Note:[ Custom test runners](https://github.com/neuralmagic/llm-compressor-testing#custom-test-runs) are down.


tests/e2e/vLLM/test_vllm.py
```diff
@@ -158,6 +168,9 @@ def test_vllm(self):
         self.tear_down()
 
     def tear_down(self):
+        import torch
+        torch.cuda.empty_cache()
```
kylesayrs, line 172: Can we remove this? `torch.cuda.empty_cache()` doesn't do anything useful except free gpu memory for other programs. I see no reason why this would be necessary for tear down

kylesayrs (changes requested):
Looks reasonable to me, nice catch

tests/e2e/vLLM/test_vllm.py
horheynm: ok

dsikka (approved):

kylesayrs (approved):

rahul-tuli (approved):

PR #1079, opened 2025-01-17
Require 'ready' label for transformers tests

dbarbuzzi:
SUMMARY:
This PR moves the `transformers-tests` job to its own workflow which only runs on a PR that has the 'ready' label. With this implementation, if the PR already has the 'ready' label when a new commit is pushed, the new commit will re-run the job. I also updated the PR comment workflow to note that the 'ready' label must be added to run this job.

There are some visual warts/caveats that come with these triggers because the filtering cannot be done by the triggers, only the jobs:

Whenever a label is added or a commit is pushed, the workflow will still be triggered. However, if the PR does not have the 'ready' label, the `transformers-tests` job will be skipped. While the job being skipped means the runner/resources will not actually be used, it also means that the workflow/job will still show up in various places in the GitHub UI (Actions tab, PR checks). It will show as skipped, but it will still allow the green check mark because that appears as long as there are no failures.

Basically, this means that reviews/people able to merge will need to pay closer attention to the jobs run on the PR to make sure that the `transformers-tests` job was run/not skipped.

TEST PLAN:
Should affect this PR.

### DIFF
--- a/.github/workflows/set-comment.yaml
+++ b/.github/workflows/set-comment.yaml
@@ -17,7 +17,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.issue.number,
-              body: '👋 Hi! Thank you for contributing to llm-compressor. Please add the ready label when the PR is ready for review.'
+              body: '👋 Hi! Thank you for contributing to llm-compressor. Please add the ready label when the PR is ready for review.\n\n**Note:** This is required to trigger the transformers tests.'
             })
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/set-comment.yaml
+++ b/.github/workflows/set-comment.yaml
@@ -17,7 +17,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.issue.number,
-              body: '👋 Hi! Thank you for contributing to llm-compressor. Please add the ready label when the PR is ready for review.\n\n**Note:** This is required to trigger the transformers tests.'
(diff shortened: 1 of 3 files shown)


dbarbuzzi:
Commenting here: At this stage, all workflows/jobs except the `transformers-tests` job have run.

dbarbuzzi:
Commenting after adding the 'ready' label: The `transformers-tests` workflow/job is now queued after adding the label.

.github/workflows/set-comment.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.issue.number,
-              body: '👋 Hi! Thank you for contributing to llm-compressor. Please add the ready label when the PR is ready for review.'
+              body: '👋 Hi! Thank you for contributing to llm-compressor. Please add the ready label when the PR is ready for review.\n\n**Note:** This is required to trigger the transformers tests.'
```
dsikka, line 20: Can we change this to 
"This is required to trigger the complete testing suite" or something along those lines?

dsikka (approved):
LGTM. One comment about the note we're providing to users

.github/workflows/set-comment.yaml
dsikka: Can we also note something about only setting the label when you're code complete and have tested locally?

dbarbuzzi: Sure, should I make those changes to the "Note" that occurs on the separate line? Something like this, perhaps:

**Note:** This is required to complete the testing suite, please only add the label once the PR is code complete and local testing has been performed.

Let me know whatever changes to that/whatever else you want it to say and I’ll make the changes.

dsikka: The note in your comment is perfect, thanks!

dbarbuzzi: Just pushed a commit with the new comment.

dsikka (approved):

rahul-tuli (approved):
LGTM!

PR #1080, opened 2025-01-17
Add group act order case to lm_eval test

dsikka:
SUMMARY:
- Requires next transformers release/ when the following commit is released: https://github.com/huggingface/transformers/pull/35615
- Adds group act order case which previously could not be loaded through the AutoModel pathway due to incorrect substring replacement of the `weight_g_idx` parameter  

Testing:
- Test passes locally with transformers 4.49


kylesayrs (approved):
Yay

brian-dellabetta (approved):

PR #1032, opened 2025-01-03
VLM: Phi3 Vision Example

kylesayrs:
## Purpose ##
* Provide example for quantizing phi3_vision model

## Changes ##
* Add phi3 example script
  * When tokenizing, we have to set `writer_batch_size=1`, otherwise we encounter https://github.com/pola-rs/polars/issues/862
  * My suspicion is that this is due to the phi processor outputs being especially long
* Add phi3 vision data collator

## Testing ##
Model | Dataset | Scheme | Runtime | Winogrande |
-- | -- | -- | -- | --
Phi-3-vision-128k-instruct | flickr | W4A16 | 51m, 1xA100 | 0.7151

PR #1087, opened 2025-01-20
[VLM] Multimodal Data Collator

kylesayrs:
## Purpose ##
* Move data collators to example script as per @mgoin's suggestion https://github.com/vllm-project/llm-compressor/pull/1032#discussion_r1922807072

## Changes ##
* Remove data collator definitions in LC
* Add data collators in examples with a comment indicating that this is for multimodal inputs

## Testing ##
Ran all multimodal vision models
* Qwen2
* Pixtral
* Mllama
* Llava
* Phi3_vision


src/llmcompressor/transformers/utils/data_collator.py
```diff
@@ -1,59 +0,0 @@
-import torch
```
dsikka, line 1: I think this is just removing the defs, not the actual file?
We should just remove the file if empty

dsikka (approved):
nit about removing empty file otherwise LGTM

horheynm (approved):

src/llmcompressor/transformers/utils/data_collator.py
kylesayrs: The file is removed
https://github.com/vllm-project/llm-compressor/blob/kylesayrs/generalized_data_collator/src/llmcompressor/transformers/utils/data_collator.py

dsikka: oh sorry, showed up weirdly on my diff. thanks!

PR #1095, opened 2025-01-24
Add Idefics3/SmolVLM quant support via traceable class

leon-seidel:
SUMMARY:
Adding a traceable Idefics3 class following the new [guide](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/transformers/tracing/GUIDE.md) to allow W4A16 quants of Idefics3 and SmolVLM (which share the same architecture). Idefics3 seems to require a max_sequence_length of 4096 and I copied the example from the Phi 3 Vision example as the dataset loading approach from the Llava example led to OOM on 64 GB RAM. 

TEST PLAN:
Tested on A100 with Idefics3 @512 samples and on a 4060 Ti with SmolVLM @128 samples.


src/llmcompressor/transformers/tracing/idefics3.py
```diff
@@ -0,0 +1,423 @@
+# coding=utf-8
```
kylesayrs, line 1:
```suggestion
# flake8: noqa
# coding=utf-8
```

kylesayrs (commented):
Thank you for your contribution @leon-seidel! This looks great to me, I look forward to getting this landed!

kylesayrs:
Hi @leon-seidel

I've run your example end to end and it looks good! All that's left is to fix the quality tests and this is good to land!
```
pip install -e ./[dev]
make style
make quality
```

kylesayrs:
Base
```
hf-multimodal (pretrained=HuggingFaceM4/Idefics3-8B-Llama3,dtype=bfloat16,add_bos_token=True,convert_img_format=True), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: 1
|     Tasks      |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|----------------|------:|------|-----:|------|---|-----:|---|-----:|
|Computer Science|      0|none  |     0|acc   |↑  |0.3333|±  |0.0875|
```

W4A16
```
hf-multimodal (pretrained=Idefics3-8B-Llama3-W4A16-G128,dtype=bfloat16,add_bos_token=True
,convert_img_format=True), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: 1
|     Tasks      |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|----------------|------:|------|-----:|------|---|----:|---|-----:|
|Computer Science|      0|none  |     0|acc   |↑  |  0.3|±  |0.0851|
```

leon-seidel:
Quality should be fixed now, thanks for your help! Also great tutorial on making the models traceable in the first place. I tried it on my own finetuned Idefics3 model and can't see any deterioration in the outputs!

kylesayrs (dismissed):

examples/multimodal_vision/idefics3_example.py
```diff
@@ -0,0 +1,118 @@
...
+    return processor(
+        text=sample["text"],
+        images=sample["images"],
+        padding=False,
+        max_length=MAX_SEQUENCE_LENGTH,
+        truncation=True,
+    )
+
+
+# long data lengths produced by the phi3_vision processor
```
kylesayrs, line 80: Small nit here to remove this comment and investigate whether the `writer_batch_size` argument is required. This can be done after landing as well

leon-seidel: Did some further investigation on this:
With SmolVLM `writer_batch_size` does not seem to be required and the approach from the Llava example works.
With Idefics3  get the following error without `writer_batch_size` at 512 samples and a `max_sequence_length` of 4096:

`ArrowInvalid: offset overflow while concatenating arrays, consider casting input from list<item: list<item: list<item: list<item: list<item: float>>>>> to list<item: list<item: list<item: list<item: large_list<item: float>>>>> first.
`
A lower `max_sequence_length` doesn't have this problem but crashes with the following after a few layers:

`RuntimeError: shape mismatch: value tensor of shape [2197, 4096] cannot be broadcast to indexing result of shape [1937, 4096]`

Will remove the comment regarding the `phi3_vision` processor!

kylesayrs: Nice, your experiment with  sequence lengths confirms the suspicion that this is a limitation with `datasets`' map function. The `ArrowInvalid` error you encountered in the same error that I encountered when running the phi3 model without `writer_batch_size`.

Thanks for your insight and changes!

kylesayrs (approved):

dsikka (approved):

PR #1096, opened 2025-01-24
[Bug Fix] Fix test that requre GPU

horheynm:
SUMMARY:
Nightly tests uses gpus, ci tests dont. 
In the set up of `tests/llmcompressor/transformers/obcq/test_consecutive_runs.py`, a bug was found where `quantization_config` was always passed. This will result in an error if the model from the config file is not optimized. 
We are passing in dense model, so it fails.

TEST PLAN:
Run the failing test, make sure it passes!

### DIFF
--- a/tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
+++ b/tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
@@ -8,6 +8,7 @@ from parameterized import parameterized_class
 from transformers import AutoModelForCausalLM
 from transformers.utils.quantization_config import CompressedTensorsConfig
 
+from llmcompressor.transformers.utils import is_model_ct_quantized
 from llmcompressor.transformers.utils.helpers import infer_recipe_from_model_path
 from tests.testing_utils import parse_params, requires_gpu
 
@@ -137,10 +138,17 @@ class TestConsecutiveRunsGPU(TestConsecutiveRuns):
     def setUp(self):
         from transformers import AutoModelForCausalLM
 
+        kwargs = {}
+        # if optimized model is passed, then must be using Linear Modules
+        if is_model_ct_quantized(self.model):
+            kwargs["quantization_config"] = CompressedTensorsConfig(
+                run_compressed=False
+            )
+
         self.model = AutoModelForCausalLM.from_pretrained(
             self.model,
             device_map=self.device,
-            quantization_config=self.quantization_config,
+            **kwargs,
         )
 
         self.output = "./oneshot_output"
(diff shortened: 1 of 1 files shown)


tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -137,10 +138,15 @@ class TestConsecutiveRunsGPU(TestConsecutiveRuns):
     def setUp(self):
         from transformers import AutoModelForCausalLM
 
+        kwargs = {}
+        # if optimized self.model is passed, then must be using Linear Modules
+        if is_model_ct_quantized_from_path(self.model):
```
dsikka, line 143: This is fine but likely overkill as the models for this test won’t be quantized to start

kylesayrs (approved):
Looks good

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
horheynm: Ok I can add an assertion instead. Its good to start from dense i agree

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
```diff
@@ -137,10 +138,15 @@ class TestConsecutiveRunsGPU(TestConsecutiveRuns):
...
+            kwargs["quantization_config"] = self.quantization_config
```
brian-dellabetta, line 144: i know this is a test, but curious if we make use of `**kwargs` heavily in the code? i'm old and was taught this is not good practice, but huggingface does it everywhere. do we have any standardized guide for this sort of thing? An alternative would be something like

```python
        self.model = AutoModelForCausalLM.from_pretrained(
            self.model,
            device_map=self.device,
            quantization_config=self.quantization_config if s_model_ct_quantized_from_path(self.model) else None,
            )
```

brian-dellabetta (approved):
left a comment but LGTM otherwise

tests/llmcompressor/transformers/obcq/test_consecutive_runs.py
horheynm: I never heard of kwargs being a bad actor before, but makes sense -- hard to keep in track of all the variables. 
We don't have a standard guide for this.
The alternate looks good, I just got rid of the kwargs so ill keep that in mind going forward!

brian-dellabetta (approved):

rahul-tuli (approved):
LGTM

dsikka (approved):

PR #1097, opened 2025-01-24
Replace readme paths with urls

kylesayrs:
## Purpose ##
* Files with the `.md` extension are not listed in the [MANIFEST.in](https://github.com/vllm-project/llm-compressor/blob/main/MANIFEST.in), meaning that they will not be included in the LLM Compressor pypi package. This means that references to these files are left dangling for users who have installed from the pypi package. Rather than including `.md` in the package and having to also ship all the large images files associated with them, this PR moves the references to urls hosted by github
  * While the github url paths may change between versions, this solution works in lieu of a dedicated readthedoc build for each version
  * This solution also aligns with the practice of other libraries which point to hosted urls rather than file paths
* Note that this does not apply to files which are themselves `.md` files, as these files will not be included in the pypi distribution
  * `src/llmcompressor/transformers/finetune/README.md`
  * `src/llmcompressor/pipelines/sequential/README.md`

## Changes ##
* Replace readme file paths with urls
* Small change to `DisableQuantization` to better catch cases where exceptions such as tracing exceptions are triggered

## Testing ##
* N/A

### DIFF
--- a/src/llmcompressor/utils/helpers.py
+++ b/src/llmcompressor/utils/helpers.py
@@ -1091,9 +1091,11 @@ def DisableQuantization(model: torch.nn.Module):
     """
     Disable quantization from QuantizationModifier
     """
-    model.apply(disable_quantization)
-    yield
-    model.apply(enable_quantization)
+    try:
+        model.apply(disable_quantization)
+        yield
+    finally:
+        model.apply(enable_quantization)
 
 
 @contextlib.contextmanager
--- a/src/llmcompressor/modifiers/quantization/gptq/base.py
+++ b/src/llmcompressor/modifiers/quantization/gptq/base.py
@@ -247,7 +247,8 @@ class GPTQModifier(Modifier, HooksMixin):
                 warnings.warn(
                     f"Failed to trace {model_name} with inputs {input_names}. For more "
                     "information on tracing with the sequential pipeline, see "
-                    "`src/llmcompressor/transformers/tracing/GUIDE.md`"
+                    "https://github.com/vllm-project/llm-compressor/blob/main/"
+                    "src/llmcompressor/transformers/tracing/GUIDE.md"
                 )
             if isinstance(exception, unfixable_errors):
                 raise exception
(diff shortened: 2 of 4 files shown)


rahul-tuli (approved):

src/llmcompressor/utils/helpers.py
```diff
@@ -1091,9 +1091,11 @@ def DisableQuantization(model: torch.nn.Module):
     """
     Disable quantization from QuantizationModifier
     """
-    model.apply(disable_quantization)
-    yield
-    model.apply(enable_quantization)
+    try:
+        model.apply(disable_quantization)
+        yield
+    finally:
```
horheynm, line 1097: model.apply(enable_quantization)
will always runs regardless of try being success or fail. Is this intended. 
Also if it fails I would add a log

kylesayrs: > model.apply(enable_quantization) will always runs regardless of try being success or fail. Is this intended.

Yes, that is exactly the intention of this change

> Also if it fails I would add a log

If an exception is raised, the user will see the raised exception

horheynm: sorry why do we want to run `model.apply(enable_quantization)` if it raises exception? Wouldn't it just exit?

And as long as the exception contains info that we can get from this function, then it should be ok. Ex. have input args useful to log causing the error.

kylesayrs: > sorry why do we want to run model.apply(enable_quantization) if it raises exception? Wouldn't it just exit?

For example, if the sequential pipeline fails, an exception will be thrown and then caught, in order to fall back to the layer-wise sequential pipeline. This is example of where an exception is raised but caught, and I do not want do have to think about the weird interactions that would result from this context not being properly exited.

This "try finally" pattern is, in general, good for contexts. It allows the user to not have to worry about if the context was exited properly.

rahul-tuli: Could we move this code change to a separate PR, ideally I would like the scope of this PR to be limited to updating readme paths

mgoin (approved):

PR #1098, opened 2025-01-24
Add compressed-tensors appropriate code from llm-compressor

horheynm:
SUMMARY:
* Move `is_model_from_path_quantized` from LC to CT
* Apply renamed methods from CT to LC

TEST PLAN:
Pass tests locally

Note: need first to merge 
https://github.com/neuralmagic/compressed-tensors/pull/246


kylesayrs:
I don't really see why this is necessary? If it is deemed necessary/useful, I ask that 1. the original definition stay in LC with a @deprecation warning and 2. this PR stay in draft until the corresponding CT change lands and releases

PR #1103, opened 2025-01-27
[Oneshot Refactor] dataclass Arguments

horheynm:
ORDER OF REVIEWS:
1. https://github.com/vllm-project/llm-compressor/pull/1108
2. https://github.com/vllm-project/llm-compressor/pull/1103 <- current PR
3. https://github.com/vllm-project/llm-compressor/pull/1109
4. https://github.com/vllm-project/llm-compressor/pull/1110

SUMMARY:
Refactor dataclass used for llm-compressor entrypoints (oneshot, train, apply) to decouple non-relevant attributes from the existing dataclass. Ex. recipe in training_args. Recipe is contained in a session, not in the trainer that training_args govern. 

Dataclass refactor details are in
https://docs.google.com/document/d/1YbR1dTQmCzqhGk74m5msBzqoPHQgB6dVxDtf6cTmetc/edit?usp=sharing

Note: 
https://github.com/vllm-project/llm-compressor/pull/1110 takes care of using a new entrypoint that will prohibit the post_train / oneshot call to use training_args. Current entrypoint will need training_args for oneshot to function - this PR is just for refactoring the dataclass. 

Before:
ModelArguments: https://github.com/vllm-project/llm-compressor/blob/6fa5a5eecc7d363ec73474d011d40135b6374179/src/llmcompressor/transformers/finetune/model_args.py#L6
DataTrainingArguments: https://github.com/vllm-project/llm-compressor/blob/6fa5a5eecc7d363ec73474d011d40135b6374179/src/llmcompressor/transformers/finetune/data/data_args.py#L70
TrainingArguments: https://github.com/vllm-project/llm-compressor/blob/6fa5a5eecc7d363ec73474d011d40135b6374179/src/llmcompressor/transformers/finetune/training_args.py#L10

After:
ModelArguments: https://github.com/vllm-project/llm-compressor/pull/1103/files#diff-58fd0f7ae4564317960ae0d4d4b2cdb97b9588c1915f062915e74ecf51b5502cR6
DatasetArguments:  https://github.com/vllm-project/llm-compressor/pull/1103/files#diff-5e43f74ba5d8327b937adada3c7f30a7efb13f9a44cb3fdb5e1a2a12b8b8ea27R70
RecipeArguments: https://github.com/vllm-project/llm-compressor/pull/1103/files#diff-0ff9c048a4deb55e5459054bdc61a5d8c81da9c94588ec2355e6b2c2ec8675d1R6
TrainingArguments: https://github.com/vllm-project/llm-compressor/pull/1103/files#diff-249ee96763dd50956a7309f898eda4bcaa91c6af653474568fbda10b5a39c817R12

TEST PLAN:
* Pass all existing tests
* Search dataclass arguments using `pygrep`
```bash
  3 function pygrep() {
  2     local search_term="$1"
  1     shift
126     local search_dirs="${*:-src examples tests}"                           
  1     grep -rn --include="*.py" -E "$search_term" $search_dirs
  2 }
```


src/llmcompressor/transformers/finetune/session_mixin.py
```diff
@@ -428,31 +431,6 @@ def predict(self, *args, **kwargs):
 
         return output
 
-    def one_shot(
```
horheynm, line 431: remove not here

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -65,27 +70,33 @@ def train(**kwargs):
...
 def oneshot(**kwargs):
+    from llmcompressor.transformers.calibration.oneshot import Oneshot
+
     """
     CLI entrypoint for running oneshot calibration
     """
-    model_args, data_args, training_args = parse_args(**kwargs)
-    training_args.do_oneshot = True
-    main(model_args, data_args, training_args)
+    oneshot = Oneshot(**kwargs)
```
horheynm, line 97: make changes in a different pr

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -111,60 +125,100 @@ def compress(**kwargs):
...
+            import sys
+
+            output_dir = None
+            if "--output_dir" in sys.argv:
+                index = sys.argv.index("--output_dir")
+                sys.argv.pop(index)
+                if index < len(sys.argv):  # Check if value exists afer the flag
+                    output_dir = sys.argv.pop(index)
+
+            return output_dir
```
horheynm, line 175: make changes in a different pr. this is for cli having --output_dir

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -177,38 +231,42 @@ def initialize_model_from_path(
...
+    if training_args is not None:
+        teacher_config = (
+            AutoConfig.from_pretrained(
+                model_args.distill_teacher,
+                use_auth_token=True if model_args.use_auth_token else None,
+                tie_word_embeddings=model_args.tie_word_embeddings,
+                trust_remote_code=model_args.trust_remote_code_model,
+            )
+            if model_args.distill_teacher
+            else None
```
horheynm, line 246: make changes in a different pr

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -177,38 +231,42 @@ def initialize_model_from_path(
...
+        set_seed(training_args.seed)
 
     model_path = (
         last_checkpoint or model_args.model
         if hasattr(model_args, "model")
         else model_args.model_name_or_path
     )
 
-    # Set seed before initializing model.
-    set_seed(training_args.seed)
```
horheynm, line 198: restore

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -218,15 +276,7 @@ def initialize_model_from_path(
         "device_map": device_map,
         "trust_remote_code": model_args.trust_remote_code_model,
     }
-    teacher_device_map = None if fsdp_enabled else "auto"
```
horheynm, line 221: all these changes in a different PR

dsikka:
@horheynm Can you touchbase with Mark on the data arguments he shared during the meeting this morning?

kylesayrs:
Is this ready for review? I'm a little confused what the comment "make changes in a different pr" means

horheynm:
> Is this ready for review? I'm a little confused what the comment "make changes in a different pr" means

It means it still needs to be broken down, it should only contain dataargs only.

kylesayrs:
@horheynm Is this ready for review now?

src/llmcompressor/transformers/utils/arg_parser/utils.py
```diff
@@ -0,0 +1,30 @@
...
+from .model_arguments import ModelArguments
+from .recipe_arguments import RecipeArguments
+from .training_arguments import TrainingArguments
+
+__all__ = [
+    "get_dataclass_as_dict",
+]
+
+
+def get_dataclass_as_dict(
```
kylesayrs, line 14: This function isn't actually used in this PR.

1. I think we can avoid parsing, expanding and parsing again. See [my comment the main PR](https://github.com/vllm-project/llm-compressor/pull/1041/files/f81ca9d29d4415badff610b6d8b5fd9dccc383ca#r1937941783)
2. I think this function is likely overly complex? [Here's an example](https://github.com/vllm-project/llm-compressor/blob/317c10ca901901c43f767c50aac8362e2c930448/src/llmcompressor/pipelines/cache.py#L137-L139) of iterating through a data class which does not require knowing the dataclass_class

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -256,16 +264,13 @@ def run_sequential_stages(self, checkpoint: Optional[str] = None):
                 self.train(checkpoint=checkpoint, stage=stage_name)
             checkpoint = None
 
-            if (
-                self._training_args.output_dir
-                != TrainingArguments.__dataclass_fields__["output_dir"].default
-            ):
+            if self._training_args.output_dir != DEFAULT_OUTPUT_DIR:
```
kylesayrs, line 267: I know that this is unrelated to this PR, but holy crap does this seem like a dumb line. If I'm reading it correctly, if the user passes in exactly output_dir="./output", the model just won't save???

kylesayrs: I've confirmed my suspicions. This is deeply horrifying.

kylesayrs: The best solution would to be set `output_dir` default to `None`, then populate it at the Trainer/Oneshot. Unfortunately because strings are primitives in python, we cannot differentiate between a user passed string and a default string of the same value using `is`.

horheynm: This is something we decided on as group. We need a good argument to make changes. 
Not saving was already considered.

horheynm: We can set a bool before parsing it into dataclass, we pop it anyway. 
use that bool to be used for saving logic.

dsikka: I’ve definitely ran models and been able to save without specifying the output dir as exactly that when running finetune? Or is this different

horheynm: The parser is a shared method, so should be the same.
Maybe in finetune --output_dir was set?

horheynm: Probably bc output is set to "./output" in training args as default and it is used in trainer. So training logic using that is probably being saved. 

So if output dir is not set 
Then in outshot -> not save
If in fine tune -> always save

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -65,27 +69,36 @@ def train(**kwargs):
...
 
 
 def oneshot(**kwargs):
     """
     CLI entrypoint for running oneshot calibration
     """
-    model_args, data_args, training_args = parse_args(**kwargs)
+    # TODO: Get rid of training args when Oneshot refactor comes in
+    model_args, data_args, recipe_args, training_args, output_dir = parse_args(
+        include_training_args=True, **kwargs
```
kylesayrs, line 96: If the purpose of separating training args is to throw an error if the user passes training args not relevant to oneshot, then shouldn't we not include training args for oneshot?

kylesayrs: The change where you set `include_training_args=False` is not in the main refactor PR. I don't know when that change is supposed to be made.

horheynm: You are right that we should make the main function for oneshot not dependent on training_args. Making those changes will really be just for oneshot, which will be refactored in the follow up PR anyways. So if i make changes it will only last until the Oneshot main refactor lands.

kylesayrs (changes requested):
Missing changes

```
grep -r 'DataTrainingArguments' src/ tests/ examples/ --exclude="*.pyc"                
src/llmcompressor/transformers/tracing/debug.py:from llmcompressor.transformers import DataTrainingArguments, TextGenerationDataset
src/llmcompressor/transformers/tracing/debug.py:    data_args = DataTrainingArguments(**get_dataset_kwargs(modality))
src/llmcompressor/transformers/finetune/README.md:* DataTrainingArguments: `src/llmcompressor/transformers/finetune/data/data_training_args.py`
```

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -111,53 +127,66 @@ def compress(**kwargs):
...
 
-    # silently assign tokenizer to processor
-    if model_args.tokenizer:
-        if model_args.processor:
-            raise ValueError("Cannot use both a tokenizer and processor")
-        model_args.processor = model_args.tokenizer
-    model_args.tokenizer = None
-
-    return model_args, data_args, training_args
+    _validate_model_args_tokenizer(model_args)
```
kylesayrs, line 188: This pattern of validating arguments as soon as they're received will be a common one. Can we rename this to `validate_arguments` and have it handle more than just model_args?

horheynm:
> grep -r 'DataTrainingArguments' src/ tests/ examples/ --exclude="*.pyc"

I see thats from new merge. Thanks

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -432,20 +466,62 @@ def main(
...
+
+    return dataclass_arguments
+
+
+def _unwrap_recipe_args(recipe_args: Dict[str, Any]):
+    """Convert recipe arguments to a dictionary if needed"""
+    if isinstance(recipe_args, dict):
+        return recipe_args
+
+    return {key: value for arg in recipe_args for key, value in [arg.split("=")]}
```
kylesayrs, line 523: This is overly complicated and kind of misleading.
```
{
    key: value
    for arg in recipe_args
    for key, value in [arg.split("=")]
}
```
Assuming that `arg.split("=")` only yields 2 values (which is the previous assumption), we get a list [(key, value)]. Technically you could use `for` to iterate through this singleton list to unpack it, but doing so is kind of misleading and unnecessary.

If you had to do it as a list comprehension, you could do it like this.
```suggestion
    {
        key: value
        for key, value in (arg.split("=") for arg in training_args.recipe_args)
    }
```

But in my opinion, the original for loop is the clearest.

Side note: Even if this code was written with the intention of supporting multiple kv pairs, it would yield [(k1, v1, k2, v2)] which does not unpack correctly.

src/llmcompressor/transformers/finetune/text_generation.py
kylesayrs: Why can't that change be logically grouped into this PR. That way you know that the changes in this PR are worth it because someone can use this branch to test that the desired feature actually works.

kylesayrs: > Making those changes will really be just for oneshot, which will be refactored in the follow up PR anyways. So if i make changes it will only last until the Oneshot main refactor lands.

I thought the overall purpose of these changes was to prepare for the oneshot main PR? If you plan on overwriting the changes anyways, then why are you proposing them?

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: I dont understand.

src/llmcompressor/transformers/finetune/text_generation.py
kylesayrs: I'll note that there used to be a *purpose* section in the PR description but has since been deleted

kylesayrs: > Making those changes will really be just for oneshot, which will be refactored in the follow up PR anyways. So if i make changes it will only last until the Oneshot main refactor lands.

I don't understand why these changes are being proposed if you think they're not relevant to the oneshot main refactor

horheynm: Yeah im actually going to review the new logic here

kylesayrs (changes requested):
```
examples/trl_mixin/sft_trainer.py
examples/trl_mixin/ex_trl_constant.py
```

```
ImportError: cannot import name 'TrainingArguments' from 'llmcompressor.transformers'
```

This is caused by the change in import path in `src/llmcompressor/transformers/finetune/__init__.py`. Since these are public primitives, I'd rather not change these paths unless it's necessary.

I did not find any issues by grepping for `ModelArguments` or `DatasetArguments` imports.

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -256,16 +261,13 @@ def run_sequential_stages(self, checkpoint: Optional[str] = None):
                 self.train(checkpoint=checkpoint, stage=stage_name)
             checkpoint = None
 
-            if (
-                self._training_args.output_dir
-                != TrainingArguments.__dataclass_fields__["output_dir"].default
-            ):
+            if self._training_args.output_dir:
```
kylesayrs, line 264: This check only checks if the user passes an empty string. If the user passes an empty string, the intended behavior should be that the user gets an error when they try to save (ideally earlier).

As coded now, we silently don't save the model. Please remove this line.

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -65,27 +70,31 @@ def train(**kwargs):
     """
     CLI entrypoint for running training
     """
-    model_args, data_args, training_args = parse_args(**kwargs)
+    model_args, data_args, recipe_args, training_args, _ = parse_args(**kwargs)
```
kylesayrs, line 73: Don't we need the `output_dir` argument for training?

kylesayrs (changes requested):
I think the problem with defaulting `output_dir` to a string is the two different behaviors we want from the two different pathways.

We want oneshot to *not* save the model if no `output_dir` is specified.
We want train to default to "./outdir" if no `output_dir` is specified.

I recommend setting the default to `None`, then during validation, populate it with `"./outdir"` for train and `None` for oneshot, perhaps with a warning.

All of this should be explained in the help text as well.

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: its inside training_args.

src/llmcompressor/transformers/finetune/runner.py
horheynm: self._training_args.output_dir always have a default value. it will save.

horheynm:
Changing saving logic and blocking this pr is out of scope. Any logic change outside of the dataclass argument population will be handled in the subsequent pr

tests/llmcompressor/transformers/compression/test_quantization.py
```diff
@@ -62,7 +62,6 @@ def _run_oneshot(model, recipe, dataset, output_dir):
         oneshot(
             model=model,
             dataset=dataset,
-            overwrite_output_dir=True,
```
kylesayrs, line 65: Note: figure out why this was removed

kylesayrs: nm, added back

kylesayrs (approved):
LGTM!

dsikka:
Please resolve conflicts

tests/llmcompressor/transformers/gptq/test_oneshot.py
```diff
@@ -74,8 +74,8 @@ def test_oneshot_application(self):
         oneshot(
             model=self.model,
             dataset=self.dataset,
-            output_dir=self.output,
             overwrite_output_dir=True,
+            output_dir=self.output,
```
dsikka, line 78: nit: unnecessary

tests/llmcompressor/transformers/oneshot/test_cli.py
```diff
@@ -49,6 +49,7 @@ def test_one_shot_cli(self):
         if len(self.additional_args) > 0:
             cmd.extend(self.additional_args)
         res = run_cli_command(cmd)
+
```
dsikka, line 52: nit: unnecessary

dsikka:
https://docs.google.com/document/d/1YbR1dTQmCzqhGk74m5msBzqoPHQgB6dVxDtf6cTmetc/edit?usp=sharing

Does this cover everything? 
Seems like we changed DataTrainingArguments to DatasetArguments but the doc just lists DataArguments?

src/llmcompressor/transformers/utils/arg_parser/training_arguments.py
```diff
@@ -0,0 +1,32 @@
...
+
+@dataclass
+class TrainingArguments(HFTrainingArgs):
+    """
+    Training arguments specific to LLM Compressor Transformers workflow using
+    HFTrainingArgs as base class
+
+    """
+
+    do_oneshot: Optional[bool] = field(
```
dsikka, line 19: Not specific to your PR but seems weird that we have no actual training arguments in our TrainingArugments (i.e lr, optimizer, etc)

src/llmcompressor/transformers/utils/arg_parser/model_arguments.py
```diff
@@ -44,17 +46,7 @@ class ModelArguments:
         default=None,
         metadata={"help": "Where to store the pretrained data from huggingface.co"},
     )
-    use_fast_tokenizer: bool = field(
```
dsikka, line 47: Are we moving this anywhere?

src/llmcompressor/transformers/utils/arg_parser/__init__.py
```diff
@@ -0,0 +1,6 @@
+# flake8: noqa
+
```
dsikka, line 2: I like that we're moving this all from under finetune. good change

dsikka: We should move this outside from `transformers`

`src/llmcompressor/args`

src/llmcompressor/transformers/finetune/README.md
```diff
@@ -74,9 +74,10 @@ train(
 
 Finetuning arguments are split up into 3 groups:
```
dsikka, line 75: We should add a new generic README related to this arguments outside from under `transformers`
Could be same location as where the args get moved to, such as `src/llmcompressor/args`

dsikka (changes requested):
Overall looks like a good change but I would go one step further and move the args from under `transformers` altogether plus add docs for the restructured args.

src/llmcompressor/transformers/utils/arg_parser/training_arguments.py
kylesayrs: These are inherited from [HFTrainingArgs](https://huggingface.co/docs/transformers/v4.48.2/en/main_classes/trainer#transformers.TrainingArguments)

src/llmcompressor/transformers/utils/arg_parser/model_arguments.py
kylesayrs: This arg is unused. Transformers doesn't have a coherent pattern w.r.t. handling fast versions of processors atm, so we replaced it with a try/catch where it tries to load the fast version by default, then falls back to the slow version otherwise.

We can consider adding this arg back to give users the ability to use slow processors through our api.

src/llmcompressor/arg_parser/README.md
```diff
@@ -0,0 +1,45 @@
+# Argument parser for `oneshot`, `train`, `eval` entrypoints
+
+Argument parsers define the input arguments required by any entrypoint in `llm-compressor`. Supported entrypoint includes: `oneshot`, `train` or `eval`.
+
+The logic is carried out with respect to the given input arguments, such as `model`, `recipe` and `dataset`.
```
rahul-tuli, line 5: What does this mean?

src/llmcompressor/arg_parser/README.md
```diff
@@ -0,0 +1,45 @@
...
+    ..., 
+    output_dir=...,
+)
+```
+
+These input arguments can be overloaded into the function signature and will be parsed using Hugging Face's [argument parser](https://github.com/huggingface/transformers/blob/main/src/transformers/hf_argparser.py). The parsers define the acceptable inputs; therefore any arguments to be passed in must be defined.
+
+`llm-compressor` uses four parsers, located in `llm_compressor/arg_parser`:
+* ModelArguments
+* DataArguments
```
rahul-tuli, line 29:
```suggestion
* DatasetArguments
```

src/llmcompressor/arg_parser/README.md
```diff
@@ -0,0 +1,45 @@
...
+* RecipeArguments
+* TrainingArguments
+
+
+## ModelArguments
+Handles model loading and saving. For example, `ModelArguments.model` can be a Hugging Face model identifier or an instance of `AutoModelForCausalLM`. The `save_compressed` flag is a boolean that determines whether the model is saved in compressed safetensors format to minimize disk usage.
+
+## DataArguments
```
rahul-tuli, line 37:
```suggestion
## DatasetArguments
```

src/llmcompressor/arg_parser/recipe_arguments.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+
+@dataclass
+class RecipeArguments:
```
brian-dellabetta, line 6: is the user-configured dataclass that gets "compiled" into something like `GPTQModifier` or `QuantizationModifier`? I'm wondering if these concepts can be merged?

src/llmcompressor/arg_parser/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+`llm-compressor` uses four parsers, located in `llm_compressor/arg_parser`:
+* ModelArguments
+* DataArguments
+* RecipeArguments
+* TrainingArguments
+
+
+## ModelArguments
+Handles model loading and saving. For example, `ModelArguments.model` can be a Hugging Face model identifier or an instance of `AutoModelForCausalLM`. The `save_compressed` flag is a boolean that determines whether the model is saved in compressed safetensors format to minimize disk usage.
```
rahul-tuli, line 35:
```suggestion
Handles model loading and saving. For example, `ModelArguments.model` can be a Hugging Face model identifier or an instance of `AutoModelForCausalLM`. The `save_compressed` flag is a boolean that determines whether the model is to be saved in compressed safetensors format to minimize disk usage.
```

brian-dellabetta (commented):
few comments for ya

src/llmcompressor/arg_parser/__init__.py
```diff
@@ -0,0 +1,6 @@
+# flake8: noqa
+
+from .data_arguments import DatasetArguments
```
rahul-tuli, line 3: rename data_arguments file to dataset_arguments?

src/llmcompressor/arg_parser/model_arguments.py
```diff
@@ -5,7 +5,9 @@
 @dataclass
 class ModelArguments:
     """
-    Arguments pertaining to which model/config/tokenizer we are going to fine-tune from
+    Model variables used for oneshot calibration, finetuning and
```
rahul-tuli, line 8: nit: Could we not use Model variables?

src/llmcompressor/arg_parser/model_arguments.py
```diff
@@ -83,3 +75,18 @@ class ModelArguments:
             "repositories you trust and in which you have read the code"
         },
     )
+    save_compressed: Optional[bool] = field(
+        default=True,
+        metadata={"help": "Whether to compress sparse models during save"},
+    )
+    oneshot_device: Optional[str] = field(
+        default="cuda:0",
```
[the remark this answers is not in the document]
brian-dellabetta: nm, after looking at the google doc it looks like this is pulled from another file that defaults to `cuda:0`. better to not change in that case

src/llmcompressor/arg_parser/data_arguments.py
```diff
@@ -67,10 +67,10 @@ class CustomDataTrainingArguments(DVCDatasetTrainingArguments):
 
 
 @dataclass
-class DataTrainingArguments(CustomDataTrainingArguments):
+class DatasetArguments(CustomDatasetArguments):
```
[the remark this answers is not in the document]
horheynm: Users will not use this class so should not need

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -672,7 +672,9 @@ def test_correct_compressor_inferred(
     if is_24:
         weights = _make_24_sparse(weights)
     else:
-        weights[0, :] = torch.ones((4,))  # guarantee not 24 sparse
+        weights[0, :] = torch.ones(
```
rahul-tuli, line 675: seems like needs a rebase?

rahul-tuli (changes requested):

src/llmcompressor/arg_parser/recipe_arguments.py
horheynm: Can you elaborate, I dont understand

examples/quantization_w4a16/llama3_example.py
```diff
@@ -5,7 +5,7 @@
 from llmcompressor.transformers import oneshot
 
 # Select model and load it.
-MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
+MODEL_ID = "TinyLlama/TinyLlama-1.1B-Chat-v0.1"
```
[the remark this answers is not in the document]
horheynm: Yes you are right

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
horheynm: Its from lint. should be fine

src/llmcompressor/arg_parser/model_arguments.py
horheynm: Why?

src/llmcompressor/arg_parser/__init__.py
horheynm: ok

src/llmcompressor/arg_parser/README.md
horheynm: If its confusing, I can rephrase

src/llmcompressor/arg_parser/data_arguments.py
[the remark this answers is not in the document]
dsikka: Wdym they wont use this class?

src/llmcompressor/arg_parser/recipe_arguments.py
brian-dellabetta: Discussed with @horheynm that is refactored from having `recipe` and `recipe_args` as inputs elsewhere, and not a net-new abstraction

brian-dellabetta (dismissed):
LGTM, but should re-request review from Dipika or Rahul before merging in

rahul-tuli (approved):

src/llmcompressor/arg_parser/data_arguments.py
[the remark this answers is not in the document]
horheynm: They wont import this class and use it in their scripts. Its embedded in oneshot callable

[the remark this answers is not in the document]
dsikka: users can do whatever they want. this is exactly the class that they will be looking at when setting up arguments for oneshot, whether they pass in the arguments from the class or the class itself.

src/llmcompressor/arg_parser/README.md
```diff
@@ -0,0 +1,45 @@
+# Argument parser for `oneshot`, `train`, `eval` entrypoints
+
+Argument parsers in `llm-compressor` define the input arguments required for various entrypoints, including `oneshot`, `train`, and `eval`.
```
dsikka, line 3: arg_parser makes it sound like it's a cli tool.
I would just call this args

dsikka, line 3: Why are we calling these argument parsers?

src/llmcompressor/arg_parser/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Each entrypoint carries out logic based on the provided input arguments, such as `model`, `recipe`, and `dataset`.
```
dsikka, line 5: What does this mean?

dsikka (changes requested):

src/llmcompressor/arg_parser/README.md
horheynm: Bc they are argument parsers no?

horheynm: Isn't this an arg parser?
https://github.com/huggingface/transformers/blob/main/src/transformers/hf_argparser.py

src/llmcompressor/arg_parser/README.md
horheynm: It means entrypoints (oneshot, train, etc), carries out their logic based on the provided input args, like model, recipe, ..

src/llmcompressor/arg_parser/README.md
dsikka: They are actually the parsing the args in that file. 
The files in this folder are just dataclasses. 

We should update the name

PR #1106, opened 2025-01-28
[Audio] Whisper Example and Readme

kylesayrs:
## Purpose ##
* Show example of quantizing whisper audio model

## Changes ##
* Add whisper audio model example
* Add traceable whisper definition (only need to comment out a value error check)
* The embedded audio is achieved using [github attached files](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files). While there's no official word on how long these files are maintained, if it is found that the file is deleted at some point, then we can replace it with a link to the file uploaded to the repo.

## Testing ##
Successfully quantized whisper models and generated reasonable sample outputs
* https://huggingface.co/nm-testing/whisper-tiny-W4A16-G128
* https://huggingface.co/nm-testing/whisper-large-v2-W4A16-G128


mgoin (approved):
Very nice work!!

PR #1108, opened 2025-01-28
[Oneshot Refactor] Rename get_shared_processor_src to get_processor_name_from_model

horheynm:
ORDER OF REVIEWS:
1. https://github.com/vllm-project/llm-compressor/pull/1108 <- current PR
2. https://github.com/vllm-project/llm-compressor/pull/1103
3. https://github.com/vllm-project/llm-compressor/pull/1109
4. https://github.com/vllm-project/llm-compressor/pull/1110

SUMMARY:
* Rename `get_shared_processor_src` to `get_processor_from_model`
* Appropriate signature on `initialize_processor_from_path`, where `teacher` should be optinal

TEST PLAN:
* Pass all existing tests
* Search `get_shared_processor_src` using pygrep
```bash
  3 function pygrep() {
  2     local search_term="$1"
  1     shift
126     local search_dirs="${*:-src examples tests}"                           
  1     grep -rn --include="*.py" -E "$search_term" $search_dirs
  2 }
```

PR #1109, opened 2025-01-28
[Oneshot refactor] Refactor initialize_model_from_path

horheynm:
ORDER OF REVIEWS:
1. https://github.com/vllm-project/llm-compressor/pull/1108
2. https://github.com/vllm-project/llm-compressor/pull/1103 
3. https://github.com/vllm-project/llm-compressor/pull/1109 <- current PR
4. https://github.com/vllm-project/llm-compressor/pull/1110

SUMMARY:
Refactor `initialize_model_from_path` to decouple `training_args` dependent logic and oneshot (non-training_args) logic. 

TEST PLAN:
* Pass all tests
* search `initialize_model_from_path` using `grep`


src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -177,38 +177,42 @@ def initialize_model_from_path(
...
-        AutoConfig.from_pretrained(
-            model_args.distill_teacher,
-            use_auth_token=True if model_args.use_auth_token else None,
-            tie_word_embeddings=model_args.tie_word_embeddings,
-            trust_remote_code=model_args.trust_remote_code_model,
+
+    last_checkpoint = None
+
+    if training_args is not None:
+        teacher_config = (
```
dsikka, line 184: Why not just combine all teacher-model related code in one section, i.e line 241 onwards?

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -162,9 +163,8 @@ def parse_args(**kwargs):
 
 def initialize_model_from_path(
     model_args: ModelArguments,
-    training_args: TrainingArguments,
+    training_args: Optional[TrainingArguments] = None,
```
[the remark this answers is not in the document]
horheynm: There is the Oneshot class pathway and main pathway. 
Its optional bc Oneshot main will use this method, but wont have training_args.

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: Yeah we can do taht

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -177,38 +177,70 @@ def initialize_model_from_path(
...
         )
-        if model_args.distill_teacher
-        else None
-    )
+
+        # Detect last checkpoint
+        last_checkpoint = detect_last_checkpoint(training_args, model_args=model_args)
+
+        # Set seed before initializing model
+        set_seed(training_args.seed)
```
kylesayrs, line 201: Seeds are used for more than just initializing models. This should be moved outside of this function

kylesayrs: And in general, seeds should be a data argument, not a training argument

kylesayrs (changes requested):
Please move `set_seed` outside of the `initialize_model_from_path` function. Otherwise lgtm.

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: Its a part of training args
https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L1044

horheynm:
> Please move `set_seed` outside of the `initialize_model_from_path` function. Otherwise lgtm.

Its a part of training args
https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L1044

horheynm:
> > Please move `set_seed` outside of the `initialize_model_from_path` function. Otherwise lgtm.
> 
> Its a part of training args, https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L1044

 and in the original code. 
https://github.com/vllm-project/llm-compressor/pull/1109/files#diff-4b4df7cb1f6042ff11af61f87b5d6dffd8e07a7ef182a8aa659566a5ee77fc04L198

If there is an argument to move it outside, we can do it in a follow up pr.

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -247,18 +270,7 @@ def initialize_model_from_path(
...
-            **teacher_kwargs,
-        )
-        if model_args.distill_teacher is not None
-        else None
-    )
-    if teacher is not None and "sequence_length" in teacher_kwargs:
-        teacher.seqlen = teacher_kwargs["sequence_length"]
-
-    return teacher, model_path, model
+    return model, teacher
```
brian-dellabetta, line 273: by changing the signature of this function, should we be concerned about breaking changes for any users?

brian-dellabetta (approved):
things look good to me, just breaking change concerns but looks like you addressed those in the slack thread with @robertgshaw2-redhat

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: Good question, we never use the model_path and theres only one location using it in text_geneation.py so should be ok!

kylesayrs (approved):

PR #1115, opened 2025-01-29
Return empty sparsity config if targets and ignores are empty

rahul-tuli:
This PR fixes an issue where a sparsity configuration could end up being empty under certain conditions. Specifically, if the global sparsity is greater than 0.05, but no individual layer has a sparsity greater than 0.5, we end up with an empty sparsity config.

To address this, we now ensure that an empty sparsity config is not added in such cases

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209272443107638


brian-dellabetta (approved):
👍

horheynm:
Would making a test for this `.from_pretrained` func take a while?

dsikka (approved):

PR #1116, opened 2025-01-30
[VLM] Update pixtral data collator to reflect latest transformers changes

kylesayrs:
## Purpose ##
* In transformers==4.48.0, the Pixtral processor was updated to not add an additional layer of wrapping for `pixel_values` (https://github.com/huggingface/transformers/pull/34801). This is more inline with how other processors handle multimodal inputs
  * Because previously the data_collator was being used to unwrap this unnecessary wrapping, attempting to quantize pixtral with transformers>=4.48.0 fails

## Changes ##
* Update pixtral data collator to match latest transformers version
* Add comment for those who want to use transformers<4.48.0

## Testing ##
* Ran pixtral example to completion, @shubhra ran pixtral large

### DIFF
--- a/examples/multimodal_vision/pixtral_example.py
+++ b/examples/multimodal_vision/pixtral_example.py
@@ -16,18 +16,20 @@ processor = AutoProcessor.from_pretrained(model_id, trust_remote_code=True)
 
 # Oneshot arguments
 DATASET_ID = "flickr30k"
-DATASET_SPLIT = {"calibration": "test[:512]"}
-NUM_CALIBRATION_SAMPLES = 512
+NUM_CALIBRATION_SAMPLES = 1  # 512
+DATASET_SPLIT = {"calibration": f"test[:{NUM_CALIBRATION_SAMPLES}]"}
 MAX_SEQUENCE_LENGTH = 2048
 
 
 # Define a oneshot data collator for multimodal inputs.
+# NOTE: for transformers<4.48.0, please squeeze the first dimension of `pixel_values`
+# by appending `[0]` to the end of line 32
 def data_collator(batch):
     assert len(batch) == 1
     return {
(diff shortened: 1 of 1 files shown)


examples/multimodal_vision/pixtral_example.py
```diff
@@ -16,18 +16,20 @@
...
 DATASET_ID = "flickr30k"
-DATASET_SPLIT = {"calibration": "test[:512]"}
 NUM_CALIBRATION_SAMPLES = 512
+DATASET_SPLIT = {"calibration": f"test[:{NUM_CALIBRATION_SAMPLES}]"}
 MAX_SEQUENCE_LENGTH = 2048
 
 
 # Define a oneshot data collator for multimodal inputs.
+# NOTE: for transformers<4.48.0, please squeeze the first dimension of `pixel_values`
+# by appending `[0]` to the end of line 32
```
brian-dellabetta, line 26: Does this mean we should require `transformers>4.48.0,>5.0` in setup.py?

kylesayrs: The current policy is to leave the transformers version largely unbounded and to instead ask users to upgrade to the latest transformers version

brian-dellabetta: Definitely creates a wrinkle for us if transformers introduced a breaking change between 4.48 and 4.49. if we hard-code an expectation they're not expected to follow, could cause downstream errors. Does that mean we need to gnarly `if transformers.__version__ < "4.49"` logic in here? or maybe as a note up top that this example should only be run with 4.49+, the surface area is pretty limited to an example file

brian-dellabetta (approved):
👍

dsikka (approved):

dsikka (approved):

PR #1120, opened 2025-02-03
[DataInputType] Decouple downstream func from data_args

horheynm:
SUMMARY:
Decouple `data_args` from `get_raw_dataset`. 

`get_raw_dataset` currently depends on hardcoded attributes of `data_args` - ex. trust_remote_code, streaming - which is only used for `load_dataset`

Currently, 'data_args` governs 
1. raw dataset
2. processed dataset
3. data loader

CHANGES:
* Decouple data_args to to 1, 2, 3 above.  For 1. Introduce `LoadDatasetArguments` dataclass and make the current code compatible. 
* Add tests to for `load_dataset` for 1. hf_stub, 2. hf_cache_path 3. local dataset path.

Notes:
Precursor PR for Dataset refactor. The attributes of DatasetArguments attributes will be organized into their own dataclass. 

TEST PLAN:
Pass existing tests

### DIFF
--- a/src/llmcompressor/datasets/utils.py
+++ b/src/llmcompressor/datasets/utils.py
@@ -20,9 +20,12 @@ def get_raw_dataset(
         local file directory in csv, json, parquet, etc.
         If local path is provided, it must be
             1. Download path where HF dataset was downloaded to
-            2. Path containing (train, test, validation) with the same extention.
-            Supported extentions are json, jsonl, csv, arrow, parquet, text,
-            and xlsx,
+            2. File path containing any of train, test, validation in its name
+            with the supported extentions: json, jsonl, csv, arrow, parquet, text,
+            and xlsx. Ex. foo-train.csv, foo-test.csv
+
+            If a custom name is to be used, its mapping can be specified using
+            `data_files` input_arg.
 
     :return: the requested dataset
 
--- a/src/llmcompressor/datasets/utils.py
+++ b/src/llmcompressor/datasets/utils.py
@@ -1,0 +1,33 @@
+from typing import Union
+
+from datasets import (
+    Dataset,
+    DatasetDict,
+    IterableDataset,
+    IterableDatasetDict,
+    load_dataset,
+)
+
+
+def get_raw_dataset(
+    path: str,
+    **kwargs,
+) -> Union[DatasetDict, Dataset, IterableDatasetDict, IterableDataset]:
+    """
+    Load HF alias or a dataset in local path.
+
+    :param path: Path or name of the dataset. Accepts HF dataset stub or
+        local file directory in csv, json, parquet, etc.
+        If local path is provided, it must be
+            1. Download path where HF dataset was downloaded to
+            2. Path containing (train, test, validation) with the same extention.
+            Supported extentions are json, jsonl, csv, arrow, parquet, text,
+            and xlsx,
+
+    :return: the requested dataset
+
+    """
+    return load_dataset(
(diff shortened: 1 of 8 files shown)


src/llmcompressor/datasets/utils.py
```diff
@@ -0,0 +1,36 @@
...
+from datasets import (
+    Dataset,
+    DatasetDict,
+    IterableDataset,
+    IterableDatasetDict,
+    load_dataset,
+)
+
+
+def get_raw_dataset(
```
kylesayrs, line 12: Why even maintain this function if it just transparently calls `load_dataset`?

kylesayrs:
I'm still a little confused as to why this is necessary? It comes at the cost of adding another data class, I don't really understand the benefit.

> get_raw_dataset is downstream and should not depend on parsing to data_args first

As implemented now (and I'm guessing in the future), we always parse data_args before downloading datasets. Is there a case where we would *not* want to depend on parsing data_args first?

horheynm:
> I'm still a little confused as to why this is necessary? It comes at the cost of adding another data class, I don't really understand the benefit.
> 
> > get_raw_dataset is downstream and should not depend on parsing to data_args first
> 
> As implemented now (and I'm guessing in the future), we always parse data_args before downloading datasets. Is there a case where we would _not_ want to depend on parsing data_args first?

Where is the data class?
Were parsing the input arg to a more compatible format to `load_dataset`. 
Currently the func depends on data_args, where some are hardcoded to be included vs the new support which will be used in `load_dataset_args`

kylesayrs (commented):
1. Please rename `load_dataset_args` to `load_dataset_kwargs`
2. Please do not create a new data class `LoadDatasetArguments`, instead put `load_dataset_kwargs` directly on `DatasetArguments`
4. Please remove `raw_kwargs`, as it's now been replaced by `load_dataset_kwargs`

src/llmcompressor/datasets/utils.py
horheynm, line 12: remove

kylesayrs: The imported `datasets.load_dataset` function can be our "central location" that loads the dataset. This function adds nothing.

src/llmcompressor/transformers/finetune/data/data_args.py
```diff
@@ -4,6 +4,23 @@
 from transformers import DefaultDataCollator
 
 
+@dataclass
+class LoadDatasetArguments:
+    """
+    Arguments for using load_dataset
+    """
+
+    load_dataset_args: Dict[str, Any] = field(
```
kylesayrs, line 13:
```suggestion
    load_dataset_kwargs: Dict[str, Any] = field(
```

src/llmcompressor/datasets/utils.py
horheynm: There should be a reason why it exists currently. Keeping that structure should be respected. 

Looks like this func had other roles and it was broken down to its current format.

horheynm:
> 1. Please rename `load_dataset_args` to `load_dataset_kwargs`
> 2. Please do not create a new data class `LoadDatasetArguments`, instead put `load_dataset_kwargs` directly on `DatasetArguments`
> 3. Please remove `raw_kwargs`, as it's now been replaced by `load_dataset_kwargs`

For 2. The base is populated with tons of args, would not know which arg is used where. Better to keep it organized by having a class.

PR #1121, opened 2025-02-04
Remove smoothquant from examples

kylesayrs:
## Purpose ##
* When writing the multimodal blog, I initially included smoothquant in the examples. However I later decided to remove this and focus on W4A16 in order to maintain conciseness for the blog. This PR updates the examples to reflect this


dsikka:
We do not need to tie examples with the blog. The examples on main can (and likely should) be an extension

kylesayrs:
@dsikka This is also because smoothquant smooths activation outliers, which is not relevant for W4A16. We should remove examples that use this combination.

dsikka:
Oh sorry, I thought we were showing w8a8

dsikka (approved):

PR #1124, opened 2025-02-04
Use unique test names in `TestvLLM`

dbarbuzzi:
SUMMARY:
This PR updates the `TestvLLM` class to pass the config file being tested as a parameter to the test. Functionally, this is no different than the existing behavior with regards to the test, but practically, it means that the config file will be included in the generated test name making it much easier to identify which config file is being tested across the board (in console output, in Testmo/reporting tools, etc.).

As a result, there is no longer a need for the previously added fixture that embedded the config file into the test suite metadata for downstream processing.

TEST PLAN:
A custom test run executing the e2e tests against this branch is here: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/13138573134/job/36659921426
Reviewing the output of the 'run e2e tests' steps, the test names are updated:
* Old: `TestvLLM.test_vllm`
* New: `TestvLLM_0_tests_e2e_vLLM_configs_fp8_dynamic_per_token_yaml.test_vllm`

**Note:** The failures (`RuntimeError: Failed to infer device type`) are expected due to a recent change (seemingly on the vLLM side).

### DIFF
--- a/tests/e2e/vLLM/test_vllm.py
+++ b/tests/e2e/vLLM/test_vllm.py
@@ -2,12 +2,12 @@ import os
 import re
 import shutil
 from pathlib import Path
-from typing import Callable
 
 import pytest
 import yaml
 from huggingface_hub import HfApi
 from loguru import logger
+from parameterized import parameterized_class
 
 from llmcompressor.core import active_session
 from tests.e2e.e2e_utils import run_oneshot_for_e2e_testing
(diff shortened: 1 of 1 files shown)


brian-dellabetta (approved):
I didn't know you could do this, awesome!

rahul-tuli (approved):

PR #1125, opened 2025-02-05
Unpin `pynvml` to fix e2e test failures with vLLM

dsikka:
SUMMARY:
- vLLM throws a runtime error if pynvml < 12.0 is installed 
- Unpin pynvml version to prevent error and allow e2e tests to run

TEST PLAN:
- e2e tests now run


kylesayrs (approved):
Makes sense, we're not using any super volatile features of this library anyways

PR #1126, opened 2025-02-05
Properly restore training mode with `eval_context`

kylesayrs:
## Purpose ##
* Remove side effect of leaving a model in eval mode when calibrating

## Changes ##
* Use `eval_context`, which stores the original training state of the model
* Fix some type hints


src/llmcompressor/utils/helpers.py
```diff
@@ -1062,7 +1064,7 @@ class DisableKVCache:
     ...     output = model(input)
     """
 
-    def __init__(self, model: torch.nn.Module):
+    def __init__(self, model: PreTrainedModel):
```
brian-dellabetta, line 1067: why this change here? Can `torch.nn.Module`s and `transformers.PreTrainedModel`s be used interchangeably?

brian-dellabetta (dismissed):
cool! one question but good to merge after sanity checking

src/llmcompressor/utils/helpers.py
kylesayrs: `PreTrainedModel` inherits from `torch.nn.Module`. This type narrowing relates to the fact that this context is only meant to be used with HF Transformers `PreTrainedModel`s, as it references capabilities specific to models such as having a config and disabling the HF cache

src/llmcompressor/utils/helpers.py
```diff
@@ -1087,31 +1089,42 @@ def __exit__(self, _exc_type, _exc_val, _exc_tb):
...
+
+@contextlib.contextmanager
+def calibration_forward_context(model: PreTrainedModel):
     """
     Context in which all calibration forward passes should occur.
 
     - Remove gradient calculations
     - Disable the KV cache
     - Disable quantization from QuantizationModifier
+    - Disable train mode and enable eval mode
```
dsikka, line 1122: Since we're updating this, can we update the comment about `Disable quantization from QuantizationModifier` to disable QDQ during forward passes? Slightly more accurate in what it disables.

kylesayrs: Good call

src/llmcompressor/utils/helpers.py
```diff
@@ -1087,31 +1089,42 @@ def __exit__(self, _exc_type, _exc_val, _exc_tb):
...
-        model.apply(enable_quantization)
+        module.apply(enable_quantization)
 
 
 @contextlib.contextmanager
-def calibration_forward_context(model: torch.nn.Module):
+def eval_context(module: torch.nn.Module):
+    restore_value = module.training
+    try:
+        module.train(False)  # equivalent to eval()
```
rahul-tuli, line 1107: But why not just call `model.eval()` if we're restoring anyway?

src/llmcompressor/utils/helpers.py
```diff
@@ -1087,31 +1089,42 @@ def __exit__(self, _exc_type, _exc_val, _exc_tb):
...
+    - Disable quantization during forward pass
+    - Disable train mode and enable eval mode
     """
-    model.eval()
-
     with (
         torch.no_grad(),
         DisableKVCache(model),
         DisableQuantization(model),
+        eval_context(model),
```
rahul-tuli, line 1128: I love this context manager chaining in python!

src/llmcompressor/utils/helpers.py
```diff
@@ -1087,31 +1089,42 @@ def __exit__(self, _exc_type, _exc_val, _exc_tb):
 
 
 @contextlib.contextmanager
-def DisableQuantization(model: torch.nn.Module):
+def DisableQuantization(module: torch.nn.Module):
```
horheynm, line 1092:
```suggestion
def distable_quantization(...)
```
I do prefer module here

src/llmcompressor/utils/helpers.py
kylesayrs: I pref `.train(False)` just because it's just the most clear and verbose, but as mentioned, it's exactly equivalent to `eval()`

src/llmcompressor/utils/helpers.py
kylesayrs: @horheynm This comment is very confusing.

1. If you prefer `module`, then that's good because it has been changed to `module.
2. Why did you change the style of the function name? (it has a typo btw)

horheynm:
In general is there a bug or feature that this pr is fixing? Whats the problem we have now to include this feature?

src/llmcompressor/utils/helpers.py
horheynm: 1. module from model is better
2. You have a context manager that uses both capitalized and snakecase. Why is this capitalized?

kylesayrs:
@horheynm I cannot point to specific issue which is caused by not restoring the training mode, but we're moving towards LLM Compressor being used in downstream training scripts, so it's good to harden out these side effects early before we encounter issues caused by them later.

src/llmcompressor/utils/helpers.py
horheynm: Usually capitalized context managers are for classes that have the entry and exit ex. ExitStack
https://asyncstdlib.readthedocs.io/en/stable/source/api/contextlib.html#asyncstdlib.contextlib.ExitStack

src/llmcompressor/utils/helpers.py
brian-dellabetta: oh i didn't know that! ok cool. yeah i like what you've done for var names, `model: PreTrainedModel` and `module: torch.nn.Module`

src/llmcompressor/utils/helpers.py
kylesayrs: > You have a context manager that uses both capitalized and snakecase. Why is this capitalized?

Please read the comment I just sent

dsikka (approved):

src/llmcompressor/utils/helpers.py
horheynm: This doesnt explain anything

PR #1130, opened 2025-02-07
[Bugfix] SparseGPT, Pipelines

kylesayrs:
## Purpose ##
* SparseGPT
  * Fix behavior where `targets` specifies which modules to sparsity, not which layers to target
  * Fix broken behavior with `_infer_owl_layer_sparsity` and add test
  * Fix owl argument validation
  * Add type hints and abstract methods for clarity
* Pipelines
  * Fix bug revealed by decorators added to the llama model definition in the latest transformers release
    * https://github.com/huggingface/transformers/pull/35757
    * For the sequential pipeline, this revealed a bug in torch.fx._symbolic_trace where wrapped functions were not being handled properly
    * Future work could involve upstreaming a bug fix
  * Fix issue caused by changes to llama model definition
    * https://github.com/huggingface/transformers/pull/34858
    * For the layer sequential pipeline, this challenges the assumption that each layer input is the previous layer's output (which was known to be a fragile assumption)
  * Fix issue related to basic pipeline slowdowns and inaccuracy

## Changes ##
* SparseGPT
  * Fully separate `targets` and `sequential_targets`
    * Modify hooks adding logic to reflect this change
  * Fix behavior of `_infer_owl_layer_sparsity` and add test
  * Code clarity
    * Add additional type hints
    * Designate `calibrate_module` as an abstract method on the sgpt mixin
* Pipelines
  * Sequential pipeline: unwrap model forward function to avoid issues with pytorch function patching
  * Layer Sequential Pipeline: Add `maybe_inject_pos_embeddings` to sequential pipeline to hackily support models with `position_embeddings`
  * Basic Pipeline: Fix `on_sequential_batch_end` to call on the end of epoch, rather than every batch
    * Calling every batch was likely causing slowdowns

## Followups ##
* Remove deprecated `sequential_update` option from examples and tests

## Testing ##
* Added `tests/llmcompressor/transformers/obcq/test_obcq_owl.py`
* Tested OBCQ+llama with sequential, layer sequential, and basic pipelines independently

## Regression Evaluations ##
Models were compressed using `examples/sparse_2of4_quantization_fp8/llama3_8b_2of4.py` without fp8 option

<details><summary>sparsegpt</summary>

Main
```
vllm (pretrained=/home/kyle/llm-compressor/Meta-Llama-3-8B-InstructSparseGPTModifierMAIN,dtype=bfloat16,add_bos_token=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 1
|  Tasks   |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|winogrande|      1|none  |     5|acc   |↑  |0.6243|±  |0.0136|
```

This branch

```
vllm (pretrained=/home/kyle/llm-compressor/Meta-Llama-3-8B-InstructSparseGPTModifierFEATURE,dtype=bfloat16,add_bos_token=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 1
|  Tasks   |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|winogrande|      1|none  |     5|acc   |↑  |0.6306|±  |0.0136|
```
</details>

To test wanda, the `SparseGPTModifier` was replaced with the `WandaPruningModifier`

<details><summary>wanda</summary>

Main
```
vllm (pretrained=/home/kyle/llm-compressor/Meta-Llama-3-8B-InstructWandaPruningModifierMAIN,dtype=bfloat16,add_bos_token=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 1
|  Tasks   |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|winogrande|      1|none  |     5|acc   |↑  |0.5912|±  |0.0138|
```

This branch
```
vllm (pretrained=/home/kyle/llm-compressor/Meta-Llama-3-8B-InstructWandaPruningModifierFEATURE,dtype=bfloat16,add_bos_token=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 1
|  Tasks   |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|winogrande|      1|none  |     5|acc   |↑  |0.5817|±  |0.0139|
```
</details>


src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -79,6 +80,9 @@ def run_pipeline(
                     output = layer(**inputs)
 
                     if layer_index < num_layers - 1:
-                        output = to_next_layer_kwargs(output, layers[layer_index + 1])
+                        next_layer = layers[layer_index + 1]
+                        output = to_next_layer_kwargs(output, next_layer)
```
dsikka, line 84: slightly unrelated but general question: considering one of the assumptions of this pipeline is the flow of inputs/outputs, what was the need for adding `to_next_layer_kwargs` initially?

src/llmcompressor/pipelines/layer_sequential/helpers.py
```diff
@@ -126,3 +131,27 @@ class EarlyStopException(Exception):
...
     _kwargs: Dict[str, Any]
+
+
+def maybe_inject_pos_embeddings(
+    output: Dict[str, Any],
+    next_layer: Module,
+    inputs: Dict[str, Any],
+) -> Dict[str, Any]:
+    """
+    As of https://github.com/huggingface/transformers/pull/34858, positional embeddings
```
dsikka, line 142: annoying

src/llmcompressor/modifiers/obcq/base.py
```diff
@@ -85,7 +85,8 @@ class SparseGPTModifier(SparsityModifierMixin, Modifier):
     # data pipeline arguments
     sequential_update: Optional[bool] = False  # deprecated
     sequential_targets: Union[str, List[str], None] = None
-    targets: Union[str, List[str], None] = None  # alias sequential_targets
+    targets: Union[str, List[str]] = ["Linear"]
+    ignore: List[str] = Field(default_factory=list)
```
dsikka, line 89: We're replicating `ignore` for code clarity?

src/llmcompressor/modifiers/pruning/wanda/base.py
```diff
@@ -69,8 +69,9 @@ class WandaPruningModifier(SparsityModifierMixin, Modifier):
 
     # data pipeline arguments
     sequential_update: Optional[bool] = False  # deprecated
-    sequential_targets: Union[str, List[str], None] = None
-    targets: Union[str, List[str], None] = None  # alias sequential_targets
+    sequential_targets: Union[str, List[str]] = None
+    targets: Union[str, List[str], None] = ["Linear"]
+    ignore: List[str] = Field(default_factory=list)
```
dsikka, line 74: same comment

src/llmcompressor/utils/helpers.py
```diff
@@ -1128,3 +1129,12 @@ def calibration_forward_context(model: PreTrainedModel):
         eval_context(model),
     ):
         yield
+
+
+@contextlib.contextmanager
```
dsikka, line 1134: I'm not sure I understand what this contextmanager is accomplishing

dsikka (changes requested):
I don't agree with the notion that adding the same variables through inheritance helps with code clarity or user readability.

Updating the docstring in the inherited classes is sufficient rather than relying on redundant code.

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -40,6 +40,6 @@ def run_pipeline(
             batch = tensors_to_device(batch, model_device)
             model(**batch)
 
-            # TODO: replace with a lifecycle event
-            if callback_modifier:
-                callback_modifier.on_sequential_batch_end()
+        # TODO: replace with a lifecycle event
+        if callback_modifier:
+            callback_modifier.on_sequential_batch_end()
```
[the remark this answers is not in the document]
kylesayrs: I agree, I'll name it something like "sequential_epoch_end" when callbacks are properly implemented

src/llmcompressor/pipelines/layer_sequential/pipeline.py
kylesayrs: For the sake of standardization, we only pass activations as kwargs. `to_next_layer_kwargs` simply assigns kwarg names to the unnamed but ordered args from the previous layer

src/llmcompressor/pipelines/layer_sequential/helpers.py
kylesayrs: This comment? The transformer changes? Both?

dsikka: Transformer change lol

src/llmcompressor/modifiers/obcq/base.py
kylesayrs: Yes, the goal being that a user can look at a modifier and see its arguments (and therefore how to use it) without having to check all of its inheritances.

This could be less necessary in the future when we have dedicated buildthedocs

src/llmcompressor/utils/helpers.py
kylesayrs: As in you don't see its usefulness as a general utility, or you don't understand how/why it's being used for this PR?

src/llmcompressor/pipelines/layer_sequential/helpers.py
kylesayrs: Yes, I wish they had kept the backwards compatibility though

src/llmcompressor/utils/helpers.py
dsikka: Sorry, I was reading it independent of where it was actually being used.

Seems like this is to maintain the original forward method, in the example that it is being used in?

Can you expand on what you mean by "unwrapping the function wrappers"? This is unwrapping the forward method?

src/llmcompressor/modifiers/obcq/sgpt_mixin.py
```diff
@@ -34,9 +36,15 @@ class SparsityModifierMixin(HooksMixin):
     # data pipeline arguments
     sequential_update: Optional[bool] = False  # deprecated
     sequential_targets: Union[str, List[str], None] = None
-    targets: Union[str, List[str], None] = None  # alias sequential_targets
+    targets: Union[str, List[str]] = ["Linear"]
```
dsikka, line 39: Just based on your PR description, can you point to where targets and/or sequential_targets are being used incorrectly?

src/llmcompressor/utils/helpers.py
kylesayrs: https://docs.python.org/3/library/inspect.html#inspect.unwrap
> inspect.unwrap(func, *, stop=None)
> Get the object wrapped by func. It follows the chain of __wrapped__ attributes returning the last object in the chain.

https://github.com/huggingface/transformers/blob/main/src/transformers/utils/deprecation.py#L37
> Returns:
> Callable:
> A wrapped function that handles the deprecated keyword arguments according to the specified parameters.

The bug is basically that torch.fx._symbolic_trace has bugged logic where it accidentally tries to replace partial code from the unwrapped function (forward) with partial code from the the wrapping function (deprecate_kwarg).

https://github.com/pytorch/pytorch/blob/2a55311773bfb9e569aa672ac3322d21abc1af32/torch/fx/_symbolic_trace.py#L677

I think this should be `root_fn = _patch_function(fn_for_analysis, len(args))`. In any case, unwrapping prior to starting the trace fixes this issue.

src/llmcompressor/modifiers/obcq/sgpt_mixin.py
kylesayrs: https://github.com/vllm-project/llm-compressor/pull/1130/files#diff-1f12d6a6c392af55d1a18a3e4f6e790a537c6011897b699819edc547f2f0e388L37

I initially implemented `sequential_targets` and `targets` as aliases of each other because I thought this was how the original implementation treated them, in reality they are treated separately (as they should be)

https://github.com/vllm-project/llm-compressor/pull/1130/files#diff-1f12d6a6c392af55d1a18a3e4f6e790a537c6011897b699819edc547f2f0e388L124-R130

Again here I implemented sequential targets and targets as being synonymous, but I reverted the behavior to be the same as before removing the layer compressor, where layer sparsities map to `targets`, not `sequential_targets`.

rahul-tuli (approved):
LGTM!

dsikka (approved):

PR #1137, opened 2025-02-11
Add timer support for testing

dsikka:
SUMMARY:
- Add timer to our e2e testing to provide support when we see regression across oneshot, compressed saving or vLLM generation
- The csv produced at the end of each test case has the same name as the model produced, apart from the -e2e substring. For example:

For the following model produced from one test case: https://huggingface.co/nm-testing/TinyLlama-1.1B-Chat-v1.0-FP8_DYNAMIC-e2e

The following csv will be produced:  `TinyLlama-1.1B-Chat-v1.0-FP8_DYNAMIC.csv` and the names will be unique as a result.

The contents of the csv:

| _load_model_and_tokenizer| _run_oneshot | _save_compressed_model | _run_vllm |
|------------------------------| ----------------| -------------| -------------|
| 1.7816431522369385|1.51918625831604|7.086644649505615|26.71957492828369| 

- For each function that has its timing recorded, we have a column (e.g in the e2e test, we record timings for _load_model_and_tokenizer, _run_oneshot, _save_compressed_model, and, _run_vllm).
- The values of the columns indicate the time it took to execute the function, each time it was called. 
- In our e2e tests, we only call each of the four functions once, so we only have 1 row.
- The csv is just saved to the directory defined by the `TIMINGS_DIR` env variable

TEST PLAN:
- Ran e2e test cases locally and csv produced expected timing breakdown
- csv saves all logged timings to disk as part of the teardown

### DIFF
--- a/tests/e2e/e2e_utils.py
+++ b/tests/e2e/e2e_utils.py
@@ -21,7 +21,7 @@ def _load_model_and_tokenizer(
 
 
 @log_time
-def _run_onshot(device: str, **oneshot_kwargs):
+def _run_oneshot(device: str, **oneshot_kwargs):
     oneshot(
         **oneshot_kwargs,
         oneshot_device=device,
@@ -70,6 +70,6 @@ def run_oneshot_for_e2e_testing(
 
     # Apply quantization.
     logger.info("ONESHOT KWARGS", oneshot_kwargs)
-    _run_onshot(device=device, **oneshot_kwargs)
+    _run_oneshot(device=device, **oneshot_kwargs)
 
     return oneshot_kwargs["model"], tokenizer
--- a/tests/e2e/e2e_utils.py
+++ b/tests/e2e/e2e_utils.py
@@ -7,7 +7,28 @@ from llmcompressor.transformers import oneshot
 from tests.testing_utils import preprocess_tokenize_dataset
 from test.timer.timer_utils import log_time
 
+
 @log_time
+def load_model_and_tokenizer(
+    model: str,
+    device: str,
+):
+    loaded_model = AutoModelForCausalLM.from_pretrained(
+        model, device_map=device, torch_dtype="auto"
+    )
+    tokenizer = AutoTokenizer.from_pretrained(model)
+    return loaded_model, tokenizer
+
+@log_time
+def run_onshot(
+    device: str, **oneshot_kwargs
+):
+    oneshot(
+        **oneshot_kwargs,
+        oneshot_device=device,
+    )
+
+
 def run_oneshot_for_e2e_testing(
     model: str,
     device: str,
@@ -22,10 +43,11 @@ def run_oneshot_for_e2e_testing(
 ):
     # Load model.
     oneshot_kwargs = {}
-    loaded_model = AutoModelForCausalLM.from_pretrained(
-        model, device_map=device, torch_dtype="auto"
+   
+    load_model_and_tokenizer = load_model_and_tokenizer(
+        model=model,
+        device=device
     )
-    tokenizer = AutoTokenizer.from_pretrained(model)
 
     if dataset_id:
         ds = load_dataset(dataset_id, name=dataset_config, split=dataset_split)
@@ -52,8 +74,9 @@ def run_oneshot_for_e2e_testing(
 
     # Apply quantization.
     logger.info("ONESHOT KWARGS", oneshot_kwargs)
-    oneshot(
-        **oneshot_kwargs,
-        oneshot_device=device,
+    run_onshot(
+        device=device,
+        **oneshot_kwargs
     )
+
     return oneshot_kwargs["model"], tokenizer
--- a/tests/e2e/e2e_utils.py
+++ b/tests/e2e/e2e_utils.py
@@ -5,11 +5,11 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 from llmcompressor.modifiers.quantization import GPTQModifier, QuantizationModifier
 from llmcompressor.transformers import oneshot
 from tests.testing_utils import preprocess_tokenize_dataset
-from test.timer.timer_utils import log_time
+from tests.test_timer.timer_utils import log_time
 
 
 @log_time
-def load_model_and_tokenizer(
+def _load_model_and_tokenizer(
     model: str,
     device: str,
 ):
@@ -19,10 +19,9 @@ def load_model_and_tokenizer(
     tokenizer = AutoTokenizer.from_pretrained(model)
     return loaded_model, tokenizer
 
+
 @log_time
-def run_onshot(
-    device: str, **oneshot_kwargs
-):
+def _run_onshot(device: str, **oneshot_kwargs):
     oneshot(
         **oneshot_kwargs,
         oneshot_device=device,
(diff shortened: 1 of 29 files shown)


horheynm (dismissed):
LGTM

tests/e2e/e2e_utils.py
```diff
@@ -4,9 +4,30 @@
...
+):
+    loaded_model = AutoModelForCausalLM.from_pretrained(
+        model, device_map=device, torch_dtype="auto"
+    )
+    tokenizer = AutoTokenizer.from_pretrained(model)
+    return loaded_model, tokenizer
+
+
+@log_time
+def _run_onshot(device: str, **oneshot_kwargs):
```
rahul-tuli, line 24: nit:
```suggestion
def _run_oneshot(device: str, **oneshot_kwargs):
```

tests/test_timer/timer.py
```diff
@@ -0,0 +1,69 @@
...
+        if self.measurements.get(func_name) is None:
+            self.measurements[func_name] = [time_elapased]
+        else:
+            self.measurements[func_name].append(time_elapased)
+
+    def _compute_average(self):
+        """
+        Display averages time for each logged function. After the average is
+        calculated, clear the measurements.
+
```
rahul-tuli, line 63:
```suggestion
```

tests/test_timer/timer_utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def log_time(func):
+    """
+    Decorator to time functions. Times for the function are stored using
+    the class and function names.
+    """
+
+    def wrapper(*args, **kwargs):
+        if args and isinstance(args[0], object):
```
rahul-tuli, line 23: I think everything is an object? so the isinstance check is a no-op

```python
>>> isinstance(1, object)
True
```

tests/test_timer/timer_utils.py
```diff
@@ -0,0 +1,39 @@
...
+        Timer._instance = Timer(enable_logging=enable_logging)
+    return Timer._instance
+
+
+def log_time(func):
+    """
+    Decorator to time functions. Times for the function are stored using
+    the class and function names.
+    """
+
```
rahul-tuli, line 21: could we wrap it in `functool.wraps(...)` it's super helpful when looking at stack traces when debugging

tests/test_timer/timer.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+
+    _instance = None
+
+    def __init__(self, enable_logging: bool):
+        """
+        :param enable_logging: whether or not time logging is enabled
+        """
+        self.enable_logging = enable_logging
+        self.measurements = {}
```
rahul-tuli, line 25: nit: We could make it a `defaultdict`, but we can ignore since these are tests

tests/e2e/vLLM/test_vllm.py
```diff
@@ -3,6 +3,7 @@
 import shutil
 from pathlib import Path
 
+import pandas as pd
```
rahul-tuli, line 6: nit: Thoughts on adding to dev requirements? we should be fine for now as `datasets` installs pandas

rahul-tuli (approved):
Great Job!

tests/test_timer/timer_utils.py
dsikka: Yeah I’m gonna clean this up. Realized there’s some leftover code as well with the func_name

tests/e2e/vLLM/test_vllm.py
horheynm: is dev and test the same?

dbarbuzzi: We don't have a `test` extra, only `dev` which includes all the test requirements.

dsikka: Yeah we could create a test extra.

tests/test_timer/timer.py
horheynm: This will be helpful to instantiate the default diict with a list for appending the tim elapsed.

tests/e2e/vLLM/test_vllm.py
dsikka: Adding to dev requirements for now. Will separate our test and dev requirements in a follow-up as slightly bigger change/will impact ci workflows.

tests/test_timer/timer.py
```diff
@@ -0,0 +1,70 @@
...
+import time
+from collections import defaultdict
+from contextlib import contextmanager
+from threading import RLock
+
+import numpy
+
+__all__ = ["Timer"]
+
+_LOGGER = logging.getLogger(__name__)
```
kylesayrs, line 11: nit: I think we're standardizing on loguru

tests/test_timer/timer_utils.py
```diff
@@ -0,0 +1,35 @@
+from functools import wraps
+
+from tests.test_timer import Timer
+
+__all__ = ["log_time", "get_singleton_manager"]
+
+
+def get_singleton_manager(enable_logging: bool = True):
```
kylesayrs, line 8: nit: can we use this
```
class Timer:
    _instance = None

    @staticmethod
    def getTimer(cls) -> Timer:
        if cls._instance is None:
            cls._instance = Timer()
        return cls._instance
```

tests/test_timer/timer_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Return the Timer. If not has not yet been initialized, initialize and
+    return. If it has, return the existing Timer.
+    """
+    if Timer._instance is None:
+        Timer._instance = Timer(enable_logging=enable_logging)
+    return Timer._instance
+
+
+def log_time(func):
```
kylesayrs, line 18: nit

```python3
class Timer:
    _instance = None

    @staticmethod
    def log_time(cls, func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            TIMER_MANAGER = cls.getTimer()
            func_name = func.__name__

            if not TIMER_MANAGER.enable_logging:
                return func(*args, **kwargs)

            with TIMER_MANAGER.time(func_name):
                return func(*args, **kwargs)

        return wrapper
```

dbarbuzzi:
Can you expand on the CSV file contents and location (directory, filename)?

Regarding location, we currently run the script that runs all the various configs for this test in sequence. I _assume_ that this will not cause any existing files to be overwritten (e.g. each test's CSV file(s) will be uniquely named), but it also seems like they might end up scattered across different directories? Since they seem more like a test artifact than a model/etc. one, it seems like it might make more sense to gather them in a consolidated directory (which would also make finding them all programmatically after the tests conclude much simpler.

As far as contents, just an example of the contents after a single test would be helpful.

dsikka:
> Can you expand on the CSV file contents and location (directory, filename)?
> 
> Regarding location, we currently run the script that runs all the various configs for this test in sequence. I _assume_ that this will not cause any existing files to be overwritten (e.g. each test's CSV file(s) will be uniquely named), but it also seems like they might end up scattered across different directories? Since they seem more like a test artifact than a model/etc. one, it seems like it might make more sense to gather them in a consolidated directory (which would also make finding them all programmatically after the tests conclude much simpler.
> 
> As far as contents, just an example of the contents after a single test would be helpful.

Yup, let me update the PR description to add more colour to what we're producing

dbarbuzzi:
> we can update this to keep one defined location

Customizable (e.g., via env var) would be nice, but even just a dedicated root folder (maybe something like `timings`) would be ideal, and if we were to expand, we could use subfolders of the root folder (e.g., the `test_vllm` timings might go into `timings/e2e-test_vllm` or something — not to get _too_ nested).

I think aside from that, this all looks good to me.

dbarbuzzi (approved):
Thanks for adding the ability to control the timing CSV output dir via env var, that will be very helpful, especially if we expand the usage of this timer to other tests.

Looks good to me!

dhuangnm:
This is awesome, thanks for adding it!

PR #1140, opened 2025-02-12
[VLM] Update mllama traceable definition

kylesayrs:
## Purpose ##
* Ensure compatibility with transformers>=4.50 by replacing `num_logits_to_keep` with `logits_to_keep`

https://github.com/huggingface/transformers/blob/1fae54c7216e144b426e753400abdc1299d4fc74/src/transformers/models/mllama/modeling_mllama.py#L2025-L2028

## Changes ##
* Update function definition

## Testing ##
* Ran mllama example

### DIFF
--- a/src/llmcompressor/transformers/tracing/mllama.py
+++ b/src/llmcompressor/transformers/tracing/mllama.py
@@ -37,7 +37,6 @@ logger = logging.get_logger(__name__)
 
 
 # TRACING: This function is not traceable
-@torch.fx.wrap
 def _prepare_cross_attention_mask(
     cross_attention_mask: torch.Tensor,
     num_vision_tokens: int,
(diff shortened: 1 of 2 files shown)


src/llmcompressor/transformers/tracing/mllama.py
```diff
@@ -37,14 +37,16 @@
...
 def _prepare_cross_attention_mask(
     cross_attention_mask: torch.Tensor,
     num_vision_tokens: int,
     dtype: str,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
     # reshape so it can be used by attn module
-    batch_size, text_total_length, *_ = cross_attention_mask.shape
+    # TRACING: cannot unpack cross_attention_mask with arbitrary number of args
+    # batch_size, text_total_length, *_ = cross_attention_mask.shape
+    batch_size, text_total_length = cross_attention_mask.shape[:2]
```
brian-dellabetta, line 48: i haven't dug into the `HFTracer` class to see how it works, but it must be pretty gnarly if it doesn't allow for things in python code that python allows?

brian-dellabetta (approved):
👍

src/llmcompressor/transformers/tracing/mllama.py
kylesayrs: It's actually not too complicated, I can explain some of the systems when we meet next.

The error here is that, because tracing tries to be (somewhat) agnostic to the shapes of the inputs, attempts to get the `shape` property of inputs are replaced by calls to the `size` method in the captured graph, which returns a proxy object.
https://github.com/huggingface/transformers/blob/main/src/transformers/utils/fx.py#L650-L652
```python3
    @property
    def shape(self):
        return self.tracer.create_proxy("call_method", "size", (self,), {})
```

Proxy objects cannot be iterated because their values cannot be determined at trace time. However, they can be unpacked to a finite set of variables

dsikka (approved):

PR #1141, opened 2025-02-12
Offload Cache Support torch.dtype

kylesayrs:
## Purpose ##
* Squelch warning about not supporting offloading dtype

```
UserWarning: Offloading not implemented for type <class 'torch.dtype'>.
```


brian-dellabetta (approved):

brian-dellabetta (approved):

src/llmcompressor/pipelines/cache.py
```diff
@@ -134,29 +134,25 @@ def _onload_value(self, intermediate: IntermediateValue) -> Any:
...
                 v = getattr(value, field.name)
                 setattr(value, field.name, self._onload_value(v))
 
             return value
 
-        elif isinstance(value, tuple):
+        if isinstance(value, tuple):
             return tuple(self._onload_value(v) for v in value)
 
-        elif isinstance(value, (int, str, float, bool)) or value is None:
```
rahul-tuli, line 147: 💯

tests/llmcompressor/pipelines/test_cache.py
```diff
@@ -28,6 +28,7 @@ def sample_cache(sample_dataloader):
     )
 
 
+@pytest.mark.unit
```
rahul-tuli, line 31: A side question: what does this marker do? Do we run a subset of tests based on the marker?

rahul-tuli (approved):

tests/llmcompressor/pipelines/test_cache.py
kylesayrs: I don’t think we use the marker now, but it allows us in the future to skip the unit tests if we thunk they’re taking too long

PR #1144, opened 2025-02-13
Remove CLI

dsikka:
Summary
- Remove cli support

Testing:
- Need to make sure everything is removed


src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -65,106 +63,6 @@
 from llmcompressor.utils.fsdp.helpers import is_fsdp_model
 
 
-def train(**kwargs):
```
horheynm, line 68: This is also used for the entrypoints for python api, so needed

dsikka: Not all, but I will clean this up and confirm.

brian-dellabetta (approved):

kylesayrs (approved):
Normally I'd suggest leaving some deprecation warnings, but in this case I think these endpoints are really unused and save to remove

PR #1147, opened 2025-02-13
[Audio] Support Whisper V3

kylesayrs:
## Purpose ##
* Support Whisper V3 model

## Changes ##
* Change default whisper model to v3
* Modify preprocessing function to be simpler
* Add dtype conversion to preprocessing function
  * Note that this is only required for feature extractor processors, as they return values which are float types (not just token ids, which work regardless of model dtype)

## Follow-ups ##
* Dtype conversion should theoretically be injected into prebaked dataset pathways as well, although I consider this low priority since we push users towards writing their own data processing functions

## Testing ##
* Quantized Whisper v3 model
* Note that you may have to add `ds.cleanup_cache_files()` to line 40 in order to overwrite any existing mapping caches

### DIFF
--- a/examples/multimodal_audio/whisper_example.py
+++ b/examples/multimodal_audio/whisper_example.py
@@ -7,12 +7,12 @@ from llmcompressor.transformers import oneshot
 from llmcompressor.transformers.tracing import TraceableWhisperForConditionalGeneration
 
 # Select model and load it.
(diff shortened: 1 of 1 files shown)


brian-dellabetta (approved):
❤️

dsikka (approved):

PR #1149, opened 2025-02-13
Skip lm_eval tests if no config file is present

kylesayrs:
## Purpose ##
* Towards allowing users and contributors to run tests locally

### DIFF
--- a/tests/e2e/vLLM/test_lmeval.py
+++ b/tests/e2e/vLLM/test_lmeval.py
@@ -45,7 +45,12 @@ class TestLMEval:
     """  # noqa: E501
 
     def set_up(self):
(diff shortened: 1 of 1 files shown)


dsikka (changes requested):
This change doesn't really make sense. 

The entrypoint for these test cases is meant to be the bash script which if given an empty folder for configs, will not run anyway.

I would rather have Brian make that clear in a readme as he's working on an update for users while setting a default config for `TEST_DATA_FILE`.

kylesayrs:
@dsikka So these tests run when running the command `python3 -m pytest tests`, which is the standard testing pathway. Unless we want to move this file outside of the tests folder, this is likely the best way to skip the test

dsikka:
I actually think these tests being moved out would be better since generally speaking, the overall compute and memory required is significantly larger than any of our other tests and we never want users running it by default.

PR #1150, opened 2025-02-13
lmeval tests multimodal

brian-dellabetta:
SUMMARY:
In order to add multimodal tests to lm_eval, I had to expand the data model and moved some hard-coded logic into conditionals.

This includes a folder re-org, so that lm-eval tests were moved out of `tests/e2e` and into their own `tests/lmeval` folder (they are integration+regression tests). 

We'll have to make sure QA team is aware of the re-org once this lands, as they might have to change the github ci/cd accordingly. This adds roughly 35 minutes of testing to the weekly tests

TEST PLAN:
No new source code


tests/lmeval/test_lmeval.py
```diff
@@ -7,10 +7,20 @@
 import yaml
 from loguru import logger
 
+from pydantic import BaseModel
 from llmcompressor.core import active_session
 from tests.e2e.e2e_utils import run_oneshot_for_e2e_testing
 from tests.examples.utils import requires_gpu_count
 
+
+class LmEvalConfig(BaseModel):
```
brian-dellabetta, line 16: I could refactor the entire model config to this Pydantic style, with defaults and docstrings it might help for readers / people extending in future

tests/e2e/e2e_utils.py
```diff
@@ -1,14 +1,28 @@
...
+)
+from transformers.models.qwen2_vl import Qwen2VLProcessor
 
 from llmcompressor.modifiers.quantization import GPTQModifier, QuantizationModifier
 from llmcompressor.transformers import oneshot
+from llmcompressor.transformers.tracing import TraceableQwen2VLForConditionalGeneration
 from tests.testing_utils import preprocess_tokenize_dataset
 
+# Map from huggingface model_type to pre-trained model class to load from
+HF_CLASS_MAP: dict[str, tuple[PreTrainedModel, AutoTokenizer | AutoProcessor]] = {
```
kylesayrs, line 17: As mentioned earlier, it's likely best if we add the model class to the config (this way we can theoretically support other model tasks).

https://github.com/vllm-project/llm-compressor/blob/98a7ae6c1503268323ab97692bd9dba04ed03342/src/llmcompressor/transformers/tracing/debug.py#L98-L103

tests/lmeval/test_lmeval.py
kylesayrs: I don't see a huge need, as the config should populate most of these and we should throw an error if the non-defaulting ones are skipped.

I do see a need to use 4-space indent though

tests/e2e/e2e_utils.py
kylesayrs: This also reduces the changes required to add more configs.

brian-dellabetta: Is there a similar way to get processor/tokenizer class?

tests/e2e/e2e_utils.py
```diff
@@ -21,19 +24,33 @@ def run_oneshot_for_e2e_testing(
...
     if dataset_id:
         ds = load_dataset(dataset_id, name=dataset_config, split=dataset_split)
         ds = ds.shuffle(seed=42).select(range(num_calibration_samples))
-        ds = preprocess_tokenize_dataset(ds, tokenizer, max_seq_length)
+        ds = process_dataset(ds, processor, max_seq_length)
         oneshot_kwargs["dataset"] = ds
         oneshot_kwargs["max_seq_length"] = max_seq_length
         oneshot_kwargs["num_calibration_samples"] = num_calibration_samples
 
+        # TODO better conditional on when multimodal data-collator should be added
```
brian-dellabetta, line 45: taking suggestions for a better way to handle this, or I can just drop the TODO

kylesayrs: I think this is fine. All multimodal models can be used with text-only, but in those cases you don't need the custom collator.

Therefore, using the custom collator is only dependent on there being multimodal input data

rahul-tuli (approved):
LGTM!

kylesayrs (dismissed):
So good

src/llmcompressor/transformers/tracing/__init__.py
```diff
@@ -8,13 +8,15 @@
...
 )
 from .idefics3 import (
-    Idefics3ForConditionalGeneration as TraceableIdefics3ForConditionalGeneration
+    Idefics3ForConditionalGeneration as TraceableIdefics3ForConditionalGeneration,
 )
 from .whisper import (
-    WhisperForConditionalGeneration as TraceableWhisperForConditionalGeneration
+    WhisperForConditionalGeneration as TraceableWhisperForConditionalGeneration,
 )
+from .debug import get_model_class
```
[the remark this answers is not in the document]
brian-dellabetta: You mean by adding it to `__all__`? I do that below

tests/lmeval/test_lmeval.py
```diff
@@ -104,26 +114,26 @@ def test_lm_eval(self):
...
-        logger.info(exact_match_flex)
-        assert numpy.isclose(exact_match_strict, self.exact_strict, rtol=0.05)
-        assert numpy.isclose(exact_match_flex, self.exact_flex, rtol=0.05)
+        metrics = results["results"][self.lmeval.task]
+        for metric, expected_val in self.lmeval.metrics.items():
+            actual_val = metrics.get(metric)
+            logger.info(
+                f"Comparing {metric}: Expected {expected_val}, Got {actual_val}"
+            )
+            assert numpy.isclose(expected_val, actual_val, rtol=0.05)
```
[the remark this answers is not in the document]
brian-dellabetta: this is what was used previously, what's the motivation for the switch?

tests/e2e/e2e_utils.py
```diff
@@ -1,23 +1,31 @@
...
     device: str,
 ):
-    loaded_model = AutoModelForCausalLM.from_pretrained(
-        model, device_map=device, torch_dtype="auto"
+    pretrained_model_class = get_model_class(model_class)
+    loaded_model = pretrained_model_class.from_pretrained(
+        model,
+        device_map=device,
+        torch_dtype="auto",
+        trust_remote_code=True,
```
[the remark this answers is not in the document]
brian-dellabetta: I can probably remove these, that was when i was trying the phi model. i'll check

src/llmcompressor/transformers/tracing/__init__.py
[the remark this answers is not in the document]
brian-dellabetta: added to `debug.py`

tests/lmeval/test_lmeval.py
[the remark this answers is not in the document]
brian-dellabetta: leaving as `numpy.isclose`, pytest approx api is a little wonky and this was how it was previously

kylesayrs (dismissed):

tests/e2e/e2e_utils.py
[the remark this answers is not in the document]
brian-dellabetta: removed

rahul-tuli (dismissed):

dsikka (approved):

tests/e2e/vLLM/lm_eval_configs/fp8_dynamic_per_token.yaml
```diff
@@ -1,8 +0,0 @@
-cadence: "weekly"
-model: meta-llama/Meta-Llama-3-8B-Instruct
```
dsikka, line 2: Why did we delete this file?

tests/lmeval/configs/fp8_dynamic_per_token.yaml
```diff
@@ -0,0 +1,13 @@
+cadence: "weekly"
+model: meta-llama/Meta-Llama-3-8B-Instruct
+scheme: FP8_DYNAMIC
+lmeval:
+  model_args:
+    dtype: bfloat16
```
dsikka, line 6: why did we have to add this?

dsikka (changes requested):

tests/e2e/vLLM/lm_eval_configs/fp8_dynamic_per_token.yaml
brian-dellabetta: I moved it to `lmeval/configs`, i'm not sure why it isn't showing as renamed instead of deleted and added, i used `git mv`. Maybe a squash/rebase messed it up

tests/lmeval/configs/fp8_dynamic_per_token.yaml
brian-dellabetta: @kylesayrs  suggested having it, I think it will load up the float32 weights in bfloat16 to be used during QDQ

tests/e2e/vLLM/lm_eval_configs/fp8_dynamic_per_token.yaml
brian-dellabetta: you can see it here -- https://github.com/vllm-project/llm-compressor/pull/1150/files#diff-86b82c071ca84fd078c05fc92a76d869784fd4a37ff59ab703f28b7b8fd9078cR1

dsikka (approved):
Looks good overall.
Just one quick question and I'm assuming we have ran a few text models as well, post changes?

tests/lmeval/test_lmeval.py
```diff
@@ -6,11 +6,23 @@
...
 import yaml
 from loguru import logger
+from pydantic import BaseModel
 
 from llmcompressor.core import active_session
 from tests.e2e.e2e_utils import run_oneshot_for_e2e_testing
 from tests.examples.utils import requires_gpu_count
 
+
+class LmEvalConfig(BaseModel):
```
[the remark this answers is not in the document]
brian-dellabetta: Yeah, `BaseModel` allows for validation during deserialization, that's all i'm using it for additionally here. otherwise it's the same as `@dataclass` at least for how i'm using it here.

kylesayrs (dismissed):
Looks great. While the ignore list could be tied to the model rather than the recipe, I think implementing that is more complexity than it's worth. Thanks for the changes

kylesayrs (approved):

tests/e2e/e2e_utils.py
```diff
@@ -43,16 +48,27 @@ def run_oneshot_for_e2e_testing(
...
+        ds = process_dataset(ds, processor, max_seq_length)
         oneshot_kwargs["dataset"] = ds
         oneshot_kwargs["max_seq_length"] = max_seq_length
         oneshot_kwargs["num_calibration_samples"] = num_calibration_samples
 
+        # Define a data collator for multimodal inputs.
+        if "flickr30k" in dataset_id:
+
+            def data_collator(batch):
+                assert len(batch) == 1
```
horheynm, line 67: why does the batch need to be 1 here? 
memory constraint? Could you add the error message also

tests/e2e/vLLM/recipes/INT8/recipe_int8_channel_weight_dynamic_per_token.yaml
```diff
@@ -3,7 +3,7 @@ quant_stage:
     SmoothQuantModifier:
       smoothing_strength: 0.8
     GPTQModifier:
-      ignore: [lm_head]
+      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:vision_model.*"]
```
horheynm, line 6: dynamic per token is fast to process to completion. Why are we ignoring these now?

tests/e2e/vLLM/test_vllm.py
```diff
@@ -71,6 +71,7 @@ def set_up(self):
             pytest.skip("Skipping test; cadence mismatch")
 
         self.model = eval_config["model"]
+        self.model_class = eval_config.get("model_class", "AutoModelForCausalLM")
```
horheynm, line 74: is this supposed to be a string or a callable?

tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
```diff
@@ -0,0 +1,19 @@
...
+  model_args:
+    dtype: bfloat16
+    add_bos_token: True
+    convert_img_format: True
+  task: mmmu_val_economics
+  num_fewshot: 0
+  limit: 1000
+  metrics:
+    acc,none: 0.4
+  batch_size: 4
```
horheynm, line 19: batchsize here is different than the batch size for multimodal in the asserts above?
its a flickr30k dataset, so wouldn't assert len(batch) == 1 trigger?

tests/e2e/e2e_utils.py
brian-dellabetta: ah! good eye, that should be `>=1`, my bad

tests/e2e/vLLM/recipes/INT8/recipe_int8_channel_weight_dynamic_per_token.yaml
brian-dellabetta: we want to avoid these layers when quantizing, as it affects model performance (as in accuracy, as opposed to runtime considerations)

tests/e2e/vLLM/test_vllm.py
brian-dellabetta: it should be a string, we use `getattr` in the code to map it to the actual class/callable

horheynm: Oh I see, eval_config is a dict.

tests/e2e/e2e_utils.py
kylesayrs: @brian-dellabetta the input to the data collator is a list whose length is the batch size. This collator indeed only works for batch_size = 1 because we haven't yet developed a universal multimodal padding function, although such a thing exists in vllm

brian-dellabetta: @kylesayrs ah ok, should I use `batch size=1` in lmeval then? I've got it set to 4 or 8 usually, it must work because they also have a compatible padding function?
where does batch size even get set in the call to oneshot? I don't see it on `DatasetArguments` or `TrainingArguments`

kylesayrs: @brian-dellabetta No, the data collator defined here is only passed into oneshot, not eval

brian-dellabetta: Right, does that mean they also have this data collation function? wondering how hard it would be to pull it, as part of a separate issue from this

brian-dellabetta: Marking this as resolved, @kylesayrs let's chat in our 1:1 about it as a potential next step though?

dsikka (approved):

PR #1153, opened 2025-02-14
Fix: Re-enable Sparse Compression for 2of4 Examples

rahul-tuli:
This PR restores sparse compression for our `2of4` examples, which was previously disabled due to a bug in the vLLM Cutlass integration.  

#### Background  
A bug in the Cutlass integration caused certain sparse-only compressed models to produce gibberish results. To mitigate this issue, we temporarily turned off sparse compression for our `2of4` examples.  

The bug has since been fixed by @tlrmchlsmth in [vllm-project/vllm#13198](https://github.com/vllm-project/vllm/pull/13198). With this fix in place, we can safely re-enable sparse compression for these examples.  

#### Changes  
- Re-enable sparse compression for `2of4` examples.  

#### Testing  
- Verified that sparse-only compressed models now produce expected outputs.

### DIFF
--- a/examples/sparse_2of4_quantization_fp8/llama3_8b_2of4.py
+++ b/examples/sparse_2of4_quantization_fp8/llama3_8b_2of4.py
@@ -116,7 +116,5 @@ print(tokenizer.decode(output[0]))
 print("==========================================\n")
 
 # Save compressed model and tokenizer
-model.save_pretrained(
-    save_dir, save_compressed=args.fp8, disable_sparse_compression=True
-)
+model.save_pretrained(save_dir, save_compressed=args.fp8)
 tokenizer.save_pretrained(save_dir)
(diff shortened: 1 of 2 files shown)


kylesayrs (approved):
Consider adding a comment about prev vllm versions

tlrmchlsmth (approved):
🥳

brian-dellabetta (dismissed):
🔥

horheynm (approved):

dsikka (changes requested):
Can we update the E2E test case as well?

dsikka (commented):
Should we consider adding a comment to the example about disable_sparse_compression (just that it exists/can be used)?

rahul-tuli:
> Can we update the E2E test case as well?

Done and tested!

brian-dellabetta (approved):

dsikka (approved):

PR #1159, opened 2025-02-17
Fix CPU Offloading 

dsikka:
SUMMARY:
- When updating the training args, `place_model_on_device` was missed and as a result, when creating the trainer (which we really should not be doing during oneshot...) the default value is left as True and the trainer tries to move the model to a gpu, if it is available.
- We want this argument to be False as we handle the device map and model initialization based on the calibration needs 

TEST PLAN:
- `cpu_offloading_fp8.py` ran to completion without issue
- `mult_gpus_int8_device_map` made it past the error and is running

### DIFF
--- a/src/llmcompressor/args/training_arguments.py
+++ b/src/llmcompressor/args/training_arguments.py
@@ -30,3 +30,7 @@ class TrainingArguments(HFTrainingArgs):
             "checkpoints will be written."
         },
     )
+
+    @property
+    def place_model_on_device(self):
+        return False


src/llmcompressor/args/training_arguments.py
```diff
@@ -30,3 +30,7 @@ class TrainingArguments(HFTrainingArgs):
             "checkpoints will be written."
         },
     )
+
+    @property
+    def place_model_on_device(self):
```
brian-dellabetta, line 35: a user would never want to set this to `True`? wondering if it's better off as an optional input that defaults to False

brian-dellabetta (approved):

src/llmcompressor/args/training_arguments.py
brian-dellabetta: > We want this argument to be False as we handle the device map and model initialization based on the calibration needs

nm, LGTM

rahul-tuli (approved):

src/llmcompressor/args/training_arguments.py
dsikka: No we dont want them controlling this as of yet. 

We provide options to update the device map and change how the model is mapped when running oneshot and in general for llmcompressor.

We then pass that to the Trainer so we're essentially already doing the device map heavy lifting.

I think the downside is that we're exposing this property at all which can be confusing for somebody looking through the args, but this is not super urgent to fixc as we're not yet focused on improving training workflows.

src/llmcompressor/args/training_arguments.py
```diff
@@ -30,3 +30,7 @@ class TrainingArguments(HFTrainingArgs):
             "checkpoints will be written."
         },
     )
+
+    @property
```
kylesayrs, line 34: Consider adding a comment explaining why this is immutable

kylesayrs (approved):

PR #1160, opened 2025-02-17
[Callbacks] Remove pre_initialize_structure

kylesayrs:
## Purpose ##
* Remove pre_initialize_structure to simplify codebase
* Fix recipe appending for appending a recipe to a model which already has a recipe 
* Remove misleading logging messages

```
2025-02-17T17:48:38.477750-0500 | _check_create_state | INFO - State created for compression lifecycle
2025-02-17T17:48:38.478670-0500 | pre_initialize_structure | INFO - Compression lifecycle structure pre-initialized for 0 modifiers
2025-02-17T17:48:38.478836-0500 | pre_initialize_structure | INFO - Compression lifecycle structure pre-initialized for 0 modifiers
```

## Prerequisites ##
* #1168

## Follow-ups ##
* Remove double initialization

## Changes ##
The preinitialization step used to fulfill a few purposes
* Construct the lifecycle state
  * This is now done by the dataclass directly
```python3
- state: Optional[State] = None
+ state: Optional[State] = field(default_factory=State)
```

* Populate state with model and recipe
  * This is now done (and has always been done) by `initialize`
  * Some functions such as Trainer.init_model attempt to access the model through the session before `initialize` is called. In these cases, we can pass the model directly
```python3
trainer = Trainer(
-     model_init=get_session_model,
+     model_init=lambda: model,
```

* Prepend recipes to the recipe.yaml if the model has already been compressed once
  * Move this logic from preinitialization to the save_pretrained function
  * Consolidate all save pathways to use the the same wrapped method

```python3
def save_pretrained_wrapper(...):
    update_and_save_recipe(model.name_or_path, save_directory)
```

* Provide a way for modifiers to influence the model after they have already been applied
  * This can still be a enacted via recipe validation, but likely no longer has a use case and shouldn't be done automatically, at most the LLM Compressor should warn if the recipe configuration is invalid / requires modification

* Create quantization modifier on GPTQ
  * This is now done within the `on_initialize` function
  * In the future, this should be done by a high-level recipe validation step
```python3
def on_initialize(...)
-     self.on_initialize_structure(state, **kwargs)
+     self._maybe_build_quant_modifier(state.model)
````

* Remove `EventType.order()` method which is unused

* Extend the `Recipe.simplify_recipe` class method to support strings

## Lifecycle ##
1. `create_session()` (doesn't do much and can be hidden behind `initialize`)
2. `initialize(model=..., recipe=...)`
    1. Maybe `start` modifiers
3. `LifecycleCallback.event(...)`
    1. Maybe `start/end` modifiers
4. `finalize()`

## Regression Evaluation ##
Main
```
vllm (pretrained=/home/kyle/llm-compressor/Meta-Llama-3-8B-Instruct-W4A16-G128,dtype=bfloat16,add_bos_token=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 1
|  Tasks   |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|winogrande|      1|none  |     5|acc   |↑  |0.7482|±  |0.0122|
```

This branch
```
vllm (pretrained=/home/kyle/llm-compressor/Meta-Llama-3-8B-Instruct-W4A16-G128,dtype=bfloat16,add_bos_token=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: 1
|  Tasks   |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|winogrande|      1|none  |     5|acc   |↑  |0.7482|±  |0.0122|
```

### CODE REVIEW

**Review by horheynm — APPROVED** [review]
Good job XD

File: src/llmcompressor/transformers/utils/helpers.py [hunk 1]
```diff
@@ -140,20 +140,9 @@ def recipe_from_huggingface_model_id(
...
         logger.info(f"Found recipe: {recipe_file_name} for model ID: {hf_stub}.")
-    except Exception as e:
-        logger.error(
-            (
-                f"Unable to find recipe {recipe_file_name} "
-                f"for model ID: {hf_stub}: {e}."
-                "Skipping recipe resolution."
-            )
-        )
+    except Exception:  # TODO: narrow acceptable exceptions
```
brian-dellabetta, line 145: we should at least retain logging of the error right?

src/llmcompressor/recipe/recipe.py
```diff
@@ -571,6 +580,11 @@ def _get_yaml_dict(self) -> Dict[str, Any]:
         return yaml_recipe_dict
 
 
+RecipeInput = Union[str, List[str], Recipe, List[Recipe], Modifier, List[Modifier]]
+RecipeStageInput = Union[str, List[str], List[List[str]]]
+RecipeArgsInput = Union[Dict[str, Any], List[Dict[str, Any]]]
```
brian-dellabetta, line 585: are we adding type definitions for inputs that were previously defined? wondering if this is just retaining status quo or adding something that would be very confusing to a user looking to build a recipe

brian-dellabetta (approved):
looking good, but i'm going off an assumption that none of this deleted code we actually need

src/llmcompressor/transformers/utils/helpers.py
kylesayrs: Because this function runs each time a model is saved (to check if a recipe already exists), and in most case a recipe will not already exist, I initially removed this to avoid confusion.

I think you're right, I've added it back as an debug statement.

src/llmcompressor/recipe/recipe.py
kylesayrs: Yes, these type defs were already defined [here](https://github.com/vllm-project/llm-compressor/pull/1160/files#diff-c4b97a607cab1f048cc5ee2b9d3a2e3bc79c939230cf62a13d78f2eaa1c93588L27-L31)

kylesayrs:
@brian-dellabetta The PR description lists all of the function that was needed by preinitialize and where that functionality now lives

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -254,6 +254,7 @@ def _pre_process(self):
         if isinstance(self.model_args.model, (str, PosixPath)):
             self.model_args.model, _ = initialize_model_from_path(self.model_args)
 
+        breakpoint()
```
brian-dellabetta, line 257:
```suggestion
```

horheynm (approved):

dsikka (approved):
great work!

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -253,10 +255,17 @@ def get_model_compressor(
 
 
 def update_and_save_recipe(model_path: str, save_directory: str):
```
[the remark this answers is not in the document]
kylesayrs: Good call, implemented [here](https://github.com/vllm-project/llm-compressor/pull/1201)

brian-dellabetta (approved):

PR #1168, opened 2025-02-18
[Callbacks] Consolidate Saving Methods

kylesayrs:
## Purpose ##
* Simplify all methods of saving into one point, namely the wrapped `save_pretrained` function
* Precursor to #1160
  * Needed for having a single point for saving on top of existing recipes

## Background ## 
All the things needed to be done during saving
1. Save the model weights, potentially compressed
2. Save the processor
3. Update the recipe checkpoint
4. Copy any necessary python files from the model cache
5. Only save on the main process

After these changes, (1, 2, 3, 4) will be done within the `save_pretrained` function, and (5) will be the responsibility of the caller. (3) will be implemented by #1160 so as not to conflict with existing logic in pre_init

All of the places where a model is saved are
* If an output dir is specified, at the end of the main function
* Between stages of the stage runner
* Between epochs of the HF Trainer
* By the user after oneshot/training completes

After these changes, all of these will be replaced by a single `save_checkpoint` function which calls `save_pretrained` to do all the necessary things

## Changes ##
* Remove `save_model_and_recipe`
  * Saving recipes is now done by `save_pretrained` function
* Implement `save_checkpoint`
  * Single entrypoint for saving a model and its processor
  * Performs actions (1, 2, 4)
* Replace all locations where a model is saved with `save_checkpoint`
  * All applicable callers with only saving on the main process (5)
* Remove support for `modify_fsdp_model_save_pretrained` and `unwrap_and_export_model`, to be added back in a future release


src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -26,81 +27,11 @@
...
-from llmcompressor.typing import Processor
-from llmcompressor.utils.fsdp.helpers import (
-    find_and_move_state_dicts_to_cpu,
-    unwrap_and_export_model,
-)
-
-__all__ = ["modify_save_pretrained", "modify_fsdp_model_save_pretrained"]
-
-
-def modify_fsdp_model_save_pretrained(trainer, processor: Processor):
```
brian-dellabetta, line 38: does this break terribly from the other changes you've made? Wondering why we want to remove code only to add it back later on

kylesayrs: @brian-dellabetta I'm essentially removing this function because I think it's approach is likely wrong. Instead of creating a separate save_pretrained function for FSDP models (and now having to maintain two functions), instead we should simply use a context manager

Something like
```python3
with maybe_unwrap_fsdp(model):
    original_save_pretrained(model, ...)

# alternatively
with unwrap_fsdp(model) if is_fsdp(model) else nullcontext():
     original_save_pretrained(model, ...)
```

This consists of using the code from `unwrap_and_export_model` and turning it into a context manager

brian-dellabetta (dismissed):

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -418,7 +418,10 @@ def main(
 
     # wrap model.save_pretrained
     if is_fsdp_model(model):
-        modify_fsdp_model_save_pretrained(trainer, processor)
+        raise NotImplementedError(
```
horheynm, line 421: We should put this in after the oneshot / stage runner refac. Currently this is an ok pipeline.

horheynm (changes requested):
Having one central location to carry out saving logic sounds great!

Could you map when the current saving logic is carried out using which pathway, and how the new changes takes over the saving logic? Ex. when do the many different saving logic now gets carried out? 

For FSDP, we do support it currently. Once stage runner is removed, then the assumption that any oneshot pathway will not have fsdp support will be valid.

kylesayrs:
@horheynm All the answers to your questions are in the PR description

horheynm (dismissed):

kylesayrs:
@dsikka This is ready for merge

brian-dellabetta (approved):

dsikka:
Can you fix conflicts

horheynm (approved):
Nice!

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -321,3 +250,13 @@ def get_model_compressor(
         sparsity_config=sparsity_config,
         quantization_format=quantization_format,
     )
+
+
+def update_and_save_recipe(model_path: str, save_directory: str):
```
dsikka, line 255: You mentioned in your PR description: 

(3) will be implemented by https://github.com/vllm-project/llm-compressor/pull/1160 so as not to conflict with existing logic in pre_init 

where (3) is updating the recipe checkpoint.  When you say update the recipe checkpoint, do you mean something beyond saving to disk, as is being done by this function currently?

dsikka (approved):
LGTM.
One clarification question.

brian-dellabetta (approved):

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
kylesayrs: @dsikka Yes, I'm referring specifically to reading any recipes that exist and appending the current recipe to them.

You can see that implementation [here](https://github.com/vllm-project/llm-compressor/pull/1160/files#diff-501f4cff84bb9a978fd637579af87e759c79d1e1f2366709a0369c176c8fee44R257-R270)

PR #1173, opened 2025-02-19
Deprecate `oneshot_device` arg in favor of `device_map`

kylesayrs:
## Purpose ##
* Generalize and clarify arg name

## Follow Ups ##
* Add deprecation information to metadata, and use the metadata to generalize deprecation warnings

### DIFF
--- a/src/llmcompressor/pytorch/model_load/helpers.py
+++ b/src/llmcompressor/pytorch/model_load/helpers.py
@@ -16,7 +16,6 @@ __all__ = [
     "initialize_recipe",
     "save_model_and_recipe",
     "copy_python_files_from_model_cache",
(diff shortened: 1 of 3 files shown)


src/llmcompressor/pytorch/model_load/helpers.py
```diff
@@ -87,22 +86,6 @@ def save_model_and_recipe(
     copy_python_files_from_model_cache(model, save_path)
 
 
-def fallback_to_cpu(device: str) -> str:
```
horheynm, line 90: Add a deprecation warning for logic change.

kylesayrs: Done

PR #1174, opened 2025-02-19
Remove unused/duplicated/non-applicable utils from pytorch/utils/helpers

kylesayrs:
## Purpose ##
* Remove utils, scoped to mostly `src/llmcompressor/pytorch/utils/helpers.py`
* Remove any utils which are related to fine tuning and therefore non-applicable
* Remove any utils which are duplicated by other utils files
* Keep any utils that may be worth using in the future

## Changes ##
* Remove any unused/duplicated/non-applicable from pytorch/utils/helpers
* Keep any utils that may be worth using in the future
* Add `pytest.mark.unit` to helpers unit tests

## Testing ##
* `grep -r "from llmcompressor.pytorch.utils" src/ tests/ examples/`
* Automated test coverage


markurtz (commented):
General note, I do worry with how much we are able to remove here and how much means we have restricted functionality, so would like to look through in more detail to designate what may be due to restricted functionality we'll need to add back in the future and what is fully migraged functionality either due to newer framework version or migration to compressed tensors

kylesayrs:
@markurtz I've kept any utils functions that I saw as might having some use in the future. Even if we remove a function that is used in the future, it can be added back when it's needed.

dbarbuzzi:
> Add `pytest.mark.unit` to helpers unit tests

❤️

rahul-tuli (approved):

brian-dellabetta (approved):
another one

Screenshot 2025-02-27 at 10 54 38 AM

PR #1184, opened 2025-02-23
[Bugfix] Raise error for processor remote code

kylesayrs:
## Purpose ##
* Raise error when processor cannot load due to required remote code option


rahul-tuli (approved):

rahul-tuli (approved):
Always appreciate diffs like this 🥇

brian-dellabetta (approved):
nice!

PR #1186, opened 2025-02-24
Add qwen moe w4a16 example

mgoin:
Made this moe for easy testing with 60 experts in vLLM https://huggingface.co/nm-testing/Qwen1.5-MoE-A2.7B-Chat-quantized.w4a16

### DIFF
--- a/examples/quantizing_moe/qwen_moe_w4a16.py
+++ b/examples/quantizing_moe/qwen_moe_w4a16.py
@@ -90,37 +90,3 @@ input_ids = tokenizer("Hello my name is", return_tensors="pt").input_ids.to("cud
 output = model.generate(input_ids, max_new_tokens=20)
 print(tokenizer.decode(output[0]))
 print("==========================================")
-
-
-# Run the model on vLLM
-try:
-    from vllm import LLM, SamplingParams
-
-    vllm_installed = True
-except ImportError:
-    vllm_installed = False
-
-if vllm_installed:
-    print("vLLM installed, running using vLLM")
-    sampling_params = SamplingParams(temperature=0.80, top_p=0.95)
-    llm = LLM(
-        model=SAVE_DIR,
-        tensor_parallel_size=2,
-        trust_remote_code=True,
-        max_model_len=1042,
-        dtype=torch.half,
-    )
-    prompts = [
-        "The capital of France is",
-        "The president of the US is",
-        "My name is",
-    ]
-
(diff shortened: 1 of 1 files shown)


kylesayrs (approved):
This example seems merited given the need to ignore the mlp gate, thanks

examples/quantizing_moe/qwen_moe_w4a16.py
```diff
@@ -0,0 +1,122 @@
...
+
+# Run the model on vLLM
+try:
+    from vllm import LLM, SamplingParams
+
+    vllm_installed = True
+except ImportError:
+    vllm_installed = False
+
+if vllm_installed:
```
dsikka, line 99: We typically don't run vLLM in our examples since vLLM isn't even listed as an optional install anywhere.

examples/quantizing_moe/qwen_moe_w4a16.py
```diff
@@ -0,0 +1,122 @@
...
+
+# select a Mixture of Experts model for quantization
+MODEL_ID = "Qwen/Qwen1.5-MoE-A2.7B-Chat"
+
+# adjust based off number of desired GPUs
+# if not enough memory is available, some layers will automatically be offloaded to cpu
+device_map = calculate_offload_device_map(
+    MODEL_ID,
+    reserve_for_hessians=True,
+    num_gpus=2,
```
dsikka, line 17: This should use the number of gpus available, not a hardcoded number

dsikka (changes requested):

examples/quantizing_moe/qwen_moe_w4a16.py
mgoin: All of the other uses of calculate_offload_device_map seems to use hardcoded values - what should we do here?

examples/quantizing_moe/qwen_moe_w4a16.py
mgoin: I'll remove it, I duplicated the deepseek example

examples/quantizing_moe/qwen_moe_w4a16.py
dsikka: Actually I think it's fine, if 2 is the suggested number that we'd want users to have available.

kylesayrs (approved):

dsikka (approved):

PR #1187, opened 2025-02-24
[Utils] Replace `preserve_attr` with `patch_attr`

kylesayrs:
## Purpose ##
* Provide explicit patching functionality to `preserve_attr`
* This function is very similar to `unittest.mock.patch`, except that using this functionality does not require using the `unittest` library in source code

## Changes ##
* Replace `preserve_attr` with `patch_attr`
* Replace usage in `src/llmcompressor/pipelines/sequential/helpers.py`, which helps with clarity
* Add unit test mark to utils/helpers pytests

## Testing ##
* Added tests


src/llmcompressor/utils/helpers.py
```diff
@@ -1132,9 +1132,39 @@ def calibration_forward_context(model: PreTrainedModel):
 
 
 @contextlib.contextmanager
-def preserve_attr(base: object, attr: str):
-    value = getattr(base, attr)
+def patch_attr(base: object, attr: str, *args, **kwargs):
```
brian-dellabetta, line 1135: why use *args and **kwargs here? It seems like you can just have a `patch_value: Any = None` input rather than trying to infer based on `len(args)` or `"value" in kwargs`

kylesayrs: It's a little pedantic, but this allows the user to pass `None` as a valid replacement value. Otherwise, there's no way to tell the difference between a user-passed `None` and a default `None`

brian-dellabetta: That might be confusing for a user. Sounds like we want to use a sentinel value to distinguish user-passed-in None vs a default None. Have you heard of the concept? Here's a PEP that kinda explains it -- https://peps.python.org/pep-0661

kylesayrs: Yeah, using a sentinel or custom type definition are both good, although my personal taste is to have `patch_attr` match the same function signature as `getattr`

kylesayrs: Any are good

brian-dellabetta: So desired behavior is if a user explicitly passes in none, we want
```python
base[attr]=None
```
and if a user doesn't pass in anything, we want
```python
del base[attr]
```
?

kylesayrs: @brian-dellabetta No, the value is not deleted, only restored upon exit

kylesayrs: I realize that patch_attr with no argument doesn't really have a use case, so I removed that functionality for the sake of clarity and ease of use

src/llmcompressor/utils/helpers.py
```diff
@@ -75,6 +75,7 @@
 RECIPE_METADATA_KEY = "__metadata__"
 FRAMEWORK_METADATA_KEY = "framework_metadata"
 ROOT_PATH = Path(__file__).resolve().parents[1]
+_sentinel = object()
```
brian-dellabetta: do we need both this and `_sentinel = object()` on L1070?

brian-dellabetta (approved):

src/llmcompressor/utils/helpers.py
brian-dellabetta: looks like you have since removed

kylesayrs: Yeah thanks for the catch :)

rahul-tuli (approved):

PR #1197, opened 2025-02-26
Make `transformers-tests` job conditional on files changed

dbarbuzzi:
SUMMARY:
Make `transformers-tests` job conditional on files changed by way of using the `tj-actions/changed-files@v45` action to check which files changed (ignoring files/folders based on a separate trial: #1194).

TEST PLAN:
Verify this PR:
1. Correctly results in skipping the `transformers-tests` job
2. Allows merging with the skipped job

### DIFF
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -15,9 +15,33 @@ env:
   CLEARML_API_SECRET_KEY: ${{ secrets.CLEARML_API_SECRET_KEY }}
 
 jobs:
+  detect-changes:
+    runs-on: ubuntu-latest
+
+    outputs:
+      changes-present: ${{ steps.changed-files.outputs.any_modified }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          files_ignore: |
+            examples/**
+            tests/e2e/**
+            tests/lmeval/**
+            tests/examples/**
+            **/*.md
+            .github/workflows/*
+
   transformers-tests:
+    needs: [detect-changes]
     runs-on: gcp-k8s-vllm-l4-solo
-    if: contains(github.event.pull_request.labels.*.name, 'ready') || github.event_name == 'push'
+    if: (contains(github.event.pull_request.labels.*.name, 'ready') || github.event_name == 'push') && needs.detect-changes.outputs.changes-present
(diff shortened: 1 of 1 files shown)


dbarbuzzi:
@dsikka If this PR's current state works as expected, before merging, I want to make one change to how files are matched (change below). The change here is to include the `test-check-transformers.yaml` workflow file specifically, as any changes to this workflow should result in it being executed. As with other testing for this PR, I have tested the below changes in an external repo. It seems using the `!` prefix does not work the same way with `files_ignored` which demands the switch to the `files` input approach instead.

```diff
         uses: tj-actions/changed-files@v45
         with:
-          files_ignore: |
-            examples/**
-            tests/e2e/**
-            tests/lmeval/**
-            tests/examples/**
-            **/*.md
-            .github/workflows/*
+          files: |
+            **
+            !examples/**
+            !tests/e2e/**
+            !tests/lmeval/**
+            !tests/examples/**
+            !**/*.md
+            !.github/**
+            .github/workflows/test-check-transformers.yaml
```

kylesayrs (approved):

dsikka (approved):

dsikka (approved):

rahul-tuli (approved):

.github/workflows/test-check-transformers.yaml
```diff
@@ -15,9 +15,41 @@ env:
...
+        with:
+          files: |
+            **
+            !examples/**
+            !tests/e2e/**
+            !tests/lmeval/**
+            !tests/examples/**
+            !**/*.md
+            !.github/**
+            .github/workflows/test-check-transformers.yaml
```
brian-dellabetta, line 41: Does this mean check every file (**) but exclude `examples/`, ..., exclude `.github/`, but re-include `.github/workflows/test-check-transformers.yaml`? so the order matters?

brian-dellabetta (approved):

.github/workflows/test-check-transformers.yaml
dbarbuzzi: > Does this mean check every file (**) but exclude `examples/`, ..., exclude `.github/`, but re-include `.github/workflows/test-check-transformers.yaml`?

Basically, yes.

> so the order matters?

I _think_ so, based on my reading through the docs. Notably, starting with v13, they switched/supported glob pattern matching:

https://github.com/tj-actions/changed-files/releases/tag/v13

Following the link about the glob pattern matching in those release notes, it seems to suggest order is relevant:

> `!`: At the start of a pattern makes it negate previous positive patterns. It has no special meaning if not the first character.

PR #1198, opened 2025-02-26
[Callbacks] Remove `MagnitudePruningModifier.leave_enabled`

kylesayrs:
## Purpose ##
* Simplify the modifier lifecycle by removing the ability for modifiers to affect the model after the modifier's `end` event
* This allows the `on_event` method to be removed in a future change

## Background ##
* The `leave_enabled` option was originally intended as a shortcut to simplify recipes which used magnitude pruning during the iterative pruning, then needed the masks to stay enabled during stabilization SFT
  * This change proposes making the recipe clearer by requiring the ConstantPruningModifier after the MagnitudePruningModifier becomes inactive

## Changes ##
* Remove `MagnitudePruningModifier.leave_enabled` with a deprecation warning

### DIFF
--- a/src/llmcompressor/modifiers/pruning/magnitude/base.py
+++ b/src/llmcompressor/modifiers/pruning/magnitude/base.py
@@ -1,4 +1,7 @@
+import warnings
 from typing import Any, Dict, List, Union
+
+from pydantic import field_validator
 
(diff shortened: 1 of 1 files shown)


rahul-tuli (approved):
Thank you for this change!

brian-dellabetta (approved):

PR #1200, opened 2025-02-26
Use KV cache constant names provided by compressed tensors 

kylesayrs:
## Purpose ##
* Harden code related to the kv_cache parameter names


rahul-tuli (approved):

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -194,8 +198,10 @@ def calibrate_kv_cache_output_hook(module: Module, _args: Any, _output: torch.Te
     Hook to update k_scale and v_scale parameters when running kv_cache quantization.
     """
     kv_cache = getattr(module, "kv_cache")
-    update_parameter_data(module, kv_cache.k_scales[module.layer_idx], "k_scale")
-    update_parameter_data(module, kv_cache.v_scales[module.layer_idx], "v_scale")
+    k_scale = kv_cache.k_scales[module.layer_idx]
+    v_scale = kv_cache.v_scales[module.layer_idx]
+    update_parameter_data(module, k_scale, KVCacheScaleType.KEY.value)
```
rahul-tuli, line 203: nit: While it's not too bad, the `.value` looks ugly, is there a way we can circumvent that? Would inheriting `KVCacheScaleType` from `str` help us with that?

brian-dellabetta (approved):

brian-dellabetta (approved):

PR #1143, opened 2025-02-12
[Train/Finetune] Decouple training pipeline from main

horheynm:
SUMMARY:
* Add `train` pathway in `src/llmcompressor/entrypoints/train.py`.
* Add `src/llmcompressor/entrypoints/utils.py` for dataclass parsing logic, model and processor instantiation logic shared between `train` and `oneshot`.
* Add info on `README` for training flows, `oneshot` then `train` and `oneshot` then knowledge distillation with end-to-end script.
---
* Update `data_args` to `dataset_args`.
* Update `oneshot` to use the shared (dataset | preprocessing | postprocessing) functions in `entrypoints/utils.py`
* Update `src/llmcompressor/transformers/finetune/README.md` to use `from llmcompressor import train`, not `from llmcompressor.transformers import train`.
* Update `src/llmcompressor/transformers/finetune/session_mixin.py:: initialize_session` to directly call `active_session().initialize(...)`,  not `initialize`.
* Update processed dataset and dataloader logic from `src/llmcompressor/transformers/finetune/data/data_helpers.py` to `src/llmcompressor/datasets/utils.py`.
* Update `parse_args`, model + processor loading logic from `src/llmcompressor/transformers/finetune/text_generation.py` to `entrypoints/utils.py`
---
* Delete `train` in StageRunner.
* Delete  `train` logic  in `text_generation.py::main`.
* Delete `initialize` from `src/llmcompressor/core/session_functions.py`, which is only used in `train` in `text_generation.py`
---

Entrypoints: 
```python3
from llmcompressor import train
train(**kwargs) 
```
TEST PLAN:
Pass all tests and examples.

PR #1202, opened 2025-02-26
Stage run

horheynm:
DRAFT - MERGED FROM https://github.com/vllm-project/llm-compressor/pull/1143

SUMMARY:

TEST PLAN:


src/llmcompressor/entrypoints/utils.py
```diff
@@ -79,20 +84,28 @@ def post_process(
...
-    )
+    else:
+        logger.warning(
+            "Optimized model is not saved. To save, please provide",
+            "`output_dir` as input arg.",
+            "Ex. `oneshot(..., output_dir=...)`",
+        )
+
+    # Reset the one-time-use session upon completion
+    reset_session()
```
brian-dellabetta, line 108: is this a big thing? if so, we may want to call it out in docstring as a side effect?

brian-dellabetta (approved):

src/llmcompressor/entrypoints/utils.py
horheynm: There are places where the info of the session is used to retrieve info. 
We can reset once it finishes running - which i think should be default, but the above need to be addressed. Ill do that in the follow up pr. 
For now resets if the user wants to reset

tests/llmcompressor/transformers/finetune/test_oneshot_and_finetune.py
```diff
@@ -17,39 +17,46 @@
...
         )
-        config_ft = ModelCompressor.parse_sparsity_config(
+        config_finetune_applied = ModelCompressor.parse_sparsity_config(
             AutoConfig.from_pretrained(
-                os.path.join(self.output, "stage_test_oneshot")
+                os.path.join(self.output, "test_train_stage")
             ).quantization_config
         )
-        assert config_ft["global_sparsity"] >= config_os["global_sparsity"]
+        breakpoint()
```
brian-dellabetta:
```suggestion
```

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -456,7 +455,7 @@ def test_sparse_24_compressor_is_lossless(model_stub, recipe, sparse_format, tmp
     splits = {"calibration": "train[:10%]"}
     empty_model = AutoModelForCausalLM.from_pretrained(model_stub, torch_dtype="auto")
 
-    oneshot(
+    model = oneshot(
```
brian-dellabetta, line 460: so `oneshot` both modifies the model passed in and returns it here? That might confuse users

horheynm: model meaning PretrainedModel? What do you mean model here?

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -63,179 +43,3 @@ def train(**kwargs):
     from llmcompressor import train
 
     train(**kwargs)
-
-
-def apply(**kwargs):
```
kylesayrs, line 68: Please raise a deprecation error?

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -63,179 +43,3 @@ def train(**kwargs):
...
-
-    training_args.run_stages = True
-    report_to = kwargs.get("report_to", None)
-    if report_to is None:  # user didn't specify any reporters
-        # get rid of the reporters inferred from hugging face
-        training_args.report_to = []
-    main(model_args, dataset_args, recipe_args, training_args)
-
-
-def compress(**kwargs):
```
kylesayrs, line 86: Please raise a deprecation error?

kylesayrs (approved):
Looks great, I much prefer this interface

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: ok

src/llmcompressor/transformers/finetune/text_generation.py
horheynm: ok

kylesayrs (approved):

src/llmcompressor/entrypoints/train.py
```diff
@@ -9,7 +11,7 @@
 from .utils import post_process, pre_process
 
 
-def train(**kwargs):
+def train(**kwargs) -> PreTrainedModel:
```
brian-dellabetta, line 14: train returns a model but `oneshot` does not, right? it modifies in place? wondering if we should be consistent across the top-level functions

src/llmcompressor/entrypoints/utils.py
```diff
@@ -79,20 +84,29 @@ def post_process(
     Raises:
         ValueError: If saving fails due to an invalid `output_dir` or other issues.
     """
-    if output_dir is not None:
+    if model_args is not None and output_dir is not None:
+        if recipe_args is not None and getattr(recipe_args, "stage", None) is not None:
+            output_dir = os.path.join(output_dir, recipe_args.stage)
```
brian-dellabetta, line 89: if the user only has one stage, does it still get saved in output_dir / stage?

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
brian-dellabetta: the main flow i've seen is user creates model
```python
model = AutoModelForCausalLM....
```
and passes that into oneshot
```python
oneshot(model, ds, recipe)
```
and then access the compressed model which has been modified in place
```
model.save_pretrained....
```
rather than doing `model = oneshot(....)`. I didn't realize `oneshot` also returned the model. I think this makes more sense, I just didn't realize this worked.

brian-dellabetta (dismissed):
few clarifying questions but i think is ready to roll!

src/llmcompressor/entrypoints/train.py
brian-dellabetta: Never mind, looks like `oneshot` also returns a model, based on your code elsewhere in this PR

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
brian-dellabetta: i'll mark this as resolved, i didn't realize we had `def oneshot(**kwargs) -> PreTrainedModel:`

src/llmcompressor/entrypoints/train.py
brian-dellabetta: marking as resolved, i misunderstood the `oneshot` function signature

src/llmcompressor/entrypoints/utils.py
horheynm: no, only if `recipe_args` has stage attr. That is passed in as oneshot(..., stage="foo-stage").

dsikka (changes requested):
This is currently failing the multi-stage example: `llama7b_sparse_w4a16.py` under `llm-compressor/examples/quantization_2of4_sparse_w4a16` - seems to get get past the sparsity stage currently and then fails during finetune.

@kylesayrs can you take a look?

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
```diff
@@ -33,31 +33,61 @@
...
-    bf16=bf16,
-    output_dir=output_dir,
+    num_calibration_samples=num_calibration_samples,
+    preprocessing_num_workers=preprocessing_num_workers,
     splits=splits,
+    output_dir=output_dir,
+)
+
+training_kwargs = dict(
+    bf16=bf16,
```
dsikka, line 49: Why do we need this argument?

brian-dellabetta (approved):

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
kylesayrs: This is a HF training argument
https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L722

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
```diff
@@ -16,6 +13,9 @@
 # uses LLM Compressor's built-in preprocessing for ultra chat
 dataset = "ultrachat-200k"
 
+# Select the recipe for 2 of 4 sparsity and 4-bit activation quantization
+recipe = "2of4_w4a16_recipe.yaml"
```
brian-dellabetta, line 17: do we need an absolute path here? Something like this would allow it to run from any working directory
```suggestion
from pathlib import Path
recipe = str(Path(__file__).parent / "2of4_w4a16_recipe.yaml")
```

brian-dellabetta (approved):

PR #1208, opened 2025-02-27
Update finetune tests to decrease execution time

dsikka:
Summary
- Update such that we're using less data or fewer epochs
- Decreases finetune test time from 20 minutes to about 4


kylesayrs (approved):
If we run into hessian invertibility issues due to hardware differences, we can increase the dampening factor

horheynm (approved):

PR #1220, opened 2025-03-04
fixing reproducibility of lmeval tests

brian-dellabetta:
SUMMARY:
LM Eval weekly tests are failing, this resolves two issues
1. installs pillow, which I had locally through vllm but is not installed as part of llm-compressor
2. adds a random seed to the lmeval tests, which seems after a good amount of testing to resolve the issue. it is entirely during calibration/quantization, lm-eval behavior is deterministic as they always set a seed. It is a bit surprising that it can have such a drastic effect, but these are 2B vision-language models and a difficult multiple choice dataset, not too far away from random guessing.

TEST PLAN:
no new src code


setup.py
```diff
@@ -60,9 +60,12 @@
         "datasets",
         "accelerate>=0.20.3,!=1.1.0",
         "pynvml",
-        "compressed-tensors"
-        if version_info.build_type == "release"
-        else "compressed-tensors-nightly",
+        "pillow",
```
brian-dellabetta, line 63: version was unpinned in vllm, so i left it the same here

dsikka (dismissed):

tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
```diff
@@ -10,7 +10,6 @@ lmeval:
     convert_img_format: True
   task: mmmu_val_economics
   num_fewshot: 0
-  limit: 1000
```
dsikka, line 13: are there > 1000 samples here?

tests/lmeval/configs/vl_int8_w8a8_dynamic_per_token.yaml
```diff
@@ -13,7 +13,6 @@ lmeval:
     convert_img_format: True
   task: mmmu_val_economics
   num_fewshot: 0
-  limit: 1000
```
dsikka, line 16: same as above

tests/lmeval/test_lmeval.py
```diff
@@ -57,6 +59,10 @@ class TestLMEval:
     """  # noqa: E501
 
     def set_up(self):
+        random.seed(42)
+        numpy.random.seed(42)
+        torch.manual_seed(42)
```
dsikka, line 64: torch.manual_seed() should be enough?

dsikka (changes requested):

tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
brian-dellabetta: the default is 1000, so removing this doesn't change the config passed to lm eval. the mmmu economics dataset doesn't have more than 1000 samples though, so this is just using all of them

tests/lmeval/test_lmeval.py
brian-dellabetta: i didn't check, but following the pattern lmeval has in their setup, setting these seeds. should i drop the other 2 and test?

tests/lmeval/configs/vl_int8_w8a8_dynamic_per_token.yaml
brian-dellabetta: see reply [here](https://github.com/vllm-project/llm-compressor/pull/1220#discussion_r1983522773). i'll mark this as resolved

tests/lmeval/test_lmeval.py
brian-dellabetta: I ran with just torch seed set and it failed to reproduce, so leaving all three set

horheynm (dismissed):

rahul-tuli (approved):
GG!

dsikka (changes requested):
rebase?

setup.py
kylesayrs, line 63: Can we make this a dev dependency rather than a base dependency?

tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
```diff
@@ -2,6 +2,7 @@ cadence: weekly
 model: Qwen/Qwen2-VL-2B-Instruct
 model_class: TraceableQwen2VLForConditionalGeneration
 scheme: FP8_DYNAMIC
+seed: 42 #compressed model is sensitive to random seed
```
kylesayrs, line 5: nit
```suggestion
seed: 42  # compressed model is sensitive to random seed
```

tests/lmeval/test_lmeval.py
```diff
@@ -73,6 +75,12 @@ def set_up(self):
         self.quant_type = eval_config.get("quant_type")
         self.save_dir = eval_config.get("save_dir")
 
+        seed = eval_config.get("seed", None)
+        if seed is not None:
```
kylesayrs, line 79: nit: use `from transformers import set_seed`
https://github.com/huggingface/transformers/blob/94ae1ba5b55e79ba766582de8a199d8ccf24a021/src/transformers/trainer_utils.py#L92

kylesayrs (dismissed):

tests/lmeval/test_lmeval.py
brian-dellabetta: thanks for the reference. I think i will leave as is though so it remains explicit

setup.py
brian-dellabetta: i've leveraged optional dependencies a lot in the past, it's a mixed bag though where users are unsure which optional deps they might need and there isn't a whole lot of benefit in not installing `pillow` which appears pretty lightweight and minimal overhead, and is commonly used. I didn't see any precedent for optional deps so i left as required for now. I think it's something we could/should leverage in future though if we have clear separation of optional dep groups

dsikka: This is used by the processor right?

dsikka (approved):

setup.py
brian-dellabetta: yes, i think any of the vision language processors will likely need it

kylesayrs (approved):

setup.py
brian-dellabetta: leaving as required for now, we may want to split all things multimodal into an optional dep later on but for now it's pretty lightweight

PR #1233, opened 2025-03-07
[Quantization] Channel-wise Output Activation Quantization for Attention QKV Modules

horheynm:
SUMMARY:
Quantize the output activation of the attention layers. 

Attention we need to worry about is the QKV. O/Up/down is not quantized.

Math:
x is the input vector -> tokenized + embedding
weight for QKV is Linear modules
output is the forward call of QKV with x
```bash
# x
(Pdb) hidden_states.shape -> torch.Size([1, 1930, 4096]) -> [batch, seq_len, hidden_size]

# weight
(Pdb) self.q_proj.weight.shape -> torch.Size([4096, 4096]) -- [hidden_size, hidden_size]
(Pdb) self.k_proj.weight.shape -> torch.Size([1024, 4096]) -- [num_key_value_heads * head_dim, hidden_size]
(Pdb) self.v_proj.weight.shape -> torch.Size([1024, 4096]) -- [num_key_value_heads * head_dim, hidden_size]

# output
(Pdb) self.q_proj(hidden_states).shape -> torch.Size([1, 1930, 4096]) -> [batch, seq_len, hidden_size]
(Pdb) self.k_proj(hidden_states).shape -> torch.Size([1, 1930, 1024]) -> [batch, seq_len, num_key_value_heads * head_dim]
(Pdb) self.v_proj(hidden_states).shape -> torch.Size([1, 1930, 1024]) -> [batch, seq_len, num_key_value_heads * head_dim]


```
Expected output scales and zp for output activations

```bash
k_proj activations -> [4096]
k_proj activations -> [1024]
v_proj activations -> [1024]

```

The observer will output the vectors in the same ndim as the given output activation tensor (ie.  torch.Size([1, 1930, 1024]), then outputs  torch.Size([1, 1, 1024])). Squeeze it to just get torch.Size([1024]), so ndim of 1.

TEST PLAN:
* Pass tests
* Pass eval


horheynm:
Next todo is to add support for group quantization

horheynm:
Will break down kv-cache logic to a different PR

PR #1237, opened 2025-03-10
[Hotfix] use pytorch bin files for stories models

kylesayrs:
## Purpose ##
* Hot fix issue https://huggingface.co/Xenova/llama2.c-stories15M/discussions/4#67cb757f8a3f9443b8a9ff74


markurtz:
@kylesayrs given how many places the logic for whether to use safe tensors or not is propagated, I think it would make sense to make a convenience function that can be centrally updated in case that logic changes towards the future.

Given the scope is only under tests, though, and I believe this is a temp workaround, not a huge blocker

brian-dellabetta (approved):
claude ftw! :)

kylesayrs:
This solution was more involved than I initially expected. There still exist instances of loading the stories model without use_safetensors=False within the src folder, which are too involved to fix as part of a hotfix. Closing in favor of #1239

PR #1238, opened 2025-03-10
[Bugfix] Staged 2of4 example

kylesayrs:
## Purpose ##
* Fix staged 2of4 example

## Background ##
* When #1160 landed, this change introduced a bug in the recipe container which meant that the recipe was not recompiled after `append`ing. This caused sgpt to initialize twice and gptq to never initialize, leading to a sparsity-only quantization config
* At some point, a changed was introduced which causes previous stages to become reconstructed after recipe recompilation. This means that without resetting the session in between stages, previous stages will initialize twice.
  * In order to avoid this issue, this PR introduces `session.reset()` in between stages
    * This change has the consequence of creating `recipe.yaml` files which do not have the full recipe history. However, I believe this is acceptable for the time being, as the stage runner and this work flow will be removed in the next release.


src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -183,6 +182,10 @@ def run_sequential_stages(
                     "the stage name."
                 )
 
+            # skip stages which have already been applied
```
kylesayrs, line 185: This does not actually affect this example, but is a good change to make. This guards against applying the same stage multiple times.

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -222,6 +225,7 @@ def run_sequential_stages(
                     recipe_stage=stage_name,
                 )
             elif run_type is StageRunType.TRAIN:
+                self.trainer.model = model
```
kylesayrs, line 228: This also does not affect this example, but makes model passing slightly clearer

horheynm (approved):

brian-dellabetta (approved):

src/llmcompressor/transformers/finetune/runner.py
```diff
@@ -248,11 +252,10 @@ def run_sequential_stages(
 
             # setup for next stage
             session = active_session()
-            session.reset_stage()
+            session.reset()
 
             # synchronize and clean up memory
             self.trainer.accelerator.wait_for_everyone()
-            self.trainer.model = get_session_model()
```
dsikka, line 255: was this code just redundant to start?

src/llmcompressor/transformers/finetune/runner.py
dsikka: Oh I see

dsikka (approved):

dsikka (commented):
Just so I understand, we're removing the functionality to track past stages? Which I think landed not too long ago

kylesayrs:
@dsikka That functionality breaks only for this example

src/llmcompressor/transformers/finetune/runner.py
kylesayrs: No, I just find it to be clearer in the other location

kylesayrs: esp since we're trying to remove dependence on the trainer

PR #1239, opened 2025-03-10
Replace Xenova model stub with nm-testing model stub

kylesayrs:
## Purpose ##
* Fix tests which break as a result of https://github.com/vllm-project/llm-compressor/pull/new/kylesayrs/replace-self-hosted

## Changes ##
* Use self-hosted model stub


brian-dellabetta (approved):

dsikka (approved):
We still need the other model loading change right?

kylesayrs:
@dsikka No, as of 5 minutes ago this self-hosted model fixes the bug which was present on the original Xenova model

dsikka:
> @dsikka No, as of 5 minutes ago this self-hosted model fixes the bug which was present on the original Xenova model

lolz

PR #1246, opened 2025-03-12
Fix run_compressed tests

dsikka:
Summary
- These tests were previously approved but something weird happened when they were landing and they were reverted to their old cases...not sure how
- This brings back the appropriate cases
- This would have caught the lm-eval test issue that Dan is reporting as well

### DIFF
--- a/tests/llmcompressor/transformers/compression/test_run_compressed.py
+++ b/tests/llmcompressor/transformers/compression/test_run_compressed.py
@@ -1,134 +1,182 @@
-import copy
 import shutil
 import tempfile
 import unittest
 
 import torch
-from compressed_tensors import QUANTIZATION_CONFIG_NAME
-from compressed_tensors.compressors import ModelCompressor
-from compressed_tensors.quantization import QuantizationStatus
+from compressed_tensors.linear.compressed_linear import CompressedLinear
+from compressed_tensors.quantization.utils import iter_named_leaf_modules
 from parameterized import parameterized_class
-from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
+from transformers import AutoModelForCausalLM, AutoTokenizer
 from transformers.utils.quantization_config import CompressedTensorsConfig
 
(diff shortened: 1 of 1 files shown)


horheynm (approved):
Looks good, thanks!

tests/llmcompressor/transformers/compression/test_run_compressed.py
```diff
@@ -1,134 +1,182 @@
...
+COMPRESSED_LINEAR_CONFIG_DIR = (
+    "tests/llmcompressor/transformers/compression/run_compressed_configs"
+)
 
 
 @requires_gpu
-@parameterized_class(parse_params(CONFIG_DIR))
-class TestDecompression(unittest.TestCase):
+@parameterized_class(parse_params(COMPRESSED_LINEAR_CONFIG_DIR))
+class Test_Decompressed_Linear_Uncompressed_Linear(unittest.TestCase):
```
brian-dellabetta, line 21: Any reason we're not using CamelCase here?
```suggestion
class TestDecompressedLinearUncompressedLinear(unittest.TestCase):
```

tests/llmcompressor/transformers/compression/test_run_compressed.py
```diff
@@ -1,134 +1,182 @@
...
+    def tearDownClass(cls):
+        shutil.rmtree(cls.test_dir)
+        del cls.decompressed_model
+        del cls.uncompressed_model
+        torch.cuda.empty_cache()
+
+
+@requires_gpu
+@parameterized_class(parse_params(COMPRESSED_LINEAR_CONFIG_DIR))
+class Test_Compressed_CompressedLinear_Decompressed_Linear(unittest.TestCase):
```
brian-dellabetta, line 98: same as above
```suggestion
class TestCompressedCompressedLinearDecompressedLinear(unittest.TestCase):
```

brian-dellabetta (approved):

PR #1251, opened 2025-03-13
update minimal python to 3.10

### LINKED ISSUE #1250: Should the minimal python version be 3.10?
Hi folks, it seems that the minimal python version should be 3.10. 
Otherwise I am getting a syntax error here
https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/obcq/base.py#L122-L126 
```python
            with (
                torch.no_grad(),
                align_module_device(module),
                CompressionLogger(module) as comp_logger,
            ):
```

I got the error when running with python 3.8. Installing it in 3.10 it got resolved. 

Or am I missing sth?

jmkuebler:
Resolves #1250 

SUMMARY:
Chained context managers were introduced by python3.10

TEST PLAN:
N/A

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,7 @@ setup(
             "llmcompressor.trace=llmcompressor.transformers.tracing.debug:main",
         ]
     },
-    python_requires=">=3.8",
+    python_requires=">=3.10",
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Programming Language :: Python :: 3",


kylesayrs (approved):
Thank you for your contribution!

dsikka (changes requested):
I think we still want to support 3.9 for the time being. We should update the incorrect lines

PR #1260, opened 2025-03-17
fix lm eval test reproducbility issues

brian-dellabetta:
SUMMARY:
lm-eval multimodal tests were failing to reproduce across different versions of compressed tensors. After upgrading the models from 2B to 7B, the tests appear to be reproducing across compressed tensors 0.9.1, 0.9.2 and nightly. I ran extensively for the fp8 config across different versions of CT, and it always returned the same result. 

I also removed the random seed from the configs. after running several of each of the 3 configs, i did not see any change in result. this may cause errors during ci/cd testing but I'd like to see if it does, i feel that is a better e2e test anyway.

The Qwen2.5 7B model was erroring out during GPTQ, there was some error during the Hessian calculation. I confirmed the visual part of the model is excluded, but it was occurring in the `model.layers[_].mlp.down_proj` matrix which is a `Linear(in_features=18944, out_features=3584, bias=False)` layer. Maybe the strange shape was causing an issue? The llava model runs fine. @kylesayrs is this cause for concern and deeper analysis?

TEST PLAN:
no new src code, just fixing tests

### DIFF
--- a/tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
+++ b/tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
@@ -1,8 +1,7 @@
 cadence: weekly
-model: Qwen/Qwen2-VL-2B-Instruct
-model_class: TraceableQwen2VLForConditionalGeneration
+model: Qwen/Qwen2.5-VL-7B-Instruct
+model_class: TraceableQwen2_5_VLForConditionalGeneration
 scheme: FP8_DYNAMIC
-seed: 42  # compressed model is sensitive to random seed
 lmeval:
   model: "hf-multimodal"
   model_args:
--- a/tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
+++ b/tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
@@ -1,11 +1,10 @@
 cadence: "weekly"
-model: Qwen/Qwen2-VL-2B-Instruct
-model_class: TraceableQwen2VLForConditionalGeneration
+model: llava-hf/llava-1.5-7b-hf
+model_class: TraceableLlavaForConditionalGeneration
 recipe: tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight.yaml
 dataset_id: lmms-lab/flickr30k
 dataset_split: "test[:512]"
 scheme: W4A16_actorder_group
-seed: 42 #compressed model is sensitive to random seed
 lmeval:
   model: "hf-multimodal"
   model_args:
@@ -15,5 +14,5 @@ lmeval:
   task: mmmu_val_economics
   num_fewshot: 0
   metrics:
-    acc,none: 0.366
+    acc,none: 0.233
   batch_size: 4
--- a/tests/lmeval/configs/vl_int8_w8a8_dynamic_per_token.yaml
+++ b/tests/lmeval/configs/vl_int8_w8a8_dynamic_per_token.yaml
@@ -5,7 +5,6 @@ scheme: INT8_dyn_per_token
 recipe: tests/e2e/vLLM/recipes/INT8/recipe_int8_channel_weight_dynamic_per_token.yaml
 dataset_id: lmms-lab/flickr30k
 dataset_split: "test[:512]"
-seed: 42 #compressed model is sensitive to random seed
 lmeval:
   model: "hf-multimodal"
   model_args:


dsikka (commented):
We should look into the error you're seeing for GPTQ.
We were seeing 0.233 for llava, the entire time we've had this test running?

kylesayrs:
@brian-dellabetta The down_proj is the most likely to fail hessian inversion, since it is the weight with the largest input size.

This is likely a normal hessian inevitability issue, which can be fixed by shuffling the dataset/ using an image dataset

brian-dellabetta:
Hitting several issues around reproducibility and slowness of lm-eval (takes about a minute for each of the 30 samples to run). Will continue this next week after resolving more urgent tasks

tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
```diff
@@ -1,19 +1,20 @@
...
-model: Qwen/Qwen2-VL-2B-Instruct
-model_class: TraceableQwen2VLForConditionalGeneration
-recipe: tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight.yaml
+model: Qwen/Qwen2.5-VL-7B-Instruct
+model_class: TraceableQwen2_5_VLForConditionalGeneration
+recipe: tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight_dampfrac1e-1.yaml
 dataset_id: lmms-lab/flickr30k
-dataset_split: "test[:512]"
+dataset_split: "test[:256]"
 scheme: W4A16_actorder_group
```
dbarbuzzi, line 7: I can see that this line isn’t touched in this PR, but _should_ the `scheme` be `W4A16_actorder_weight` instead? If so, please update it accordingly to help avoid timings CSV file clashes.

brian-dellabetta: Should be resolved now, `scheme` is never used if `recipe` is set, so i got rid of any references when that was the case. PR should be ready for review now. We had to move to a bigger model to have more robust results, so i changed one of the tests to not run GPTQ (it's run in another config). So i did change some file names, do we need to update anything for the timings CSVs?

The 3 tests ran added run in about an hour, maybe 1h15m.

brian-dellabetta: Reverted this, please let me know if things still look amiss

dbarbuzzi: Looks like `scheme` is removed now, rather than updated, which may cause issues with the naming for the CSV timings during the test. My understanding from skimming the test code is that the model folder and CSV filename that get saved during the test execution is pulled from the config in such a way that:

* If `save_dir` is defined, just use that, otherwise
* Join the `model_name` and `scheme` properties together to create the context-specific folder/filename

```
self.save_dir = eval_config.get("save_dir")
[…]
if not self.save_dir:
    self.save_dir = self.model.split("/")[1] + f"-{self.scheme}"
```

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight_dampfrac1e-1.yaml
```diff
@@ -0,0 +1,15 @@
+quant_stage:
+  quant_modifiers:
+    GPTQModifier:
+      dampening_frac: 0.1
```
kylesayrs, line 4: Just make sure that, for qwen, the inputs and model quantization looks correct. Hessian inversion issues can sometimes be the result of bad data or trying to quantize modules you didn't mean to quantize

tests/lmeval/test_lmeval.py
```diff
@@ -160,7 +163,11 @@ def _run_lm_eval(self):
             logger.info(
                 f"Comparing {metric}: Expected {expected_val}, Got {actual_val}"
             )
-            assert numpy.isclose(expected_val, actual_val, rtol=0.05)
+            if self.lmeval.use_stderr_atol:
+                std_err = metrics.get(metric.replace(",none", "_stderr,none"))
+                assert numpy.isclose(expected_val, actual_val, atol=std_err)
```
kylesayrs, line 168: Here it seems like you're using the standard error of the quantized model as a dynamic metric for determining absolute tolerance. This may be flawed, as if a model produces garage results, its std_err will be very large, and will therefore be less likely to trigger the threshold.

If anything, we should be using static tolerances determined from the standard error of the unquantized model on this task

> Okay, but you'd have to use the stderr of the base model

tests/lmeval/test_lmeval.py
```diff
@@ -160,7 +163,11 @@ def _run_lm_eval(self):
             logger.info(
                 f"Comparing {metric}: Expected {expected_val}, Got {actual_val}"
             )
-            assert numpy.isclose(expected_val, actual_val, rtol=0.05)
+            if self.lmeval.use_stderr_atol:
+                std_err = metrics.get(metric.replace(",none", "_stderr,none"))
```
kylesayrs, line 167: nit: This is a little bit hard to understand

tests/lmeval/test_lmeval.py
brian-dellabetta: I updated this to use stderr as absolute tolerance only if provided in metrics. Lmk what you think

tests/lmeval/test_lmeval.py
brian-dellabetta: I refactored the config and this logic to just use user-provided stderr vals, take a look and lmk what you think

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight_dampfrac1e-1.yaml
brian-dellabetta: This may no longer be necessary for the larger model. I'll try it quick

brian-dellabetta: This seems to work fine now with the larger model. I got rid of this recipe and switched back to `recipe_w4a16_actorder_weight.yaml`

kylesayrs: Okay, as long as we know that this recipe is working as expected

tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
```diff
@@ -0,0 +1,18 @@
...
+scheme: FP8_DYNAMIC
+lmeval:
+  model: "hf-multimodal"
+  model_args:
+    dtype: bfloat16
+    add_bos_token: True
+    convert_img_format: True
+  task: mmmu_val_literature
+  num_fewshot: 0
+  use_stderr_atol: True
```
kylesayrs, line 13: Seems like this is not used?

tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
```diff
@@ -0,0 +1,18 @@
...
+  batch_size: 8
+  # dense model achieves accuracy of 0.9 +/ 0.0557
+  metrics:
+    acc,none: 0.8667
+    acc_stderr,none: 0.0557
```
kylesayrs, line 18: This might be easier to parse
```suggestion
    acc,none: (0.8667, 0.0557)
```

```python3
value, std = value if isinstance(value, tuple) else value, 0.05
```

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight_dampfrac1e-1.yaml
brian-dellabetta: It's the same recipe, except for dampening_frac

kylesayrs (dismissed):

tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
```diff
@@ -1,19 +1,20 @@
 cadence: "weekly"
-model: Qwen/Qwen2-VL-2B-Instruct
-model_class: TraceableQwen2VLForConditionalGeneration
+model: Qwen/Qwen2.5-VL-7B-Instruct
+model_class: TraceableQwen2_5_VLForConditionalGeneration
+scheme: W4A16_actorder_group
```
rahul-tuli, line 4: Seems like this scheme still says `actorder_group` should be updated to `actorder_weight`?

rahul-tuli (commented):
LGTM pending one typo!

tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
brian-dellabetta: Must've been a typo left over from the previous PR. Updated, thanks for catching that

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight_dampfrac1e-1.yaml
brian-dellabetta: Removed this in favor of original recipe that doesn't set dampening_frac from default. Things seem to be running fine for the larger model

tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
dbarbuzzi: This looks good, thanks! I mostly wanted to make sure the `scheme` was correct (based on filename), because we've had collisions in the past related to different configs incorrectly having the same scheme (due to being overlooked, presumably after a copy-paste/starting from duplicated file).

brian-dellabetta:
> LGTM. Can you verify with Domenic that there will not be any conflicts with the existing tests that we're now tracking here? https://fantastic-adventure-plymqoj.pages.github.io/timings/lmeval/
> 
> I think it should be fine since we use the config name but just in case

Confirmed with domenic, I will ask rahul to approve and then merge this in

PR #1263, opened 2025-03-18
[Performance] Sequential onloading

### LINKED ISSUE #1078: Add support for W8A8 quantization with CPU weight offloading
I always face problems when trying to run oneshot calibration with CPU offloaded tensors, this limits GPUs with 24GiB VRAM from quantizing models larger than 7~8B.

**Describe the solution you'd like**
Add support for block swap / layer-wise loading

**Describe alternatives you've considered**
accelerate's offloading (device_map = "auto") and DeepSpeed ZeRO stage 3 doesn't work when I tried it with oneshot.

**Additional context**
Doing block swap / layer-wise loading with probably slow-down SmoothQuantModifier calibration by many times, but that enables running calibrations with longer context / bigger models.

### LINKED ISSUE #1122: Whole model gets offloaded to the CPU
I'm running the following code to calculate an offset map:
```
     MODEL_ID = "deepseek-ai/DeepSeek-R1"
     device_map = calculate_offload_device_map(MODEL_ID, num_gpus=8, reserve_for_hessians=True, trust_remote_code=True)
    
     print("calculated device map", device_map)
```
After it finishes it decides to offload the whole model to my CPU which results in a very slow compression. 

```
calculated device map OrderedDict({'model.embed_tokens': 'cpu', 'model.layers.0': 'cpu', 'model.layers.1': 'cpu', 'model.layers.2': 'cpu', 'model.layers.3': 'cpu', 'model.layers.4': 'cpu', 'model.layers.5': 'cpu', 'model.layers.6': 'cpu', 'model.layers.7': 'cpu', 'model.layers.8': 'cpu', 'model.layers.9': 'cpu', 'model.layers.10': 'cpu', 'model.layers.11': 'cpu', 'model.layers.12': 'cpu', 'model.layers.13': 'cpu', 'model.layers.14': 'cpu', 'model.layers.15': 'cpu', 'model.layers.16': 'cpu', 'model.layers.17': 'cpu', 'model.layers.18': 'cpu', 'model.layers.19': 'cpu', 'model.layers.20': 'cpu', 'model.layers.21': 'cpu', 'model.layers.22': 'cpu', 'model.layers.23': 'cpu', 'model.layers.24': 'cpu', 'model.layers.25': 'cpu', 'model.layers.26': 'cpu', 'model.layers.27': 'cpu', 'model.layers.28': 'disk', 'model.layers.29': 'disk', 'model.layers.30': 'disk', 'model.layers.31': 'disk', 'model.layers.32': 'disk', 'model.layers.33': 'disk', 'model.layers.34': 'disk', 'model.layers.35': 'disk', 'model.layers.36': 'disk', 'model.layers.37': 'disk', 'model.layers.38': 'disk', 'model.layers.39': 'disk', 'model.layers.40': 'disk', 'model.layers.41': 'disk', 'model.layers.42': 'disk', 'model.layers.43': 'disk', 'model.layers.44': 'disk', 'model.layers.45': 'disk', 'model.layers.46': 'disk', 'model.layers.47': 'disk', 'model.layers.48': 'disk', 'model.layers.49': 'disk', 'model.layers.50': 'disk', 'model.layers.51': 'disk', 'model.layers.52': 'disk', 'model.layers.53': 'disk', 'model.layers.54': 'disk', 'model.layers.55': 'disk', 'model.layers.56': 'disk', 'model.layers.57': 'disk', 'model.layers.58': 'disk', 'model.layers.59': 'disk', 'model.layers.60': 'disk', 'model.norm': 'disk', 'lm_head': 'disk'})
```

vllm 0.7.1
transformers 4.48.2
accelerate 1.0.1

Running on a node with 8xH100

### LINKED ISSUE #1216: Lazy Loading of Weights for Large Model Quantization
I noticed that LLM-Compressor needs to fully load the model weights onto the CPU/GPU before starting the pruning or quantization process. However, when dealing with very large models on the CPU, the pruning process becomes extremely slow and could take an impractically long time.

Would it be possible to implement a layer-by-layer pruning approach on the GPU to improve efficiency and reduce memory overhead? This would significantly speed up the process and make it more feasible for large-scale models.

### LINKED ISSUE #1228: Quantization Memory Requirements
Hello!

I was trying the various quantization recipes for quantizing a 70B Llama 3 based model to FP8, INT8, INT4(A16) precisions as mentioned in the [quantization docs by vLLM](https://docs.vllm.ai/en/stable/features/quantization/index.html). 

1. Could you help me understand the memory requirements for the quantization recipes, i.e SmoothQuant (SmoothQuantModifier), GPTQ (GPTQModifier) and RTN (QuantizationModifier). A calculation/formula would help, for example, like the one we have for calculating kv cache:
```
memory in bytes for kv cache = 80 (layers) * 8 (kv heads) * 128 (head_dim) * 8192 (seq length) * 2 (k and v) * 2 (fp16)
```
I understand that the `calculate_offload_device_map` creates a custom device map by reserving memory for 
GPTQ (reserve_for_hessians), but I would still like to understand the memory requirements to be able to utilize the GPU memory efficiently, to understand where all the GPU memory is consumed and to ensure that there are no bugs. 

2. Also, I understand that currently, for quantization of big models, the model is split in a pipeline parallel way on multiple GPUs available on the instance. 
- Since the GPU which is being used at any given time is the one which has the model layer that is being quantized at that time, would the time taken to quantize the model be similar to using a single GPU to quantize the model vs using multiple GPUs? 
- Is it possible to split the model in a tensor parallel way? 
- I understand that 'non-sequential GPTQ ' is deprecated, but how much memory is required for a non-sequential GPTQ? I think the above memory calculation would help. Also, how much speed up would we see using the non-sequential approach (compared to the sequential one)?

Thank you!

### LINKED ISSUE #1383: Load the model to CPU but quantize using the GPU
Some models are too large to fully load into the GPU, but (GPTQ and AWQ in particular) quantization is too slow on the CPU. Is it possible to load the model to CPU but quantize layer by layer, and before quantizing each layer move that layer to the GPU to do the quantization, then offload the layer after quantization finishes.
Most other quantization frameworks support this, for example in [GPTQModel](https://github.com/ModelCloud/GPTQModel/releases/tag/v1.5.1):
```
Added QuantizeConfig.device to clearly define which device is used for quantization: default = auto. Non-quantized models are always loaded on cpu by-default and each layer is moved to QuantizeConfig.device during quantization to minimize vram usage.
```

### LINKED ISSUE #1483: [Qwen2-72B-Instruct] CUDA out of memory when quant model with gptq
**Describe the bug**
I used the gptq method to quantize the Qwen2-72B-Instruct model on two A100*80GB nvidia card, 
Although I used **calculate_offload_device_map**, I still get CUDA out of memory.

**Expected behavior**
Ability to use gptq normal quantization Qwen2-72B-Instruct model

**Environment**
1. OS [e.g. Ubuntu 20.04]: centos 7
2. Python version [e.g. 3.7]: 3.10
3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]: 0.5.1
4. ML framework version(s) [e.g. torch 2.3.1]: torch 2.5.1
5. Other Python package versions [e.g. vLLM, compressed-tensors, numpy, ONNX]: vllm 0.7.3 compressed-tensors 0.9.4
6. Other relevant environment information [e.g. hardware, CUDA version]: cuda 12.2

**To Reproduce**
Exact steps to reproduce the behavior:
```
import torch


MODEL_ID = "/workspace/models/Qwen/Qwen2-72B-Instruct"
DATASET_ID = "/data/alpaca_data_cleaned.json"
DATASET_SPLIT = "train"
NUM_CALIBRATION_SAMPLES = 256
MAX_SEQUENCE_LENGTH = 256
OUTPUT_DIR = MODEL_ID.split("/")[-1] + "-gptq-int4"

    MODEL_ID,
    reserve_for_hessians=True,
    num_gpus=2,
    torch_dtype=torch.bfloat16,
    trust_remote_code=True,
)

    MODEL_ID, device_map=device_map, torch_dtype="auto"
)
recipe = GPTQModifier(targets="Linear", scheme="W4A16", ignore=["lm_head"], sequential_update=True)

def get_calib_dataset(tokenizer):

    ds = load_dataset(
        "json",
        data_files=DATASET_ID,
        split=f"{DATASET_SPLIT}[:{NUM_CALIBRATION_SAMPLES*100}]",
    )
    
    def preprocess(example):
        return {
            "text": f"Instruction:\n{example['instruction']}\nInput:\n{example['input']}\nOutput:\n{example['output']}".strip()
        }
    
    def tokenize(sample):
        return tokenizer(
            sample["text"],
            padding=False,
            truncation=True,
            add_special_tokens=False,
        )

    ds = (
        ds.map(preprocess)
        .map(tokenize, remove_columns=ds.column_names)
        .select(range(NUM_CALIBRATION_SAMPLES))
    )

    return ds


oneshot(
    model=model,
    dataset=get_calib_dataset(tokenizer=tokenizer),
    recipe=recipe,
    output_dir=OUTPUT_DIR,
)

print("Done! model saved to", OUTPUT_DIR)


```

**Errors**
```
2025-05-28T06:31:37.546360+0000 | compress | METRIC - error 2163.18
2025-05-28T06:31:37.546673+0000 | compress | METRIC - GPU 0 | usage: 97.51% | total memory: 85 GB
2025-05-28T06:31:37.546757+0000 | compress | METRIC - GPU 1 | usage: 98.13% | total memory: 85 GB
2025-05-28T06:31:37.546870+0000 | compress | METRIC - Compressed module size: 490.119168 MB
2025-05-28T06:31:37.547185+0000 | on_sequential_batch_end | INFO - Quantizing model.layers.37.mlp.down_proj using 256 samples
2025-05-28T06:31:41.172376+0000 | compress | METRIC - time 3.62s
2025-05-28T06:31:41.174574+0000 | compress | METRIC - GPU 0 | usage: 97.51% | total memory: 85 GB
2025-05-28T06:31:41.174859+0000 | compress | METRIC - GPU 1 | usage: 96.74% | total memory: 85 GB
2025-05-28T06:31:41.174978+0000 | compress | METRIC - Compressed module size: 490.119168 MB
  File "/workspace/models/llm_compressor_gptq.py", line 64, in <module>
    oneshot(
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/entrypoints/oneshot.py", line 179, in oneshot
    one_shot()
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/entrypoints/oneshot.py", line 131, in __call__
    self.apply_recipe_modifiers(
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/entrypoints/oneshot.py", line 173, in apply_recipe_modifiers
    session.initialize(**session_kwargs)
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/core/session.py", line 118, in initialize
    mod_data = self._lifecycle.initialize(
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/core/lifecycle.py", line 105, in initialize
    data = mod.initialize(state=self.state, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/modifiers/stage.py", line 83, in initialize
    modifier.initialize(state, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/modifiers/modifier.py", line 90, in initialize
    self.initialized_ = self.on_initialize(state=state, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/modifiers/quantization/gptq/base.py", line 252, in on_initialize
    raise exception
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/modifiers/quantization/gptq/base.py", line 234, in on_initialize
    run_sequential(
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/pipelines/sequential/pipeline.py", line 71, in run_pipeline
    callback_modifier.on_sequential_batch_end()
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/modifiers/quantization/gptq/base.py", line 345, in on_sequential_batch_end
    loss, quantized_weight, scale, zero_point, g_idx = quantize_weight(
  File "/usr/local/lib/python3.10/dist-packages/llmcompressor/modifiers/quantization/gptq/gptq_quantize.py", line 161, in quantize_weight
    H = torch.cholesky_inverse(H)
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 3.26 GiB. GPU 1 has a total capacity of 79.15 GiB of which 2.61 GiB is free. Process 966714 has 76.53 GiB memory in use. Of the allocated memory 74.06 GiB is allocated by PyTorch, and 1.95 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)

```
**Additional context**

kylesayrs:
# Sequential Onloading #
<p align="center">Screenshot 2025-06-05 at 22 53 01</p>

```
(25/33): Calibrating:   0%|                                                                                                                                  | 0/512 [00:00<?, ?it/s]
<class 'transformers.models.llama.modeling_llama.LlamaRMSNorm'>.weight -> cuda
<class 'torch.nn.modules.linear.Linear'>.weight -> cuda
<class 'torch.nn.modules.linear.Linear'>.weight_scale -> cuda
<class 'torch.nn.modules.linear.Linear'>.weight_zero_point -> cuda
...
(25/33): Calibrating: 100%|█████| 512/512 [00:23<00:00, 21.91it/s]
2025-06-03T17:29:15.536963-0400 | compress_modules | INFO - Quantizing model.layers.24.self_attn.q_proj using 512 samples
2025-06-03T17:29:17.328720-0400 | compress | METRIC - time 1.79s
2025-06-03T17:29:17.329265-0400 | compress | METRIC - error 8948.54
2025-06-03T17:29:17.329781-0400 | compress | METRIC - GPU 0 | usage: 5.41% | total memory: 85 GB
2025-06-03T17:29:17.330248-0400 | compress | METRIC - Compressed module size: 33.947648 MB
...
(25/33): Propagating: 100%|█████| 512/512 [00:03<00:00, 131.16it/s]
<class 'transformers.models.llama.modeling_llama.LlamaRMSNorm'>.weight -> meta
<class 'torch.nn.modules.linear.Linear'>.weight -> meta
<class 'torch.nn.modules.linear.Linear'>.weight_scale -> meta
<class 'torch.nn.modules.linear.Linear'>.weight_zero_point -> meta
...
```

## Purpose ##
* Reduce hardware requirements for calibrating large models
* Reduce runtime caused by excess device movement when calibrating offloaded models

## Prerequisites ##
* https://github.com/neuralmagic/compressed-tensors/pull/354
* https://github.com/neuralmagic/compressed-tensors/pull/355
* https://github.com/neuralmagic/compressed-tensors/pull/356
* https://github.com/neuralmagic/compressed-tensors/pull/357

## Related Issues ##
* Resolves #1383
* Resolves #1228
* Resolves #1122
* Resolves #1078
* Resolves #1216
* Resolves #1483

## Changes ##
* Keep layer parameters onloaded during the entire sequential calibration + compression + propagation step
  * This is achieved through the `keep_onload_context`, which disables offloaded until the context is exited
* Dispatch model within each calibration pipeline
  * Sequential pipeline offloads the model to CPU, and executes on the first cuda device
  * Deprecate passing sequential_targets via modifiers, instead prefer passing via oneshot argument
* Use sequential pipeline as default pipeline (basic pipeline is never used)
  * Deprecate passing sequential_targets via modifiers, instead prefer passing via oneshot argument
* Dispatch model before sample generation
  * The model is dispatched exactly as it would be if it was loaded with `device_map="auto"`

### Examples ###
* Models are loaded onto cpu before oneshot (rather than being dispatched across GPUs)
* Model is reloaded from disk in order to redispatch onto "auto" device map
  * In my opinion, this is a better flow anyways, since models can raise errors / take a very long time during generation, which can cause the entirely compression job to go to waste
  * The alternative is to either call `accelerate.remove_hooks(model)` and `accelerate.dispatch_model(model)` before generating, or get rid of sample generation entirely. One of these may be required if compressed_linear isn't reliable enough to add to our examples

<details><summary>New example script</summary>

```python3

from llmcompressor.utils.dev import dispatch_for_generation

# Load model (on cpu)
model_id = "meta-llama/Meta-Llama-3-8B-Instruct"
model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype="auto")  # model is loaded on cpu

# Define recipe

# Apply oneshot (model execution device is set to cuda, model stays on cpu)
oneshot(
    model=model,
    dataset="ultrachat_200k",
    recipe=recipe,
    max_seq_length=2048,
    num_calibration_samples=512,
)

# Perform sample generation
print("\n\n")

# Save to disk before generating
SAVE_DIR = model_id.split("/")[1] + "-W4A16-G128"
```
</details>

## Testing ##
* Calibrated and GPTQ-compressed one layer of Deepseek-V3 with a single H100 in 50 seconds
  * 4.5x Improvement over original 236 seconds
  * Peak memory of ~40 GB, which can be further reduced by increasing the granularity of sequential targets
* Not offloading activations did not result in a performance improvement
* TODO: Test all example models can be reloaded and run


src/llmcompressor/utils/helpers.py
```diff
@@ -1138,3 +1140,18 @@ def preserve_attr(base: object, attr: str):
         yield
     finally:
         setattr(base, attr, value)
+
+
+@contextlib.contextmanager
+def align_modules(modules: Iterable[torch.nn.Module]):
```
dsikka, line 1146: Why not keep this in compressed tensors with the other cpu offloading tools?

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -63,26 +63,29 @@ def run_pipeline(
...
-
-            # TODO: replace with a lifecycle event
-            if callback_modifier:
-                callback_modifier.on_sequential_batch_end()
-
-            # this pass does not trigger modifier hooks
-            # and is only used for capturing outputs from the newly compressed modules
-            with HooksMixin.disable_hooks():
-                for batch_index in tqdm.tqdm(range(len(dataloader)), desc=prop_desc):
+            with align_modules([layer]):
```
dsikka, line 66: Seems like all we're doing is wrapping the forward passes in this context manager, if I'm reading this correctly?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -380,3 +385,8 @@ def match_modules(model: Module, target_names: List[str]) -> Set[Module]:
         for name, module in model.named_modules()
         if find_name_or_class_matches(name, module, target_names)
     )
+
+
+def get_subgraph_modules(subgraph: Graph, parent_graph: GraphModule) -> List[Module]:
```
dsikka, line 390: Docstring

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -310,11 +313,13 @@ def partition_graph(model: Module, partitions: List[List[Node]]) -> List[Subgrap
         # save the subgraph for this partition
         graph.lint()
         input_names = set(node.name for node in graph.nodes if node.op == "placeholder")
+        modules = get_subgraph_modules(graph, parent_graph)
```
dsikka, line 316: Do you mind explaining what we're changing in our graph partition here?

brian-dellabetta (approved):

dsikka (changes requested):

brian-dellabetta (commented):
sorry, i approved this thinking it was the one-liner removing clear-ml, will have to take a closer look

src/llmcompressor/utils/helpers.py
kylesayrs: Yep! Implementing here before the next CT release

src/llmcompressor/pipelines/layer_sequential/pipeline.py
kylesayrs: Yes. Rather than onloading then discarding for each of the 512 forward passes, we onload once for the layer and keep it onloaded through compression and propagation.

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: The graph partition doesn't change, this change just collects all the modules used by this subgraph for use in onloading/offloading by the sequential pipeline.

brian-dellabetta (dismissed):
I am understanding this for the most part -- very cool!

src/llmcompressor/utils/helpers.py
kylesayrs: https://github.com/neuralmagic/compressed-tensors/pull/282

src/llmcompressor/utils/helpers.py
```diff
@@ -1059,3 +1061,28 @@ def preserve_attr(base: object, attr: str):
         yield
     finally:
         setattr(base, attr, value)
+
+
+@contextlib.contextmanager
+def align_modules(
+    modules: Iterable[torch.nn.Module], execution_device: Optional[torch.device] = None
```
brian-dellabetta, line 1068: and this will be removed?

brian-dellabetta (dismissed):

src/llmcompressor/utils/helpers.py
kylesayrs: Yes, after the next CT release

brian-dellabetta: ok, if this PR is intended to be merged in as is, a TODO to indicate that would be helpful
```suggestion
# TODO remove after https://github.com/neuralmagic/compressed-tensors/pull/282
@contextlib.contextmanager
def align_modules(
    modules: Iterable[torch.nn.Module], execution_device: Optional[torch.device] = None
```

brian-dellabetta (approved):

kylesayrs:
```python3
def DoNotOffloadContext():
	  to_offload = {}
    def patched(module):
        to_offload.add(module)
        return None

    with patch_attr(AlignDeviceHook, "post_forward", patched)
        yield
        
    # offload on exit
    for module in to_offload:
        module.post_forward()
        
        
for subgraph in subgraphs():
    with DoNotOffloadContext():
        subgraph(**inputs)
```

src/llmcompressor/pipelines/registry.py
```diff
@@ -18,6 +18,7 @@
 __all__ = ["CalibrationPipeline"]
 
 SEQUENTIAL_MODIFIERS = (AWQModifier, GPTQModifier, SparsityModifierBase)
+NEED_DATA = (SmoothQuantModifier, *SEQUENTIAL_MODIFIERS)
```
brian-dellabetta: What about `CALIBRATED_MODIFIERS` to make name consistent with `SEQUENTIAL_MODIFIERS`?
```suggestion
CALIBRATED_MODIFIERS = (SmoothQuantModifier, *SEQUENTIAL_MODIFIERS)
```

kylesayrs: 100%

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -123,6 +126,17 @@ def __init__(
...
         pre_process(model_args)
 
+        # offload to cpu if possible
+        if "cuda" in str(model_args.oneshot_device) and torch.cuda.is_available():
+            # TODO: consider renaming function similar to "offload_dispatch_model"
+            # TODO: modify function to remove any hooks if they already exist (making
+            # sure to move to cpu when removing hook
+            # TODO: remove hook in util
+            remove_hook_from_module(model_args.model, recurse=True)
+            force_cpu_offload(model_args.model, model_args.oneshot_device)
```
brian-dellabetta: I like the `offload_dispatch_model` name over `force_cpu_offload`, possibly even making `devce=torch.device("cpu")` an input parameter if we want to offload to disk or some other device in future.

Do these other `TODO`s still hold?

src/llmcompressor/utils/module.py
```diff
@@ -0,0 +1,29 @@
+from typing import Callable, Union
+
+import tqdm
+from torch.nn import Module
+
+
+def module_bfs(
```
brian-dellabetta: docstring would be helpful here, and do we want to expose this in `utils/__init__.py`?

kylesayrs: This was a mistake and shouldn't have been committed

kylesayrs:
Spoke with @brian-dellabetta, we imagine an optimal example script might look something like this

```python3
model = load(..., device_map="cpu")
model = oneshot(model)
compressed_tensors.compress(model)
model.save_pretrained(...)

compressed_tensors.decompress(model)
dispatch_model(model, device_map="auto")
... run sample generation
```

Or, using a context
```python3
model = load(..., device_map="cpu")
model = oneshot(model)
with compressed_tensors.compress(model):
    model.save_pretrained(...)

dispatch_model(model, device_map="auto")
... run sample generation
```

src/llmcompressor/entrypoints/oneshot.py
kylesayrs: No, the other todos should be removed

logos-mhr:
Excellent job! Sequential onloading is definitly essential for larger models.
But I encountered a potential bug when trying to apply GPTQ to DeepSeek v2-lite, which made` weight_scale` and `weight_zero_point` of the quantized Linear modules meta tensors.
It seems this may be caused by the hook of  `force_cpu_offload`.  When applying that hook, the Linear modules does not have ` weight_scale` and `weight_zero_point` attributes. Could that be the potiential cause? Or could it be solved by simply adding place holding tensors?

logos-mhr:
> Excellent job! Sequential onloading is definitly essential for larger models. But I encountered a potential bug when trying to apply GPTQ to DeepSeek v2-lite, which made` weight_scale` and `weight_zero_point` of the quantized Linear modules meta tensors. It seems this may be caused by the hook of `force_cpu_offload`. When applying that hook, the Linear modules does not have ` weight_scale` and `weight_zero_point` attributes. Could that be the potiential cause? Or could it be solved by simply adding place holding tensors?

Apply `remove_hook_from_module` and `force_cpu_offload`  after `session.initialize` in `one_shot.py` will do the job. This may indicate my assumption above is correct.

kylesayrs:
Hi @logos-mhr!

> But I encountered a potential bug when trying to apply GPTQ to DeepSeek v2-lite, which made weight_scale and weight_zero_point of the quantized Linear modules meta tensors.

Could you specify where exactly in the code you encounter this bug? Those parameters (indeed all parameters) are meant to stay on the `meta` device, until they are onloaded within the `disable_offloading` context, so observing that the parameters are on the meta device is expected behavior, unless it's during calibration and quantization steps.

```
# the parameters start offloaded to the meta device
<class 'torch.nn.modules.linear.Linear'>.weight -> meta
<class 'torch.nn.modules.linear.Linear'>.weight_scale -> meta
<class 'torch.nn.modules.linear.Linear'>.weight_zero_point -> meta

# calibration begins
(1/33): Calibrating:   0%|                                                                                                                                   | 0/512 [00:00<?, ?it/s]

# the parameters are onloaded and stay onloaded until the layer is finished
<class 'torch.nn.modules.sparse.Embedding'>.weight -> cuda
<class 'transformers.models.llama.modeling_llama.LlamaRMSNorm'>.weight -> cuda
<class 'torch.nn.modules.linear.Linear'>.weight -> cuda
<class 'torch.nn.modules.linear.Linear'>.weight_scale -> cuda
```

kylesayrs:
Hi @logos-mhr, I was able to replicate the issue I believe you're referring to, namely that the qparams are on the meta device after the hooks are removed. I'll put up a fix soon

EDIT: Fix is here: https://github.com/neuralmagic/compressed-tensors/pull/356, thanks again!

examples/awq/qwen3_moe_example.py
```diff
@@ -3,13 +3,12 @@
```
dsikka, line 3: Instead of deleting the entire big_models_with_accleerate folder, I would keep a readme highlighting/summarizing how the models are handled now  i.e cover the basics of your PR description

src/llmcompressor/utils/dev.py
```diff
@@ -106,3 +112,14 @@ def patch_transformers_logger_level(level: int = logging.ERROR):
     transformers_logger.setLevel(level=level)
     yield
     transformers_logger.setLevel(level=restore_log_level)
+
+
+def dispatch_for_generation(model: PreTrainedModel) -> PreTrainedModel:
```
dsikka, line 118: docstring

src/llmcompressor/transformers/compression/helpers.py
```diff
@@ -111,36 +104,6 @@ def infer_sparsity_structure_from_model(model: torch.nn.Module) -> Optional[str]
...
-                    max_column_size[no_split_name] = column_size
-
-    max_total_hessian_elems = max(total_hessian_elems.values())
-    overall_max_column_size = max(max_column_size.values())
-    bytes_per_weight = 32 // 8  # hessians are float32
-    inverse_reserved = overall_max_column_size * overall_max_column_size
-    return (max_total_hessian_elems + inverse_reserved) * bytes_per_weight
-
-
 def quantization_memory_requirement(model: torch.nn.Module) -> int:
```
dsikka: do we still need quantization_memory_requirements?

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -123,6 +126,13 @@ def __init__(
         # initialize the model and processor
         pre_process(model_args)
 
+        # offload to cpu if possible
+        if "cuda" in str(model_args.oneshot_device) and torch.cuda.is_available():
+            remove_hook_from_module(model_args.model, recurse=True)
```
dsikka: What hooks would we be removing at this point?

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -123,6 +126,13 @@ def __init__(
...
+            offloaded_dispatch(model_args.model, model_args.oneshot_device)
```
dsikka: Use keyword args

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -123,6 +126,13 @@ def __init__(
         # initialize the model and processor
         pre_process(model_args)
 
+        # offload to cpu if possible
+        if "cuda" in str(model_args.oneshot_device) and torch.cuda.is_available():
```
dsikka: Why not move this to preprocess?

src/llmcompressor/entrypoints/utils.py
```diff
@@ -105,6 +106,9 @@ def post_process(
             "Ex. `oneshot(..., output_dir=...)`"
         )
 
+    # Remove any existing hooks (maybe added by oneshot sequential onloading)
+    remove_hook_from_module(model_args.model, recurse=True)
```
dsikka: this would be removing AlignDevicesHook?

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -73,29 +85,34 @@ def __call__(
...
-                    layer(**inputs)
-
-                LifecycleCallbacks.sequential_epoch_end()
-
-                # this pass does not trigger modifier hooks
-                # and is only used for capturing outputs from newly compressed modules
-                with HooksMixin.disable_hooks():
-                    for batch_idx in tqdm.tqdm(range(len(dataloader)), desc=prop_desc):
+                # reduce memory movement by keeping modules onloaded
+                with disable_offloading():
```
dsikka, line 84: does this mean all activations stay onloaded as well? or will the cache still offload them?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -59,37 +70,39 @@ def __call__(
...
         subgraphs = trace_subgraphs(model, sample_input, sequential_targets, ignore)
+        num_subgraphs = len(subgraphs)
 
         LifecycleCallbacks.calibration_epoch_start()
 
         with calibration_forward_context(model), DisableQuantization(model):
             # prepare intermediates cache
             model_device = get_execution_device(model)
-            intermediates = IntermediatesCache.from_dataloader(dataloader, model_device)
+            activations = IntermediatesCache.from_dataloader(dataloader, model_device)
```
dsikka: sorry, comment about offloading activations was for here

src/llmcompressor/pipelines/registry.py
```diff
@@ -75,12 +76,7 @@ def _validate_infer_pipeline(modifiers: List[Modifier]) -> str:
             quant_modifier = active_qmods[0]
             config = quant_modifier.resolve_quantization_config()
             if config.requires_calibration_data():
-                return "basic"
-            else:
-                return "datafree"
-
-        if any(isinstance(modifier, SmoothQuantModifier) for modifier in modifiers):
-            return "basic"
+                return "sequential"
```
dsikka: So datafree pipeline, which is not sequenitally onloaded, should be just for the QuantizationModifier at this point, when doing weight-only quantization?

src/llmcompressor/pipelines/registry.py
```diff
@@ -60,7 +61,7 @@ def from_modifiers(
```
dsikka, line 55: Seems like we can remove the basic pipeline file?

dsikka (commented):
Overall LGTM.
Some questions - will approve once we've gone through and tested the modifiers 
Kicking off llama 3.3 70b runs with 1 GPU over the weekend would be a good way to evlauate.

src/llmcompressor/entrypoints/oneshot.py
dsikka: We offload the model before setting up the qparams, correct?
And qparams are initialized on the execution device https://github.com/neuralmagic/compressed-tensors/blob/3fb2844be33459e80ab813d90f0b7f98c3a9ce9a/src/compressed_tensors/quantization/lifecycle/initialize.py#L154? 

How do we make sure they're also offloaded?

src/llmcompressor/entrypoints/oneshot.py
kylesayrs: Theoretically none if users only follow the examples exactly, but this is to foolproof against any previous dispatches
FYI, the `remove_hook_from_module` call has been moved to the `offloaded_dispatch` function def
https://github.com/neuralmagic/compressed-tensors/pull/357

src/llmcompressor/transformers/compression/helpers.py
kylesayrs: No, thanks for the catch

src/llmcompressor/entrypoints/oneshot.py
kylesayrs: In the case of training, we do not want to offload the model. Another way of saying this is, `model_args.oneshot_device` should only apply to oneshot, not training

src/llmcompressor/entrypoints/utils.py
kylesayrs: I moved this to within `dispatch_for_generation`.

In summary yes, we want to remove hooks for sequential calibration before adding hooks for generation

src/llmcompressor/pipelines/layer_sequential/pipeline.py
kylesayrs: activation caching via the `IntermediatesCache` is unaffected

src/llmcompressor/entrypoints/oneshot.py
kylesayrs: Newly registered parameters are offloaded if applicable
https://github.com/neuralmagic/compressed-tensors/blob/3fb2844be33459e80ab813d90f0b7f98c3a9ce9a/src/compressed_tensors/utils/offload.py#L212-L213

src/llmcompressor/pipelines/registry.py
kylesayrs: It's still used by [sparsegpt](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/obcq/sgpt_base.py#L252) for the OWL feature. I suspect this is also worth maintaining as an experimental research option.

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15670690251

lyzhongcrd:
Great job! I tried to quantize Qwen2.5-VL-32B on an A100 (failed due to OOM) and found this PR. When shall it be released? Can't wait to use this new feature.

src/llmcompressor/entrypoints/oneshot.py
kylesayrs: EDIT: I've since removed `oneshot_device` functionality. Instead, the model is dispatched within the pipelines

kylesayrs:
@lyzhongcrd We expect to release later this week or early next week

examples/big_models_with_sequential_onloading/README.md
```diff
@@ -0,0 +1,12 @@
+## Big Modeling with Sequential Onloading ##
+### What is Sequential Onloading? ###
+Sequential onloading is a memory-efficient approach for compressing large language models (LLMs) using only a single GPU. Instead of loading the entire model into memory—which can easily require hundreds of gigabytes—this method loads and compresses one layer at a time. The outputs are offloaded before the next layer is processed, dramatically reducing peak memory usage while maintaining high compression fidelity.
```
brian-dellabetta, line 3: 🚀

examples/multimodal_vision/llava_example.py
```diff
@@ -47,10 +45,12 @@ def data_collator(batch):
     num_calibration_samples=NUM_CALIBRATION_SAMPLES,
     trust_remote_code_model=True,
     data_collator=data_collator,
+    sequential_targets=["LlamaDecoderLayer"],
```
brian-dellabetta, line 48: we should make sure this is reflected in oneshot args after #1445 lands

examples/quantizing_moe/deepseek_moe_w4a16.py
```diff
@@ -4,7 +4,7 @@
 from transformers import AutoModelForCausalLM, AutoTokenizer, __version__
 
 from llmcompressor import oneshot
-from llmcompressor.transformers.compression.helpers import calculate_offload_device_map
+from llmcompressor.utils.dev import dispatch_for_generation
```
brian-dellabetta: since this is used heavily, can we move it to just `llmcompressor.utils` or higher-scoped submodule?

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -186,3 +186,13 @@ class DatasetArguments(CustomDatasetArguments):
             "{module}.{method_name} or {function_name}"
         },
     )
+    sequential_targets: Optional[List[str]] = field(
```
brian-dellabetta, line 189: i would think this would be on recipe arguments, but since `pipeline` is in DatasetArguments as well, it should live on the same dataclass as that.

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -37,6 +38,7 @@ def __call__(
         :param dataloader: loads data for calibration
         :param dataset_args: dataset arguments relevant to pipelines
         """
+        dispatch_for_generation(model)
```
brian-dellabetta: why `dispatch_for_generation` here? The calibration has yet to occur

brian-dellabetta (approved):
Changes look good! some nits/comments

src/llmcompressor/args/dataset_arguments.py
kylesayrs: Yeah, it's mostly because we pass dataset arguments to the pipelines, not any other arguments. We could at some point break out "pipeline arguments" in the future

examples/multimodal_vision/llava_example.py
kylesayrs: Good point

examples/quantizing_moe/deepseek_moe_w4a16.py
kylesayrs: Makes sense

src/llmcompressor/pipelines/basic/pipeline.py
kylesayrs: The dispatch for generation is identical to the dispatch for basic calibration; I'll add a comment

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15687111832

src/llmcompressor/pipelines/registry.py
kylesayrs: Yes, by nature of the fact that that is the only configuration where `modifier not in CALIBRATION_MODIFIERS` and `not config.requires_calibration_data()`

kylesayrs: I've simplified this logic to make this clearer

kylesayrs:
Nightly passes
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15689881695

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -9,6 +9,7 @@
 from llmcompressor.modifiers.utils.pytorch_helpers import apply_pad_mask_to_batch
 from llmcompressor.pipelines.registry import CalibrationPipeline
 from llmcompressor.pytorch.utils.helpers import tensors_to_device
+from llmcompressor.utils.dev import dispatch_for_generation
```
brian-dellabetta: Is this also available in `llmcompresor.utils.dev`?

brian-dellabetta (approved):
awesome stuff!

src/llmcompressor/pipelines/basic/pipeline.py
kylesayrs: Thanks for the catch, updated

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15694064423

brian-dellabetta (approved):
pretty sweet that we're able to remove ~500 lines of code while adding a huge feature like this 🔥 

Screenshot 2025-06-17 at 1 50 41 PM

dsikka (approved):
Great work

PR #1264, opened 2025-03-18
[Tests] Use requires_gpu, fix missing gpu test skip, add explicit test for gpu from gha

kylesayrs:
## Purpose ##
* Update all tests to use `requires_gpu` decorator
* Add GPU mark skip for `test_compressor_stacking`, which requires a GPU
* Add an explicit GPU test for GHA, so as to unambiguously catch situations where CUDA is not properly installed on a runner


dsikka (changes requested):
Why not update existing tools?
https://github.com/vllm-project/llm-compressor/blob/cf6a3042c37e54b4b8ead574fcd24a50c7538731/tests/testing_utils.py#L40

kylesayrs:
@dsikka If you're referring to [here](https://github.com/vllm-project/llm-compressor/pull/1264/files#diff-6925aa003f1a6c657e53a26b79dda46f7ed01c9b91f3423c586b879e4e33dd9aR354) I'm just following the convention of the file. Theoretically we could consolidate these methods at some point, esp if we want to be support amd

dsikka (commented):
still not sure why we need this if these exist: tests/testing_utils.py

dsikka (approved):

brian-dellabetta (approved):

PR #1266, opened 2025-03-18
Replace tj-actions with custom action

tarukumar:
SUMMARY:
Due to recent security issue. Modify the workflow to use custom action

wait for this pr to be merged : https://github.com/neuralmagic/nm-actions/pull/46/

### DIFF
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -23,22 +23,13 @@ jobs:
           fetch-depth: 0
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v45
+        uses: neuralmagic/nm-actions/actions/changed-files@main
         with:
-          files: |
-            **
-            !examples/**
-            !tests/e2e/**
-            !tests/lmeval/**
-            !tests/examples/**
-            !**/*.md
-            !.github/**
-            .github/workflows/test-check-transformers.yaml
-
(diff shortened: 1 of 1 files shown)


dbarbuzzi:
https://github.com/neuralmagic/nm-actions/pull/46 has since been merged. @tarukumar can you please update this PR to use a non-branch tag reference to the action (either release tag or commit SHA)? If you prefer to use a tag and are unable to create tags in the `nm-actions` repo, let me know, and I’ll add one (and optionally update this PR if you prefer).

.github/workflows/test-check-transformers.yaml
```diff
@@ -12,33 +12,28 @@ env:
...
-            !.github/**
             .github/workflows/test-check-transformers.yaml
-
+          exclude-patterns: |
+            examples/**
+            tests/e2e/**
+            tests/lmeval/**
+            tests/examples/**
+            **/*.md
+            .github/**
```
dbarbuzzi, line 33: Based on the logic of how the new action works, I think this input/line specifically (`.github/**`) should be removed as it's no longer necessary and would actually prevent the one included in that file from being included, since exclusions always take precedence over inclusions.

dbarbuzzi (changes requested):
Aside from other comments on the PR, this seems like it is missing required changes. Specifically, in the `transformers-tests` job, the condition on the job itself needs to be updated according to the new action’s output (unless this action also provides a boolean `changes-present` output?).

andy-neuma (approved):
thanks

PR #1170, opened 2025-02-19
[Callbacks] Remove EventLifecycle and on_start event

kylesayrs:
## Purpose ##
* Simplify code related to callbacks
  * Remove event lifecycle
    * Callback event lifecycle
    * Optimizer event lifecycle
  * Remove the concept of a "start_event", which was originally used because initialization sometimes requires triggering on_start, and on_start requires an event in order to get an index and index-related info
    * For now, we create a dummy event on initialization which has an index of zero
    * In the future, all start events will be triggered by events (never initialize) and all event events will be triggered by events or finalization

## Prerequisites ##
* #1160

## Changes ##

* Instead of using event lifecycle as a proxy, pass events to modifiers directly
```python3
- self._check_setup_event_lifecycle(event_type)
+ event = Event(event_type=event_type)
- for event in self.event_lifecycle.events_from_type(event_type):
+ for mod in self.modifiers:
     data = mod.update_event(state=self.state, event=event, **kwargs)
```
  * This was originally needed for optimizer event lifecycle, that is now removed
* Make silent noops into errors
```python3
  if not self.initialized_:
-     return
+     raise RuntimeError("Cannot update an uninitialized modifier")

```
* Remove `check_initialized`, which was used to allow for checking in situations which required double initialization

PR #1267, opened 2025-03-18
[Bugfix] Remove constant pruning from 2of4 example, do not start on initialize

kylesayrs:
## Background ##
* When #1170 landed, an assumption was made that modifiers should be capable of starting on initialize
  * However, seems to never have been the case, as a `start_event` was required, and afaict start_events are only determined by events (after initialize)
  * Now, we make the assumption that modifiers only start on events (not initialize)


rahul-tuli (dismissed):
🚀

rahul-tuli (approved):

brian-dellabetta (approved):

PR #1270, opened 2025-03-19
Replace tj-action/changed-files

dbarbuzzi:
SUMMARY:
Replace the use of the formerly compromised `tj-action/changed-files` action.

TEST PLAN:
There are a couple of runs during development of the action itself which indicate the changes behave as expected:
- https://github.com/vllm-project/llm-compressor/actions/runs/13953464034 — similar to the changes in PR regarding patterns used/expected matches
- https://github.com/vllm-project/llm-compressor/actions/runs/13954383262/job/39061707849 — a slightly more comprehensive test with similar patterns, but more files touched to ensure only the desired files are considered a match


.github/workflows/test-check-transformers.yaml
```diff
@@ -23,28 +23,29 @@ jobs:
           fetch-depth: 0
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v45
+        uses: neuralmagic/nm-actions/actions/changed-files@57598f52d308d41e054b11ce9a52fc32a068ffc5 # v1.14.0
```
tarukumar, line 26: can we remove the comment at the end

dbarbuzzi: Is there a reason to remove it? I ask because I specifically added the comment because, when reading the file to understand it, seeing the comment which is simply the referenced version helps a lot more than having to look up the commit hash in the source repo.

tarukumar: I personally don't prefer to have the comment in the workflow, but if people are interested, they can easily look it up and find it. However, I don't have a strong opinion on it—if you'd like to keep it, that's fine.

tarukumar (approved):
/lgtm

.github/workflows/test-check-transformers.yaml
dbarbuzzi: Fair enough, I’ll leave this open to other contributors/maintainers if they prefer having it or not.

kylesayrs (approved):

brian-dellabetta (approved):

Issue #37, opened 2024-07-24
SmoothQuant Mappings Only Work When Defined in a Recipe String


Satrat:
**Describe the bug**
When a oneshot recipe is defined programmatically, the SmoothQuant mappings are not correctly parsed. This bug does not occur when the recipe is specified as a YAML string or file

**Expected behavior**
SmoothQuantModifier is correctly initialized, and oneshot runs to completion.

**Environment**
Include all relevant environment information:
1. OS [e.g. Ubuntu 18.04]: Ubuntu
2. Python version [e.g. 3.7]: 3.10.12
3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]: main, 07c1fd7
4. ML framework version(s) [e.g. torch 1.7.1]: torch 2.3.1, transformers 4.42.4
5. Other Python package versions [e.g. SparseZoo, DeepSparse, numpy, ONNX]: n/a
6. Other relevant environment information [e.g. hardware, CUDA version]: CUDA 12.3

**To Reproduce**
Example script:
```
from llmcompressor.modifiers.smoothquant.base import DEFAULT_SMOOTHQUANT_MAPPINGS
from llmcompressor.modifiers.smoothquant import SmoothQuantModifier
from llmcompressor.modifiers.quantization.gptq import GPTQModifier

from llmcompressor.transformers import SparseAutoModelForCausalLM, oneshot

recipe = [
    SmoothQuantModifier(smoothing_strength=0.8, mappings=DEFAULT_SMOOTHQUANT_MAPPINGS),
    GPTQModifier(targets="Linear", scheme="W8A8", ignore=["lm_head"], sequential_update=False),
]

model_stub = "TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T"
model = SparseAutoModelForCausalLM.from_pretrained(model_stub, device_map="auto", torch_dtype="auto")

dataset = "ultrachat-200k"
output_dir = "./test_output"
splits = {"calibration": "train_gen[:5%]"}
max_seq_length = 2048
pad_to_max_length = False
num_calibration_samples = 8

oneshot(
    model=model,
    dataset=dataset,
    recipe=recipe,
    output_dir=output_dir,
    splits=splits,
    max_seq_length=max_seq_length,
    pad_to_max_length=pad_to_max_length,
    num_calibration_samples=num_calibration_samples,
    save_compressed=True
)

```

**Errors**
oneshot fails with the following error:

```
[...]

PR #1271, opened 2025-03-20
[BugFix] Directly Convert Modifiers to Recipe Instance

rahul-tuli:
Currently, the process of recipe creation follows this sequence:

**Modifiers → String (Serialization) → Recipe Instance (Deserialization)**

This intermediate serialization and deserialization step introduces issues when dealing with more complex objects, such as **SmoothQuant mappings**, which can lead to parsing errors.

### Solution
This PR refactors the flow to directly construct the **Recipe Instance** from **Modifiers**, thereby **removing an unnecessary conversion step** and eliminating a potential source of error.

### Issue Tracking
This issue was originally surfaced in [[vllm-project/llm-compressor#37](https://github.com/vllm-project/llm-compressor/issues/37)](https://github.com/vllm-project/llm-compressor/issues/37) and is formally tracked under **[[INFERENG-358](https://issues.redhat.com/browse/INFERENG-358)](https://issues.redhat.com/browse/INFERENG-358)**.

### Testing
The issue was reproduced using the following script, which previously errored out but now runs **successfully** with this fix:

```python
from llmcompressor.modifiers.smoothquant import SmoothQuantModifier

MODEL_ID = "bigscience/bloom-3b"
DATASET_SPLIT = "train_sft"
NUM_CALIBRATION_SAMPLES = 512
MAX_SEQUENCE_LENGTH = 2048

# Load model and tokenizer
)

# Define quantization recipe
recipe = [
    SmoothQuantModifier(
        smoothing_strength=0.8,
        mappings=[
            (["re:.*query_key_value"], "re:.*input_layernorm"),
            (["re:.*dense_h_to_4h"], "re:.*post_attention_layernorm"),
        ],
    ),
    GPTQModifier(
        scheme="W8A8",
        targets="Linear",
        ignore=["lm_head"],
        dampening_frac=0.003,
    ),
]

# Load and preprocess dataset
dataset = load_dataset(DATASET_ID, split=DATASET_SPLIT)
dataset = dataset.shuffle(seed=42).select(range(NUM_CALIBRATION_SAMPLES))

def preprocess(example):
    """Formats the messages into a simple dialogue format."""
    text = "\n".join([msg["content"] for msg in example["messages"]])
    return {"text": text}

dataset = dataset.map(preprocess)

# Apply quantization
oneshot(
    model=model,
    dataset=dataset,
    recipe=recipe,
    output_dir="bloom-3b-gptq-w8a8",
)
```

With this fix, the script now runs **to completion** without errors. Automated tests have also been added to test new changes


src/llmcompressor/recipe/recipe.py
```diff
@@ -61,20 +62,29 @@ def from_modifiers(
...
+                group=group_name,
+                args=modifier.model_dump(exclude_unset=True),
+            )
+            for modifier in modifiers
+        ]
+        # assume one stage for modifier instances
+        stages: List[RecipeStage] = [
+            RecipeStage(group=group_name, modifiers=recipe_modifiers)
+        ]
+        recipe = Recipe()
```
brian-dellabetta, line 85: do you want to use `cls` here?

brian-dellabetta (commented):
looks good! couple suggestions based on the parts i understand

src/llmcompressor/recipe/modifier.py
```diff
@@ -83,13 +83,14 @@ def create_modifier(self) -> "Modifier":
...
-        modifier_type, args = list(values.items())[0]
-
-        modifier["type"] = modifier_type
-        modifier["args"] = args
-        return modifier
+        if len(values) == 2:
+            # values contains only group and the Modifier type as keys
+            group = values.pop("group")
+            modifier_type, args = values.popitem()
+            return {"group": group, "type": modifier_type, "args": args}
```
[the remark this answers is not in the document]
rahul-tuli: While I don't understand pydantic very well, after running multiple tests with different recipe configurations(with single/multiple groups); I've found this function gets invoked once per modifier and the values received are of the format

{"group" : modifier_group, "modifier_type": modifier_args}

So TLDR; we will always have exactly two values (not greater) for cases when recipe is specified as a list of modifier instances and a group is always expected to be present. I added a few assertions for this. Let me know if they look okay to you!

brian-dellabetta (approved):
cool, thanks!

src/llmcompressor/recipe/stage.py
```diff
@@ -161,26 +161,26 @@ def extract_dict_modifiers(values: Dict[str, Any]) -> List[Dict[str, Any]]:
...
-                for mod_key, mod_value in value.items():
-                    modifier = {mod_key: mod_value}
-                    modifier["group"] = group
-                    modifiers.append(modifier)
-
-        for key in remove_keys:
-            del values[key]
+
+        if "modifiers" in values:
+            modifier_values = values.pop("modifiers")
```
kylesayrs, line 166: Is there a case where values["modifiers"] is None?

kylesayrs (approved):
Nice

src/llmcompressor/recipe/stage.py
dsikka: @rahul-tuli can you address?

rahul-tuli: No it won't be!

PR #1272, opened 2025-03-20
Addback: ConstantPruningModifier for finetuning cases

rahul-tuli:
We mistakenly also removed `ConstantPruningModifier` from finetuning examples as a part of #1267

This PR adds it back for fine-tuning examples.


kylesayrs (approved):
My mistake, thanks for catching

dsikka (approved):

PR #1273, opened 2025-03-20
move failing mulitmodal lmeval tests to skipped folder

brian-dellabetta:
SUMMARY:
multi-modal lm-eval tests are failing due to a non-reproducibility issue that still needs to be resolved. In the meantime, moving those tests to a skipped folder until resolution.

Resolution can be tracked in #1260 

TEST PLAN:
no new source code


brian-dellabetta:
> we expect failures from all 3? I thought it was just the 2.

I have changes on #1260 that will move away from the 2B parameter model we are using with FP8, so all 3 configs will change. so i thought it'd be best to just disable them all for now, WDYT?

kylesayrs (approved):

dsikka (approved):

PR #1244, opened 2025-03-12
[WIP] Oneshot Callbacks

kylesayrs:
## Purpose ##
* Standardize oneshot callbacks and prepare for oneshot pipeline extraction

## Changes ##
* Implement the `sequential_epoch_end` callback event which should be called after one sequential layer has been calibrated with one epoch
  * Add callback event to sequential and layer_sequential pipelines
* Trigger module compression both on the `SEQUENTIAL_EPOCH_END` event as well as on finalize
* Prepare smoothquant for pipeline extraction
  * Replace uses of `_hf_hook. pre_forward` with `align_module_device` for clarity
  * Specify `resolved_mappings_` type hint for clarity
  * Trigger `_apply_smoothing` on the `SEQUENTIAL_EPOCH_END` as well as on finalize
  * Add a [guard](https://github.com/vllm-project/llm-compressor/pull/1244/files#diff-90bb5efcbf5f23ba1db62664a91f6b2d6492a909c387cd82c1589f45d5e8615cR285) which allows the `_apply_smoothing` function to be called multiple times per session (as is required by sequential pipeline)

## Testing ##
* Cannot test since cannot call events before initialization

PR #1279, opened 2025-03-24
Pipeline Extraction

kylesayrs:
## Purpose ##
* Extract data pipelines from modifiers to enable multiple modifiers to be active at the same time
  * This enables faster compression of larger models
  * This enables more memory efficient compression of larger models (not limited to just GPTQ/SGPT)

## Prerequisites ##
* #1351
* #1298

## Callback Changes ##
* Implement `calibration_epoch_start`
  * This callback should be called at the start of every calibration pipeline
  * This callback causes modifiers to attach hooks
* Implement `sequential_epoch_end`
  * This callback should be called after one sequential layer has been calibrated with one epoch
  * This callback triggers compression and replaces passing a `callback_modifier`
* Implement `calibration_epoch_end`
  * This callback triggers at the end of a calibration epoch, and is used to *trigger compression* in between pipelines composed using the independent pipeline and *remove hooks* in between independent pipelines

## Lifecycle Changes ##
* Oneshot modifiers implement on_end, which removes hooks when calibration finishes
  * In the future, calibration_epoch_start is treated like batch_start, where it is an opportunity for modifiers to start
  * In the future, calibration_epoch_end is treated like batch_end, where it is an opportunity for modifiers to end
  * In the future, finalize is treated like batch_end, where it is an opportunity for modifiers to end
  * Right now, these opportunities are implemented manually on each oneshot modifier, rather than being a lifecycle rule

## Data Pipeline Changes ##
* Implement data pipeline registry
  * Inferred pipeline is selected using modifiers and can be overridden by user
* Implement independent pipeline
  * This pipeline treats each modifier as a separate stage and assigns a pipeline to each modifier
  * Meant to replicate current LC behavior
  * Originally, these compression events were triggered by reaching the end of each module’s initialize function. Now a separate event is required
* Implement `session.get_modifiers`
  * In order to perform data pipeline inference and other sequential pipeline inference, these functions must get the list of active modifiers before they initialize
  * This function gets all the active modifiers across all `ModifierStages`
* Prepare smoothquant for pipeline extraction
  * Trigger `_apply_smoothing` on the `sequential_epoch_end ` and `calibration_epoch_end`
  * Add a [guard](https://github.com/vllm-project/llm-compressor/pull/1244/files#diff-90bb5efcbf5f23ba1db62664a91f6b2d6492a909c387cd82c1589f45d5e8615cR285) which allows the `_apply_smoothing` function to be called multiple times per session (as is required by sequential pipeline)

## Testing ##
* Quantized llama3-8b using both the independent (basic + sequential) and sequential pipelines
  * There was no accuracy regression from using a shared pipeline, although we keep the `independent` pipeline as the default for now
* Transformers tests pass
* https://github.com/neuralmagic/llm-compressor-testing/actions/runs/14622080074


src/llmcompressor/pipelines/registry.py
```diff
@@ -0,0 +1,49 @@
...
+__all__ = ["PIPELINES", "get_pipeline_fn"]
+
+SEQUENTIAL_MODIFIERS = (GPTQModifier, SparsityModifierMixin)
+
+PIPELINES: Dict[str, PipelineFn] = {
+    "sequential": sequential.run_pipeline,
+    "layer_sequential": layer_sequential.run_pipeline,
+    "basic": basic.run_pipeline,
+    "independent": independent.run_pipeline,
+}
```
brian-dellabetta, line 20: If we make pipelines a class satisfying an abstract `Pipeline` base class with 
```
@abstractmethod
def run_pipeline(model:PreTrainedModel, dataloader:DataLoader):...
```
would it avoid needing maps like this or types like 
```
PipelineFn = Callable[[PreTrainedModel, torch.utils.data.DataLoader], None]
```
in the typing.py file

brian-dellabetta (commented):
Definitely looks cleaner this way! Leaving comments rather than approving, as I am still getting up to speed with pipelines

src/llmcompressor/pipelines/registry.py
kylesayrs: I purposefully avoided adding a base class, since I think it adds more infrastructure than is required. Such a class would only have one abstract method, which imho doesn't justify a class definition.

brian-dellabetta (commented):
I know you're looking for feedback on this, but I'm not sure I understand it enough to approve. I do like the removal of all the try/catch code in GPTQ. Maybe we can have a deep dive session on this next week?

src/llmcompressor/pipelines/registry.py
```diff
@@ -0,0 +1,90 @@
...
+            "Please modify your recipe to use at most one quantization modifier"
+        )
+
+    if len(quant_modifiers) == 1:
+        quant_modifier = quant_modifiers[0]
+        config = quant_modifier.resolve_quantization_config()
+        if config.requires_calibration_data():
+            return "basic"
+        else:
+            return "data_free"
```
rahul-tuli, line 77: maybe typo?

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/14779544427

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/14839620945

src/llmcompressor/typing.py
```diff
@@ -1,17 +1,27 @@
...
     PreTrainedTokenizer, BaseImageProcessor, FeatureExtractionMixin, ProcessorMixin
 ]
 
 # Supported dataset types, IterableDataset is a streamed dataset
 DatasetType = Union[Dataset, DatasetDict, IterableDataset]
+
+# Pipeline callable
+PipelineFn = Callable[
+    [PreTrainedModel, torch.utils.data.DataLoader, "DatasetArguments"], None
+]
```
brian-dellabetta, line 27: nit: feels awkward to me, I know we discussed an abstract base class as being too much for this, but maybe a `Protocol` would be a nice middle-ground

```python
class Pipeline(Protocol):
    def run_pipeline(self, model: PreTrainedModel, data_loader: torch.utils.data.DataLoader, dataset_args: "DatasetArguments") -> None:
        ...
```

brian-dellabetta (approved):
really cool! excited to try this out. Should we run the e2e/lmeval tests before merging this in? Lots of moving pieces, they might catch something

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -56,72 +52,44 @@ def on_initialize(self, state: State, **kwargs) -> bool:
...
+        if event.type_ == EventType.CALIBRATION_EPOCH_END:
+            # TODO: modify lifecycle to end on calibration epoch end
+            if not self.ended_:
+                self.on_end(state, None)
 
     def on_end(self, state: State, event: Event, **kwargs):
         """
         Finish calibrating by removing observers and calibration hooks
         """
+        self.ended_ = True  # TODO: move to super call
```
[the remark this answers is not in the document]
kylesayrs: Doing this has implications for the non-oneshot modifiers, so I leave it as a follow up. The change has technically be validated and tested in #1256, but the changes are different than the goal of this PR

src/llmcompressor/typing.py
kylesayrs: I used the ABC

kylesayrs:
Reran job: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/14839620945

brian-dellabetta (dismissed):

kylesayrs:
I've validated the only failing awq test passes locally

src/llmcompressor/pipelines/basic/__init__.py
brian-dellabetta, line 1: nit: looks like several of the folders in `src/llmcompressor/pipelines` contain a single file and a single `__init__.py`. Could just have `src/llmcompressor/pipelines/basic.py`, `src/llmcompressor/pipelines/data_free.py`, etc instead

brian-dellabetta (approved):
👍

src/llmcompressor/pipelines/basic/__init__.py
kylesayrs: This is an aesthetic choice, motivated by wanting to keep similar folder structure with those pipelines which do require helpers

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -403,6 +405,46 @@ def match_modules(model: Module, target_names: List[str]) -> Set[Module]:
     )
 
 
+def get_targets_from_modifiers(
```
rahul-tuli, line 408: nit: could we update the name to reflect targets are only fetched from sequential modifiers?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -403,6 +405,46 @@ def match_modules(model: Module, target_names: List[str]) -> Set[Module]:
...
+    :return: list of sequential targets and list of modules to ignore for tracing
+    """
+    # avoid circular import
+    from llmcompressor.pipelines.registry import SEQUENTIAL_MODIFIERS
+
+    sequential_modifiers = [
+        modifier for modifier in modifiers if isinstance(modifier, SEQUENTIAL_MODIFIERS)
+    ]
+
+    if len(sequential_modifiers) >= 2:
```
rahul-tuli, line 425: After this we won't be supporting the cases where we have two GPTQ modifiers (acting on different parts of the model); now we would only allow this to be conveyed via different config_groups, is my understanding correct?

rahul-tuli (changes requested):
Looks good overall, left some minor comments.
One change/resolution/explanation requested for independent pipelines.

Generally I see a lot of similar TODO's trickled across multiple file(s); would like to address/delete/link out to ticket or issues before merge.

Great job on this!

src/llmcompressor/modifiers/modifier.py
```diff
@@ -103,8 +103,8 @@ def finalize(self, state: State, **kwargs):
         :param state: The current state of the model
         :param kwargs: Additional arguments for finalizing the modifier
         """
-        if self.finalized_ or not self.initialized_:
-            return
+        if self.finalized_:
+            raise RuntimeError("cannot finalize a modifier twice")
```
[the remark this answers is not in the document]
kylesayrs: There is no reason why a modifier should finalize twice

src/llmcompressor/pipelines/independent/pipeline.py
```diff
@@ -0,0 +1,51 @@
...
+                mod_type = str(type(modifier).__name__)
+                session.lifecycle.modifiers = [
+                    StageModifiers(modifiers=[modifier], group=mod_type, index=index)
+                ]
+
+                pipeline = CalibrationPipeline.from_modifiers([modifier])
+                pipeline_name = pipeline.__class__.__name__
+                _logger.info(f"Inferred `{pipeline_name}` for `{mod_type}`")
+
+                pipeline(model, dataloader, dataset_args)
```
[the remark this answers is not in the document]
kylesayrs: No, because the `from_modifiers` only returns the `independent` pipeline if either
1. it is passed by the user (which it is not in this case)
2. If the recipe has AWQ and two or more modifiers. The independent pipeline only uses one modifier at a time, this is not relevant in the recursive case

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: This function operates on all modifiers. For example, if no targets are given, then the targets are inferred. In the future, this will be a config external to the modifiers entirely

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: That's a good point. Yes, as currently implemented, that recipe is disallowed because those two modifiers could have conflicting `sequential_targets`.

Supporting those kinds of recipes is a larger job, since it requires either extracting the `sequential_targets` field away from modifiers (as it should be) or moving the code that resolves `sequential_targets` out of modifiers and into this function

src/llmcompressor/pipelines/registry.py
```diff
@@ -0,0 +1,102 @@
...
+    def __call__(
+        model: torch.nn.Module,
+        dataloader: DataLoader,
+        dataset_args: "DatasetArguments",
+    ):
+        raise NotImplementedError()
+
+    @classmethod
+    def from_modifiers(
+        cls, modifiers: List[Modifier], user: Optional[str] = None
```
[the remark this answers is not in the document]
kylesayrs: > :param user: pipeline name passed by user

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/14870766869

src/llmcompressor/pipelines/sequential/helpers.py
rahul-tuli: Sounds good, thank you!

src/llmcompressor/modifiers/obcq/base.py
```diff
@@ -41,9 +41,6 @@ class SparseGPTModifier(SparsityModifierMixin, Modifier):
     Lifecycle:
         - on_initialize
             - register_hook(module, calibrate_module, "forward")
-            - run_sequential / run_layer_sequential / run_basic
```
rahul-tuli, line 44: Now that SparseGPTMixin inherits from Modifier, do we need multi-inheritance? commenting here as Github won't let me comment on L26

rahul-tuli (approved):

brian-dellabetta (approved):

brian-dellabetta:
> https://github.com/neuralmagic/llm-compressor-testing/actions/runs/14870766869

Looks like tests passed, but there was a issue reporting timings, possibly expected for this manual run and unrelated to these changes. So I think we're good to go on this!

tests/llmcompressor/transformers/kv_cache/test_kv_cache.py
```diff
@@ -232,12 +226,24 @@ def test_kv_cache_gptq_model_state_dict_attr(kv_cache_fixture, tmp_path):
...
+                    symmetric: {symmetric}
     """
 
     output_dir, _ = next(kv_cache_fixture(recipe, tmp_path))
 
     with init_empty_weights():
-        model = AutoModelForCausalLM.from_pretrained(output_dir)
+        # TODO: There is a bug in `apply_quantization_config` which means that, if using
+        # CompressedLinears, the compression status is inferred to `compressed` and
+        # therefore the attention kvcache parameters never undergo initializations
```
dsikka, line 242: In other words, we can load quantized kv cache for decompression post compressing, correct? 
Not really a bug - more so this feature was never built out on purpose. We can discus on slack

kylesayrs: If this was intended behavior, then we should throw an explicit error rather than silently failing to load parameters

src/llmcompressor/modifiers/obcq/base.py
kylesayrs: https://github.com/vllm-project/llm-compressor/pull/1416

PR #1283, opened 2025-03-25
[GPTQ] Add inversion fallback

kylesayrs:
## Purpose ##
* Given the increasing size of large language models (such as DeepSeek-R1 which contains 45034 linear layers), the likelihood that any of the hessian inversions will spontaneously fail is significant
* These changes cause the GPTQ algorithm to fall back to RTN for any layers which fail hessian inversion

## Changes ##
* Implement fallback by setting hessian value to identity matrix if inversion fails


src/llmcompressor/modifiers/quantization/gptq/gptq_quantize.py
```diff
@@ -161,11 +162,12 @@ def quantize_weight(
         H = torch.linalg.cholesky(H, upper=True)
         Hinv = H
     except torch._C._LinAlgError:
-        raise torch._C._LinAlgError(
+        logger.warning(
```
brian-dellabetta, line 165: wondering if we should add a check for how many times this happens? a user may see it pass with warnings when none of the layers are successfully modified?

brian-dellabetta (dismissed):
cool!

src/llmcompressor/modifiers/quantization/gptq/gptq_quantize.py
kylesayrs: I'm hoping a user will notice if their console has multiple warnings, I don't think that case justifies the added logic complexity

rahul-tuli (dismissed):

rahul-tuli (approved):

dsikka (approved):

PR #1288, opened 2025-03-26
Bump: Min python version to 3.9

rahul-tuli:
This PR updates the minimum Python version requirement from 3.8 to 3.9 in `setup.py`. Python 3.8 reached its end-of-life (EOL) on October 14, 2024, and is no longer receiving security updates or bug fixes. Aligning with the broader ecosystem, the [vLLM project](https://github.com/vllm-project/vllm/blob/27df5199d99627e1eb101071c2155f888181bd64/pyproject.toml#L22) also requires Python 3.9+,

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ setup(
             "llmcompressor.trace=llmcompressor.transformers.tracing.debug:main",
         ]
     },
-    python_requires=">=3.8",
+    python_requires=">=3.9",
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Programming Language :: Python :: 3",


markurtz (approved):

brian-dellabetta (approved):
you had this in another PR right? are we getting this in first before the other lands?

rahul-tuli:
> you had this in another PR right? are we getting this in first before the other lands?

You are correct; the Python version upgrade has been broken out from that PR into its own separate diff here (the description on that diff was also updated accordingly). Both the diffs can land in any order they don't depend on each other!

markurtz (approved):

PR #1114, opened 2025-01-29
SQ and QM: Remove `torch.cuda.empty_cache`, use `calibration_forward_context`

kylesayrs:
## Purpose ##
* Fixes #1081
* Fixes #963
* There's really no explanation online as to why the `torch.cuda.empty_cache()` kernel sometimes fails to launch. Given that `empty_cache` does not actually free memory that wouldn't have already been freed by the python garbage collector + [pytorch caching allocator](https://zdevito.github.io/2022/08/04/cuda-caching-allocator.html), it should be safe to remove this call.

## Changes ##
* Remove `torch.cuda.empty_cache()` in `run_calibration_forward`, which only affects smoothquant and quantization modifier (sparsegpt and wanda will soon use sequential pipelines instead)
* Use `calibration_forward_context` in smoothquant and quantization modifier
* Remove use of `torch.cuda.empty_cache()` by smoothquant modiifier

## Testing ##
* Performed memory analysis with and without `torch.cuda.empty_cache` and `calibration_forward_context` independently

### Smooth Quant ###
20c0e104-2353-4a09-9556-f953075205d2

### Quantization Modifier ###
0a0451e2-108e-40fb-be5c-e9619928ab67

It was also found that removing the `empty_cache` calls in between each operation reduced the runtime of Quantization Modifier on llama3-8B by 78%

Before
```
512/512 [03:18<00:00,  2.58it/s]
Duration: 199.38174653053284
```

After
```
512/512 [00:42<00:00, 11.91it/s]
Duration: 44.374401807785034
```

PR #1299, opened 2025-03-29
Keep quantization enabled during calibration

kylesayrs:
## Purpose ##
* Revert the behavior regression introduced as a result of #1114
* When calibrating a model using the `QuantizationModifier`, quantization should be enabled when calibrating

## Changes ##
* Remove "disabling quantization" from the calibration forward pass
* Add "disabling quantization" to the sequential pipelines in order to continue to disable quantization during calibration for GPTQ and SGPT
  * When [calibration pipelines become shared between modifiers](#1279), the decision of whether to disabling quantization during calibration will have to be moved to the calibration pipelines themselves. Some work needs to be done to demonstrate that GPTQ and SGPT do not suffer accuracy regression from enabling activation quantization during calibration  (in theory, the change should increase accuracy)


dsikka (commented):
Thanks, we should get this in for the next release.

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -51,7 +51,7 @@ def run_pipeline(
     # find layers
     layers = match_modules(model, sequential_targets)
 
-    with calibration_forward_context(model):
+    with calibration_forward_context(model), DisableQuantization(model):
```
brian-dellabetta, line 54: do we need to worry about the multi-context python 3.9 issue here? i think those changes are being reverted?

brian-dellabetta (approved):

kylesayrs:
@markurtz Please see my comment in the description
> the decision of whether to disabling quantization during calibration will have to be moved to the calibration pipelines themselves. Some work needs to be done to demonstrate that GPTQ and SGPT do not suffer accuracy regression from enabling activation quantization during calibration (in theory, the change should increase accuracy)

As currently implemented, activation quantization is enabled while calibrating with `QuantizationModifier`, but is not enabled when calibrating with `GPTQModifier`, and SGPT modifiers. In the future, we hope to demonstrate that enabling activation quantization while calibrating GPTQ and SGPT leads to accuracy improvements, not regressions

markurtz:
> As currently implemented, activation quantization is enabled while calibrating with `QuantizationModifier`, but is not enabled when calibrating with `GPTQModifier`, and SGPT modifiers. In the future, we hope to demonstrate that enabling activation quantization while calibrating GPTQ and SGPT leads to accuracy improvements, not regressions

Yep, that's all fine. My main concern is around the pipelines needing to know about disabling quantization within a context and that not being handled at the modifier level.

src/llmcompressor/pipelines/layer_sequential/pipeline.py
markurtz: No, see the comments I left in the PR's and issues around that. That was incorrect information that has been propagated around. Multi context statements have been enabled since Python 3.1. The piece that was added recently was enabling parentheses to group them.

rahul-tuli: This is acceptable, since it's on a single line, it will work with both 3.9 and above; only thing we need to take care of going forward is to not use features introduced from 3.10+ onwards

dsikka (approved):

PR #1304, opened 2025-04-01
Update: Readme for fp8 support

rahul-tuli:
This PR updates the README to clarify that FP8 is only supported on GPUs with CUDA Compute Capability ≥ 9.0, such as NVIDIA's Hopper and Blackwell architectures.
GPUs based on Ada Lovelace (Compute Capability 8.9) do not support FP8.


dsikka (changes requested):
Could we move all the scheme descriptions to their own file under `docs`?

brian-dellabetta (approved):

brian-dellabetta (approved):

PR #1308, opened 2025-04-01
[Tracing] Allow sizes to be iterated

kylesayrs:
## Purpose ##
* Reduce user burden when tracing models by automatically handling the iteration of torch.Size types

The following code
```python3
input_shape = hidden_states.shape[:-1]
hidden_shape = (*input_shape, -1, self.head_dim) # iterate shape
```

gets graph compiled into
```python3
input_shape = getattr(hidden_states, "shape")
getitem = __getitem__(input_shape, slice(None, -1, None))
hidden_shape = (__getitem__(getitem, 0), __getitem__(getitem, 1), -1, head_dim)
```

## Changes ##
* Catch any attempts to iterate torch.Size proxies. Because lengths are assumed to be static during tracing, we can replace the iteration with an iteration of `__getitem__` calls of each of the indices

## Testing ##
* Can now run sequential pipeline with llama3 with `Linear` granularity without custom code definition

```python3
recipe = GPTQModifier(targets="Linear", scheme="W4A16", ignore=["lm_head"], sequential_targets="Linear")
```

### DIFF
--- a/src/llmcompressor/pipelines/sequential/helpers.py
+++ b/src/llmcompressor/pipelines/sequential/helpers.py
@@ -1,12 +1,13 @@
 import inspect
 from collections import deque
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, List, Set, Union
+from typing import Any, Callable, Dict, Iterator, List, Set, Union
 
+import torch
 from compressed_tensors import has_offloaded_params
 from compressed_tensors.quantization import find_name_or_class_matches
 from torch.fx import Graph, GraphModule, Node
-from torch.fx.proxy import Argument
+from torch.fx.proxy import Argument, Proxy
 from torch.nn import Module
 from transformers import PreTrainedModel
 from transformers.configuration_utils import PretrainedConfig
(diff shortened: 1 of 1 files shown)


src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -149,6 +142,26 @@ def trace(self, root: Union[Module, Callable], *args, **kwargs) -> Graph:
...
+            if isinstance(a, PretrainedConfig):
+                kwargs = {k: self.create_arg(v) for k, v in a.to_dict().items()}
+                return self.create_node("call_function", a.__class__, (), kwargs)
+
+            else:
+                return super().create_arg(a)
+
+        def iter(self, obj: Proxy) -> Iterator:
+            # special extension which allows torch.Sizes to be iterated, but keeps
+            # their values are dynamic
```
brian-dellabetta, line 156: maybe this?
```suggestion
            # their values as dynamic
```

brian-dellabetta (approved):
neat!

PR #1322, opened 2025-04-03
replace custom version handling with setuptools-scm

dhellmann:
SUMMARY:

Use setuptools-scm to automatically compute a version based on the
latest git tag.

Remove the old logic for reading a static file to get the version
string, but configure setuptools-scm to generate a similar file for
anyone who wants to import it to get the version information.

Replace the build_type value set in code with an environment variable
so setup.py can continue to change the versioning scheme based on the
type of build being performed.

When the BUILD_TYPE is "nightly", produce an alpha release for the
_next_ version, with the date as the alpha version. For example, when
the most recent tag is 0.4.1 the nightly release version built on 3
Apr 2025 is 0.4.2a20250403.

When the BUILD_TYPE is "release", if the commit being built has a tag
use that tag's value as the version.

When the BUILD_TYPE is "dev", or if the commit being built does not
have a tag, or the git repository has pending changes, compute a
development version combining the distance since the last tag with the
expected next tag. For example, if the most recent tag is 0.4.1 and
there have been 74 commits since that tag and the most recent commit
SHA is 5e2c1c92, then the version would be 0.4.2.dev74+g5e2c1c92. If
the local git repository has modified files that are not committed,
the date is appended in the local version field, giving
0.4.2.dev74+g5e2c1c92.d20250403.

TEST PLAN:

I tested this change locally using these commands

```
$ BUILD_TYPE=nightly python3 -m build .
$ python3 -m build .
```

And using a variety of clean, dirty, and tagged commits.


setup.py
```diff
@@ -3,28 +3,67 @@
 
 from setuptools import find_packages, setup
 
-sys.path.insert(0, os.path.abspath(os.path.dirname(__file__)))
-from utils.version_extractor import extract_version_info  # noqa isort:skip
+# Set the build type using an environment variable to give us
+# different package names based on the reason for the build.
+VALID_BUILD_TYPES = {"release", "nightly"}
```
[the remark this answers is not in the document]
dhuangnm: I guess Andy meant to add "dev" into the VALID_BUILD_TYPES list explicitly instead of having it as default when not provided?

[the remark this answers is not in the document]
dhellmann: Sure, I could do that. In dev mode, should it always add a `dev` value to the version being built?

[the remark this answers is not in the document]
dhuangnm: yeah I think so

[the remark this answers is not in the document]
dhellmann: OK, that's done. The default is now "dev", which always appends the "dev" release with a distance from the last commit. The release and nightly modes work the same way as before.

setup.py
```diff
@@ -62,7 +109,7 @@
         "pillow",
         (
             "compressed-tensors==0.9.3"
-            if version_info.build_type == "release"
+            if BUILD_TYPE == "release"
             else "compressed-tensors-nightly"
```
dhuangnm, line 113: Since we'll apply the same changes to compressed-tensors as well, should we update here to install only compressed-tensors?

dhellmann: That's a good point. Maybe we should make the change in the other repository before making it here?

dhuangnm: Yeah I'm working on it, hopefully to have it ready shortly.

dhellmann: OK, I've updated this code to not change the package name but to still change the requirement so that it is not using a pinned version of the compressed-tensors package. It seems like that will let you have the control you want for releases, then be more flexible in other situations. Let me know if you had something different in mind.

dhuangnm: Thanks, the updated code looks good to me:

        "pillow",
        (
            "compressed-tensors==0.9.3"
            if version_info.build_type == "release"
            else "compressed-tensors-nightly"
            if BUILD_TYPE == "release"
            else "compressed-tensors"
        ),

@dsikka can you please review if this looks correct to you?

dbarbuzzi (dismissed):

brian-dellabetta (approved):

dhellmann:
How do I run that linter locally so I can check the formatting before pushing?

dhuangnm:
@dhellmann I just realized we need to update the [Makefile](https://github.com/vllm-project/llm-compressor/blob/main/Makefile) as well according to the build changes. Can you please update it too? Thanks.

dhellmann:
> @dhellmann I just realized we need to update the [Makefile](https://github.com/vllm-project/llm-compressor/blob/main/Makefile) as well according to the build changes. Can you please update it too? Thanks.

What updates are needed there?

dhellmann:
> > @dhellmann I just realized we need to update the [Makefile](https://github.com/vllm-project/llm-compressor/blob/main/Makefile) as well according to the build changes. Can you please update it too? Thanks.
> 
> What updates are needed there?

Nevermind, I figured it out as soon as I got set up to run `make style`. :-)

dhellmann:
I think this is ready for testing again, if someone could approve it, please.

Makefile
```diff
@@ -1,5 +1,5 @@
 BUILDDIR := $(PWD)
-CHECKDIRS := src tests utils examples setup.py
+CHECKDIRS := src tests examples setup.py
 DOCDIR := docs
 
 BUILD_ARGS :=  # set nightly to build nightly release
```
dhuangnm, line 5: I think we also need to update the "build" section:

python3 setup.py sdist bdist_wheel $(BUILD_ARGS)

where the build command now takes BUILD_TYPE env var?

dhellmann: OK, I've updated the Makefile to set the variable to a default and allow it to be overridden when calling make. For example,

```
BUILD_TYPE=nightly make build
```

You could add some separate make targets if you want to do that, too.

setup.py
```diff
@@ -62,8 +118,8 @@
         "pillow",
         (
             "compressed-tensors==0.9.3"
-            if version_info.build_type == "release"
-            else "compressed-tensors-nightly"
+            if BUILD_TYPE == "release"
+            else "compressed-tensors"
```
dhuangnm, line 122: Actually "compressed-tensors" for else statement will install the release version of compressed-tensors. Maybe we should use "compressed-tensors>0.9.3" here?

dhellmann: Saying `>0.9.3` here won't be enough to ensure the nightly package is installed. The job will also need to set the `--pre` option for pip (or an equivalent for uv) to allow alpha versions to be installed (the default is to not install pre-release versions that are alpha, beta, or rc). When that is done, the installer should automatically pick the latest nightly, and the version specifier would be redundant.

Makefile
dhuangnm: Cool, thanks!

dhuangnm (approved):
LGTM, thanks!

dhellmann:
I have fixed the linter issue and rebased, so need approval on the workflow to run the tests again.

dhuangnm (approved):

dsikka (approved):

Issue #1293, opened 2025-03-28
Very slow saving of models due to irrelevant stats computation


eldarkurtic:
Saving large models, especially MoEs with many experts (e.g. DeepSeek-R1) takes quite a lot of time due to some unnecessary stats computations which should probably be optional.

I am already disabling the ones that are exposed through save_pretrained: `model.save_pretrained(save_dir, skip_compression_stats=True, skip_sparse_compression=True)` and these issues are not related to them.  After tracing a bit, I've detected these two lines to be the problematic ones:
1. in LLM-Compressor: https://github.com/vllm-project/llm-compressor/blob/8f6a0b5fc9397a0ed3baa012850c96e007ba7dea/src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py#L212
2. in compressed-tensors: https://github.com/neuralmagic/compressed-tensors/blob/574299879e90df3d6381dab9d24f1abf92d2cae6/src/compressed_tensors/quantization/quant_config.py#L235

For a small 8x7B Mixtral MoE, these two lines add ~10 mins of additional wait time during saving of the checkpoint. And this grows to an order of hours with DeepSeek-R1 which has many more layers. Disabling these is not exposed externally so for my local runs, I'm just disabling them with `sparsity_structure = None` and `compression_ratio = None`, but we should find a better solution to handle them properly.

For the first one, I think we should try to make it a bit more efficient by looking only at a couple of first layers to detect sparsity pattern (I think @markurtz had a suggestion for this in Slack).
For the second one, I feel like it could be removed completely as it already has a note in the code that it's performing an incorrect calculation: https://github.com/neuralmagic/compressed-tensors/blob/574299879e90df3d6381dab9d24f1abf92d2cae6/src/compressed_tensors/quantization/quant_config.py#L233C9-L234C66

dsikka:
Opened on Jira: https://issues.redhat.com/browse/INFERENG-542

dsikka:
This has been addressed https://github.com/vllm-project/llm-compressor/commit/517a3efa7dc16514cd56d0d5b3ec980f22bad4ca

PR #1332, opened 2025-04-08
[Compression] Update sparsity calculation lifecycle when fetching the compressor

dsikka:
# Summary
- Updating the lifecycle based on feedback from research and meta
- #1293 is addressed @eldarkurtic 

### 1. No longer always infer sparsity 
- If there is existing sparsity in the model that should be accounted for while compressing the model, the user is required to indicate this by providing a sparsity config. This is not a common case and primarily something we've seen from research. As a result, it does not need to be the default pathway
- If no sparsity config is provided, only check for sparsity if a recipe with sparsity was applied OR the user explicitly would like to apply a sparsity compressor by setting `skip_compression_stats` parameter
- Based on these changes, we no longer use the global_sparsity to dictate whether or not a sparse compressor is applied

### 5. Fix log outputs to prevent cutting off

# Follow-ups:
- Need to update passing saving arguments properly - they are currently not being passed apart from save_compressed and there is no clear API on what the arguments are/how they should be passed, both to save_pretrained and when providing an output dir to oneshot

# Testing
- Tested a sparse-only model
- Tested a model with both sparsity and quantization
- Tested a model with existing sparsity
- Tested a model with only quantization

For all cases, the model compresses to the expected format.


src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -58,7 +58,7 @@ def save_pretrained_wrapper(
             quantization_format: Optional[str] = None,
             save_compressed: bool = True,
             safe_serialization: bool = True,
-            skip_compression_stats: bool = False,
+            skip_sparsity_compression_stats: bool = False,
```
rahul-tuli, line 61:
```suggestion
            skip_sparsity_compression_stats: bool = True,
```

Since it's True in get_model_compressor as well

rahul-tuli (commented):
Nice, the functionality seems to be all there! LGTM pending a small nit and some clean up (docstrings and such)

brian-dellabetta (dismissed):

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
```diff
@@ -537,7 +542,7 @@ def __init__(self, weights, weight_scale=None, weight_zero_point=None):
 
         # Linear layer without bias
         self.linear = nn.Linear(in_features, out_features, bias=False)
-        self.linear.weight = nn.Parameter(weights, requires_grad=False)
+        self.linear.weight = nn.Parameter(weights, requires_grad=True)
```
brian-dellabetta, line 545: we need gradients here?

brian-dellabetta (dismissed):

brian-dellabetta (approved):

tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
dsikka: When generating a sparsity config, we calculate the global_sparsity, which looks at trainable parameters. The test was skipping this step before so it did not matter previously.

src/llmcompressor/transformers/finetune/session_mixin.py
```diff
@@ -358,7 +358,13 @@ def train(self, *args, stage: Optional[str] = None, **kwargs):
 
         return output
 
-    def save_model(self, output_dir: str, _internal_call: bool = False):
+    # TODO: support all save args, not just skip_sparsity_compression_stats
+    def save_model(
+        self,
+        output_dir: str,
+        _internal_call: bool = False,
+        skip_sparsity_compression_stats: Optional[bool] = False,
```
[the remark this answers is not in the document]
dsikka: We don’t support passing in saving args to the train pathway right now. I will enable this in a follow up. So we need this param to be set to True while training. However for one shot, we support this. Hence different defaults. Will be consistent when I had the saving params

brian-dellabetta (approved):

rahul-tuli (approved):

kylesayrs (approved):

PR #1333, opened 2025-04-08
[Model] Llama4 Support

kylesayrs:
## Purpose ##
* Support quantization of Llama4 with GPTQ


kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/1438

PR #1334, opened 2025-04-08
[Utils] Add `skip_weights_download` for developers and testing

kylesayrs:
## Purpose ##
* Follow up to #1188
* Add utilities which can be used by developers as well as used during testing of model architectures

## Prerequisites ##
* #1187

## Changes ##
* Add `skip_weights_download` which allows a model to be initialized and dispatched without downloading the weights
* Add `patch_transformers_logger_level` which is used by `skip_weights_download` to reduce warning verbosity


src/llmcompressor/utils/dev.py
```diff
@@ -0,0 +1,107 @@
...
+    with tempfile.TemporaryDirectory() as tmp_dir, patch_attr(
+        model_class, "from_pretrained", patched
+    ), skip_weights_initialize(), patch_transformers_logger_level():
+        yield
+
+
+@contextlib.contextmanager
+def skip_weights_initialize(use_zeros: bool = False):
+    """
+    Very similar to `transforms.model_utils.no_init_weights`, except that torch.Tensor
```
brian-dellabetta, line 74:
```suggestion
    Very similar to `transformers.model_utils.no_init_weights`, except that torch.Tensor
```

src/llmcompressor/utils/dev.py
brian-dellabetta, line 1: as in utils only a developer would need? Not sure what would be a better name

brian-dellabetta (dismissed):
cool!

src/llmcompressor/utils/dev.py
kylesayrs: It's not my favorite name either, but essentially yes, these are utils for rapid development and testing

brian-dellabetta (approved):

dsikka (approved):
This looks good but I'm wondering if we should add some doc somewhere else so users are aware that these utils exist. Maybe an example using them under the examples folder?

mgoin:
FYI in vLLM we do this through the `load_format="dummy"` argument, if you think that might be a better name

https://github.com/vllm-project/vllm/blob/98d01d3ce2a4d06e85348e375e726c40bee0bdf0/vllm/config.py#L1347-L1359

PR #1351, opened 2025-04-15
Implement `QuantizationMixin`

kylesayrs:
## Purpose ## 
* Abstract functionality which allows modifiers to act as quantization configs into a mixin called `QuantizationMixin`
  * This gives #1279 an interface to properly infer which pipeline to use based on the recipe (if a recipe contains modifiers requires calibration, then use the "basic" or "sequential" pipelines)
  * This enables future modifiers to act as quantization modifiers (in the same way that GPTQ does now)
* Related to #1354 where previous logic would attempt to add a QuantizedKVCache for dynamic kv_quant

## Changes ##
* Implement `QuantizationMixin` which implements five public methods
  * Lifecycle methods
    * `initialize_quantization` is used to apply a config and attach observers to a model
      * quantization is disabled so that modules aren't quantized before they're calibrated
    * `start_calibration` is used to initialize calibration hooks and status
      * quantization is enabled, since we currently quantize as we calibrate, although this decision is somewhat arbitrary
    * `end_calibration` is used to remove calibration hooks and apply the frozen status
      * quantization remains enabled, since we want future forward passes to simulate quantization
  * Recipe-related methods
    * `has_config` returns true if a config was specified, used for checking against duplicate configs in the recipe
    * `resolve_quantization_config` returns the quantization config specified by the modifier fields
* `QuantizationModifier` inherits from `QuantizationMixin`
* `GPTQModifier` inherits from `QuantizationMixin`
  * Unlike QMod, GPTQ disables quantization during calibration. As noted before, this is a somewhat arbitrary choice but one which matches the current implementation

* Calibration utils
  * Replace `set_unset_kv_cache` with `initialize_quantized_kv_cache` and `freeze_module_quantization`
    * Treat the `QuantizedKVCache` as analogous to another observer
  * Pull setting the calibration status out of`update_weight_zp_scale`
    * This better matches the lifecycle detailed in `QuantizationMixin` description
  * Implement `reset_quantization_status` which is used to remove any existing quantization configs before the current config is applied by `initialize_quantization`

## Remove Support ##
* Removing support for recipe with multiple quantization modifiers active at the same time (a check for this will be added by #1279)
* Remove `num_calibration_steps`, `quantize`, `disable_quantization_observer_epoch` and `min_tokens_per_module`
  * `num_calibration_steps` is already controlled by https://github.com/vllm-project/llm-compressor/blob/42b62f5283d0234b26623fe1f1bf02a77c6e4019/src/llmcompressor/datasets/utils.py#L106
  * `quantize` was implemented as a workaround for GPTQ's modifier builder. Similar functionality may be require to support SpinQuant + GPTQ, but such functionality should exist at a higher level
  * `disable_quantization_observer_epoch` seems to implement functionality where a model's observers are removed but quantization remains active. This functionality is maintained by setting an "end" epoch for qmod
  * `min_tokens_per_module` requires that the modifier have references to the calibration dataset, which is disallowed by #1279. This information is already printed in GPTQ's logs. If research still wants this tool specifically for `QuantizationModifier`, then it can be reimplemented to avoid using references to the calibration dataset
  
## Testing ##
* Updated tests to reflect new mixin
* Ran a set of GPTQ and QuantizationModifier examples to completion
* CI tests pass


brian-dellabetta (dismissed):
This LGTM! slowly understanding more

coolKeen:
> ## Purpose
> * Abstract functionality which allows modifiers to act as quantization configs into a mixin called `QuantizationMixin`
>   
>   * This gives [Pipeline Extraction #1279](https://github.com/vllm-project/llm-compressor/pull/1279) an interface to properly infer which pipeline to use based on the recipe (if a recipe contains modifiers requires calibration, then use the "basic" or "sequential" pipelines)
>   * This enables future modifiers to act as quantization modifiers (in the same way that GPTQ does now)
> * Related to [problem cause when doing w8a8kvfp8 for qwen2.5-vl #1354](https://github.com/vllm-project/llm-compressor/issues/1354) where previous logic would attempt to add a QuantizedKVCache for dynamic kv_quant
> 
> ## Changes
> * Implement `QuantizationMixin` which implements four public methods
>   
>   * `attach_scheme_and_observers` is used to apply the quantization config to the model and attach observers
>   * `register_calibration_hooks` add calibration hooks which calibrate the observers which calibrate the scales
>   * `resolve_quantization_config` returns the quantization config specified by the modifier fields
>   * `has_config` returns true if a config was specified, used for checking against duplicate configs
> * `QuantizationModifier` inherits from `QuantizationMixin`
>   
>   * The scheme and observers are attached on initialization
>   * The activations and weights are calibrated on start via hooks and `update_weight_zp_scale`
>     
>     * Add a tqdm for weight calibration, which is useful for larger models
>   * The observers and hooks are removed on finalize
> * `GPTQModifier` inherits from `QuantizationMixin`
>   
>   * Implements a similar lifecycle to  `QuantizationModifier`
>   * Remove attached `QuantizationModifier` logic
> * Replace `set_unset_kv_cache` with `initialize_quantized_kv_cache` and `freeze_module_quantization`
>   
>   * Treat the `QuantizedKVCache` as analogous to another observer
> * Pull setting the calibration status out of`update_weight_zp_scale`
>   
>   * This better matches the lifecycle detailed in `QuantizationMixin` description
> 
> ## Remove Support
> * Removing support for recipe with multiple quantization modifiers active at the same time (a check for this will be added by [Pipeline Extraction #1279](https://github.com/vllm-project/llm-compressor/pull/1279))
> * Remove `num_calibration_steps`, `quantize`, `disable_quantization_observer_epoch` and `min_tokens_per_module`
>   
>   * `num_calibration_steps` is already controlled by https://github.com/vllm-project/llm-compressor/blob/42b62f5283d0234b26623fe1f1bf02a77c6e4019/src/llmcompressor/datasets/utils.py#L106
>   * `quantize` was implemented as a workaround for GPTQ's modifier builder. Similar functionality may be require to support SpinQuant + GPTQ, but such functionality should exist at a higher level
>   * `disable_quantization_observer_epoch` seems to implement functionality where a model's observers are removed but quantization remains active. This is currently implemented by setting an "end" epoch
>   * `min_tokens_per_module` requires that the modifier have references to the calibration dataset, which is disallowed by [Pipeline Extraction #1279](https://github.com/vllm-project/llm-compressor/pull/1279). This information is already printed in GPTQ's logs. If research still wants this tool specifically for `QuantizationModifier`, then it can be reimplemented to avoid using references to the calibration dataset
> 
> ## Testing
> * Updated tests to reflect new mixin
> * Ran a set of GPTQ and QuantizationModifier examples to completion
> * CI tests pass

Hi, I see that you mention SpinQuant + GPTQ?  Does llm compressor support such recipe now?

kylesayrs:
Hi @coolKeen! Transforms support is actively being worked on, you can see the WIP PRs [here](https://github.com/neuralmagic/compressed-tensors/pulls)!

brian-dellabetta (dismissed):

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -49,10 +51,6 @@ def initialize_observer(
         # no quantization scheme nothing to do
         return
 
-    # observers have a different lifecycle for kv_cache
-    if is_attention_module(module):
```
dsikka, line 53: why is it ok to move this?

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -102,25 +100,15 @@ def update_weight_zp_scale(module: Module):
...
-        # not set to initialize; no scales/zp to update
-        return
-    if status != QuantizationStatus.INITIALIZED:
+    if getattr(module, "quantization_status", None) != QuantizationStatus.CALIBRATION:
         logger.warning(
-            f"Attempting set module with status {status} to calibration mode. "
-            f"but status is not {QuantizationStatus.INITIALIZED} - you may "
-            "be calibrating an uninitialized module which may fail or attempting "
-            "to re-calibrate a frozen module"
+            "Attempting to calibrate weights of a module not in calibration mode"
```
dsikka, line 108: Why are changing this warning?

dsikka (dismissed):

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -140,81 +131,39 @@ def validate_sequential_update(cls, value: bool) -> bool:
...
-    def _check_build_quant_modifier(self, model: torch.nn.Module):
-        """
-        Check the model's quantization state matches that expected by this modifier,
-        adding a default quantization scheme if needed
-
-        # TODO: build modifier during recipe validation
-
-        :param state: session state storing input model and calibration data
-        """
-        quantization_already_active = qat_active(model)
```
rahul-tuli, line 152: I see that this line was removed? Is that intentional?

Does your diff work for cases when we have both GPTQModifier and a QuantizationModifier present in a recipe? Could we add a test for that case as well

rahul-tuli (commented):
LGTM pending comment; Thank you, nice work!

src/llmcompressor/modifiers/quantization/calibration.py
kylesayrs: This is moved [here](https://github.com/vllm-project/llm-compressor/pull/1351/files#diff-33303ae48e185b2fbb14dc45c2052805837deb5723248367b9579321c4c4e974R209)

src/llmcompressor/modifiers/quantization/calibration.py
kylesayrs: `update_weight_zp_scale` no longer changes the calibration status. Setting the calibration status is done outside of this function now for lifecycle clarity.

https://github.com/vllm-project/llm-compressor/pull/1351/files#diff-33303ae48e185b2fbb14dc45c2052805837deb5723248367b9579321c4c4e974R36-R45

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: This removal is intentional, as this behavior allows a user to provide two configs, and have the second one be ignored.

This behavior is replaced by [this](https://github.com/vllm-project/llm-compressor/pull/1351/files#diff-1c9652a451ec6e28d99279b8731d5e0b3c4b172e22a06aba331ad70b9674341bR141), where we only prepare the model for quantization if a config is provided

.github/workflows/test-check-transformers.yaml
```diff
@@ -95,7 +95,3 @@ jobs:
         if: (success() || failure()) && steps.install.outcome == 'success'
         run: |
           pytest -v tests/llmcompressor/transformers/obcq
-      - name: Running KV Cache Tests
```
brian-dellabetta, line 98: this is from from your other PR, might want to drop from this one since Dipika requested changes

brian-dellabetta (approved):
Just make sure your other PR goes in first, or drop the changes to kv cache tests from this one

brian-dellabetta (approved):
good stuff. always nice when adding an abstraction leads to more lines removed than added 🔥

tests/llmcompressor/modifiers/calibration/test_kv_cache.py
```diff
@@ -75,7 +75,7 @@ def test_kv_cache_quantization(config):
     config = QuantizationConfig(**config)
     config.quantization_status = QuantizationStatus.CALIBRATION
     apply_quantization_config(model, config)
-    model.apply(set_unset_kv_cache)
+    model.apply(initialize_quantized_kv_cache)
```
rahul-tuli, line 78: nice!

rahul-tuli (approved):
Synced offline on scope of testing; Changes introduced by this diff are tested by #1279 by virtue of lm_eval tests! Changes look good.

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/14779544427

brian-dellabetta (approved):

rahul-tuli (approved):

PR #1365, opened 2025-04-21
[Testing] Reduce error verbosity of cleanup

kylesayrs:
## Purpose ##
* When tests fail, they often show the error that occurs during cleanup, rather than the error that caused the tests to fail
```
ERROR tests/llmcompressor/transformers/obcq/test_consecutive_runs.py::TestConsecutiveRunsSmall_0_commit::test_consecutive_runs_small - FileNotFoundError: [Errno 2] No such file or directory: './oneshot_output'
```

## Changes ## 
* Check if a directory exists before attempting to delete it during cleanup


rahul-tuli (approved):
Thoughts on using `pathlib` for these changes?

PR #1370, opened 2025-04-22
Fix versioning for source installs

dbarbuzzi:
SUMMARY:
Update the various source installs in testing workflows to properly fetch git tags and set the `BUILD_TYPE` to nightly. This notably results in installing a nightly-style version of `compressed-tensors` for the source install to allow it to satisfy the version requirements from `llm-compressor`.

TEST PLAN:
Prior to these changes, you would find a broken `llm-compressor` version in the local install, as well as a warning about the source install version of `compressed-tensors` being incompatible:

```
Successfully installed […] llmcompressor-0.1.dev1+g67614d5 […]
...
Installing collected packages: compressed-tensors
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
llmcompressor 0.1.dev1+g8ca8fd3 requires compressed-tensors>=0.9.4a2, but you have compressed-tensors 0.1.dev1+g16e6435 which is incompatible.
Successfully installed compressed-tensors-0.1.dev1+g16e6435
```

With the changes, both packages have proper versioning and no compatibility complaints:

```
Successfully installed […] llmcompressor-0.5.1.dev28+g3b1aac94 […]
...
Installing collected packages: compressed-tensors
Successfully installed compressed-tensors-0.9.4a20250421
```

### DIFF
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -65,8 +65,10 @@ jobs:
       - name: "⚙️ Install compressed-tensors dependencies"
         id: install
         run: |
-          pip3 uninstall -y compressed-tensors compressed-tensors-nightly
-          pip3 install ./compressed-tensors/
+          pip3 uninstall -y compressed-tensors
+          export GIT_CEILING_DIRECTORIES="$(pwd)"
+          cd compressed-tensors
+          BUILD_TYPE=nightly pip3 install .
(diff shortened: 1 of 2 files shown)


markurtz (approved):
This looks fine to unblock tests for now, but I definitely want to revisit these setups over the next few days to standardize them -- we should not need to fetch tags just to create a working / acceptable version format for the builds

PR #1371, opened 2025-04-22
[Tests] Disable silently failing kv cache test

kylesayrs:
## Background ##
The current KV cache tests are silently failing because Qmod(kv) + GPTQ(weights) is not a supported recipe. This is because GPTQ was being entirely skipped (because it was being preceded by a quantization modifier with no weight schemes).

If you attempt to fix this by disallowing multi-qconfig recipes, you run into the issue that model compression with KV+weights is not supported.

## Multi-round quantization ##

Previously, the model did not have weight quantization. This means that there was no compressor. This means that the model would be saved in the frozen status, not the compressed status. This would mean that the model would be loaded, and the inferred status would be None, which means that at load time, the status would move from none to frozen, hence passing through initialization.

After fixing the recipe to run GPTQ with kv+weight quantization, you run into another issue.

## KV + Weight Compression ##

Now, the model has weight quantization, meaning there is a compressor. This means that the model will be saved in the compressed status. This means that the model would be loaded with CompressedLinear (which are in the frozen status), causing the whole model to be inferred as compressed. Because the model is already supposedly in the compressed status, then initialization does not happen and the kv_cache attention parameters are not initialized, and hence the model fails to load kv_cache qparams from the state dict.

## Ideal Solution ##

Ideally, we should be replacing with CompressedLinear as a part of apply_quantization_status, not before applying quantization status. Doing it the other way unintentionally skips all the lifecycle steps

As a side note, ideally, we should always be saving final models with the compressed status, even if the compressor is dense. [structure is initialized, checkpoints are calibration or frozen, and final models are compressed (even if nothing was applied, so long as save_compressed)], but this is out of scope of this fix.

### DIFF
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -95,3 +95,7 @@ jobs:
         if: (success() || failure()) && steps.install.outcome == 'success'
         run: |
           pytest -v tests/llmcompressor/transformers/obcq
+      - name: Running KV Cache Tests
+        if: (success() || failure()) && steps.install.outcome == 'success'
+        run: |
+          pytest -v tests/llmcompressor/transformers/kv_cache -k "not test_kv_cache_gptq_model_state_dict_attr"
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -95,7 +95,3 @@ jobs:
         if: (success() || failure()) && steps.install.outcome == 'success'
         run: |
           pytest -v tests/llmcompressor/transformers/obcq
-      - name: Running KV Cache Tests
-        if: (success() || failure()) && steps.install.outcome == 'success'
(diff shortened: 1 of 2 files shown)


brian-dellabetta (approved):

dsikka (dismissed):
Why not just skip that particular test as opposed to all the tests?
Or just update to just use one of the quant or gptq modifiers, not both.

kylesayrs:
@dsikka The larger problem is not multiround quantization, it is that loading models with kv+weight quantization is not supported in transformers.

2/5 of the tests rely on this, but clearly more work needs to be done to verify that the other 3/5 are still running as intended

kylesayrs:
@dsikka I've disabled only the offending test, the other tests seem to be valid

rahul-tuli (approved):

brian-dellabetta (approved):

PR #1374, opened 2025-04-23
[WIP] Add AWQ Asym e2e test case

dsikka:
SUMMARY:
- Add a test case for asym awq 
- Add processing steps for `mit-han-lab/pile-val-backup`
- Update incorrect logger.warning call

Testing
- Generated model which was pushed to the hub successfully and ran using vLLM
- Model: https://huggingface.co/nm-testing/TinyLlama-1.1B-Chat-v1.0-w4a16-asym-awq-e2e/tree/main

### DIFF
--- a/tests/testing_utils.py
+++ b/tests/testing_utils.py
@@ -228,6 +228,18 @@ def process_dataset(
                 "images": sample["image"],
             }
 
+    elif ds_name == "pile-val-backup":
+
+        def preprocess(example):
+            return {
+                "input_ids": processor.encode(example["text"].strip())[:max_seq_length]
+            }
+
+        ds = ds.map(preprocess, remove_columns=ds.column_names)
+        # Note: potentially swap filtering to pad for AWQ
+        ds = ds.filter(lambda example: len(example["input_ids"]) >= max_seq_length)
+        return ds
+
     else:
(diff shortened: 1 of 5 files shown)


brian-dellabetta (approved):
🔥

tests/testing_utils.py
```diff
@@ -228,6 +228,18 @@ def process(sample):
...
+    elif ds_name == "pile-val-backup":
+
+        def preprocess(example):
+            return {
+                "input_ids": processor.encode(example["text"].strip())[:max_seq_length]
+            }
+
+        ds = ds.map(preprocess, remove_columns=ds.column_names)
+        # Note: potentially swap filtering to pad for AWQ
+        ds = ds.filter(lambda example: len(example["input_ids"]) >= max_seq_length)
```
kylesayrs, line 240: Does this dataset need explicit preprocessing? Also why not truncate rather than filter?

kylesayrs (approved):

tests/testing_utils.py
dsikka: Just following practices that Brian has listed out - I think in general we should clean this up to be more consistent with the other datasets but ok for now.

PR #1375, opened 2025-04-23
[Tests] Fix test case; update structure

dsikka:
# Summary
- Fix device_map and set torch.dtype for the given model
- Move tests to a folder which makes more sense


brian-dellabetta (approved):
do we need to prune `tests/llmcompressor/transformers/finetune/finetune_generic`?

rahul-tuli (approved):

dsikka:
> do we need to prune `tests/llmcompressor/transformers/finetune/finetune_generic`?

used by actual finetune tests!

kylesayrs (approved):
Nice

PR #1386, opened 2025-04-26
AWQ: Clean up forward passes with kwargs using inspect.bind

### LINKED ISSUE #1385: AWQ -- Cleanup forward passes with kwargs using `inspect.bind`
The current implementation of AWQ currently has a [lot of helper code](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L577-L629) for passing kwargs into the `.forward` call of an arbitrary module. This can be cleaned up by using the `inspect.bind` method directly. Here's a nice github gist example -- https://gist.github.com/encukou/5060767

ved1beta:
Changes Made
- Simplified the `_sanitize_kwargs` function in `src/llmcompressor/modifiers/awq/base.py`
- Replaced manual parameter filtering with `inspect.Signature.bind_partial`

 used `bind_partial()` instead of `bind()` 
- In the AWQ context only handling keyword arguments, not all positional arguments
- The tensor input is handled separately and passed as a positional argument in `tensor_forward_with_input_args`
- `bind_partial()` allows for partial binding of arguments similar to `functools.partial` which is exactly what we need here since we're only providing some of the parameters : ) 

Fix #1385

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -599,33 +599,30 @@ class AWQModifier(Modifier):
 
 def _sanitize_kwargs(inputs_kwargs, module):
     """
-    Remove the arguments that are not supported in the module's
-    forward pass to avoid breaking behaviour between different versions
-    of transformers.
-
+    Sanitize input keyword arguments to match the module's forward method signature
+    using inspect.bind_partial.
+    
     Args:
         inputs_kwargs (`dict`):
             The input dictionary to pass to the model layer
         module (`torch.nn.Module`):
             Target module to quantize.
     """
-    params = inspect.signature(module.forward).parameters
-    sanitized_kwargs = {}
-    for k, v in inputs_kwargs.items():
-        if k in params and k != "use_cache":
-            sanitized_kwargs[k] = v
-    # In case forward pass has optional dependencies that don't default to None.
-    # This is the case for `LlamaAttention.forward` which has input
-    #  `attention_mask: Optional[torch.Tensor],` (with no `= None` default)
-    # https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L246
-    for k, v in params.items():
-        if (
-            k not in sanitized_kwargs
-            and k != "use_cache"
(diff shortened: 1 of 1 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -599,33 +599,30 @@ def _forward_input_with_kwargs(
...
-            sanitized_kwargs[k] = None
-
+    signature = inspect.signature(module.forward)
+    
+    try:
+        # Use bind_partial to allow partial binding of arguments
+        bound_args = signature.bind_partial(**inputs_kwargs)
+        # Convert bound arguments to a dictionary
+        sanitized_kwargs = bound_args.arguments
+    except TypeError:
```
kylesayrs, line 618: Is this is an acceptable error? If this error occurs, is it indicative of a bug in the code?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -599,33 +599,30 @@ def _forward_input_with_kwargs(
...
+        # If binding fails, fall back to empty dict
+        sanitized_kwargs = {}
+        
+    # Handle special case for use_cache if needed
+    if 'use_cache' in sanitized_kwargs:
+        del sanitized_kwargs['use_cache']
```
kylesayrs, line 624:
```suggestion
    sanitized_kwargs.pop("use_cache", None)
```

kylesayrs (commented):
Awesome contribution!

brian-dellabetta:
Hi @ved1beta , thanks for the contribution! I tried this branch out with `examples/awq/awq_one_shot.py`, but unfortunately we still need the edge case check, and `bind_partial` will fail out for cases when input_kwargs has fields that `module.forward` does not, which commonly occurs:
```
...
File "/usr/lib/python3.10/inspect.py", line 3175, in _bind
    raise TypeError(
TypeError: got an unexpected keyword argument 'cache_position'
```

So we unfortunately won't be able to migrate to `bind_partial` in the way we had hoped. 

I merged some of this into my #1405 PR, which has reverted more to some minor cleanup and typehints.

PR #1390, opened 2025-04-27
[Tracing] Autowrap methods by name

kylesayrs:
## Purpose ##
* Reduce model support burden by allowing users to skip untraceable methods using string matching
* This would commonly be used for skipping `_update_causal_mask`, which is the default

## TODO ##
* Update documentation to reflect new usage

## Prerequisites ##
* #1279
* #1389

## Changes ##
* Update `ignore` documentation to reflect how ignore is now used to ignore methods, not modules
* Implement `add_autowrap_methods` which finds module methods to autowrap
* Add `tracing_ignore`, which defaults to `_update_causal_mask`

## Testing ##
* Add tests for `add_autowrap_methods`

## Follow ups ##
* Catch tracing failures and suggest that users ignore methods in the stack trace
* In the future, support autowrapping functions or code


kylesayrs:
#1411

PR #1392, opened 2025-04-27
[Tracing] Trace with eager attention

kylesayrs:
## Purpose ##
* Enable higher-granularity tracing for llama-style models
* Tracing at the linear level requires tracing into the attention mechanism, which can be complex for non-eager implementations

## Changes ##
* Use eager attention implementation when tracing

## Testing ##
* After #1308 lands, you can now perform sequential calibration of llama3 models with `Linear` granularity


brian-dellabetta (approved):

kylesayrs:
#1411

PR #1401, opened 2025-04-29
[Testing] add observers test

kylesayrs:
## Purpose ##
* Test observers
* Reveal potential issue with how qparams are initialized
https://github.com/neuralmagic/compressed-tensors/blob/1068c848e420443d4d5d73fe031b78bf7a832926/src/compressed_tensors/quantization/lifecycle/initialize.py#L165

## Changes ##
* Added `test_observers_update` in `tests/llmcompressor/modifiers/calibration/test_observers.py`

### DIFF
--- a/tests/llmcompressor/modifiers/calibration/test_observers.py
+++ b/tests/llmcompressor/modifiers/calibration/test_observers.py
@@ -1,0 +1,56 @@
+import pytest
+import torch
+from compressed_tensors.quantization import (
+    QuantizationArgs,
+    QuantizationScheme,
+    initialize_module_for_quantization,
+)
+
+from llmcompressor.modifiers.quantization.calibration import initialize_observer
+
+
+@pytest.mark.parametrize(
+    "shape,group_size",
+    [
+        ((64, 64), None),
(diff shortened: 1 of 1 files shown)


brian-dellabetta (dismissed):

markurtz (dismissed):
I think we should simplify this using the built-in filter pathways for the logger.add function from Loguru. This way, all log levels and standard functions are supported, and the root logger import will work without other special/additional imports.

Something along the lines of the following:

```python
_logged_once = set()

def deduplicate_filter(record):
    message = record["message"]
    log_once = record["extra"].get("log_once", False)

    if log_once and message in _logged_once:
        return True
    
    if log_once:
        _logged_once.add(message)

    return False
    
logger.add(..., filter=deduplicate_filter)

logger.bind(log_once=True).info("This will only log once")
logger.bind(log_once=True).info("This will only log once")
```

brian-dellabetta (dismissed):
🔥

shanjiaz (approved):
Looks good to me!

dsikka (approved):

PR #1335, opened 2025-04-08
[Tracing][Testing] Add tracing tests

kylesayrs:
## Purpose ##
* Add regression testing to model tracing beyond example tests
  * These tests complete in ~1 min and can be run at a quicker cadence than example tests
  * These can also be used to test tracing capabilities beyond those in the examples, for example tracing into linear layers

## Prerequisites ##
* #1334
* #1402

## Changes ##
* Fix function signature of peoples speech dataset
* Add `trust_remote_code` argument to debugger
* Add `tests/llmcompressor/transformers/tracing/models.py`
  * I did not include phi3 because it's a very difficult model to work with programmatically. I will revisit once the major tracing improvements have landed

PR #1402, opened 2025-04-29
[VLM] Fix mllama targets

kylesayrs:
## Purpose ##
* When #1389 landed, modules being skipped by ignore were no longer being skipped. However, this requires that the sequential targets list be correct. Mllama defaults to targeting vision layers, and hence the vision tower was being traced, leading to errors.

```python3
_no_split_modules = [
    "MllamaVisionEncoderLayer",
    "MllamaCrossAttentionDecoderLayer",
    "MllamaSelfAttentionDecoderLayer",
]
```

## Changes ##
* Only target text decoder layers, not vision decoder layers

## Testing ##
* #1335 passes

### DIFF
--- a/examples/multimodal_vision/mllama_example.py
+++ b/examples/multimodal_vision/mllama_example.py
@@ -32,6 +32,7 @@ recipe = [
     GPTQModifier(
         targets="Linear",
         scheme="W4A16",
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):
LGTM!

brian-dellabetta (approved):

PR #1405, opened 2025-04-30
AWQ sanitize_kwargs minor cleanup

brian-dellabetta:
SUMMARY:
Using `inspect.bind_partial` inside the AWQ `_sanitize_kwargs` helper does provide the robust support we were hoping. We still need to explicitly add None for the edge case, and `bind_partial` will fail if the input kwargs have fields that are not in the `module.forward` signature. 

So we have to stick with what we have. This PR just has some minor touch-ups and typehints, and removes the helpers that were only used by AWQ and not needed

Resolves https://github.com/vllm-project/llm-compressor/issues/1385
Replaces https://github.com/vllm-project/llm-compressor/pull/1386 

TEST PLAN:
No net new src code, just minor cleanup and redundant code removal

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -217,7 +217,7 @@ class AWQModifier(Modifier):
         self._resolved_mappings = resolved_mappings
         return
 
-    def _setup_scale_hooks(self):
+    def _setup_scale_hooks(self) -> None:
         """
         Attach a forward hook to each activation we want to smooth. This allows us to
         calculate the dynamic range during calibration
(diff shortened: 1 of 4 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -590,43 +589,44 @@ def _forward_input_with_kwargs(
         """
         kwargs = input_kwargs or self._module_kwargs
         kwargs = _sanitize_kwargs(kwargs, module)
-        return tensor_forward_with_input_args(
-            module=module,
-            inputs=inputs,
-            input_kwargs=kwargs,
-        )[0]
+
+        inputs = inputs.to(next(module.parameters()).device)
```
kylesayrs, line 593: This will fail for offloaded modules (will return the meta device)
Instead, use `get_execution_device`. This kind of device movement can also be handled by the `IntermediatesCache`

kylesayrs (changes requested):

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: ok, updated. This is a vestige from AutoAWQ that we'll have to clean up with the pipelining refactor. We are doing `.to(device=...)` in several places in this file, which won't work for offloaded modules (ticket INFERENG-531)

kylesayrs (approved):

rahul-tuli (approved):

PR #1407, opened 2025-05-01
validation check added

### LINKED ISSUE #1226: Fail when invalid arguments are provided in the recipe for a modifier
**Describe the bug**
- You can provide extra irrelevant arguments when specifying a modifier and no error is thrown 
- This can mask issues for arguments that may no longer be supported or if they are provided incorrectly 

E.g:
```yaml
quant_stage:
    quant_modifiers:
        QuantizationModifier:
            dampen: 0.1
            ignore: ["lm_head"]
            config_groups:
                group_0:
                    weights:
                        num_bits: 8
                        type: "float"
                        symmetric: true
                        strategy: "tensor"
                        observer: "mse"
                    targets: ["Linear"]
```

- The recipe below specifies `dampen` for the QuantizationModifier, however, this is not a valid argument and no error is thrown

**Expected behavior**
- Validate recipe arguments are valid for the modifiers being specified

ved1beta:
SUMMARY:
added a validation check in create_modifier
fixes #1226

### DIFF
--- a/src/llmcompressor/recipe/modifier.py
+++ b/src/llmcompressor/recipe/modifier.py
@@ -73,6 +73,21 @@ class RecipeModifier(RecipeBase):
         """
         if not ModifierFactory._loaded:
             ModifierFactory.refresh()
+
+        modifier_class = ModifierFactory.get_class(self.type)
+    
+        if hasattr(modifier_class, "model_fields"):
+            valid_fields = set(modifier_class.model_fields.keys())
+        
+        provided_args = set(self.args_evaluated.keys() if self.args_evaluated else {})
+        
+        invalid_keys = provided_args - valid_fields
+        
+        if invalid_keys:
+            error_msg = f"Invalid arguments found for {self.type}: {', '.join(invalid_keys)}"
+            error_msg += f"\nValid arguments are: {', '.join(sorted(valid_fields))}"
+            raise ValueError(error_msg)
+
         return ModifierFactory.create(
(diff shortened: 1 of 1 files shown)


kylesayrs (commented):
Hi @ved1beta ! Could you provide more context as to when this code would be useful?

I also see here that `valid_fields` is undeclared in some cases, you may want to make sure that that variable is declared in all cases, or make explicit the assumption that all modifiers have model fields

kylesayrs:
@ved1beta I see! Have you explored Pydantic's [extra="forbid"](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra) option at all?

```python3
class Model(BaseModel):
    x: int

    model_config = ConfigDict(extra="forbid")
```

kylesayrs:
Sorry, I was actually referencing adding the forbid option directly to the `Modifier` class. This might be simpler than attempting to create a temporary base class as an intermediary

kylesayrs:
I was actually suggesting to add the extra forbid config to the [Modifier](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/modifier.py#L12) class, directly. This solution also covers cases where users create modifiers directly, rather than going through the ModifierFactory

```python3
model_config = ConfigDict(extra="forbid")
```

kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/1614

PR #1411, opened 2025-05-04
[Tracing] Code AutoWrapper

kylesayrs:
## Purpose ##
* Reduce model support burden by automatically wrapping untraceable code
  * This is a programmatic implementation of all of the rules specified by the [tracing guide](https://github.com/vllm-project/llm-compressor/blob/0.5.1/src/llmcompressor/transformers/tracing/GUIDE.md)
* Remove traceable definitions for `Idefics3ForConditionalGeneration`, `LlavaForConditionalGeneration`, `MllamaForConditionalGeneration`, `LlavaForConditionalGeneration`, `Qwen2_5_VLForConditionalGeneration`, and `Qwen2VLForConditionalGeneration`, `Gemma3ForConditionalGeneration` (all of them)

## Fixes ##
* #1457

## Autowrap Patterns ##
These patterns match syntax which is untraceable and unlikely to call sequential targets (either directly or indirectly)
<details><summary>If statements whose conditions cannot be statically evaluated</summary>

This if statement can be statically evaluated, since its value can be evaluated in the context of `{"self": LlamaModel(...)}`
```python3
if self.config._attn_implementation != "eager":
    ...
```

If the statement cannot be statically evaluated, then it is wrapped
```python3
torch.fx.wrap
def wrapped(input_ids, inputs_embeds):
    if (input_ids is None) ^ (inputs_embeds is not None):
        raise ValueError("You must specify exactly one of input_ids or inputs_embeds")

def forward(...):
    (,) = wrapped(input_ids, inputs_embeds)
```
</details>

<details><summary>Ignored functions (`_update_causal_mask`)</summary>

Any function or method names listed in the ignore list will be automatically wrapped
```python3
torch.fx.wrap
def wrapped(attention_mask, inputs_embeds, cache_position, ...):
    return self._update_causal_mask(attention_mask, inputs_embeds, cache_position, ...)

def forward(...):
    causal_mask = wrapped(attention_mask, inputs_embeds, cache_position, ...)
```
</details>

<details><summary>Starred tuple unpacking</summary>

Any use of iterated unpacking will be automatically wrapped
```python3
torch.fx.wrap
def wrapped(input_shape):
    return (*input_shape, -1, self.head_dim)

def forward(...):
    hidden_shape = wrapped(input_shape)
```
</details>

<details><summary>Starred argument unpacking</summary>

Any use of iterated unpacking into variadic args is automatically wrapped
```python3
torch.fx.wrap
def wrapped(attn_output, input_shape):
    return attn_output.reshape(*input_shape, -1)

def forward(...):
    attn_output = wrapped(attn_output, input_shape)
```
</details>

## Autowrap Implementation Details ##
<details><summary>Wrapper arguments</summary>
Autowrapping a piece of code requires determining which variable names are used by that code and which variable names are produced by that code. This is done using the `NameAnalyzer`, which determines the unbound, assigned, and conditionally assigned names for a given piece of code.

```python3
# unbound := names which are read by node before being assigned
# assigned := names which are assigned by operations in node
# cond_assigned := names which may be assigned depending on execution
analyzer = NameAnalyzer(omit=self.namespace.keys())
unbound, assigned, conditionally_assigned = analyzer.analyze(node)
```

This information is then used to determine what the args, kwargs, and return names should be for the wrapping function.

```python3
# args := names which already existed and are needed for ops or wrapped return
# kwargs := names which are needed for return but did not already exist
# returns := names which are assigned or could be assigned
args = (unbound | conditionally_assigned) & self._local_names
kwargs = conditionally_assigned - self._local_names
returns = assigned | conditionally_assigned
```

</details>

<details><summary>Wrapping methods</summary>
Some untraceable code references `self` during execution. While normally, `self` would be an argument to the wrapped function, `self` is usually a `torch.nn.Module` which is not a handled type that can be passed around the graph. Instead, we treat `self` as a variable in the compiled python module namespace, and this namespace is automatically captured and executed by `torch.fx._symbolic_trace`

</details>

<details><summary>Unwrappable code</summary>
Some code cannot be wrapped because it contains control flow statements which must exist in a certain context. For example, we cannot wrap code that contains a `continue` without also wrapping the for loop that surrounds it.

```python3
for index, layer in enumerate(self.layers):
    # ---- cannot autowrap ----
    if index <= 10:
        continue
    # ---- cannot autowrap ----
    
    hidden_states = layer(hidden_states)
```

</details>

## Future Extensions/ Improvements ##
<details><summary>Sequentially executing vision towers</summary>

Sequentially tracing vision towers is a lower priority, as the vision towers are typically fewer parameters and aren't quantization targets. However, in the event that they do become quantization targets, or memory(vision_tower + one target) > memory(one gpu), then the vision tower layers will need to be split up.

Some changes may be required to support this. Conditionally executing the vision tower is a very common pattern:
```python3
def forward(pixel_values, image_embeds, ...):
    if image_embeds is None:
        image_embeds = self.vision_tower(pixel_values)
    ...
```

Some approaches might be
1. Allowing names like `image_embeds` to be evaluated based on the sample input being passed
2. Pattern matching against `self.{module_name}()`, where module_name is determined to be a module through evaluation
3. Using type hinting analysis tools like `jedi` to track the types of all names, and to check if any names whose type is a module are called
</details>

<details><summary>Towards perfect autowrapping</summary>

As mentioned in in "Sequentially executing vision towers", it may be possible to use use type hinting analysis tools like `jedi` or `pytype` to infer whether any given code chunk calls a sequential target or target ancestor. If this can be done reliably (which will require extensions such as analysis of called functions), then all code that does not call sequential targets can be wrapped.
</details>

<details><summary>Towards removing tracing</summary>

If we can reliably determine if a given code chunk calls sequential any targets, it may be possible to define each autowrapped function as its own subgraph directly, without the need for tracing. However, this will require inference of model execution from the static code, which means unrolling loops, expanding any function/method calls, and resolving dynamic model structure (in the case of llava and idefics), which may be tricky.

For example, if you want to determine the execution of the LlamaDecoder layers of a llava model like pixtral, you'd need to evaluate `self.language_model`, then analyze the source of the caller's forward function, which is stored in a separate file.
```python3
def forward(...):
    self.language_model(...)  # the type of self.language_model is determined from the config
```

Another point of evaluation would be evaling any iteration of ModuleLists
```python3
def forward(...):
    for decoder_layer of self.layers:  # ModuleList isn't well typed and may contain different types
        decoder_layer(...)
```

The tracing system could be replaced with static code inference, both are different systems for solving the problem of determining model execution
</details>

## Testing ##
* Able to trace all models in `tests/llmcompressor/transformers/tracing/test_models.py` without requiring traceable definitions
* Verified sequentially executed outputs are correct for `LlamaForCausalLM`
* Ran `examples/quantization_w4a16/llama3_example.py` to completion


src/llmcompressor/pipelines/sequential/ast_utils/AutoWrapper.py
brian-dellabetta, line 1: any reason this is `AutoWrapper.py` and not `auto_wrapper.py`? If this was copied from somewhere it'd be good to link to it

src/llmcompressor/pipelines/sequential/ast_utils/ControlFlowAnalyzer.py
brian-dellabetta, line 1: `control_flow_analyzer.py` ?

src/llmcompressor/pipelines/sequential/ast_utils/NameAnalyzer.py
brian-dellabetta, line 1: `name_analyzer.py`?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -85,15 +89,27 @@ def trace_subgraphs(
     :return: a list of Subgraphs in order of execution
     """
     # find modules
-    sequential_targets = match_modules(model, sequential_targets)
-    ignore = match_modules(model, ignore)
+    ignore = ["_update_causal_mask"]
```
brian-dellabetta, line 92: the `ignore` passed into `trace_subgraphs` is no longer used?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -160,26 +178,9 @@ def create_arg(self, a: Any) -> Argument:
...
-
-                # we override the class method because the
-                # class method is the one being traced
-                with patch_attr(type(root), "forward", unwrapped_forward):
-                    return super().trace(root, *args, **kwargs)
-
-            else:
-                return super().trace(root, *args, **kwargs)
+            # TODO: cleanup
+            nonlocal ancestors
```
brian-dellabetta, line 182: is the cleanup to remove use of `nonlocal`?

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -112,16 +128,20 @@ def trace_subgraphs(
     graph.device = model.device
 
     # perform subgraph partition
-    partitions = topological_partition(graph, sequential_targets)
+    partitions = topological_partition(graph, targets)
     subgraphs = partition_graph(model, partitions)
     trace_consumed_names(subgraphs)
 
+    if len(subgraphs) != len(targets) + 1:
```
brian-dellabetta, line 135: why the `+ 1` here?

brian-dellabetta (commented):
Leaving a few comments of what I've reviewed so far, since i need to join a meeting soon. I still need to take a closer lok at the ast_utils module

src/llmcompressor/pipelines/sequential/ast_utils/AutoWrapper.py
kylesayrs: You're right, snake case is a better convention for file names

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: As currently implemented, the `topological_partition` function generates an extra subgraph at the beginning which does not contain a target. This adds a little more runtime, and could be removed in a separate PR

kylesayrs: I've added a note clarifying this in the code

src/llmcompressor/pipelines/sequential/helpers.py
kylesayrs: I've added `DatasetArguments.tracing_ignore` which is configurable, defaults to `"_update_causal_mask"`

brian-dellabetta (approved):
discussed further over a call, impressive work!

kylesayrs:
@dsikka I have no such attachment :)

dsikka (commented):
This LGTM but I was unfamiliar with the ast library so I used this source to try and understand the manipulation provided by the classe: https://github.com/xbeat/Machine-Learning/blob/main/Exploring%20Python's%20Abstract%20Syntax%20Tree%20Manipulation.md

We should potentially provide some sort of docs/resource in the source code as well before landing

brian-dellabetta (approved):
awesome stuff!

kylesayrs:
Seems like there's an unrelated problem specifically with loading llama4 processors  https://huggingface.co/meta-llama/Llama-4-Scout-17B-16E-Instruct/discussions/45 🫠

brian-dellabetta (approved):

PR #1413, opened 2025-05-06
[WIP] Adding a readthedocs docs build for llm-compressor

aireilly:
Minimally build a RTD docs site pulling in md files from examples/


markurtz:
Nice! Thanks @aireilly, still need to go through the sphinx setup and review code/intro docs and will do that soon, but it'd be great to get in this diff or a follow-up diff, some automation to build and push to GitHub pages for the repo, that way we can keep everything local to a GitHub flow and not need to worry about account setup / credentials. Let me know if you'd like some help on that CI/CD side. Another future goal that would be great to get in is compiling the source code docs, I have some prior examples towards that and can build on top a bit later or point you in the direction for those!

aireilly:
@markurtz Hey, yeah happy to help with whatever. Meeting with @dsikka to discuss next week if you want to join. I used readthedocs here to mirror what's on https://github.com/vllm-project/vllm based on work started by @Chr1st1anSears

markurtz:
> readthedocs has some nice features, eg., PR build previews, etc. For example: [aireilly#5](https://github.com/aireilly/llm-compressor/pull/5)

That's definitely fair, but with a bit of work, we could also enable it through GitHub Pages. My core concerns with going through RTD are the following:
- Default ads on pages, detracting and potentially pulling traffic away from where we want
- Reliance on an external platform and a build system from an external platform where we can't as easily control when to build and update
- Limited customizability as well as limited advanced feature support (issues around notebooks and things like that have been common in the past)
- Minimal options for analytics and tracking controls
- Limited control over the domain name / one that doesn't map as easily directly into the project
- Limited control over routing and redirects in case we need to control that for future use for rerouting versions / deleted doc pages

aireilly:
@markurtz I fully agree with what you're saying. Standardizing on GitHub pages seems pragmatic to me. PR previews will need some outside solution however. GitHub does not have out of the box PR previews. I have used https://surge.sh/help/getting-started-with-surge in the past and it works well, no ads, lightweight. We use it here: https://github.com/redhat-documentation/vale-at-red-hat

It's just a deployment tool, and isn't tied to any particular site build.

aireilly:
Not required

PR #1414, opened 2025-05-06
removing RecipeMetadata and references

shanjiaz:
SUMMARY:
Removed RecipeMedata and its references since it's not actively being used. Removed `_set_model_layer_prefix` method from lifecycle. I think this should be safe. 
1. `_set_model_layer_prefix` has no references outside of `core/lifecycle`
2. Checked `compiled_recipe` from main and the `metadata` field is None. This is an example:
```
version=None 
args={} 
stages=[RecipeStage(group='default', args={}, enabled=True, modifiers=[RecipeModifier(type='GPTQModifier', group='default', args={'targets': ['Linear'], 'ignore': ['lm_head'], 'scheme': 'W4A16'}, args_evaluated={'targets': ['Linear'], 'ignore': ['lm_head'], 'scheme': 'W4A16'})], exclude_default=False, args_evaluated={})] 
metadata=None 
args_evaluated={}
```

TEST PLAN:
Tested with `examples/quantization_w4a16/llama3_example.py`. Successfully completed. 

Any concerns with removing it? 
Any suggestions for more testing?


src/llmcompressor/core/lifecycle.py
```diff
@@ -98,7 +98,6 @@ def initialize(
         logger.debug("Initializing compression lifecycle")
         self.recipe_container.append(recipe, recipe_stage, recipe_args)
         self.modifiers = self.recipe_container.get_modifiers()
-        self._set_model_layer_prefix()
```
markurtz, line 101: We should double check this is safe to remove, especially with integrations. Specifically, the plan for this pathway is to enable a prefix that is generally added by wrapping modules for distributed pathways to ensure those are ignored when matching on layers for the recipe. As long as we're accounting for this by stripping out the standard wrapper names / detecting that distributed wrapping in our module/parameter retrieval logic, then this should be fine to strip out

markurtz (commented):
Left one thing to double check on, otherwise it looks good. I think we should open up a future issue, though, to store some more detailed metadata on compilation of the saved recipe/config to enable better future debugging including local framework versions and things like that

src/llmcompressor/core/lifecycle.py
shanjiaz: Will double check!

shanjiaz: `layer_prefix` doesn't appear to be actively used in our codebase at the moment. The [only references](https://github.com/search?q=repo%3Avllm-project%2Fllm-compressor%20layer_prefix&type=code) are in initialization code and aren't functionally applied in modifiers. If we later need to support distributed pathways that rely on consistent prefix handling, we can reintroduce layer_prefix as part of the Recipe class or standardize prefix normalization in the matching logic.

tests/lmeval/configs/vl_fp8_dynamic_per_token.yaml
brian-dellabetta, line 1: If you rebase against main these changes from my other PR should be removed from the diff

brian-dellabetta (commented):
Glad we can get rid of a lot of these abstractions! ❤️

kylesayrs (dismissed):
Great start!

brian-dellabetta (approved):
LGTM, pending confirmation of `self._set_model_layer_prefix` removal

rahul-tuli (approved):
LGTM; great work!

shanjiaz:
tests pass locally, failing on a flaky one:
`FAILED tests/llmcompressor/pytorch/utils/test_helpers.py::test_tensor_sparsity[tensor17-dim17-expected_sparsity17]`

Will retry!

kylesayrs (approved):

rahul-tuli (approved):

PR #1416, opened 2025-05-07
Rename SparsityModifierMixin to SparsityModifierBase

kylesayrs:
## Purpose ##
* Follow up to https://github.com/vllm-project/llm-compressor/pull/1279/files/5e8496bdf117b2e9cc2ab7bda437f2a8d2d0c601#r2077598645

## Changes ##
* Remove extra Modifier inheritance on `SparseGPTModifier` and `WandaPruningModifier`
* Rename `SparsityModifierMixin` -> `SparsityModifierBase` since it now acts more like an ABC rather than a mixin


brian-dellabetta (approved):
This doesn't get referenced in any integration, right? Seems like axolotl etc would use the modifiers that subclass `SparsityModifierBase`

kylesayrs:
@brian-dellabetta I'm not really sure what concern you're referring to. This would only be an issue if any downstream users used `SparsityModifierBase`, which I just checked, axlotl doesn't

brian-dellabetta:
> @brian-dellabetta I'm not really sure what concern you're referring to. This would only be an issue if any downstream users referenced `SparsityModifierMixin`, which I just checked, axlotl doesn't

Sounds good, just wanted to confirm. LGTM

shanjiaz (approved):
Looks good to me!

PR #1424, opened 2025-05-12
[GPTQ] Add `actorder` option to modifier

kylesayrs:
## Purpose ##
* Make actorder option more intuitive for users
* Enable easier adjustment of actorder default #1425
* This change is conceptually intuitive because activation ordering is a concept that only applies to the GPTQ algorithm (the only algorithm for which quantization group order matters)

## Changes ##
* Add `actorder` argument to `GPTQModifier`
* Override `resolve_quantization_config` method to resolve config groups with `actorder` argument
* (Misc) rearrange method order to match the typical order in which they are called in the modifier lifecycle

## Testing ##
* Ran llama w4a16 example to completion


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -120,6 +128,29 @@ def validate_sequential_update(cls, value: bool) -> bool:
...
 
+    def resolve_quantization_config(self) -> QuantizationConfig:
+        config = super().resolve_quantization_config()
+
+        # Resolve config with `self.actorder`
+        for scheme in config.config_groups.values():
+            assert isinstance(scheme, QuantizationScheme)  # (1)
+            if scheme.weights is not None:
+                existing = scheme.weights.actorder
+                assert isinstance(existing, (ActivationOrdering, type(None)))  # (2)
```
brian-dellabetta, line 139: why the `type(None)` here?

brian-dellabetta (approved):
LGTM, pending clarification on one question

src/llmcompressor/modifiers/quantization/gptq/base.py
markurtz: I don't think we should generally check on None types in place of an `is None` check, since the latter is much more explicit. Also, MyPy has standards around this where the expectation is to check is None and not on the type for these primary reasons: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type

kylesayrs: @markurtz This style exists purely for ease of reading and to avoid overly-verbose line wrapping. mypy is still able to track the type through this style of check

```python3
from typing import Any

def f(my_var: Any) -> None:
    if isinstance(my_var, type(None)):
        reveal_type(my_var)  # still infers to None
```
```bash
mypy --strict file.py
file.py:5: note: Revealed type is "None"
Success: no issues found in 1 source file
```

kylesayrs: @brian-dellabetta Ah yeah, I think it's a nice syntax, rather that having to chain a bunch of calls
https://docs.python.org/3/library/functions.html#isinstance

> If classinfo is a tuple of type objects (or recursively, other such tuples) or a Union Type of multiple types, return True if object is an instance of any of the types

markurtz: @kylesayrs, definitely agree that code will run correctly, and typing libraries will read it correctly. The main question from my side is whether that is a good standard pathway to follow, and that's why I pointed out to MyPy's docs that they recommend the is None check rather than a type check for None, so we check on value for None's since they can only be one, and then types when we need to ensure something is compatible with an assumed interface. 

It's a NIT from my side, so not something to block on, but it does look weird IMO to do type(None) rather than chain an is None onto the if statement as it's not something I've run across in other code bases

rahul-tuli (approved):

PR #1425, opened 2025-05-12
[GPTQ] Change actorder default to "static"

kylesayrs:
## Purpose ##
* Use best defaults for GPTQ quantization

## Prerequisites ##
* https://github.com/vllm-project/llm-compressor/pull/1424

## Changes ##
* Set gptq actorder default to "static"

## Testing ##
* Ran llama w4a16 example to completion

### DIFF
--- a/src/llmcompressor/modifiers/quantization/gptq/base.py
+++ b/src/llmcompressor/modifiers/quantization/gptq/base.py
@@ -75,8 +75,9 @@ class GPTQModifier(Modifier, QuantizationMixin):
     :param block_size: Used to determine number of columns to compress in one pass
     :param dampening_frac: Amount of dampening to apply to H, as a fraction of the
         diagonal norm
-    :param actorder: order in which weight columns are quantized. For more information,
-        on actorder options, see https://github.com/vllm-project/vllm/pull/8135
+    :param actorder: order in which weight columns are quantized. Defaults to "static"
+        activation ordering, which achieves best accuracy recovery with no runtime cost.
+        For more information, see https://github.com/vllm-project/vllm/pull/8135
     :param offload_hessians: Set to True for decreased memory usage but increased
         runtime.
 
@@ -109,7 +110,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
     sequential_targets: Union[str, List[str], None] = None
     block_size: int = 128
     dampening_frac: Optional[float] = 0.01
-    actorder: Optional[ActivationOrdering] = None
+    actorder: Optional[ActivationOrdering] = ActivationOrdering.STATIC
     offload_hessians: bool = False
 
     # private variables


brian-dellabetta (dismissed):

dsikka:
nice!

markurtz:
@kylesayrs quick question from my side. Since the old default was None, do we risk defaulting to an incorrect value for older recipes that don't include it? Especially for ones that specified it in the quantization scheme?

brian-dellabetta (dismissed):

rahul-tuli (dismissed):

brian-dellabetta (approved):

brian-dellabetta (approved):

rahul-tuli (dismissed):

brian-dellabetta (dismissed):

dsikka (changes requested):
FYI this will incorrectly save models for our e2e testing:
https://github.com/vllm-project/llm-compressor/tree/main/tests/e2e/vLLM/configs (specifically, essentially remove the cases where there is no act order).

Same for lm-eval tests:
https://github.com/vllm-project/llm-compressor/tree/main/tests/lmeval/configs

kylesayrs:
@dsikka

> specifically, essentially remove the cases where there is no act order while duplicating some cases

Turning all all non-specified actorder cases into actorder cases is the intention of this PR.
As for duplicating tests, I don't think any are actually duplicated? There's only one non-specified gptq test [kv_cache_gptq_tinyllama.yaml](https://github.com/vllm-project/llm-compressor/blob/main/tests/e2e/vLLM/configs/kv_cache_gptq_tinyllama.yaml), and that uses kv cache

kylesayrs:
@dsikka

> specifically, essentially remove the cases where there is no act order while duplicating some cases

Turning all non-specified actorder cases into actorder cases is the intention of this PR.

As for duplicating tests, I don't think any are actually duplicated? In order for a duplication to occur, it must conflict with an existing weight actorder config

```
w4a16_actorder_weight.yaml
w4a16_actorder_weight_qwen.yaml

vl_w4a16_actorder_weight.yaml
w4a16_actorder_weight.yaml
```

These are all w4a16. If you grep for non-specified w4a16 configs, you get
```
w4a16_channel_quant.yaml  # channelwise, not duplicating
w4a16_channel_quant_qwen.yaml  # channelwise, not duplicating
w4a16_grouped_quant.yaml  # different calibration and dataset, not duplicating
w4a16_grouped_quant_asym_awq.yaml  # AWQ, not duplicating
w4a16_2of4_channel_quant.yaml  # 2of4, not duplicating
w4a16_2of4_grouped_quant.yaml  # 2of4, not duplicating
```

All of the other w4a16 configs test for different things, except for maybe `w4a16_grouped_quant.yaml` which essentially tests for the same thing but has slightly different model and dataset

dsikka:
> w4a16_actorder_weight.yaml

1. We lose all non-act order cases. We still need to test this in e2e and lm-eval
2. We need to update the naming in the model produced/pushed to nm-testing so that we indicate act order being present 
3. We have two identical lm-eval cases?

`tests/lmeval/configs/w4a16_actorder_weight.yaml` and `tests/lmeval/configs/w4a16_grouped_quant.yaml`

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15934262421

kylesayrs:
> We lose all non-act order cases. We still need to test this in e2e and lm-eval

Just posted a run above

> We need to update the naming in the model produced/pushed to nm-testing so that we indicate act order being present

Since actorder is now the default, we do not need to include it in model names, since the convention is that default arguments are not included in the model name

> We have two identical lm-eval cases?

As I mentioned previously, these two tests are slightly different. After the above nightly finishes, I'll delete `tests/lmeval/configs/w4a16_actorder_weight.yaml`

dsikka:
> > We lose all non-act order cases. We still need to test this in e2e and lm-eval
> 
> Just posted a run above
> 
> > We need to update the naming in the model produced/pushed to nm-testing so that we indicate act order being present
> 
> Since actorder is now the default, we do not need to include it in model names, since the convention is that default arguments are not included in the model name
> 
> > We have two identical lm-eval cases?
> 
> As I mentioned previously, these two tests are slightly different. After the above nightly finishes, I'll delete `tests/lmeval/configs/w4a16_actorder_weight.yaml`

This is for our internal testing. Please keep the naming convention until we have adaptability of this case being our default. 

And as I mentioned, we still need to test a case with no act order.

kylesayrs:
@dsikka I've updated all of the tests to explicitly retain original actorder behavior

brian-dellabetta (approved):

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_group.yaml
```diff
@@ -2,6 +2,7 @@ quant_stage:
   quant_modifiers:
     GPTQModifier:
       ignore: ["lm_head"]
+      actorder: "group"
```
dsikka, line 5: Can remove from line 14?

kylesayrs: We can, the resolution logic is compatible with both

kylesayrs:
@dsikka 

> I think this file was added after and may have been missed?
https://github.com/vllm-project/llm-compressor/blob/main/tests/e2e/vLLM/recipes/WNA16/recipe_w4a16_awq_sym.yaml

The test you linked to is an AWQ recipe, not a GPTQ recipe?

> For test cases that do not use a recipe / use a scheme directly, they will have act order. So we may want to update to use a recipe

I've updated the `e2e_utils` to explicitly set `actorder=None` in order to maintain behavior.

> We definitely do want to maintain at least one test with no act order

Not sure why this has taken me so long to add 🙃. I added `recipe_w4a16_actorder_weight.yaml`

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_group.yaml
kylesayrs: Removed

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_none.yaml
```diff
@@ -0,0 +1,14 @@
+quant_stage:
```
dsikka, line 1: Thanks - we should add a config / test to use this in a follow-up

dsikka (dismissed):

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_none.yaml
kylesayrs: Let me hook this up now

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -134,15 +135,15 @@ def resolve_actorder(existing):
...
-                    return self.actorder
-                raise ValueError(
-                    "Cannot resolve activation ordering when both "
-                    "`GPTQModifier.actorder` and `QuantizationScheme.actorder` "
-                    "are provided and differ. Either set `GPTQModifier.actorder = "
-                    "None` or remove `actorder` from config groups."
-                )
+            if existing is None or self.actorder == existing:
+                return self.actorder
+            raise ValueError(
```
brian-dellabetta, line 142: is this indent correct? if so we have unreachable code `return existing` just below this

kylesayrs: Thanks for the catch. This ValueError indendentation is correct, and is only reached if `existing is not None and self.actorder != existing`

Issue #1410, opened 2025-05-04
[Feature] Log/info/Save/Restore quantization steps


mratsim:
Inspired by trying to push through #1409 or gptq experiments.

It's incredibly frustrating to fix quantization workflows to have new failures force you to redo hours of quantization.

Some stories:

1. AWQ quantization is done on CPU then GPU, I tried to workaround #1409 by increasing swap space but:
    1. When killed by Linux OOM, the whole terminal is killed so there is no trace of what sample #ID led to OOM. I.e. if left unattended, I have to do guesswork because there is no log available anymore.
    2. After the CPU OOM is fixed, I get a GPU OOM and so have to redo 40min of CPU computation (on overclocked Ryzen 9 9950X). It would be much much much more user-friendly to do a GPU test run at the very beginning, ensuring ahead of time that all hardware requirements are met instead of being surprised after people committed time.
2. GPTQ quantization can have errors due to numerical instability and finding no solution with hessian.
    1. Those should definitely be logged to disk.
    2. It would be great time savings to be allowed to restart GPTQ from layers that converged and allow more samples or more seq lengths or a different dampening fraction for the follow-up layers that failed.

brian-dellabetta:
Hi @mratsim , thank you for the report, this is helpful for us to understand pain points of using llm-compressor.

Regarding your OOM errors, #1426  and https://github.com/neuralmagic/compressed-tensors/pull/301 should go a long way to resolving these. The latter has some nice memory usage charts to show we shouldn't hit OOM when saving.

Regarding GPTQ, I think the concern with offloading hessians is the large storage requirement, but maybe a few could be logged, cc @kylesayrs 

Regarding checkpointing, that could be a good feature to add to the sequential pipelines, but we'd have to think about how to expose it. Hopefully these latest enhancements will make the remove the need for a lot of this

brian-dellabetta:
Hi @mratsim , we have released a 0.6.0 version of llm-compressor -- https://pypi.org/project/llmcompressor/

This should resolve the OOM errors you were hitting with AWQ, and model saving at the end will not require more memory than the current model in memory. Some key PRs that have landed since your post:

- #1419 
- #1444 
- #1426 
- #1435 
- #1263 

i will mark this as resolved for now, please re-open if you hit further issues. We have some GPTQ improvements in-flight as well:
* #1561

Issue #1406, opened 2025-05-01
AWQ Qwen3-235B-A22B and Qwen3-30B-A3B


ehartford:
**Describe the bug**
When I try to AWQ these models, it hangs forever.

**Expected behavior**
I expect it to quantize the model

**Environment**
Nvidia DGX A100

**To Reproduce**

I used examples/awq/awq_one_shot.py and modified it:

```
from compressed_tensors.quantization import (
    QuantizationArgs,
    QuantizationScheme,
    QuantizationStrategy,
    QuantizationType,
)
from transformers import AutoModelForCausalLM, AutoTokenizer

from llmcompressor import oneshot
from llmcompressor.modifiers.awq import AWQModifier
from llmcompressor.modifiers.quantization import QuantizationModifier

MODEL_ID = "Qwen/Qwen3-30B-A3B"
DATASET_ID = "mit-han-lab/pile-val-backup"
DATASET_SPLIT = "validation"
NUM_CALIBRATION_SAMPLES = 256
MAX_SEQUENCE_LENGTH = 512
OUTPUT_DIR = MODEL_ID.split("/")[-1] + "-awq-asym"


def get_calib_dataset(tokenizer):
    from datasets import load_dataset

    ds = load_dataset(
        DATASET_ID,
        split=f"{DATASET_SPLIT}[:{NUM_CALIBRATION_SAMPLES*100}]",
    )

    def preprocess(example):
        return {
            "input_ids": tokenizer.encode(example["text"].strip())[:MAX_SEQUENCE_LENGTH]
        }

    ds = (
        ds.shuffle(seed=42)
        .map(preprocess, remove_columns=ds.column_names)
        .filter(lambda example: len(example["input_ids"]) >= MAX_SEQUENCE_LENGTH)
        .select(range(NUM_CALIBRATION_SAMPLES))
    )

    return ds


if __name__ == "__main__":
    recipe = [
        AWQModifier(bits=4, symmetric=False),
        QuantizationModifier(
            # Ignore these layers during quantization
            ignore=[
                "lm_head",                          # Output layer
                ".*layernorm.*weight",              # All layernorms
                ".*norm.*weight",                   # All norms (input, post attention)
                ".*self_attn.*_norm.*weight",       # Q/K norms
                ".*mlp.gate.weight",                # Expert routing gates
            ],
            config_groups={
                "group_0": QuantizationScheme(
                    targets=["Linear"],
                    weights=QuantizationArgs(
[...]

Issue #1368, opened 2025-04-22
How to run AWQ-W4Afp8 quantization?


wanzhenchn:
How to run AWQ-W4Afp8 quantization on MoE models?

I have run awq-w4afp8 quantization on [Qwen1.5-MoE-A2.7B](https://huggingface.co/Qwen/Qwen1.5-MoE-A2.7B), however,  the ValueError occurred below

![Image](https://github.com/user-attachments/assets/ef53936b-ccc0-44eb-8b12-aeda46ec3a42)

```python
# llmcompressor and compressed-tensors are both installed from source code of main barch.

from compressed_tensors.quantization import (
    QuantizationArgs,
    QuantizationScheme,
    QuantizationStrategy,
    QuantizationType,
)

from llmcompressor.modifiers.awq import AWQModifier
from llmcompressor.modifiers.quantization import QuantizationModifier

model_path="Qwen/Qwen1.5-MoE-A2.7B"

tokenizer = AutoTokenizer.from_pretrained(
model_path, trust_remote_code=True)

model = AutoModelForCausalLM.from_pretrained(
model_path, device_map="auto", torch_dtype="auto",
trust_remote_code=True)

ignore_layers = ["lm_head", "re:.*mlp.gate", "re:.*mlp.shared_expert_gate"]

recipe = [
    AWQModifier(bits=4, symmetric=False),
    QuantizationModifier(
        ignore=ignore_layers,
        config_groups={
            "group_0": QuantizationScheme(
                targets=["Linear"],
                weights=QuantizationArgs(
                    num_bits=4,
                    type=QuantizationType.INT,
                    dynamic=False,
                    symmetric=False,
                    strategy=QuantizationStrategy.GROUP,
                    group_size=128,
                ),
                input_activations=QuantizationArgs(
                    num_bits=8,
                    type=QuantizationType.FLOAT,
                    strategy=QuantizationStrategy.TENSOR,
                    dynamic=False,
                    symmetric=True,
                ),
            ),
        },
    ),
]


oneshot(
    model=model,
    tokenizer=tokenizer,
    dataset="open_playpus",
    recipe=self.recipe,
    max_seq_length=2948,
    num_calibration_samples=2,
    save_compressed=True,
    trust_remote_code_model=True,
    output_dir=self.saved_path,
)

tokenizer.save_pretrained
```

brian-dellabetta:
[...]

PR #1426, opened 2025-05-12
AWQ QuantizationMixin + SequentialPipeline

### LINKED ISSUE #1369: OOM (host) when running AWQ
**Describe the bug**
Running out of host memory when calibrating using AWQModifier.
The model I use is llama-3-instruct. Host memory is 126G.
Is this a exsisting problem that calibrating will consume too much memory? Thanks!

```
(py310_new) zjnyly@ric-MS-7D25:~/LLMs$ python awq_one_shot.py
INFO 04-22 23:14:37 __init__.py:190] Automatically detected platform cuda.
2025-04-22:23:14:38,762 INFO     [modeling.py:957] We will use 90% of the memory on device 0 for storing the model, and 10% for the buffer to avoid OOM. You can set `max_memory` in to a higher value to use more memory (at your own risk).
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████| 4/4 [00:02<00:00,  1.85it/s]
Repo card metadata block was not found. Setting CardData to empty.
2025-04-22:23:14:42,634 WARNING  [repocard.py:108] Repo card metadata block was not found. Setting CardData to empty.
2025-04-22T23:14:44.900211+0800 | reset | INFO - Compression lifecycle reset
Logging all LLM Compressor modifier-level logs to sparse_logs/22-04-2025_23.14.44.log
2025-04-22:23:14:44,900 INFO     [logger.py:391] Logging all LLM Compressor modifier-level logs to sparse_logs/22-04-2025_23.14.44.log
2025-04-22T23:14:44.900749+0800 | from_modifiers | INFO - Creating recipe from modifiers
2025-04-22T23:14:46.201364+0800 | _calibrate | INFO - Running AWQModifier calibration with 256 samples...
 56%|█████████████████████████████████████████████▏                                   | 143/256 [00:53<00:43,  2.62it/s] 
Killed
```

**Environment**
1. OS ubuntu 18.04
2. Python version 3.10.16
3. LLM Compressor version 0.5.1.dev26+g76af7af2 
4. compressed-tensors  0.9.4a20250414

### LINKED ISSUE #1409: [AWQ] Insane memory requirement: over 900GB for 32B model
I tried to quantized GLM-4-0414-32B: https://huggingface.co/THUDM/GLM-4-32B-0414

Recipe:
```python
recipe = [
    AWQModifier(
        bits=4,
        symmetric=False,
        # Read input->output from https://github.com/huggingface/transformers/blob/v4.51.3/src/transformers/models/glm4/modeling_glm4.py
        # which are somewhat easier than vllm ones as it's all in a single file
        mappings=[
            AWQMapping("re:.*input_layernorm", ["re:.*q_proj", "re:.*k_proj", "re:.*v_proj"]),
            AWQMapping("re:.*v_proj", ["re:.*o_proj"]),
            AWQMapping("re:.*post_attention_layernorm", ["re:.*gate_up_proj"]),
            AWQMapping("re:.*gate_up_proj", ["re:.*down_proj"]),
        ]
    ),
    QuantizationModifier(
        ignore=ignore_layers,
        config_groups={
            "group_0": QuantizationScheme(
                targets=["Linear"],
                weights=QuantizationArgs(
                    num_bits=4,
                    type=QuantizationType.INT,
                    dynamic=False,
                    symmetric=False,
                    group_size=128,
                ),
            ),
        },
    )
]
```

I tried using 128 samples as suggested in those slides ("Calibration set"): https://minjiazhang.github.io/courses/fall24-resource/slides/awq.pdf

However every sample memory usage grew by 1~5 GB leading in the end to over 900GB before I decided to give up on AWQ. Even with a swapfile, the time was spent in kernel swap in/out and IO and compute was slow to then frustratingly crash on Cuda OOM once that CPU part was solved.

Screenshot:
Image

 Side-note: couldn't the calibration be made multi-threaded?

brian-dellabetta:
SUMMARY:
- Add QuantizationMixin to AWQModifier so we don't have redundant inputs (num_bits, symmetric, group_size)
- Move AWQModifier to sequential pipelines, to avoid huge memory requirements of caching all activations at once.

Regression test results are acceptable, results are all roughly the same, and within stderr, see test plan below.

Resolves #1409 
Resolves #1369 
Related to #1383
Related to #1406 
Related to #1368 
Related to #1410 

More improvements split into #1435

TEST PLAN:
- [x] Rerun tests to validate
No regression in tests, comparing against those reported in [original AWQ PR](https://github.com/vllm-project/llm-compressor/pull/1177#issuecomment-2773113040). All gsm8k results are within stderr:

| Type            | gsm8k       | wikitext
| ------          | ------      | ----- 
| Old AWQ+QuantModifier Sym          | .1054, .1069     | 9.1931 
| New AWQ+QuantMixin Sym        | .1077, .1084 | 9.1841
| Old AWQ+QuantModifier Asym             | .1274, .1281 | 9.0281
| New AWQ+QuantMixin Asym        | .1312, .1350 | 9.0288

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -176,7 +176,25 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         model._group_size = next(iter(group_size_set))
 
-        # TODO confirm no activation quantization
+        in_num_bits_set = set(
+            group.input_activations.num_bits
+            for group in config.config_groups.values()
+            if group.input_activations is not None
+        )
+        assert len(in_num_bits_set) == 0 or in_num_bits_set == {16}, (
+            "AWQ activations must be 16-bit precision, "
+            f"input activations {in_num_bits_set} not allowed"
+        )
+
+        out_num_bits_set = set(
+            group.output_activations.num_bits
+            for group in config.config_groups.values()
+            if group.output_activations is not None
+        )
+        assert len(out_num_bits_set) == 0 or out_num_bits_set == {16}, (
+            "AWQ activations must be 16-bit precision, "
+            f"output activations {out_num_bits_set} not allowed"
+        )
 
         return model
 
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -90,6 +90,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                 - raise error if any unused activations remain
         - on_end
             - re-run logic of sequential epoch end (in case of basic pipeline)
+            - set scales and zero points
             - remove activation hooks
         - on_finalize
             - clear resolved mappings and captured activations
@@ -139,11 +140,10 @@ class AWQModifier(Modifier, QuantizationMixin):
         if not model.config_groups and not model.scheme:
             raise ValueError("AWQ requires either a config_groups or a scheme")
 
-        # TODO better way to do this?
+        # TODO better way to do this? model.resolve_config() ??
         config_groups = model.config_groups or {
             "group_0": preset_name_to_scheme(model.scheme, model.targets)
         }
-        # config = model.resolve_config()
 
         num_bits_set = set(
             group.weights.num_bits
@@ -183,6 +183,7 @@ class AWQModifier(Modifier, QuantizationMixin):
     def on_initialize(self, state: State, **kwargs) -> bool:
         """
         Initialize AWQ on the given state
+        Initialize quantization, resolve mappings, cache module kwargs
 
         :param state: state to run AWQ on
         :return: True on a successful run, False otherwise
@@ -234,16 +235,15 @@ class AWQModifier(Modifier, QuantizationMixin):
 
     def on_end(self, state: State, event: Event, **kwargs):
         """
-        Finish calibrating by removing observers and calibration hooks
+        Finish calibrating by setting scales and zero-points,
+         removing observers and calibration hooks
         """
-        # confirm all activations have been used
-        if len(self._activations) > 0:
-            raise RuntimeError("Some cached activations were not used")
+        self._assert_all_activations_consumed()
 
         self.ended_ = True
 
         modules = list(state.model.modules())
-        for module in tqdm(modules, desc="Calibrating weights", leave=False):
+        for module in tqdm(modules, desc="Calibrating weights"):
             update_weight_zp_scale(module)
 
         QuantizationMixin.end_calibration(self, state.model)
@@ -384,9 +384,7 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         :param model: model to apply smoothing to
         """
-        for mapping in tqdm(
-            self._resolved_mappings, desc="Smoothing activation scales", leave=False
-        ):
+        for mapping in tqdm(self._resolved_mappings, desc="Smoothing"):
             # When using SequentialPipeline, not all the mappings will have
             # activations not found in this segment
             if mapping.smooth_name not in self._activations:
@@ -491,9 +489,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                     smooth(layer)
                 smooth(smooth_layer)
 
-        # confirm all activations have been used
-        if len(self._activations) > 0:
-            raise RuntimeError("Some cached activations were not used")
+        self._assert_all_activations_consumed()
 
     def _compute_best_scale(
         self,
@@ -624,6 +620,14 @@ class AWQModifier(Modifier, QuantizationMixin):
         loss /= num_elements
 
         return loss
+
+    def _assert_all_activations_consumed(self):
+        """
+        Confirm all activations have been consumed
+        If not, something has gone wrong
+        """
+        if len(self._activations) > 0:
+            raise RuntimeError("Some cached activations were not used")
 
     def _set_module_kwargs(self, model, dataloader) -> None:
         _, modules = next(iter(get_layers("re:.*layers", model).items()))
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -217,9 +217,6 @@ class AWQModifier(Modifier, QuantizationMixin):
             if not self.started_:
                 self.on_start(state, None)
 
-        elif event.type_ == EventType.SEQUENTIAL_EPOCH_START:
-            pass
-
         elif event.type_ == EventType.SEQUENTIAL_EPOCH_END:
             # Run smoothing in case of sequential pipeline
             self._apply_smoothing(state.model)
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -137,6 +137,11 @@ class AWQModifier(Modifier, QuantizationMixin):
 
     @model_validator(mode="after")
     def validate_model_after(model: "AWQModifier") -> "AWQModifier":
+        """
+        Confirm only one configuration for group_size, symmetric, and num_bits,
+        as AWQ algorithm depends on it
+        Confirm no activation quantization, as AWQ only works with WNA16
+        """
         if not model.config_groups and not model.scheme:
             raise ValueError("AWQ requires either a config_groups or a scheme")
 
@@ -178,6 +183,8 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         model._group_size = next(iter(group_size_set))
 
+        # TODO confirm no activation quantization
+
         return model
 
     def on_initialize(self, state: State, **kwargs) -> bool:
@@ -195,8 +202,7 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         self._set_resolved_mappings(state.model)
 
-        with calibration_forward_context(state.model):
-            self._set_module_kwargs(state.model, state.data.calib)
+        self._set_module_kwargs(state.model, state.data.calib)
 
         return True
 
@@ -222,13 +228,11 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         elif event.type_ == EventType.SEQUENTIAL_EPOCH_END:
             # Run smoothing in case of sequential pipeline
-            with calibration_forward_context(state.model), HooksMixin.disable_hooks():
-                self._apply_smoothing(state.model)
+            self._apply_smoothing(state.model)
 
         elif event.type_ == EventType.CALIBRATION_EPOCH_END:
             # Run smoothing in case of basic pipeline
-            with calibration_forward_context(state.model), HooksMixin.disable_hooks():
-                self._apply_smoothing(state.model)
+            self._apply_smoothing(state.model)
 
             if not self.ended_:
                 self.on_end(state, None)
@@ -385,8 +389,8 @@ class AWQModifier(Modifier, QuantizationMixin):
         :param model: model to apply smoothing to
         """
         for mapping in tqdm(self._resolved_mappings, desc="Smoothing"):
-            # When using SequentialPipeline, not all the mappings will have
-            # activations not found in this segment
+            # NOTE: When using SequentialPipeline, not all the mappings
+            # will have cached activations in the segment being udpated
             if mapping.smooth_name not in self._activations:
                 continue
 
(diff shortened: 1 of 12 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -96,53 +112,138 @@ class AWQModifier(Modifier):
...
+
+        elif event.type_ == EventType.SEQUENTIAL_EPOCH_END:
+            self._concat_collected_activations()
+
+            with calibration_forward_context(state.model), HooksMixin.disable_hooks():
+                self._apply_smoothing(state.model)
+
+            self._activations.clear()
+
+        elif event.type_ == EventType.CALIBRATION_EPOCH_END:
```
kylesayrs, line 222: In order to support the basic pipeline, do `self._apply_smoothing(state.model)`

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -96,53 +112,138 @@ class AWQModifier(Modifier):
...
+        # This choice does not seem to have a meaningful impact on accuracy
+        state.model.apply(disable_quantization)
+        self._setup_activation_cache_hooks()
+
+    def on_event(self, state: State, event: Event, **kwargs):
+        if event.type_ == EventType.CALIBRATION_EPOCH_START:
+            if not self.started_:
+                self.on_start(state, None)
+
+        elif event.type_ == EventType.SEQUENTIAL_EPOCH_START:
```
kylesayrs, line 211: How we typically avoid having to have an epoch start event is by deleting accumulations as they're used

```python3
def sequential_epoch_end
   for activation in achivations:
       ... use activation
       del self._activations[...]
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -220,64 +324,47 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
         """
 
-        def create_hook_fn(layer_name):
-            def hook_fn(module, inp, out):
-                inp = inp[0].cpu().detach()
-
-                if layer_name in self._scales:
-                    self._scales[layer_name].append(inp)
+        def create_cache_activation_hook(layer_name):
+            def cache_activation_hook_fn(
```
kylesayrs, line 334: nit: You can just define this hook as a class method, rather than creating a new instance of the function every time

src/llmcompressor/modifiers/utils/pytorch_helpers.py
```diff
@@ -34,54 +28,6 @@ def apply_pad_mask_to_batch(batch: Dict[str, torch.Tensor]) -> Dict[str, torch.T
     return batch
 
 
-def run_calibration_forward(
```
kylesayrs, line 37: sweet

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -96,53 +112,138 @@ class AWQModifier(Modifier):
...
         """
 
+        # apply config to model and prepare calibration hooks
+        if QuantizationMixin.has_config(self):
+            QuantizationMixin.initialize_quantization(self, state.model)
+
         self._set_resolved_mappings(state.model)
 
         with calibration_forward_context(state.model):
             self._set_module_kwargs(state.model, state.data.calib)
```
kylesayrs, line 191: Can we get rid of this?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -96,53 +112,138 @@ class AWQModifier(Modifier):
...
+            if not self.started_:
+                self.on_start(state, None)
+
+        elif event.type_ == EventType.SEQUENTIAL_EPOCH_START:
+            self._activations.clear()
+
+        elif event.type_ == EventType.SEQUENTIAL_EPOCH_END:
+            self._concat_collected_activations()
+
+            with calibration_forward_context(state.model), HooksMixin.disable_hooks():
```
kylesayrs, line 217: Consider moving this calibration_forward_context to within _apply_smoothing so that it only covers the area that you need it to

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -96,53 +112,138 @@ class AWQModifier(Modifier):
...
+        elif event.type_ == EventType.CALIBRATION_EPOCH_END:
+            if not self.ended_:
+                self.on_end(state, None)
+
+    def on_end(self, state: State, event: Event, **kwargs):
+        """
+        Finish calibrating by removing observers and calibration hooks
+        """
+        self.ended_ = True
+
```
kylesayrs, line 231: I typically add a warning here to check if there are any `_activations` that haven't been used to scale yet

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: thanks, added

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: updated to delete activations as they are used, and raise error if unused activations exist

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -220,64 +324,47 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                layer, create_cache_activation_hook(mapping.smooth_name), "forward"
             )
-
-        run_calibration_forward(
-            model,
-            calibration_dataloader,
-        )
+            self._activation_hooks.add(hook)
 
     def _concat_collected_activations(self) -> None:
```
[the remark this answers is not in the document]
brian-dellabetta: removed

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I'm hoping so, but we need it to run forward passes during smoothing. Bit different than GPTQ, AWQ needs the full output as well as the quantized output for the loss function. Maybe we can just cache output activations as well?

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: added

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I think a better solution is dropping forward passes from smoothing entirely, by caching output activations. but let's chat about that

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: Is there a performance hit to doing that? If not, I prefer this as more scoped

brian-dellabetta: Also note the `layer_name` input, i think we'll have to create a function every time because of this

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -96,53 +116,147 @@ class AWQModifier(Modifier):
...
+    _activations: Dict[str, List[torch.Tensor]] = PrivateAttr(default_factory=dict)
+    _activation_hooks: Set[RemovableHandle] = PrivateAttr(default_factory=set)
+    _module_kwargs: Dict = PrivateAttr(default_factory=dict)
+
+    @model_validator(mode="after")
+    def validate_model_after(model: "AWQModifier") -> "AWQModifier":
+        if not model.config_groups and not model.scheme:
+            raise ValueError("AWQ requires either a config_groups or a scheme")
+
+        # TODO better way to do this?
```
kylesayrs, line 141: You can use `config = self.resolve_config()` and then validate that

brian-dellabetta: thanks!

src/llmcompressor/modifiers/awq/base.py
kylesayrs: Ah yeah, in other implementations I curry this using `partial` but this is good too

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: hmm, got a link? I don't see any `resolve_config` defined anywhere in the code

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I think we'll need to keep the forward passes for now, so I moved this down into a lower scope in _apply_smoothing and _compute_loss

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: 🍛

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: found `.resolve_quantization_config`, thanks!

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: Unfortunately, we need this for forward passes through the parent modules, which are not always Linear layers. So let's leave as is for now, and we can tackle this with more hooks to inputs to each of the smoothing layers in a follow-up?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -49,32 +53,50 @@ class AWQModifier(Modifier):
...
+                - raise error if any unused activations remain
+        - on_end
+            - re-run logic of sequential epoch end (in case of basic pipeline)
+            - set scales and zero points
+            - remove activation hooks
         - on_finalize
             - clear resolved mappings and captured activations
 
+    :param sequential_targets: list of layer names to compress during AWQ, or
+        '__ALL__' to compress every layer in the model
```
kylesayrs, line 99:
```suggestion
    :param sequential_targets: list of module names to compress in the same calibration pass
```

There might be a better way to describe this in english

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -96,53 +118,147 @@ class AWQModifier(Modifier):
...
 
+        # apply config to model and prepare calibration hooks
+        if QuantizationMixin.has_config(self):
+            QuantizationMixin.initialize_quantization(self, state.model)
+
         self._set_resolved_mappings(state.model)
 
-        with calibration_forward_context(state.model):
-            self._set_module_kwargs(state.model, state.data.calib)
+        self._set_module_kwargs(state.model, state.data.calib)
```
kylesayrs, line 199: This `_set_module_kwargs` function still needs work. This code, as currently written, still runs through the entire dataset (with massive batch size). This means that large number of calibration samples will still trigger oom.

kylesayrs (changes requested):

src/llmcompressor/modifiers/awq/base.py
kylesayrs: It aught to be possible to capture these kwargs during the calibration step

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: Will use your suggestion, thanks! I pulled it directly from [GPTQ](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/gptq/base.py#L68). Should I change it there as well?

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: It only does it for the first module, and errors out after kwargs are cached. certainly not the greatest, and we can improve, but I'd like to do that in a follow-up

kylesayrs (approved):
This modifier still has extreme memory constraints due to how we're still megabatching everything for `_set_module_kwargs` and `_apply_smoothing`

This followup PR gets part way towards removing that, but still batches outputs in its current state (grep for `torch.cat`): https://github.com/vllm-project/llm-compressor/pull/1435

Let's land this as is and add accumulation in a follow up to reduce memory requirements

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: We discussed that we'll address this in https://github.com/vllm-project/llm-compressor/pull/1435, so i'm going to resolve this comment

rahul-tuli (approved):
Lgtm! Good updates!

PR #1429, opened 2025-05-13
Remove RecipeArgs class & its references

shanjiaz:
SUMMARY:
Removed Recipe Arg class. 

1. It is now a dict `Dict[str, Any]` and being stored in `modifier.args`, `stage.args` and `recipe.args`. 
2. Removed all references of `calculate_start`, `calculate_end` and `shift`, as they're not actively being used.
3. Removed all references of `evaluate` and `args_evaluated` since we're not currently using them.

TEST PLAN:
Tested `examples/quantization_2of4_sparse_w4a16` and results is saved [here](https://www.notion.so/Recipe-classes-cleanup-1ea30c7e73f380398eade1d2806a3097?pvs=4)


src/llmcompressor/recipe/utils.py
brian-dellabetta, line 1: Were these all pulled from one of the other deleted files? Do we need them? I think it'd be good to remove any eval calls from recipe parsing, using Jinja or some other templating system instead if needed

shanjiaz: Yes! The utils came from `recipe/args.py`. Unfortunately we call them [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/recipe/modifier.py#L60) and this is how we parse modifier arguments. Good point! I'll explore some options for templating!

brian-dellabetta: Do we still want to support those? Maybe we can just deprecate them and remove them entirely, none of our examples would rely on this, not sure if many people are using it out in the wild.

src/llmcompressor/recipe/stage.py
```diff
@@ -16,80 +15,36 @@ class RecipeStage(RecipeBase):
...
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     group: Optional[str] = None
-    args: Optional[RecipeArgs] = None
+    args: Optional[Dict[str, Any]] = None
     enabled: bool = True
     modifiers: List[RecipeModifier] = Field(default_factory=list)
     exclude_default: bool = False
-    args_evaluated: Optional[RecipeArgs] = None
+    args_evaluated: Optional[Dict[str, Any]] = None
```
kylesayrs, line 33: Do we still need args_evaluated? Isn't `args` enough?

shanjiaz: yeah true! will remove them

brian-dellabetta (dismissed):
This looks good to me! Worth confirming with @dsikka that we don't want to support any of this functionality moving forward, but I don't think any of it is needed

brian-dellabetta (approved):

kylesayrs (approved):
Nice job. We should definitely consider adding this functionality back once the Recipe class is in a simpler state

brian-dellabetta:
> Nice job. We should definitely consider adding this functionality back once the Recipe class is in a simpler state

@kylesayrs is there any functionality removed here that we would want to keep?

src/llmcompressor/recipe/stage.py
```diff
@@ -16,80 +15,34 @@ class RecipeStage(RecipeBase):
 
     :param group: Name of the current stage
     :param run_type: Whether this is a oneshot or training stage
-    :param args: Optional RecipeArgs to use for this stage
+    :param args: Optional recipe args to use for this stage
```
dsikka, line 18: Is this docstring still correct?

dsikka:
> > Nice job. We should definitely consider adding this functionality back once the Recipe class is in a simpler state
> 
> @kylesayrs is there any functionality removed here that we would want to keep?

We likely want to add RecipeArgs but we'd likely rewrite the class when we do.
I dont think there's any need to bring back calculcate_start, calculate_end, or evaluate related args

@kylesayrs @brian-dellabetta

dsikka (changes requested):

src/llmcompressor/recipe/stage.py
shanjiaz: args are still being used for passing in targets, ignore, etc. Stage and Modifier class will eventually be removed though.

PR #1431, opened 2025-05-14
[Logging] Support logging once

kylesayrs:
## Purpose ##
* Add ability to log messages only once

## Changes ##
* Add `support_log_once` filter
* Integrate `_initialize_metric_logging` function to reduce verbosity

## Testing ##
```python3
from llmcompressor.logger import logger
logger.bind(log_once=True).info("This will only log once")
# 2025-05-14T10:14:30.987512-0400 | <module> | INFO - This will only log once
logger.bind(log_once=True).info("This will only log once")
logger.bind(log_once=True).info("This will only log once")

logger.bind(log_once=True).info("This is unique")
# 2025-05-14T10:14:42.987610-0400 | <module> | INFO - This is unique
logger.bind(log_once=True).warning("This is unique")
# 2025-05-14T10:14:57.907901-0400 | <module> | WARNING - This is unique
```


src/llmcompressor/logger.py
```diff
@@ -93,30 +97,47 @@ def configure_logger(config: Optional[LoggerConfig] = None) -> None:
...
     """
-    Initialize metric logging for loguru and metric-related libraries
-    Defaults to stdout
-    usage:
-        `logger.log("METRIC", "foo description, bar result")`
+    Support logging only once using `.bind(log_once=True)`
 
+    ```
+    logger.bind(log_once=True).info("This will only log once")
+    logger.bind(log_once=True).info("This will only log once")
```
brian-dellabetta, line 127: redundant line, maybe you meant to show the converse?
```suggestion
    logger.bind(log_once=False).info("This will log multiple times")
```

brian-dellabetta (approved):
Might be worth adding this to a logger invocation somewhere in this PR, so we can see an example in the code

src/llmcompressor/logger.py
kylesayrs: Not really, I've updated with more exampling

brian-dellabetta (approved):
very nice!

markurtz (approved):

PR #1435, opened 2025-05-15
[AWQ] Support accumulation for reduced memory usage

kylesayrs:
> [!NOTE]  
> @brian-dellabetta updated summary here (took over this PR from @kylesayrs )

### Summary
This update removes the `Catcher` logic ported from AutoAWQ, instead using the SequentialPipeline features and a couple hooks to cache args into module forward passes, as needed to run AWQ. Should cause no significant change in results, but this should be a more accurate implementation because kwargs are cached for each parent layer, rather than re-using those of the first module.

Leveraging `IntermediatesCache` for cached values, this also exposes a new `offload_cache` bool value on `AWQModifier`, if set to True it will offload cached values at the expense of slower runtime. With `meta-llama/Llama-2-7b-hf`, offloading decreases max GPU memory from ~27GB to ~20GB, at the cost of `apply_smoothing` taking ~17 seconds per iteration as opposed to ~5 seconds. Because of this, I am leaving the default to not offload, just noting in the docstring to toggle this if users encounter OOM errors.

### Test Plan
Confirmed that these changes don't significantly alter PPL scores relative to current implementation on main.

- `meta-llama/Llama-3.2-3B-Instruct`
  - PPL 14.1523 on main, 14.081 on this branch
- `Qwen/Qwen2.5-7B-Instruct`
  - PPL 10.411 on main, 10.736 on this branch
- `meta-llama/Llama-2-7b-hf`
  - PPL 9.5075 on main, 9.503 on this branch


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -131,9 +132,11 @@ class AWQModifier(Modifier, QuantizationMixin):
...
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: List[ResolvedMapping] = PrivateAttr(default_factory=list)
-    _activations: Dict[str, List[torch.Tensor]] = PrivateAttr(default_factory=dict)
-    _activation_hooks: Set[RemovableHandle] = PrivateAttr(default_factory=set)
-    _module_kwargs: Dict = PrivateAttr(default_factory=dict)
+    _samples: Dict[Module, IntermediatesCache] = PrivateAttr(
+        default_factory=IntermediatesCache
+    )
+    _sample_means: Dict[Module, float] = PrivateAttr(default_factory=dict)
+    _num_samples: Dict[Module, int] = PrivateAttr(default_factory=dict)
```
brian-dellabetta, line 139: I know this is in GPTQ, but a field named `num_samples` indicates an `int` in any virtually context i've seen it. what bout `_sample_counts` so it's more similar to `sample_means`?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -131,9 +132,11 @@ class AWQModifier(Modifier, QuantizationMixin):
 
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: List[ResolvedMapping] = PrivateAttr(default_factory=list)
-    _activations: Dict[str, List[torch.Tensor]] = PrivateAttr(default_factory=dict)
-    _activation_hooks: Set[RemovableHandle] = PrivateAttr(default_factory=set)
-    _module_kwargs: Dict = PrivateAttr(default_factory=dict)
+    _samples: Dict[Module, IntermediatesCache] = PrivateAttr(
+        default_factory=IntermediatesCache
+    )
+    _sample_means: Dict[Module, float] = PrivateAttr(default_factory=dict)
```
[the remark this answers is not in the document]
kylesayrs: No? This is used to compute [x_mean](https://github.com/vllm-project/llm-compressor/pull/1435/files#diff-a79ed86c2a25b361e101f22c2cdcc549cf2dbc09b48de91c752f5edc4cf925a7L450)

src/llmcompressor/modifiers/awq/base.py
kylesayrs: That's good too!

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -571,9 +546,7 @@ def _compute_best_scale(
                     )
 
             # W * X
-            int_w_output = self._forward_input_with_kwargs(
-                module=module2inspect, inputs=x, input_kwargs=self._module_kwargs
-            )
+            int_w_output = self._run_samples(parent_layer)
```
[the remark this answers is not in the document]
kylesayrs: x is the dataset, right?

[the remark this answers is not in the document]
brian-dellabetta: my bad, misread what `self._samples` was. can we keep that as a different name to distinguish it from actual data?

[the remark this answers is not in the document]
kylesayrs: Yeah I'm always happy to take suggestions on names for any of the code I push 🙂

src/llmcompressor/pipelines/cache.py
```diff
@@ -127,6 +119,22 @@ def delete(self, batch_index: int, consumed_names: Optional[List[str]] = None):
...
+        self.update(batch_index, values)
+
+    def iter(
+        self, input_names: Optional[List[str]] = None
+    ) -> Generator[Any, None, None]:
+        for batch_index in self.batch_intermediates:
+            yield self.fetch(batch_index, input_names)
+
+    def __iter__(self) -> Generator[Any, None, None]:
+        yield from self.iter()
```
rahul-tuli, line 133: Why not just have __iter__?

rahul-tuli (commented):
These deletions make me happy! LGTM!

src/llmcompressor/pipelines/cache.py
kylesayrs: Saves an extra function call
```python3
for batch in cache:
    ...
```

```python3
for batch in cache.iter():
    ...
```

brian-dellabetta (approved):
approving a PR I contributed to

Screenshot 2025-05-27 at 5 07 14 PM

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -382,30 +383,50 @@ def _setup_activation_cache_hooks(self) -> None:
...
 
-        def create_cache_activation_hook(smooth_layer_name):
-            def cache_activation_hook_fn(
+        def cache_parent_kwargs_hook(
+            module: torch.nn.Module,
+            args: Tuple[torch.Tensor, ...],
+            kwargs,
+        ):
+            values = inspect.signature(module.forward).bind(*args, **kwargs)
+            self._parent_kwargs_cache[module].append(values.arguments)
```
kylesayrs, line 392: It doesn't seem like these values are offloaded to the CPU, meaning that you're storing N x L samples on the GPU, which creates a memory bottleneck. This is where using the activations_cache offloading logic might be useful

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -528,14 +520,26 @@ def smooth(module):
                     smooth(layer)
                 smooth(smooth_layer)
 
+            # remove caches needed to smooth this mapping
+            del self._smooth_activation_cache[mapping.smooth_name]
```
kylesayrs, line 524: Do we want to delete the parent kwargs cache?

src/llmcompressor/modifiers/awq/base.py
kylesayrs, line 392: It seems like this is caching more than just the kwargs, so maybe `_parent_kwargs_cache` is not an appropriate name?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -595,13 +599,12 @@ def _compute_best_scale(
...
-            int_w_output = self._forward_input_with_kwargs(
-                module=module2inspect, inputs=x, input_kwargs=self._module_kwargs
-            )
-            int_w_output = int_w_output.clip(
-                torch.finfo(int_w_output.dtype).min,
-                torch.finfo(int_w_output.dtype).max,
-            )
+            with HooksMixin.disable_hooks():
+                int_w_output = self._run_samples(parent_module)
+                int_w_output = int_w_output.clip(
```
kylesayrs, line 604: Does this line actually do anything? Isn't `int_w_output` by definition already in between the min and max values of its dtype?

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I moved that logic into `assert_all_activations_consumed`, which in hindsight is not great. I'll move it back to just below this line

brian-dellabetta: updated

src/llmcompressor/pipelines/cache.py
```diff
@@ -31,28 +34,17 @@ class IntermediatesCache:
...
-        offload_device: torch.device,
+        batch_intermediates: Optional[List[IntermediateValues]] = None,
+        offload_device: torch.device = "cpu",
     ):
-        self.batch_intermediates = batch_intermediates
+        self.batch_intermediates = batch_intermediates or []
         self.offload_device = offload_device
 
-    @classmethod
-    def empty(cls, num_batches: int, offload_device: torch.device):
```
brian-dellabetta, line 46: @kylesayrs `IntermediatesCache.empty` is used in `src/llmcompressor/pipelines/layer_sequential/helpers.py`, should we retain this?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -237,7 +241,11 @@ def on_start(self, state: State, event: Event, **kwargs):
         # This choice does not seem to have a meaningful impact on accuracy
         state.model.apply(disable_quantization)
 
-        self._setup_activation_cache_hooks()
+        self._setup_activation_cache_hooks(
+            offload_device=(
+                "cpu" if self.offload_cache else get_execution_device(state.model)
```
brian-dellabetta, line 246: @kylesayrs this seems to work but please confirm this is what you'd expect it to be

kylesayrs: This won't work for models across separate devices. For example, if the model device is gpu0, but you're capturing from a parent on gpu1. When you try to rerun samples, the devices won't match

kylesayrs: This is why I suggested adding an "offload_device=None" option to the IntermedaitesCache

src/llmcompressor/pipelines/cache.py
kylesayrs: Yes, my mistake for deleting it

kylesayrs (commented):
Sweet!

brian-dellabetta (approved):

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -134,9 +136,14 @@ class AWQModifier(Modifier, QuantizationMixin):
...
     _resolved_mappings: List[ResolvedMapping] = PrivateAttr(default_factory=list)
-    _activations: Dict[str, List[torch.Tensor]] = PrivateAttr(default_factory=dict)
-    _activation_hooks: Set[RemovableHandle] = PrivateAttr(default_factory=set)
-    _module_kwargs: Dict = PrivateAttr(default_factory=dict)
+    # Cache list of forward input args for each parent module, one dict for each batch
+    _parent_args_cache: Dict[Module, IntermediatesCache] = PrivateAttr(
+        default_factory=dict
+    )
+    # Dict[smooth layer name, (activation means, activation counts)]
+    _smooth_activation_cache: Dict[str, Tuple[torch.FloatTensor, int]] = PrivateAttr(
```
kylesayrs, line 144: This is not really an activations cache, it doesn't store activations. Instead it stores means

kylesayrs (commented):
Please change this naem

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: updated to None, with new functionality in intermediates cache to not offload if offload_device is None

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: what would be a better name here? `_smooth_activation_means_cache`?

brian-dellabetta: renamed to `_smooth_activation_means`

brian-dellabetta (approved):
also approved changes with @kylesayrs , co-owner of this PR

PR #1444, opened 2025-05-16
AWQModifier fast resolve mappings, better logging, MoE support

brian-dellabetta:
SUMMARY:
In AWQ, resolving mappings can take a while because it is traversing the entire model tree, rather than just the parent, to find the balance layers. This scopes the search to just the parent module. For MoE models, the previous implementation only found a single layer for each regex string provided in mappings. This updates that to find as many as it can, which is necessary for mappings like 

```python
AWQMapping(
    "re:.*post_attention_layernorm$",
    ["re:.*mlp.experts.*.gate_proj$", "re:.*mlp.experts.*.up_proj$"],
)
```

which have multiple gate_proj and up_proj layers, one for each expert.

gsm8k results with `Qwen/Qwen3-30B-A3B` MoE model after AWQ W4A16 Symmetric:

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.3813|±  |0.0134|
|     |       |strict-match    |     5|exact_match|↑  |0.8810|±  |0.0089|

TEST PLAN:
- [x] working with `Qwen/Qwen3-30B-A3B` with [same set of mappings used in AutoAWQ](https://github.com/casper-hansen/AutoAWQ/blob/main/awq/models/qwen3_moe.py#L24). Example included in this PR in `examples/awq/qwen3_moe_example.py`. Ran successfully in ~2 hours on a single H100 with ~70GB of 80GB used (additional memory needed during saving)
- [x] Same wikitext PPL (14.0814) as on `main` for `meta-llama/Llama-3.2-3B-Instruct`

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -767,7 +767,9 @@ def get_lowest_common_parent(names: List[str], module: Module) -> Tuple[str, Mod
             parent_name = s1[:i].rstrip(".")
             break
 
-    while parent_name != "":
+    while True:
+        if parent_name == "":
+            return "", module
         parent = get_layer_by_name(parent_name, module)
         if not isinstance(parent, torch.nn.ModuleList):
             return parent_name, parent
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -338,9 +338,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                         # exclude v_proj->o_proj mappings whose shapes are incompatible
                         # https://github.com/mit-han-lab/llm-awq/pull/67#issuecomment-1681632777
                         if (
-                            ".v_proj" in smooth_name
-                            and ".o_proj" in balance_name
-                            and isinstance(smooth_layer, torch.nn.Linear)
+                            isinstance(smooth_layer, torch.nn.Linear)
                             and isinstance(balance_layer, torch.nn.Linear)
                             and ".o_proj" in balance_name
                             and (
@@ -406,6 +404,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                 _output: torch.Tensor,
             ):
                 self._smooth_activation_means[smooth_name] = _accumulate_mean(
+                    # Assume that first argument is the input
                     args[0].cpu().detach().squeeze(),
                     self._smooth_activation_means.get(smooth_name, None),
                 )
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -412,15 +412,11 @@ class AWQModifier(Modifier, QuantizationMixin):
         def create_cache_smooth_activations_hook_fn(smooth_name):
             def cache_smooth_activations_hook(
                 _module: torch.nn.Module,
-                args: Union[torch.Tensor, Tuple[torch.Tensor, ...]],
+                args: Tuple[torch.Tensor, ...],
                 _output: torch.Tensor,
             ):
-                if not isinstance(args, Tuple):
-                    print(f"GOT unexpected args {args}")
-                inp = args[0] if isinstance(args, Tuple) else args
-
                 self._smooth_activation_means[smooth_name] = _accumulate_mean(
-                    inp.cpu().detach().squeeze(),
+                    args[0].cpu().detach().squeeze(),
                     self._smooth_activation_means.get(smooth_name, None),
                 )
 
@@ -490,10 +486,14 @@ class AWQModifier(Modifier, QuantizationMixin):
                 # [STEP 3]: Compute output of module
                 # could cache from hook, rather than recomputing here
                 fp16_output = self._run_samples(parent_module)
-                fp16_output = fp16_output.clip(
-                    torch.finfo(fp16_output.dtype).min,
-                    torch.finfo(fp16_output.dtype).max,
-                )
+                if fp16_output.shape[0] == 0:
+                    breakpoint()
+                    logger.info(
+                        f"Skipping smooth_layer {mapping.smooth_name}, no activations "
+                        "found to scale. This occurs in MoE models when calibration "
+                        "samples don't activate certain experts."
+                    )
+
                 x_mean = self._smooth_activation_means[mapping.smooth_name][0]
 
                 # [STEP 4]: Compute loss
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -29,7 +29,7 @@ from llmcompressor.utils.helpers import calibration_forward_context
 from llmcompressor.utils.pytorch.module import (
     get_layers,
     get_matching_layer,
-    get_parent_by_name,
+    get_layer_by_name,
 )
 
 __all__ = ["AWQModifier"]
@@ -330,9 +330,8 @@ class AWQModifier(Modifier, QuantizationMixin):
                 balance_layers, balance_names = [], []
                 for balance_suffix in mapping.balance_layers:
                     # find the submodule that matches the activation layer
-                    parent_name, parent_module = get_parent_by_name(
-                        layer_name=smooth_name, model=model
-                    )
+                    parent_name = ".".join(smooth_name.split(".")[:-1])
+                    parent_module = get_layer_by_name(parent_name, model)
                     balance_name, balance_layer = get_matching_layer(
                         balance_suffix,
                         smooth_name,
@@ -380,9 +379,9 @@ class AWQModifier(Modifier, QuantizationMixin):
                 else:
                     # for multiple balance layers,
                     # parent of any balance layer is the parent
-                    parent_name, parent = get_parent_by_name(
-                        layer_name=balance_name, model=model
-                    )
+                    parent_name = ".".join(balance_name.split(".")[:-1])
+                    parent = get_layer_by_name(parent_name, model)
+
                 resolved_mappings.append(
                     ResolvedMapping(
                         smooth_name,
@@ -413,14 +412,15 @@ class AWQModifier(Modifier, QuantizationMixin):
         def create_cache_smooth_activations_hook_fn(smooth_name):
             def cache_smooth_activations_hook(
                 _module: torch.nn.Module,
-                args: Tuple[torch.Tensor, ...],
+                args: Union[torch.Tensor, Tuple[torch.Tensor, ...]],
                 _output: torch.Tensor,
             ):
-                # Assume that first argument is the input
-                inp = args[0].cpu().detach().squeeze()
+                if not isinstance(args, Tuple):
+                    print(f"GOT unexpected args {args}")
+                inp = args[0] if isinstance(args, Tuple) else args
 
                 self._smooth_activation_means[smooth_name] = _accumulate_mean(
-                    inp,
+                    inp.cpu().detach().squeeze(),
                     self._smooth_activation_means.get(smooth_name, None),
                 )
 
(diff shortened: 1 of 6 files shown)


src/llmcompressor/modifiers/awq/mappings.py
```diff
@@ -25,17 +25,33 @@ class AWQMapping:
...
+
+_moe_default_mappings = [
+    AWQMapping(
+        "re:.*input_layernorm$",
+        ["re:.*q_proj$", "re:.*k_proj$", "re:.*v_proj$"],
+    ),
+    AWQMapping("re:.*v_proj$", ["re:.*o_proj$"]),
+    AWQMapping(
+        "re:.*post_attention_layernorm$",
+        ["re:.*mlp.experts.*.gate_proj$", "re:.*mlp.experts.*.up_proj$"],
```
rahul-tuli, line 50: Great job getting to the bottom of this!

rahul-tuli (dismissed):

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -736,3 +747,30 @@ def _accumulate_mean(
...
+    Given a list of names, returns the lowest-scope common parent,
+    excluding parents of type ModuleList, which don't seem to play
+    nicely with hooks.
+    Returns name of parent and pointer to parent module
+
+    Implementation is a small alteration of os.path.commonprefix
+    https://docs.python.org/3/library/os.path.html#os.path.commonprefix
+    """
+    s1 = min(names)
+    s2 = max(names)
```
kylesayrs, line 771: What is the min/max of a list of strings? Shouldn't this be a for loop over elements?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -307,22 +303,37 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
-                        )
-                        if not balance_layer:
-                            continue
+        for mapping_idx, mapping in enumerate(self.mappings):
+            smooth_layers = get_layers(mapping.smooth_layer, model)
+            smooth_names = [
+                smooth_name
+                for smooth_name in smooth_layers
+                if (
+                    smooth_name not in self.ignore
```
kylesayrs: This doesn't account for regex or module names
```suggestion
                    find_name_or_class_matches(smooth_name, module, self.ignore)
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -307,22 +303,37 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
-                    for balance_suffix in mapping.balance_layers:
-                        # find the submodule that matches the activation layer
-                        balance_name, balance_layer = get_matching_layer(
-                            balance_suffix, layer_name, model
-                        )
-                        if not balance_layer:
-                            continue
+        for mapping_idx, mapping in enumerate(self.mappings):
+            smooth_layers = get_layers(mapping.smooth_layer, model)
+            smooth_names = [
```
kylesayrs, line 311: Can we call this "smooth_paths", since it's a path to the module in the state dict?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -307,22 +303,37 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+            num_skipped_mappings = 0
+            pbar = tqdm(smooth_names)
+            for smooth_name in pbar:
+                pbar.set_description(
+                    f"Resolving mapping {mapping_idx+1}/{len(self.mappings)}"
+                    f" ({num_skipped_mappings} skipped)"
+                )
+                smooth_layer = smooth_layers[smooth_name]
+
+                smooth_parent_name = ".".join(smooth_name.split(".")[:-1])
```
kylesayrs, line 328: Consider something like this
```suggestion
                smooth_parent_name = smooth_name.rsplit(".", 1)[0]
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -332,52 +343,43 @@ def _set_resolved_mappings(self, model: Module) -> None:
                             and ".o_proj" in balance_name
                             and (
                                 (
-                                    ".v_proj" in layer_name
+                                    ".v_proj" in smooth_name
```
```suggestion
                                    smooth_name.endswith(".v_proj")
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -401,11 +403,9 @@ def cache_smooth_activations_hook(
                 args: Tuple[torch.Tensor, ...],
                 _output: torch.Tensor,
             ):
-                # Assume that first argument is the input
-                inp = args[0].cpu().detach().squeeze()
-
                 self._smooth_activation_means[smooth_name] = _accumulate_mean(
-                    inp,
+                    # Assume that first argument is the input
+                    args[0].cpu().detach().squeeze(),
```
kylesayrs, line 410: Do we actually need to move to cpu device? Isn't it better to keep on the module device?

kylesayrs: Do we actually need to squeeze? I think we should make the `_accumulate_mean` function handle shape logic, if necessary

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -473,10 +475,15 @@ def _apply_smoothing(self, model: Module) -> None:
                 # [STEP 3]: Compute output of module
                 # could cache from hook, rather than recomputing here
                 fp16_output = self._run_samples(parent_module)
-                fp16_output = fp16_output.clip(
-                    torch.finfo(fp16_output.dtype).min,
-                    torch.finfo(fp16_output.dtype).max,
-                )
+                if fp16_output.shape[0] == 0:
```
kylesayrs: Slightly more robust
```suggestion
                if fp16_output.numel() == 0:
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -536,10 +543,14 @@ def smooth(module):
...
         with align_module_device(module):
+            outputs = [
+                module(**batch_kwargs)
+                for batch_kwargs in self._parent_args_cache[module]
+            ]
             return torch.cat(
                 [
-                    module(**batch_kwargs)[0]
-                    for batch_kwargs in self._parent_args_cache[module]
+                    output[0] if isinstance(output, Tuple) else output
```
kylesayrs, line 555: Add a note about assuming that the first output is the tensor

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -736,3 +747,30 @@ def _accumulate_mean(
...
+    for i, c in enumerate(s1):
+        if c != s2[i]:
+            parent_name = s1[:i].rstrip(".")
+            break
+
+    while True:
+        if parent_name == "":
+            return "", module
+        parent = get_layer_by_name(parent_name, module)
+        if not isinstance(parent, torch.nn.ModuleList):
```
kylesayrs, line 782: Why do you think ModuleLists don't work with hooks?

src/llmcompressor/utils/pytorch/module.py
```diff
@@ -344,20 +335,13 @@ def get_no_split_params(model: PreTrainedModel) -> Union[str, List[str]]:
     return no_split_modules
 
 
-def get_parent_by_name(layer_name: str, model: Module) -> Tuple[str, Module]:
+# https://discuss.pytorch.org/t/how-to-access-to-a-layer-by-module-name/83797/8
+def get_layer_by_name(layer_name: str, module: Module) -> Module:
```
kylesayrs, line 339: huggingface uses
```
from operator import attrgetter

attrgetter(layer_path)(module)
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -736,3 +747,30 @@ def _accumulate_mean(
     new_count = prev_count + num_added
 
     return (prev_sum + sum_added) / new_count, new_count
+
+
+def get_lowest_common_parent(names: List[str], module: Module) -> Tuple[str, Module]:
```
kylesayrs, line 755: I think a lot of this logic might be handled by os.path.commonprefix

kylesayrs (approved):
Many nits, otherwise looks good

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: min will grab the first name in alphabetical order, max will grab last name in alphabetical order. when finding common prefix, those are the only values we need. Pulled from https://docs.python.org/3/library/os.path.html#os.path.commonprefix

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: cool, updated

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I have been using the convention that "name" means the full name from top-level all the way down, and prefix or suffix can be composed to create `name = f"{prefix}.{suffix}"`. 

This is the [same convention used in torch](https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.named_modules). `model.named_modules()` will return a generator with names like `model.layers.0.mlp.experts.9.down_proj`.

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I am following convention i've seen elsewhere [in llm-compressor](https://github.com/search?q=repo%3Avllm-project%2Fllm-compressor%20.split(%22.%22)%5B%3A-1%5D&type=code) as well as [in vllm](https://github.com/vllm-project/vllm/blob/f20f9f063b8a61a96f055470c03dd26de8a2c19e/vllm/prompt_adapter/models.py#L219). Seems like six of one, half a dozen of the other?

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: updated for v_proj, o_proj, qkv_proj checks

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: updated

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I think it's because the forward method isn't ever directly called on the experts ModuleList. Certain experts are selected, and only the forward pass of those is called. Example for Qwen 3 MoE -- https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_moe/modeling_qwen3_moe.py#L235-L251

Will update docstring

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: It is, i call it out in the docstring. I have to remove a lot of the logic specific to os.path, so I can't use it directly

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: updated

src/llmcompressor/utils/pytorch/module.py
brian-dellabetta: updated

src/llmcompressor/modifiers/awq/base.py
kylesayrs: 100% makes ense

rahul-tuli (approved):

Chao-Xue:
https://github.com/vllm-project/llm-compressor/blob/ceffa644072b1d440df3d99b0f98f6416a05bf2f/examples/awq/qwen3_moe_example.py#L55~L56
It looks like `re:.*mlp.shared_expert_gate$` is included in the ignore list, but Qwen3-MoE doesn’t seem to have a shared_expert_gate module. Probably not an issue tho

brian-dellabetta:
@Chao-Xue , yes that is for other qwen MoE architectures where there is a shared expert that we ignore. not all qwen MoE models have that though

PR #1445, opened 2025-05-17
oneshot entrypoint update

### LINKED ISSUE #1443: Update `oneshot` to use an explicit keyword argument instead of using `**kwargs`
Summary
- The current `oneshot` entrypoint lists all the arguments under `**kwargs`
- This can be seen here: https://github.com/vllm-project/llm-compressor/blob/2db8fab932188b11585c087e73d7812c9febf439/src/llmcompressor/entrypoints/oneshot.py#L85
- These arguments are composed of recipe args, dataset args and model args which get parsed here: https://github.com/vllm-project/llm-compressor/blob/2db8fab932188b11585c087e73d7812c9febf439/src/llmcompressor/entrypoints/oneshot.py#L103

- We want to update the function definition to explicitly write out all the relevant oneshot argument, so that there is more clarity on exactly what the inputs are when running `oneshot` instead of hiding everything under `kwargs`
- End feature should include a docstring listing all the arguments as well

ved1beta:
SUMMARY:
Updated the oneshot function to use explicit keyword arguments instead of relying solely on **kwargs.

TEST PLAN:
Changes were tested using mock tests that verify:
Correct handling of explicit arguments
Backward compatibility with kwargs
Proper behavior with mixed explicit args and kwargs
Verification that kwargs correctly override explicit parameters when duplicates exist
fixes 
#1443

TODO:
- [x] run nightly tests on branch -- https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15710803245 ✅


src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -174,8 +174,155 @@ def apply_recipe_modifiers(
...
+        trust_remote_code_data (bool): Whether to allow for datasets defined on the Hub using
+            a dataset script.
+        
+        # Output arguments
+        output_dir (Optional[str]): Path to save the output model after calibration.
+    
+    Returns:
+        PreTrainedModel: The calibrated model
+    """
+    params = {
```
kylesayrs, line 277: In the future we could use something like this to avoid duplicating information
```python3
from typing import Tuple, List, Any, Dict
import inspect

def collect_function_arguments() -> Tuple[List[Any], Dict[str, Any]]:
    caller_frame = inspect.currentframe().f_back
    args_info = inspect.getargvalues(caller_frame)
    args = []
    kwargs = {}
    for key, value in args_info.locals.items():
        if key in args_info.args:
            args.append(value)
        elif key == args_info.varargs:
            args.extend(value)
        elif key == args_info.keywords:
            kwargs.update(value)
        else:
            assert False
        
    return args, kwargs
    
def oneshot(...):
    args, kwargs = collect_function_arguments()
    one_shot = Oneshot(*args, **kwargs)
```

kylesayrs (dismissed):
This looks great! There are a few arguments missing from the signature like `dvc_data_repository`, but this is handled and by **kwargs and those arguments don't need to be explicit.

This will go a long way towards making LLM Compressor more user friendly :)

kylesayrs:
@ved1beta Could you please `make style; make quality`?

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -174,8 +174,160 @@ def apply_recipe_modifiers(
         session.finalize(**session_kwargs)
 
 
-def oneshot(**kwargs) -> PreTrainedModel:
-    one_shot = Oneshot(**kwargs)
+def oneshot(
+    # Required model parameters
+    model: str,
```
brian-dellabetta, line 179: Users can also pass in a pre-loaded model
```suggestion
    model: Union[str, PreTrainedModel],
```

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -174,8 +174,160 @@ def apply_recipe_modifiers(
...
+    # Optional model parameters
+    distill_teacher: Optional[str] = None,
+    config_name: Optional[str] = None,
+    tokenizer: Optional[str] = None,
+    processor: Optional[str] = None,
```
brian-dellabetta, line 184: These can also be the loaded up classes. I think these are the best base classes to use? cc @kylesayrs 
```suggestion
    tokenizer: Optional[Union[str, transformers.PreTrainedTokenizerBase]] = None,
    processor: Optional[Union[str, transformers.ProcessorMixin]] = None,
```

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -174,8 +174,160 @@ def apply_recipe_modifiers(
...
+    cache_dir: Optional[str] = None,
+    use_auth_token: bool = False,
+    precision: str = "auto",
+    tie_word_embeddings: bool = False,
+    trust_remote_code_model: bool = False,
+    save_compressed: bool = True,
+    oneshot_device: str = "cuda:0",
+    model_revision: str = "main",
+    # Recipe parameters
+    recipe: Optional[str] = None,
```
brian-dellabetta, line 194: This can also be a list of Modifiers

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -174,8 +174,160 @@ def apply_recipe_modifiers(
...
+    recipe_args: Optional[List[str]] = None,
+    clear_sparse_session: bool = False,
+    stage: Optional[str] = None,
+    # Dataset parameters
+    dataset: Optional[str] = None,
```
brian-dellabetta, line 199: this can be an actual loaded dataset

brian-dellabetta (changes requested):
thanks @ved1beta for taking a shot at this! I think some of the type hints need to be updated to account for actual classes being passed in other than string representations, but otherwise this is looking great!

brian-dellabetta (commented):
Codereview updates

brian-dellabetta:
Once #1493 lands and we rebase, test checks should pass

src/llmcompressor/entrypoints/oneshot.py
brian-dellabetta: I updated it to just pass them into `OneShot(...` directly. We could probably remove the OneShot class as well. lmk what you think!

brian-dellabetta (approved):

brian-dellabetta (approved):

brian-dellabetta (dismissed):

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -191,8 +194,157 @@ def apply_recipe_modifiers(
...
+    precision: str = "auto",
+    tie_word_embeddings: bool = False,
+    trust_remote_code_model: bool = False,
+    save_compressed: bool = True,
+    oneshot_device: str = "cuda:0",
+    model_revision: str = "main",
+    # Recipe arguments
+    recipe: Optional[Union[str, List[str]]] = None,
+    recipe_args: Optional[List[str]] = None,
+    clear_sparse_session: bool = False,
```
dsikka, line 216: Can we verify these are all used still?
I didn't think clear_sparse_session was still required for example

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -191,8 +194,157 @@ def apply_recipe_modifiers(
...
+    dataset: Optional[Union[str, "Dataset", "DatasetDict"]] = None,
+    dataset_config_name: Optional[str] = None,
+    dataset_path: Optional[str] = None,
+    num_calibration_samples: int = 512,
+    shuffle_calibration_samples: bool = True,
+    max_seq_length: int = 384,
+    pad_to_max_length: bool = True,
+    text_column: str = "text",
+    concatenate_data: bool = False,
+    streaming: bool = False,
```
dsikka, line 228: I dont think this is used either

dsikka (changes requested):

src/llmcompressor/entrypoints/oneshot.py
ved1beta: clear_sparse_session: After checking the codebase, I found this parameter was defined in the RecipeArguments class but not used meaningfully anywhere in the oneshot operation flow

Dataset parameters (pad_to_max_length, text_column, concatenate_data, streaming): These are handled at a lower level in the dataset loading code, but they don't need to be exposed directly in the oneshot API. They're still accessible through the DatasetArguments when needed internally.

brian-dellabetta: Thanks @ved1beta for taking a look. We discussed that we will prune the args in a follow-up PR, since we'll have to modify the `DatasetArguments` etc. as you pointed out. That way this PR will be scoped to just the entrypoint, I reverted the changes from your last commit

src/llmcompressor/entrypoints/oneshot.py
brian-dellabetta: To be pruned in follow-up PR

brian-dellabetta (approved):

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -191,8 +194,157 @@ def apply_recipe_modifiers(
...
+    overwrite_cache: bool = False,
+    preprocessing_num_workers: Optional[int] = None,
+    min_tokens_per_module: Optional[float] = None,
+    trust_remote_code_data: bool = False,
+    # Miscellaneous arguments
+    output_dir: Optional[str] = None,
+    log_dir: Optional[str] = "sparse_logs",
+    **kwargs,
+) -> PreTrainedModel:
+    """
```
rahul-tuli, line 238: nit: The docstring is in a different style than most of our codebase, should we update it for consistency? Not a blocker at all.

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -191,8 +194,157 @@ def apply_recipe_modifiers(
...
+    save_compressed: bool = True,
+    oneshot_device: str = "cuda:0",
+    model_revision: str = "main",
+    # Recipe arguments
+    recipe: Optional[Union[str, List[str]]] = None,
+    recipe_args: Optional[List[str]] = None,
+    clear_sparse_session: bool = False,
+    stage: Optional[str] = None,
+    # Dataset arguments
+    dataset: Optional[Union[str, "Dataset", "DatasetDict"]] = None,
```
rahul-tuli, line 219: nit: Why not import Dataset and DatasetDict?

brian-dellabetta: we do `if TYPE_CHECKING` pretty liberally throughout llm-compressor, in this case it avoids another import that might not ever be needed.

src/llmcompressor/entrypoints/oneshot.py
brian-dellabetta: thanks for calling out, will make it consistent

brian-dellabetta: ok updated

kylesayrs (dismissed):
This is a great change to make

brian-dellabetta (dismissed):

rahul-tuli (dismissed):

brian-dellabetta (dismissed):
Codereview updates

kylesayrs (dismissed):

brian-dellabetta (approved):

PR #1419, opened 2025-05-08
[WIP] Use model compression pathways

kylesayrs:

PR #1449, opened 2025-05-19
[Tests] Use proper offloading utils in `test_compress_tensor_utils`

kylesayrs:
## Purpose ##
* Prerequisite for #1419
  * This PR disables getting the offloaded state dict unless necessary (sparsity statistics). However, the utility function `cpu_offload` only works if the offloaded state dict is retrieved. Let's replace this with `dispatch_model`, which is the actual function used by `PretrainedModel`, not `cpu_offload`

## Changes ##
* Rename `device_map` to `device`
* Use `dispatch_model` rather than `cpu_offload`
* Use `align_module_device` and `update_offload_parameter` utilities
  * This change is necessary because, after these changes, some of these test models no longer have offloaded state dicts (which is the way it should always have been)

### DIFF
--- a/tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
+++ b/tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
@@ -4,7 +4,7 @@ import shutil
 
 import pytest
 import torch
-from accelerate import cpu_offload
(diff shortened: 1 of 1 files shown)


brian-dellabetta (approved):
💚

dsikka (approved):

PR #1450, opened 2025-05-19
[Observer] Optimize mse observer 

shanjiaz:
SUMMARY:
The calculate_mse_min_max function previously performed a full grid search across a 0.8 × 100 = 80-point space. After discussing with Alex and Eldar last week, we reduced max_shrink to 0.2 to improve performance without sacrificing accuracy.

Additionally, implemented an early stopping mechanism. The function now tracks the best quantization error seen so far and stops if no improvement is observed over 5 consecutive steps (patience = 5).

maxshrink variable is now configurable in recipe file, and patience(for early stop) can be passed in as well. 

TEST PLAN:
All lm_eval tests were run. No regressions in accuracy were observed. Performance improved significantly after maxshrink is updated.
**There's a 3-7 mins slow down per test switching from MinMax to MSE observer.**  

USAGE:
Tested the recipe by adding:
```yaml
observer: "mse"
observer_kwargs:
    maxshrink: 0.3
```

More details can be found in this [notion page](https://www.notion.so/Accuracy-test-1d930c7e73f3803bb057fd17d6d45302?pvs=4)
Raw timing data are stored [here](https://drive.google.com/drive/folders/1I69QNGKxLJJZ06k9jSw0f0BRhPchV_nt?usp=drive_link)


rahul-tuli (approved):
LGTM!

dsikka (changes requested):
Could we make sure the user can adjsut max_shrink?
This should be possible by using the obsever_kwargs in the quantization_args i believe

dsikka (commented):
What was the average change in runtime?

brian-dellabetta (approved):
Thanks for the nice detailed results! If it is consistently showing improvement over min-max without a big time hit, I think it looks good without having to try to vectorize or jit-compile the grid search

shanjiaz:
> Could we make sure the user can adjsut max_shrink? This should be possible by using the obsever_kwargs in the quantization_args i believe

Yes! Will add that today 🫡

src/llmcompressor/observers/mse.py
```diff
@@ -17,22 +17,17 @@ class MovingAverageMSEObserver(Observer):
...
-        averaging_constant: float = 0.01,
-        grid: float = 100.0,
-        maxshrink: float = 0.80,
-        norm: float = 2.4,
-    ):
+    def __init__(self, quantization_args: QuantizationArgs):
         super().__init__(quantization_args=quantization_args)
 
+        kwargs = quantization_args.observer_kwargs or {}
+        self.averaging_constant = kwargs.get("averaging_constant", 0.01)
```
kylesayrs, line 24: Sorry, why are we moving these kwargs out of the function definition?

src/llmcompressor/observers/mse.py
```diff
@@ -61,6 +56,11 @@ def calculate_mse_min_max(
         )
         min_val = torch.ones_like(absolute_min_val)
         max_val = torch.zeros_like(absolute_max_val)
+
+        # Early stopping params
+        patience = 5
```
kylesayrs, line 61: nit: You can save a variable by decrementing patience
```
patience = 5

if no_improvement:
    patience -= 1
    if patience <= 0:
        break
```

kylesayrs (changes requested):

src/llmcompressor/observers/mse.py
shanjiaz: I think maxshrink will be passed in as part of the `observer_kwargs`. Just adding this unpacking logic to make sure we can get the right maxshrink parameter when users pass them in.

src/llmcompressor/observers/mse.py
shanjiaz: sure! good point.

src/llmcompressor/observers/mse.py
shanjiaz: I see what you meant now!

kylesayrs (approved):

dsikka (commented):
Have we verified the observer_kwargs/are propagated to the mse obsever as expected?

Yes! Let me find the run result

brian-dellabetta (approved):

shanjiaz:
Verified with this following recipe:

```yaml
quant_stage:
  quant_modifiers:
    SmoothQuantModifier:
      smoothing_strength: 0.8
    GPTQModifier:
      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:vision_model.*"]
      config_groups:
        group_0:
          weights: {num_bits: 8, type: int, symmetric: true, strategy: channel, observer: "mse", observer_kwargs: {maxshrink: 0.3}}
          input_activations: {num_bits: 8, type: int, symmetric: true, strategy: token, dynamic: true}
          targets: [Linear]
```
And set a checkpoint in lifecycle to verify:

GPTQModifier(config_groups={'group_0': QuantizationScheme(targets=['Linear'], weights=QuantizationArgs(num_bits=8, type='int', symmetric=True, group_size=None, strategy='channel', block_structure=None, dynamic=False, actorder=None, observer='mse', observer_kwargs={'maxshrink': 0.3})

dsikka (approved):
This looks good. 
I think in a follow-up, we should do two things:

1. I think we ideally want to make it really obvious what arguments can be configured by the observer. It is hidden right now that we can set the maxshrink
3. I think we should generally pass in `observer_kwargs` when the observer is first created and allow the user to configure all args, not just maxshrink

PR #1453, opened 2025-05-20
[GPTQ] Fix actorder resolution, add sentinel

kylesayrs:
## Purpose ##
* Fix false assumption that `actorder` field is of enum type
  * Despite the fact that actorder passes through a [field_validator](https://github.com/neuralmagic/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_args.py#L200), `QuantizationArgs` has the [use_enum_values](https://github.com/neuralmagic/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_args.py#L128) configuration set, meaning that enum values are converted to strings.
  * This was done in relation to [this fix](https://github.com/neuralmagic/sparseml/pull/2327)
* Remove conflict with recipes which manually specify activation ordering by using a sentinel value

## Follow ups ##
* https://github.com/vllm-project/llm-compressor/pull/1425

## Testing ##
* Ran llama3 example with manually specified `actorder=group`


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -109,7 +110,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
     sequential_targets: Union[str, List[str], None] = None
     block_size: int = 128
     dampening_frac: Optional[float] = 0.01
-    actorder: Optional[ActivationOrdering] = None
+    actorder: Optional[Union[ActivationOrdering, Sentinel]] = None
```
brian-dellabetta, line 113: Doesn't this need to be set the sentinel value in order for this to work? Looking at reference [here](https://github.com/pydantic/pydantic/discussions/9943#discussioncomment-10131374).

src/llmcompressor/sentinel.py
```diff
@@ -0,0 +1,53 @@
+import inspect
+
+from pydantic_core import core_schema
+
+_registry = {}
+
+
+class Sentinel:
+    """
+    Unique sentinel values. Implements https://peps.python.org/pep-0661/
```
brian-dellabetta, line 10: probably worth calling out the pydantic suggestion since this implementation does not use the `sentinel` package referenced in the PEP

https://github.com/pydantic/pydantic/discussions/9943#discussioncomment-10131374

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: I'm only type checking that it is an instance of the `Sentinel` class, not that it is the `Sentinel("STATIC")` singleton. I think type checking the class is good enough.

kylesayrs: That reference implementation is slightly different, since they use a single `MissingSentinel` class. I implemented a `Sentinel` value order to be a little more expressive (ie, I can a `Sentinel("STATIC")` value which hints at the fact that it will default to static activation ordering

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -131,23 +132,29 @@ def validate_sequential_update(cls, value: bool) -> bool:
     def resolve_quantization_config(self) -> QuantizationConfig:
         config = super().resolve_quantization_config()
 
-        # Resolve config with `self.actorder`
+        def resolve_actorder(existing):
+            # sentinel default only overrides if existing is None
+            if self.actorder == Sentinel("STATIC"):
```
dsikka, line 137: Do you mind expanding on the motivation in using this class instead of just checking self.actorder == "STATIC" (seems to be the only usecase?)?

I think I get it but not 100%.

dsikka (changes requested):
Quick question before approving

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: Any value could be used as a sentinel. The only purpose of instantiating from the `Sentinel` class is to make it obvious to a reader that the value is being used as a sentinel

For example, using a string might imply to a user that string names of actorder types are are valid values. `Sentinel`s are also impossible to accidentally pass in as a caller.

src/llmcompressor/modifiers/quantization/gptq/base.py
brian-dellabetta: so the user will be doing
```python
GPTQModifier(..., actorder = Sentinel("STATIC"))
```
?
I'm confused where it will get used

kylesayrs: The whole point of a sentinel is that it is not (and cannot) be used by a user. It is used to differentiate between a user-passed `"static"` value and a defaulted `Sentinel("STATIC")` value

brian-dellabetta: doesn't that mean the default values here should be `Sentinel("STATIC")` and not `None`?
I don't want to hold this up if it's working as expected, so i will approve

brian-dellabetta (approved):

dsikka (approved):

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: Yes, that's what https://github.com/vllm-project/llm-compressor/pull/1425 does

I separate them into two PRs because the actual swap to static activation ordering for most recipes might have performance implications which I'd like to separate from the actual bug fix

brian-dellabetta: ah! i missed that. ok sounds good

PR #1454, opened 2025-05-20
Disable kernels during calibration (and tracing)

kylesayrs:
## Purpose ##
* Guarantee that module hooks trigger by disabling kernel acceleration

## Background ##
* As of `transformers>=4.52`, model forward functions may be overwritten with kernels
* Kernel execution can be disabled using the config ([source](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/hub_kernels.py#L84))
* It seems that HF wants to continue enabling regular execution through this disabling feature. This gives us some faith that they will not [override the cls init function](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/hub_kernels.py#L78) in a way that inhibits execution with the standard forward definition
* This only affects users who have the hf `kernels` library installed

## Changes ##
* Implement `disable_hf_kernels` context and add to `calibration_forward_context`
* Remove parenthesis around `calibration_forward_context` with in order to support python3.9
* Apply style to tests


src/llmcompressor/utils/helpers.py
```diff
@@ -1033,6 +1034,16 @@ def eval_context(module: torch.nn.Module):
         module.train(restore_value)
 
 
+@contextlib.contextmanager
+def disable_hf_kernels(model: PreTrainedModel):
```
dsikka, line 1038: Can you leave a quick comment about the relevant transformers version?

tests/examples/test_quantization_2of4_sparse_w4a16.py
```diff
@@ -36,9 +36,9 @@ def test_doc_example_command(self, example_dir: str, tmp_path: Path):
         readme = ReadMe(readme_path)
 
         command = readme.get_code_block_content(position=2, lang="shell")
-        assert command.startswith("python"), (
-            "Expected shell command to start with 'python'"
-        )
+        assert command.startswith(
```
dsikka, line 39: is this file just a change to style?

dsikka (approved):

tests/examples/test_quantization_2of4_sparse_w4a16.py
kylesayrs: Yes

brian-dellabetta: I noticed this too on my branch

brian-dellabetta (approved):

brian-dellabetta (approved):

dsikka (approved):

PR #1458, opened 2025-05-21
Pin Transformers Version

dsikka:
SUMMARY:
- Temporarily pin the transformers version as tracing tests are currently failing
- Will create a ticket to track enabling the latest version

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,7 +116,7 @@ setup(
         "requests>=2.0.0",
         "tqdm>=4.0.0",
         "torch>=1.7.0",
-        "transformers>4.0,<5.0",
+        "transformers>4.0,<4.52.1",
         "datasets",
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):
Looks good to me!

setup.py
```diff
@@ -116,7 +116,7 @@ def localversion_func(version: ScmVersion) -> str:
         "requests>=2.0.0",
         "tqdm>=4.0.0",
         "torch>=1.7.0",
-        "transformers>4.0,<5.0",
+        "transformers>4.0,<4.52.1",
```
brian-dellabetta, line 119: this needs to be `<4.52.0`, right?

dsikka: 4.42.0 is a yanked version so it should resolve to 4.51.x 
I can make it hard requirement

Issue #1457, opened 2025-05-21
Tracing tests failing with latest transformers>4.52.0


brian-dellabetta:
**Describe the bug**
transformers tests are failing with the latest version of `transformers>4.52.0`, released May 20th. For now, we will pin transformers<4.52.0 in the setup.py, and resolve in a future PR.

**Expected behavior**
Tests to pass

**Errors**
See GHA error logs for failing tests [here](https://github.com/vllm-project/llm-compressor/actions/runs/15146183391/job/42582308996?pr=1451)

PR #1459, opened 2025-05-21
Pin transformers to <4.52.0

brian-dellabetta:
SUMMARY:
See #1457 for error report. This will pin transformers version in setup.py until issue is resolved

TEST PLAN:
no new src code, tests should pass now with pin


rahul-tuli (commented):
LGTM!

PR #1460, opened 2025-05-21
Removed RecipeTuple & RecipeContainer class

shanjiaz:
SUMMARY:
Removed RecipeTuple class and downstream dependencies. Updated related docstrings. 

TEST PLAN:
Ran tests from tests/llmcompressor and passed
Fixed test_obcq_owl. It was creating event without initializing a session, which shouldn't be allowed.

CONCERNS:
I did make one assumption that there will be only one `target_stage` each time a session gets initialized. Is it true?

### DIFF
--- a/tests/llmcompressor/transformers/obcq/test_obcq_owl.py
+++ b/tests/llmcompressor/transformers/obcq/test_obcq_owl.py
@@ -3,9 +3,9 @@ import torch
 from datasets import Dataset
 from transformers import AutoModelForCausalLM
 
+from llmcompressor.core.session_functions import create_session
 from llmcompressor.datasets import format_calibration_data
 from llmcompressor.modifiers.obcq import SparseGPTModifier
-from llmcompressor.core.session_functions import create_session
 from llmcompressor.utils.pytorch.module import get_layers
 
 


brian-dellabetta (approved):
🧹 🧹

kylesayrs (changes requested):
Consider removing logic for handling multiple recipes

src/llmcompressor/core/lifecycle.py
```diff
@@ -91,13 +91,16 @@ def initialize(
         :return: List of data returned from initialization of modifiers
         :rtype: List[Any]
         """
+
         self.state.update(**kwargs)
         if self.initialized_:  # TODO: do not initialize twice
             return
-
+        import pdb; pdb.set_trace()
```
brian-dellabetta, line 98: probably leftover from debugging?
```suggestion
```

tests/lmeval/test_lmeval.py
```diff
@@ -67,9 +67,11 @@ class TestLMEval:
     def set_up(self, test_data_file: str):
         eval_config = yaml.safe_load(Path(test_data_file).read_text(encoding="utf-8"))
 
+        '''
         if os.environ.get("CADENCE", "commit") != eval_config.get("cadence"):
             pytest.skip("Skipping test; cadence mismatch")
-
+        '''
+
```
brian-dellabetta, line 74: probably left over from debugging

brian-dellabetta (changes requested):
looking good! just a couple things leftover from debugging that need to be reverted

src/llmcompressor/core/lifecycle.py
```diff
@@ -91,13 +86,14 @@ def initialize(
...
         self.state.update(**kwargs)
         if self.initialized_:  # TODO: do not initialize twice
             return
-
-        logger.debug("Initializing compression lifecycle")
-        self.recipe_container.append(recipe, recipe_stage, recipe_args)
-        self.modifiers = self.recipe_container.get_modifiers()
+        self.recipe = Recipe.simplify_recipe(recipe=recipe,
+                                             target_stage=recipe_stage,
+                                             override_args=recipe_args)
```
brian-dellabetta, line 95: looks like this needs a `make style` as well

shanjiaz: Totally!

brian-dellabetta (approved):
🥇

src/llmcompressor/recipe/recipe.py
```diff
@@ -164,99 +161,56 @@ def create_instance(
 
     @staticmethod
     def simplify_recipe(
-        recipe: Union[str, "Recipe", "RecipeTuple"], shift: Optional[int] = None
+        recipe: Optional["RecipeInput"] = None,
```
kylesayrs, line 164: Why are we accepting `None` as a valid recipe? Does/should this ever come up?

src/llmcompressor/recipe/recipe.py
```diff
@@ -164,99 +161,56 @@ def create_instance(
...
 
-        :param recipe: The Recipe or RecipeTuple instance to simplify
+        :param recipe: The Recipe instance to simplify
+        :param target_stages: The stages to target when simplifying the recipe
+        :param override_args: The arguments used to override existing recipe args
         :return: The simplified Recipe instance
         """
-        if isinstance(recipe, str):
+        if recipe is None or (isinstance(recipe, list) and len(recipe) == 0):
+            return Recipe()
```
kylesayrs, line 178: Ideally, we shouldn't be handling this case

kylesayrs (commented):
Awesome work

src/llmcompressor/recipe/recipe.py
shanjiaz: You meant the "recipe is None" case? I can try to find the test I ran into. I can raise it in `lifecycle` but if you look at the lifecycle code I think recipe is optional as well. I kept the way it is.

src/llmcompressor/recipe/recipe.py
kylesayrs: That's fine for now, if it's only for testing.

kylesayrs (approved):
In the future, let's fold the logic of `simplify_recipe` into the Recipe creation itself

Good point! Will do.

src/llmcompressor/core/session_functions.py
```diff
@@ -75,10 +75,6 @@ def event(cls, event_type: EventType, **kwargs) -> ModifiedState:
                 f"Use the corresponding method instead."
             )
 
-        # skip event callbacks if no recipe was provided
```
kylesayrs, line 78: Yay

src/llmcompressor/recipe/recipe.py
kylesayrs, line 164: @shanjiaz Do you think we can get rid of this now that the test has been updated?

brian-dellabetta (approved):
very nice!

src/llmcompressor/recipe/recipe.py
shanjiaz: I don't think this is the only place. Especially since `lifecycle` can take in None recipes [here](https://github.com/vllm-project/llm-compressor/blob/94a3e5356f1c5bf3d836d6522c2a68391761cc63/src/llmcompressor/core/lifecycle.py#L82). I can dig into this more but I think it's out of the scope of this PR.

kylesayrs (approved):

PR #1467, opened 2025-05-22
Remove duplicate model id var from awq example recipe

AndrewMead10:
SUMMARY:
I noticed the MODELID var is declared twice in the AWQ example recipe, so i removed the second instance of it

TEST PLAN:
N/A


AndrewMead10:
Cant edit the labels for some reason, but this is ready @dsikka @kylesayrs

kylesayrs (approved):
Thanks for the catch!

dsikka (approved):

PR #1468, opened 2025-05-22
[Tests] GPTQ Actorder Resolution Tests

kylesayrs:
## Purpose ##
* Clarify behavior introduced by #1453 

## Changes ##
* Remove `TestQuantizationRegistered` test, which is less necessary that it was before
* Add `test_actorder_resolution`, which demonstrates and regression tests behavior related to resolving the `actorder` argument


shanjiaz (approved):
Looks great! Would appreciate if you could add a short docstring explaining what the test is about 🫡

PR #1471, opened 2025-05-23
Set `show_progress` to True

dsikka:
Summary
- Until we're using compress_model, we need to show compression progress 
- Without it, it seems like nothing is happening


kylesayrs (dismissed):
Thanks

brian-dellabetta (dismissed):

brian-dellabetta (approved):

Issue #1475, opened 2025-05-27
Block-wise Quantization Not supported


brian-dellabetta:
_Originally reported in [#1464](https://github.com/vllm-project/llm-compressor/issues/1464)_

The `Observer` class and compressed-tensor do not currently have an implementation for BLOCK-wise quantization (see code starting [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/observers/base.py#L90)) 

Here is a minimal reproducible example:

```python
from transformers import AutoModelForCausalLM

from llmcompressor import oneshot
from llmcompressor.modifiers.quantization import QuantizationModifier
from compressed_tensors.quantization import (
    QuantizationArgs,
    QuantizationConfig,
    QuantizationScheme,
    QuantizationStrategy,
    QuantizationType,
)

MODEL_ID = "meta-llama/Llama-3.2-3B-Instruct"

# define a llmcompressor recipe for FP8 W8A8 quantization
# since the MoE gate layers are sensitive to quantization, we add them to the ignore
# list so they remain at full precision
recipe = [
    QuantizationModifier(
        ignore=["lm_head", "re:.*mlp.gate$"],
        config_groups={
            "group_0": QuantizationScheme(
                targets=["Linear"],
                weights=QuantizationArgs(
                    num_bits=4,
                    type=QuantizationType.INT,
                    dynamic=False,
                    symmetric=False,
                    strategy=QuantizationStrategy.BLOCK,
                    # group_size=128,
                    block_structure="128x128",
                ),
            )
        },
    )
]

SAVE_DIR = MODEL_ID + "-W4A16-BLOCK128"

model = AutoModelForCausalLM.from_pretrained(
    MODEL_ID, device_map="auto", torch_dtype="bfloat16", trust_remote_code=True
)


oneshot(
    model=model,
    recipe=recipe,
    save_compressed=True,
    output_dir=SAVE_DIR,
)
```

We will probably want to support block-wise quantization moving forward, but this is probably low priority. As far as I know, vllm doesn't have optimized kernels to run block-wise quantized models, unlike group-wise quantization. I will create a PR to reference this issue and raise an error if block-wise is used, to try group-wise instead

ved1beta:
working on this

robertgshaw2-redhat:
note: we do have block quantized kernels for fp8 in vllm (since this is the scheme deepseek uses to release its models)

shuxiaobo:
[...]

Issue #1464, opened 2025-05-22
quant a Moe model with block wise which is 128x128 raise the error


shuxiaobo:
**Describe the bug**
quant a Moe model with block wise which is 128x128 raise the error
recipe: 
```
recipe = [QuantizationModifier(
            ignore=["lm_head", "re:.*mlp.gate$"],
            config_groups={
                "group_0": QuantizationScheme(
                    targets=["Linear"],
                    weights=QuantizationArgs(
                        num_bits=8,
                        type=QuantizationType.FLOAT,
                        dynamic=False,
                        symmetric=True,
                        strategy=QuantizationStrategy.BLOCK,
                        #group_size=128,
                        block_structure="128x128"
                    ),
                    input_activations=QuantizationArgs(
                        num_bits=8,
                        type=QuantizationType.FLOAT,
                        strategy=QuantizationStrategy.TOKEN,
                        symmetric=True,
                        dynamic=True
                    )
                )
            }
        )]
```
error: 
```
025-05-21T20:35:02.631325+0800 | from_modifiers | INFO - Creating recipe from modifiers
2025-05-21T20:35:07.387697+0800 | _check_calibration_data | INFO - Skipping QuantizationModifier calibration, it is not required for the provided quantization config.
Traceback (most recent call last):
  File "compressor_block.py", line 108, in <module>
    oneshot(
  File "/opt/conda/envs/python3.10/lib/python3.10/site-packages/llmcompressor/entrypoints/oneshot.py", line 179, in oneshot
    one_shot()
  File "/opt/conda/envs/python3.10/lib/python3.10/site-packages/llmcompressor/entrypoints/oneshot.py", line 131, in __call__
    self.apply_recipe_modifiers(
  File "/opt/conda/envs/python3.10/lib/python3.10/site-packages/llmcompressor/entrypoints/oneshot.py", line 173, in apply_recipe_modifiers
    session.initialize(**session_kwargs)
  File "/opt/conda/envs/python3.10/lib/python3.10/site-packages/llmcompressor/core/session.py", line 118, in initialize
    mod_data = self._lifecycle.initialize(
  File "/opt/conda/envs/python3.10/lib/python3.10/site-packages/llmcompressor/core/lifecycle.py", line 105, in initialize
    data = mod.initialize(state=self.state, **kwargs)
  File "/opt/conda/envs/python3.10/lib/python3.10/site-packages/llmcompressor/modifiers/stage.py", line 83, in initialize
[...]

PR #1476, opened 2025-05-27
raise error if block quantization is used, as it is not yet supported

brian-dellabetta:
SUMMARY:
More info at #1464 and #1475, for now just raising an error if user is trying to block-quantize. This provides a more useful error than downstream error user is reporting in #1464 

TEST PLAN:
n/a


shanjiaz (approved):
Looks good to me! Thanks for adding the warning : )

kylesayrs (approved):

PR #1477, opened 2025-05-27
loosen lmeval assertions to upper or lower bound

brian-dellabetta:
SUMMARY:
`lm_eval` end-to-end tests are occasionally failing when the actual value is higher than the expected value, when we really only care about if performance has regressed (example run [here](https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15232878962/job/42842981702)). 

This PR loosens the check to only assert 
- actual value > expected value - error tolerance if higher score is better (generally the case)
- actual value < expected value + error tolerance if lower score is better (in case we have PPL checks or add in future)

TEST PLAN:
- [x] Rerun weekly lm-eval tests before merging this in -- https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15281691638

### DIFF
--- a/tests/lmeval/test_lmeval.py
+++ b/tests/lmeval/test_lmeval.py
@@ -162,7 +162,7 @@ class TestLMEval:
                 continue
             actual_val = metrics.get(metric_key)
             higher_is_better = results["higher_is_better"][self.lmeval.task].get(
-                metric_key.split[","][0], True
+                metric_key.split(",")[0], True
             )
             stderr_key = metric_key.replace(",", "_stderr,")
             std_err = self.lmeval.metrics.get(stderr_key)


tests/lmeval/test_lmeval.py
```diff
@@ -154,29 +154,46 @@ def _run_lm_eval(self):
...
+                    f"relative tolerance ±5%, Got {actual_val}. "
+                    f"Higher is better: {higher_is_better}"
                 )
-                assert numpy.isclose(expected_val, actual_val, rtol=0.05)
+                # If higher is better, assert actual val >= expected val * (1 - stderr)
+                if higher_is_better:
+                    assert actual_val >= expected_val * (0.95)
+                # If higher is worse, assert actual val <= expected val * (1 + stderr)
+                else:
+                    assert actual_val <= expected_val * (1.05)
```
rahul-tuli, line 183: nit: What do you think about this pseudo-code?

```python
if std_err is None:
   ...logging...
   std_err = 0.05

lower_bound = expected_val * (1 - std_err)
upper_bound = expected_val * (1 + std_err)

if higher_is_better:
    assert actual_val >= lower_bound, f"{actual_val} < {lower_bound}"
else:
    assert actual_val <= upper_bound, f"{actual_val} > {upper_bound}"   
   
```

My thought being it makes the logic slightly easier to follow and we don't have to repeat the same condition twice?

rahul-tuli (commented):
The changes make sense to me, left a small nit!

rahul-tuli (approved):

tests/lmeval/test_lmeval.py
```diff
@@ -154,29 +154,46 @@ def _run_lm_eval(self):
...
+        metrics: dict = results["results"][self.lmeval.task]
         for metric_key, expected_val in self.lmeval.metrics.items():
             # stderr metrics are only used as absolute tolerance
             # checks for actual values
             if "stderr" in metric_key:
                 continue
             actual_val = metrics.get(metric_key)
-            # If stderr is provided, use it as absolute tolerance
-            # Otherwise, default to a 5% relative tolerance
+            higher_is_better = results["higher_is_better"][self.lmeval.task].get(
```
dsikka, line 164: where is `higher_is_better` being defined?

brian-dellabetta: it's part of the lmeval data model, it's how they populate the table with up arrows or down arrows

PR #1487, opened 2025-05-29
[NVFP4] Expand observers to calculate gparam, support NVFP4 Activations

dsikka:
# SUMMARY:
- Add NVFP4 Example
- Update compression condition to no longer be weight only
- Support NVFP4 Activations:
     - Update observers to also provide the option to calculate gparam (global_param), not just qparams
     - Update dynamic activation condition checks to consider DynamicType.LOCAL
    
# Testing
- All test cases pass

# Next Steps:
We now have the framework to also calculate the weight global scale in llmcompressor. Will remove it from compressed-tensors and add it here onc this lands.

### DIFF
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -1,5 +1,8 @@
+import os
+from datetime import datetime
 from typing import Optional
 
+from loguru import logger
 from torch.utils.data import DataLoader
 from transformers import PreTrainedModel
 
@@ -82,6 +85,7 @@ class Oneshot:
 
     def __init__(
         self,
+        log_dir: Optional[str] = "sparse_logs",
         **kwargs,
     ):
         """
@@ -100,6 +104,15 @@ class Oneshot:
         :param output_dir: Path to save the output model after carrying out oneshot
 
         """
+        # Set up logging
+        if log_dir:
+            os.makedirs(log_dir, exist_ok=True)
+            date_str = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+            logger.add(
+                f"{log_dir}/oneshot_{date_str}.log",
+                level="DEBUG",
+            )
+
         model_args, dataset_args, recipe_args, _, output_dir = parse_args(**kwargs)
 
         self.model_args = model_args
--- a/src/llmcompressor/modifiers/stage.py
+++ b/src/llmcompressor/modifiers/stage.py
@@ -67,7 +67,6 @@ class StageModifiers(ModifierInterface, BaseModel):
                 modifier.initialize(state, **kwargs)
             if accelerator:
                 accelerator.wait_for_everyone()
-        state.loggers.system.info(tag="stage", string="Modifiers initialized")
 
     def finalize(self, state: "State", **kwargs):
         """
@@ -88,7 +87,6 @@ class StageModifiers(ModifierInterface, BaseModel):
                 accelerator.wait_for_everyone()
 
         self.applied = True
-        state.loggers.system.info(tag="stage", string="Modifiers finalized")
 
     def update_event(self, state: "State", event: "Event", **kwargs):
         """
--- a/tests/llmcompressor/transformers/gptq/test_oneshot.py
+++ b/tests/llmcompressor/transformers/gptq/test_oneshot.py
@@ -65,7 +65,7 @@ class TestGPTQOneShotWithFullScheme(unittest.TestCase):
         import torch
 
         self.output = "./oneshot_output"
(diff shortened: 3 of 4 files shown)


src/llmcompressor/observers/base.py
```diff
@@ -73,11 +79,14 @@ def post_calculate_qparams(self) -> None:
         Run any logic specific to its observers after running calculate_qparams
         """
 
+    # TODO: use a different name?
```
brian-dellabetta, line 82: rather than having the return signature of `get_qparams` change based on these bool values, what if we just had two separate methods `def get_gparams` and `def get_qparams`?

brian-dellabetta (commented):
Examples looks great! One question on naming and signature of `get_qparams`

src/llmcompressor/observers/base.py
dsikka: yeah I refactor it out a bit to clean it up

src/llmcompressor/observers/base.py
```diff
@@ -52,12 +51,12 @@ def forward(
         :return: tuple of scale and zero point based on last observed value
         """
         self.record_observed_tokens(observed)
+        if should_calculate_gparam:
+            return self.get_gparam(observed=observed)
```
brian-dellabetta, line 55: `get_gparam` returns a single Tensor, but the return signature here is `Tuple[FloatTensor, IntTensor]`. Can this be set up in a way where function signatures do not change? Is the `calculate_gparam` method entirely separable from the qparams method?

src/llmcompressor/observers/base.py
```diff
@@ -74,19 +73,39 @@ def calculate_qparams(
         """
         raise NotImplementedError(f"{self.__class__} must implement calculate_qparams")
 
+    def calculate_gparam(
```
brian-dellabetta, line 76: should this be an `@abstractmethod`?

brian-dellabetta (commented):
comments on function signatures

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -53,7 +55,10 @@ def initialize_observer(
 
     quantization_args = getattr(quantization_scheme, arg_name, None)
     # dont need observers for dynamic
-    if quantization_args is not None and not quantization_args.dynamic:
+    if quantization_args is not None and quantization_args.dynamic in (
+        False,
+        DynamicType.LOCAL,
```
shanjiaz, line 60: Just curious, what is DynamicType.LOCAL?

src/llmcompressor/observers/base.py
dsikka: Yeah they're returning two separate sets of parameters but use the same underlying min/max tacking that the observers provide

dsikka: No but I think the pattern is to use the singular forward method for all observer calls.

brian-dellabetta: ok, at the very least we should update the signature of `forward` to `-> Union[FloatTensor, Tuple[FloatTensor, IntTensor]]`

dsikka: Yeah will update. I'm ok with separating out the functionality but that would mean moving away from the forward design

src/llmcompressor/modifiers/quantization/calibration.py
rahul-tuli: I feel it is unclear even from the link above what are local quantization parameters; It would be nice if we could add another line to the comment in compressed tensors what are local quantization parameters, not a blocker for this diff

src/llmcompressor/observers/helpers.py
```diff
@@ -20,3 +26,29 @@ def get_observer_token_count(module: torch.nn.Module) -> Counter:
...
+    dtype: Optional[torch.dtype] = torch.float32,
+):
+    """
+    Generate a global scale for an entire tensor (input_tensor).
+    Goal of the scale is to ensure that the quantization (local) scale
+    falls into the approproiate dtype range.
+
+    E.g. for NVFP4, group (local) scales are in dtype FP8. The global_scale
+    attempts to use the entire FP8 dtype range while mapping a per-group max
+    to the FP4 max.
```
rahul-tuli, line 48: If we do check in this function, args would be nice

rahul-tuli (approved):
minor nit, but looks good overall great job on this

src/llmcompressor/modifiers/quantization/calibration.py
dsikka: Sounds good - yeah I can add better documentation.

PR #1493, opened 2025-05-30
skip tracing tests if token unavailable

brian-dellabetta:
SUMMARY:
some tests in `tests/llmcompressor/transformers/tracing/test_models.py` requiring access to gated HF hub models will fail for community user PRs, because `HF_TOKEN` is not injected. This just skips the tests in question.

This update allows #1446 and #1445 to pass GHA checks so that they can be merged in 

TEST PLAN:
n/a


kylesayrs (approved):
woop

PR #1500, opened 2025-06-02
[Tests] Remove duplicate test

kylesayrs:
## Background ##
* When #97 was merged, it accidentally created two test config files which are identical
  * Because the auto device is no longer tested, `llama_7b_sparse_auto.yaml` is now identical to `llama_7b_sparse.yaml`

## Purpose ##
* Reduce testing runtime by removing duplicated test


dsikka (approved):
Thanks

PR #1502, opened 2025-06-02
[Testing] Use a slightly larger model that works with group_size 128

dsikka:
SUMMARY:
- The previous model was too small for group_size 128; update to use a larger one
- The point of the test  is to validate the preset schemes (which include 128 as the group_size) so chose to update the model instead of the recipe


shanjiaz (approved):
Looks good to me!

brian-dellabetta (approved):

kylesayrs:
What was wrong with the original? Outside not being able to run with vllm?

dsikka:
> What was wrong with the original? Outside not being able to run with vllm?

Model size was [64,64] while using group_size 128. Should error out in this case

kylesayrs:
@dsikka Got it, this what you were referencing wrt a bad fallback

PR #1514, opened 2025-06-04
Removed recipeStage, recipeModifer, stageModifier classes 

shanjiaz:
SUMMARY:
Removed recipeStage, recipeModifer, stageModifier classes and dependencies. 

TEST PLAN:
Tested locally

DETAILS & CONCERNS:
1. `recipe.py` is where most changes happened. 
- Rewrote the parsing logic in `parse_from_dict`. 
- `Recipe` class now is basically just a list of `Modifiers`, all other changes are downstream dependencies. 
- All the serilization helpers and deserialization helpers now live in `utils`. Had to rewrite serialization logic. 
- Stage is stored as a string by `self.stage`.
2. Had to update a test in `test_recipe`. We support creating a recipe instance by reading a multi-stage recipe but no longer support serializing multi-stage recipes directly. The current structure assumes one recipe could only have one stage.
3. Had to rewrite `update_and_save_recipe` function since now each recipe only has one stage and therefore can't handle merging multi-stage recipes. The new setup:
- `infer_recipe_from_model_path` will find existing recipe path. 
- `yaml` method now takes in an optional argument `existing_recipe_path` and merge the existing recipe with the current recipe. Since there would only be at most one existing recipe at a time, we should be safe. 

TODO:
- [x]  Add validations for recipe class. 
- [x] Simplify `recipe.py` structure


src/llmcompressor/recipe/utils.py
brian-dellabetta, line 1: these helpers were all pulled from the deleted files? Any net-new logic in here?

tests/llmcompressor/recipe/test_recipe.py
```diff
@@ -52,8 +54,7 @@ def test_recipe_creates_correct_modifier():
     recipe_instance = Recipe.create_instance(yaml_str)
 
     stage_modifiers = recipe_instance.create_modifier()
-    assert len(stage_modifiers) == 1
-    assert len(modifiers := stage_modifiers[0].modifiers) == 1
```
brian-dellabetta, line 56: I'm surprised we had this before, isn't the `:=` syntax python 3.10+? I would think this would fail the python 3.9 tests.

brian-dellabetta (dismissed):
This is looking great! 
🤩 
Screenshot 2025-06-05 at 3 54 46 PM

Would just make sure Dipika is good with the changes, and we'll probably want to kick off a weekly/nightly test before merging when it's all ready

tests/llmcompressor/recipe/test_recipe.py
shanjiaz: I could get rid of it.

src/llmcompressor/recipe/utils.py
shanjiaz: Almost! I rewrote the last two for serialization.

tests/llmcompressor/recipe/test_recipe.py
brian-dellabetta: nm, i'm wrong. `:=` is working for me in python 3.9. my bad

```python
>>> if (a := 2+3):
...     print(a)
... 
5
```

src/llmcompressor/core/lifecycle.py
```diff
@@ -33,7 +33,7 @@ class CompressionLifecycle:
 
     state: State = field(default_factory=State)
     recipe: Recipe = field(default_factory=Recipe)
-    modifiers: List[StageModifiers] = field(default_factory=list)
+    modifiers: List[Modifier] = field(default_factory=list)
```
kylesayrs: Oh Yeah

src/llmcompressor/core/lifecycle.py
```diff
@@ -96,7 +96,7 @@ def initialize(
             recipe=recipe, target_stage=recipe_stage, override_args=recipe_args
         )
         self.modifiers = self.recipe.create_modifier()
-
+        breakpoint()
```
brian-dellabetta:
```suggestion
```

brian-dellabetta (changes requested):
looks good but got a `breakpoint()` in code. also can you remove `[WIP]` from PR header if it's ready?

src/llmcompressor/core/lifecycle.py
kylesayrs: Remove

src/llmcompressor/pipelines/independent/pipeline.py
```diff
@@ -34,18 +32,13 @@ def __call__(
         _logger = logger.patch(lambda r: r.update(function="IndependentPipeline"))
 
         session = active_session()
-        modifiers = session.get_modifiers()
-        with patch_attr(session.lifecycle, "modifiers", None):
```
kylesayrs, line 38: This is still needed, right? The idea of the independent pipeline is that each pipeline runs with only one modifier at a time. Afaict, this code says that each pipeline runs with all modifiers active

src/llmcompressor/core/lifecycle.py
shanjiaz: Ah!

src/llmcompressor/recipe/utils.py
```diff
@@ -0,0 +1,111 @@
+import json
+import re
+from typing import Any, Dict, List
+
+import yaml
+
+from llmcompressor.modifiers import Modifier
+
+
+def _load_json_or_yaml_string(content: str) -> Dict[str, Any]:
```
kylesayrs, line 10: In the future we'll move these utils and the modifiers list to exist as class methods/attributes on the Recipe class itself

src/llmcompressor/transformers/compression/helpers.py
```diff
@@ -58,21 +59,17 @@ def tensor_follows_mask_structure(tensor: torch.Tensor, mask: str = "2:4") -> bo
 
 
 def infer_sparsity_structure_from_stage_modifiers(
```
kylesayrs: nit: rename function

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -294,17 +293,10 @@ def update_and_save_recipe(model_stub: str, save_directory: str):
...
-        recipes_to_save.append(new_recipe)
+    existing_recipe = infer_recipe_from_model_path(model_stub)
 
-    recipe = Recipe.simplify_combine_recipes(recipes_to_save)
+    recipe = active_session().lifecycle.recipe
 
-    # save recipe
     recipe_path = os.path.join(save_directory, RECIPE_FILE_NAME)
-    recipe.yaml(recipe_path)
+    recipe.yaml(file_path=recipe_path, existing_recipe_path=existing_recipe)
```
kylesayrs, line 262: I thought we were going to save yaml files sequentially, rather than merging them?

src/llmcompressor/recipe/utils.py
```diff
@@ -0,0 +1,111 @@
...
+def deep_merge_dicts(d1: dict, d2: dict) -> dict:
+    """
+    Recursively merge d2 into d1.
+    """
+    result = dict(d1)  # copy base
+    for key, val in d2.items():
+        if key in result and isinstance(result[key], dict) and isinstance(val, dict):
+            result[key] = deep_merge_dicts(result[key], val)
+        else:
+            result[key] = val
```
kylesayrs, line 126: Might want to mention in the docstring that d2 overrides d1's values, and also make sure that that behavior is intended

src/llmcompressor/core/lifecycle.py
kylesayrs: I'm a little confused, do modifiers exist on the session or do they exist on the recipe? Right now we have both, it'd be nice to focus on one source of truth

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
shanjiaz: I still think this functionality is preferred by users

src/llmcompressor/recipe/utils.py
shanjiaz: Will test on this more!

src/llmcompressor/core/lifecycle.py
shanjiaz: You're right! I guess we can remove modifiers from session and just used the recipe ones.

src/llmcompressor/pipelines/independent/pipeline.py
shanjiaz: good catch!

brian-dellabetta (dismissed):
I think this is looking good!

kylesayrs (commented):
Looks like all references to modifiers are caught, good job
```
grep -r '\.modifier' src/ tests/ --include="*.py" | grep -v 'from'
```

brian-dellabetta (approved):

brian-dellabetta (approved):

src/llmcompressor/recipe/utils.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    Filter a dictionary to only include keys that match the target stage.
+
+    :param obj: The recipe dictionary to filter.
+    :param target_stage: The stage to filter by (e.g., 'test_stage').
+    :return: A new dictionary containing only the keys that match the target stage.
+    """
+    if not target_stage:
+        return obj
+    return {k: v for k, v in obj.items() if k.startswith(target_stage)}
```
kylesayrs, line 111: Can we make this an exact match rather than startswith in order to avoid confusion?

src/llmcompressor/recipe/recipe.py
```diff
@@ -135,7 +136,7 @@ def create_instance(
             )
             logger.debug(f"Input string: {path_or_modifiers}")
             obj = _load_json_or_yaml_string(path_or_modifiers)
-            return Recipe.model_validate(obj)
+            return Recipe.model_validate(filter_dict(obj, target_stage=target_stage))
```
kylesayrs: Did you mean to pass a recipe_stage argument to Recipe here?

Also, why use model_validate rather than create_instance?

src/llmcompressor/recipe/recipe.py
```diff
@@ -157,77 +158,43 @@ def create_instance(
...
-            combined.version = simplified.version
-            combined.stages.extend(simplified.stages)
-            combined.args.update(simplified.args)
-
-        return combined
 
-    version: str = None
-    args: Dict[str, Any] = Field(default_factory=dict)
-    stages: List[RecipeStage] = Field(default_factory=list)
+            return Recipe.model_validate(filter_dict(obj, target_stage=target_stage))
```
kylesayrs: Same here, did you mean to pass stage to reipce creation and why not create using create_instance?

src/llmcompressor/recipe/recipe.py
```diff
@@ -31,6 +34,13 @@ class Recipe(RecipeBase):
     when serializing a recipe, yaml will be used by default.
     """
 
+    version: Optional[str] = None
```
kylesayrs: If this isn't being used then we should just remove it

src/llmcompressor/recipe/utils.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def deep_merge_dicts(d1: dict, d2: dict) -> dict:
```
kylesayrs: Since this util has logic specific to recipe stage conflicts, can we name it something like `append_recipe_dict`?

src/llmcompressor/recipe/utils.py
```diff
@@ -0,0 +1,141 @@
...
+        'test_stage_0', 'test_stage_1', etc.
+
+    Always starts numbering from 0 even for the first occurrence.
+    """
+    result = dict(d1)
+    for key, val in d2.items():
+        if key not in result:
+            result[key] = val
+        else:
+            # Stage key conflict — apply suffixes to both entries
```
kylesayrs, line 128: FYI it seems like ordering of the stages is not guaranteed when dumping and loading, since `yaml` uses dicts, not odicts. Just something to keep in mind

kylesayrs, line 128: Nice job considering stage name conflicts this

src/llmcompressor/recipe/recipe.py
```diff
@@ -238,295 +205,82 @@ def create_modifier(self) -> List["StageModifiers"]:
...
 
-        dict_["stages"] = stages
+        return get_yaml_serializable_dict(modifiers=self.modifiers, stage=self.stage)
 
-        return dict_
-
-    def yaml(self, file_path: Optional[str] = None) -> str:
+    def yaml(
+        self,
+        file_path: Optional[str] = None,
```
kylesayrs, line 210: When will this function ever be called with `file_path=None`?

src/llmcompressor/recipe/recipe.py
```diff
@@ -238,295 +205,82 @@ def create_modifier(self) -> List["StageModifiers"]:
...
+            stage=self.stage,
+        )
 
-        ret = yaml.dump(
-            yaml_dict,
+        # Deep merge — keep both recipe contents
+        merged_dict = deep_merge_dicts(existing_dict, self_dict)
+
+        # Dump YAML
+        file_stream = None if file_path is None else open(file_path, "w")
```
kylesayrs, line 238: See previous comment. When is file_path ever nullable?

src/llmcompressor/recipe/utils.py
```diff
@@ -0,0 +1,141 @@
...
+
+    if not isinstance(ret, dict):
+        raise ValueError(
+            f"Could not parse recipe from string {content}. If you meant load from "
+            "a file, please make sure that the specified file path exists"
+        )
+    return ret
+
+
+def _parse_recipe_from_md(file_path, yaml_str):
```
kylesayrs, line 29: Thanks for breaking these utils out

src/llmcompressor/recipe/recipe.py
```diff
@@ -66,27 +76,18 @@ def from_modifiers(
...
-        recipe.stages = stages
+        recipe.stage = group_name
+        recipe.modifiers = modifiers
         return recipe
 
     @classmethod
     def create_instance(
         cls,
         path_or_modifiers: Union[str, Modifier, List[Modifier], "Recipe"],
         modifier_group_name: Optional[str] = None,
```
kylesayrs, line 88: In the future we should kill modifier groups, but out of scope for here

src/llmcompressor/recipe/recipe.py
```diff
@@ -66,27 +76,18 @@ def from_modifiers(
...
-            )
-            for modifier in modifiers
-        ]
         # assume one stage for modifier instances
-        stages: List[RecipeStage] = [
-            RecipeStage(group=group_name, modifiers=recipe_modifiers)
-        ]
         recipe = cls()
-        recipe.stages = stages
+        recipe.stage = group_name
```
kylesayrs, line 80: This looks wrong. Shouldn't the stage just be `default`, in which case you don't need to override with the modifier group name?

Modifier group name and stage are different. Having the stage name become the group name is kind of unexpected behavior

src/llmcompressor/recipe/recipe.py
```diff
@@ -66,27 +76,18 @@ def from_modifiers(
...
+        recipe.modifiers = modifiers
```
kylesayrs, line 81:
```suggestion
        return cls(modifiers=modifiers)
```

src/llmcompressor/recipe/recipe.py
```diff
@@ -66,27 +76,18 @@ def from_modifiers(
```
kylesayrs, line 75: The validation above seems to contradict the type hints in the function definition

src/llmcompressor/pipelines/independent/pipeline.py
```diff
@@ -34,18 +33,15 @@ def __call__(
...
+        modifiers = session.lifecycle.recipe.modifiers
+        with patch_attr(session.lifecycle.recipe, "modifiers", None):
             for index, modifier in enumerate(modifiers):
```
kylesayrs:
```suggestion
            for modifier in modifiers:
```

tests/llmcompressor/recipe/test_recipe.py
```diff
@@ -26,8 +26,10 @@ def test_recipe_create_instance_accepts_valid_recipe_file(recipe_str):
 @pytest.mark.parametrize("recipe_str", valid_recipe_strings())
 def test_serialization(recipe_str):
     recipe_instance = Recipe.create_instance(recipe_str)
+    recipe_instance.create_modifier()
```
kylesayrs: Instead of having to call `create_modifier()` after every `create_instance()` call, can we just create the modifiers when we create the instance? Right now the type hint is `modifiers: List[Dict[str, Any]]`, which seems inconsistent with how `create_modifier` is used.

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -57,8 +57,9 @@ def __call__(
         model_device = get_execution_device(model)
 
         # prepare to trace subgraphs
-        modifiers = session.get_modifiers()
+        modifiers = session.lifecycle.recipe.modifiers
         sequential_targets = get_sequential_targets(modifiers, model, dataset_args)
+
```
kylesayrs, line 62: nit: remove

src/llmcompressor/pipelines/layer_sequential/pipeline.py
```diff
@@ -63,8 +63,9 @@ def __call__(
         model_device = get_execution_device(model)
 
         # find layers
-        modifiers = session.get_modifiers()
+        modifiers = session.lifecycle.recipe.modifiers
         sequential_targets = get_sequential_targets(modifiers, model, dataset_args)
+
```
kylesayrs: nit: remove

tests/llmcompressor/helpers.py
```diff
@@ -32,13 +32,6 @@ def valid_recipe_strings():
                     start: 0
                     end: 5
                     targets: __ALL_PRUNABLE__
-        test2_stage:
-                MagnitudePruningModifier:
-                    start: 5
-                    end: 10
-                    init_sparsity: 0.1
-                    final_sparsity: 0.5
-                    targets: __ALL_PRUNABLE__
```
kylesayrs, line 41: Why is this test removed?

tests/llmcompressor/recipe/test_recipe.py
```diff
@@ -52,8 +54,7 @@ def test_recipe_creates_correct_modifier():
     recipe_instance = Recipe.create_instance(yaml_str)
 
     stage_modifiers = recipe_instance.create_modifier()
-    assert len(stage_modifiers) == 1
-    assert len(modifiers := stage_modifiers[0].modifiers) == 1
+    assert len(modifiers := stage_modifiers) == 1
```
kylesayrs, line 55: Do you know what this assertion test is really doing? Do you think we can remove it?

src/llmcompressor/recipe/utils.py
shanjiaz: Sure!

tests/llmcompressor/recipe/test_recipe.py
shanjiaz: Yeah, this line `assert len(modifiers := stage_modifiers[0].modifiers) == 1` is just unwrapping the `stage_modifier` structure, and since we don't have `stage_modifier` anymore, it's safe to remove.

tests/llmcompressor/helpers.py
shanjiaz: This is the only test that assumes recipe should work with multiple stages. I could rewrite the test, at the moment I thought removing the second stage is faster.

src/llmcompressor/recipe/utils.py
shanjiaz: Huh this could be bad, I'll look into this.

src/llmcompressor/recipe/utils.py
shanjiaz: I think the issue is sometimes we use `finetune_stage`, and in some examples we use only `finetune`, for example. This was how we handled it before, so I'm preserving this behavior.

src/llmcompressor/recipe/recipe.py
shanjiaz: This is inside the `create_instance`, if I call `create_instance` here it'll end up in infinite recursive calls?

src/llmcompressor/recipe/recipe.py
shanjiaz: same as above.

tests/llmcompressor/recipe/test_recipe.py
shanjiaz: You're right! Will merge these two functions.

src/llmcompressor/recipe/utils.py
kylesayrs: Okay

src/llmcompressor/recipe/recipe.py
kylesayrs: I mean that the recipe which you have created does not know which stage it is

Maybe you want something like this
```python3
recipe_dict = _load_json_or_yaml_string(path_or_modifiers)
recipe_dict = filter_dict(obj, target_stage=target_stage)
Recipe.create_recipe(modifiers=recipe_dict, stage=target_stage)
```

EDIT: ignore my comment about `create_recipe`

src/llmcompressor/recipe/utils.py
kylesayrs: I don't think it's a big deal, and by default python with preserve the ordering, and if the order isn't preserved it's not a big deal.

It's just a limitation due to how the `yaml` library seems to return dicts instead of odicts. Not much we can do about it (I don't think) and not a big deal either way

src/llmcompressor/recipe/recipe.py
```diff
@@ -157,376 +157,100 @@ def create_instance(
...
+                            modifier = ModifierFactory.create(
+                                mod_type,
+                                group=group,
+                                allow_registered=True,
+                                allow_experimental=True,
+                                **mod_args,
+                            )
+                            modifiers.append(modifier)
+
+        return Recipe(
```
kylesayrs, line 195:
```suggestion
        return cls(
```

brian-dellabetta (approved):
Exciting!

kylesayrs (approved):
Nice job! Just nits left on my end, but we can land those in follow ups

Issue #1510, opened 2025-06-04
Better error/messaging when attempting to use smoothquant modifier without a calibration dataset


kylesayrs:
The error when attempting to quantize with smoothquant is misleading, as it does not properly indicate that a calibration dataset is needed

```
site-packages/llmcompressor/modifiers/smoothquant
/base.py", line 140, in on_initialize
    self._calibrate(state.model, calibration_dataloader)
site-packages/llmcompressor/modifiers/smoothquant
/base.py", line 249, in _calibrate
    f"{len(calibration_dataloader)} samples..."
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: object of type 'NoneType' has no len()
```

dsikka:
Similar case for any case requiring a dataset (eg Quantization Modifier calibration of activations).

PR #1526, opened 2025-06-09
error msg for using smoothquant without calibration dataset

ved1beta:
SUMMARY:
check for calibration dataset not present 
#1510


kylesayrs (approved):
Thank you for the contribution @ved1beta !

In the future, this may be incorporated into an earlier recipe validation step, but for now this will go a long way to helping users use smoothquant!

kylesayrs (changes requested):
Waiting for post release before merging

PR #1528, opened 2025-06-09
[Examples] [Bugfix] skip sparsity stats when saving checkpoints

kylesayrs:
## Purpose ##
* Fix example failure where sparsity stats were attempted to be computed on a compressed model
  * This happened because the `get_model_compressor` cannot compute sparsity stats of a compressed model

## Prerequisites ##
* https://github.com/neuralmagic/compressed-tensors/pull/345

## Changes ##
* When saving checkpoints, call `get_model_compressor` on the decompressed model (before it is compressed, not after)
* Do not save checkpoints with sparsity stats by default, due to high runtime

## Testing ##
* Ran example to completion with reduced training duration

### DIFF
--- a/src/llmcompressor/transformers/finetune/session_mixin.py
+++ b/src/llmcompressor/transformers/finetune/session_mixin.py
@@ -363,7 +363,7 @@ class SessionManagerMixIn:
         self,
         output_dir: str,
         _internal_call: bool = False,
-        skip_sparsity_compression_stats: Optional[bool] = False,
+        skip_sparsity_compression_stats: Optional[bool] = True,
     ):
         """
         Override of the save_model function and expects it to exist in the parent.
@@ -388,6 +388,8 @@ class SessionManagerMixIn:
             self.model.prepare_for_save()  # TODO: move to finalize
 
         # save checkpoint
+        # note that skip_sparsity_compression_stats
+        # is True by default to avoid high runtime cost
         self.save_state()
         if self.accelerator.is_main_process:
             processor = getattr(self, "processing_class", self.tokenizer)
--- a/src/llmcompressor/pytorch/model_load/helpers.py
+++ b/src/llmcompressor/pytorch/model_load/helpers.py
@@ -45,6 +45,16 @@ def save_checkpoint(
         get_model_compressor,  # avoid circular import
     )
 
+    # used for decompression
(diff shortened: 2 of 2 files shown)


rahul-tuli:
Lgtm, could you also update the save_pretrained readme in the docs folder

src/llmcompressor/transformers/finetune/session_mixin.py
```diff
@@ -363,7 +363,7 @@ def save_model(
         self,
         output_dir: str,
         _internal_call: bool = False,
-        skip_sparsity_compression_stats: Optional[bool] = False,
+        skip_sparsity_compression_stats: Optional[bool] = True,
```
dsikka, line 366: These were set to True by default because of the training flows which don’t allow saving arguments to be passed atm, apart from save_compressed fyi 


I would test the nightly if we’re changing default behaviour and make sure checkpoints make sense

kylesayrs: @dsikka
> These were set to True by default

I think you mean set to False by default?

> training flows which don’t allow saving arguments to be passed atm, apart from save_compressed

Makes sense. Because we agreed to not save as compressed, I don't think computing sparsity stats for each checkpoint should be the default (it incurs a large runtime cost and doesn't really make sense from a model lifecycle perspective)

> I would test the nightly if we’re changing default behaviour and make sure checkpoints make sense

https://github.com/neuralmagic/llm-compressor-testing/actions/runs/15540941320

kylesayrs:
@rahul-tuli I've read through the docs and don't see the section you want updated? Afaict these docs are for users who want help calling `save_pretrained` themselves. I'm not sure how these changes are related or relevant to that reader?

rahul-tuli:
> @rahul-tuli I've read through the docs and don't see the section you want updated? Afaict these docs are for users who want help calling `save_pretrained` themselves. I'm not sure how these changes are related or relevant to that reader?




My bad @kylesayrs meant to request a change on #1530 
And, missed the filename, thought you were making changes to `save_pretrained_wrapper` but this is a different pathway altogether. This is good to go from my side!

rahul-tuli (approved):

brian-dellabetta (approved):

PR #1530, opened 2025-06-09
[Examples] [Bugfix]

kylesayrs:
## Purpose ##
* Fix failing examples

## Changes ##
* Save model after generation in all examples
  * Previously, models would be saved before generation, causing generation to fail because we do not fully support generating with compressed models atm

## Future ##
* In the future, we can define a better API around compressing and decompressing models which does not require so many arguments
* In the future, we can standardize around reloading (and redispatching) the model before generation, as suggested here #1263
* In the future, we can remove the sample generation step

### DIFF
--- a/examples/quantization_w8a8_fp8/gemma2_example.py
+++ b/examples/quantization_w8a8_fp8/gemma2_example.py
@@ -20,12 +20,10 @@ recipe = QuantizationModifier(
 )
 
 # 3) Apply quantization and save in compressed-tensors format.
-OUTPUT_DIR = MODEL_ID.split("/")[1] + "-FP8-Dynamic"
 oneshot(
     model=model,
     recipe=recipe,
     tokenizer=tokenizer,
-    output_dir=OUTPUT_DIR,
 )
 
 # Confirm generations of the quantized model look sane.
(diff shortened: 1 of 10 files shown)


kylesayrs:
@shanjiaz The model id strings are hardcoded, therefore there is no risk of the string index missing

rahul-tuli (commented):
We need this change because our compression logic now modifies the model directly (potentially deleting the required params), whereas previously the model state remained unchanged.

That said, I think it would be helpful to sync offline to better understand the broader implications. It might be worth focusing on the root issue rather than addressing it here in isolation.

From a usability perspective, it's a bit unexpected for inference to break after a call to save_pretrained, regardless of whether compression was involved. Also, given that PyTorch inference can sometimes be slow, we should consider how this affects the overall user experience.

kylesayrs:
@rahul-tuli I agree, but this was already discussed when these changes were being planned. As called out in the PR description:

> In the future, we can define a better API around compressing and decompressing models which does not require so many arguments

This PR is simply a bug fix for changes which have already landed

rahul-tuli:
> @rahul-tuli I agree, but this was already discussed when these changes were being planned. As called out in the PR description:
> 
> 
> 
> > In the future, we can define a better API around compressing and decompressing models which does not require so many arguments
> 
> 
> 
> This PR is simply a bug fix for changes which have already landed

Discussed this offline with @kylesayrs, approving this PR for now since the flow is already broken on main; we will create an issue and link it here

rahul-tuli (approved):

PR #1540, opened 2025-06-12
[Memory] Allow user configuration of quantization memory requirements

dsikka:
SUMMARY:
- Allow the user to set the group_size and quantization memory


kylesayrs:
I'm thinking of removing all of these utilities in sequential onloading, since the sequential pipeline will become the default https://github.com/vllm-project/llm-compressor/pull/1263

PR #1541, opened 2025-06-12
fix qwen 2.5 VL multimodal example

brian-dellabetta:
SUMMARY:
Fix a failing qwen 2.5 VL example, just had incorrect dataset split configuration.

TEST PLAN:
Runs on H100 on latest llm-compressor


shanjiaz (approved):
Looks good to me!

PR #1544, opened 2025-06-13
[Examples] Use more robust splits

kylesayrs:
## Purpose ##
* Make tests more robust to custom model ids that don't have a "/" in the name
  * Many users replace the model ids with local paths. These local paths do not necessarily have "/"s in them, and will fail at the very end of compression, which is decidedly a bad time

## Changes ##
* `model_id.split("/")[1]` -> `model_id.split("/")[-1]`

## Testing ##
* `grep -r -i 'MODEL_ID.split("/")\[1\]' src examples/ tests`


shanjiaz (approved):
Kyle's great surrender to gemini lol

brian-dellabetta (approved):

kylesayrs:
@shanjiaz It was a rough battle, but I concede defeat here

vegeta

PR #1554, opened 2025-06-16
[NVFP4] Fix global scale update when dealing with offloaded layers

dsikka:
SUMMARY:

- Updating the global scale using the `align_module` context does not persist the scale parameter
- Update outside of the context so that the offloaded dict is upadated as well

Testing
- Resolves CPU offloading issues seen with a Llama 70b FP4

### DIFF
--- a/src/llmcompressor/modifiers/utils/helpers.py
+++ b/src/llmcompressor/modifiers/utils/helpers.py
@@ -2,11 +2,7 @@ from typing import List
 
 import torch
 from compressed_tensors.quantization import QuantizationStrategy
-from compressed_tensors.utils import (
-    align_module_device,
-    align_modules,
-    update_parameter_data,
-)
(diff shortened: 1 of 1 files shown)


src/llmcompressor/modifiers/utils/helpers.py
```diff
@@ -72,10 +76,16 @@ def _valid_tensor_group_quant(layer_list: List[Linear]):
                 )
             ).reshape([1])
 
-            update_parameter_data(submodule.q_proj, global_scale, "weight_global_scale")
+        with align_module_device(submodule.k_proj):
```
kylesayrs: You don't need to onload the module before updating its parameters. Simply call `update_parameter_data` directly

kylesayrs (changes requested):

PR #1557, opened 2025-06-16
AWQ minor performance improvements to smoothing

brian-dellabetta:
SUMMARY:
I wanted to investigate how the runtime for `apply_smoothing` in AWQ could be decreased with torch.compile and removing python for loops. These changes reduce the runtime of `apply_smoothing` by about 8-10%, by vectorizing the loss computation.

TEST PLAN:
Confirmed wikitext PPL scores are virtually the same when running AWQ on `"meta-llama/Llama-3.2-3B-Instruct"` and `"Qwen/Qwen2.5-7B-Instruct"`

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -480,9 +480,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                     )
                     del self._smooth_activation_means[mapping.smooth_name]
                     continue
-                if not all(
-                    [fp16_output.isfinite().all() for fp16_output in fp16_outputs]
-                ):
+                if not fp16_output.isfinite().all():
                     logger.warning(
                         f"Skipping smooth_layer {mapping.smooth_name}, NaN or inf "
                         "outputs found during forward pass of the parent module "


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -541,25 +541,35 @@ def smooth(module):
             v.batch_intermediates.clear()
         self._assert_all_activations_consumed()
 
-    def _run_samples(self, module: Module) -> List[torch.Tensor]:
+    def _get_flattened_output(self, module: Module) -> torch.Tensor:
+        """
+        Returns output of running cached batch inputs through module
+        Output tensor is 1D, as shapes aren't necessary for calculating loss
+        """
```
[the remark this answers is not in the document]
brian-dellabetta: added

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -677,6 +659,17 @@ def _assert_all_activations_consumed(self):
             raise RuntimeError("Some cached activations were not used")
 
 
+@torch.no_grad()
+@torch.compile()
+def _compute_loss(
+    fp16_output: torch.Tensor,
+    int_w_output: torch.Tensor,
+) -> torch.Tensor:
+    """Compute MSE loss for each batch"""
```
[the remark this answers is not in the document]
brian-dellabetta: added

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -677,6 +660,19 @@ def _assert_all_activations_consumed(self):
...
+@torch.no_grad()
+@torch.compile()
+def _compute_loss(
+    fp16_output: torch.Tensor,
+    int_w_output: torch.Tensor,
+) -> torch.Tensor:
+    """
+    Compute MSE loss over the flattened output of all batches
+    """
+    return (fp16_output - int_w_output).view(-1).float().pow(2).mean()
```
kylesayrs: Using `torch.nn.MSE()` here might be faster and likely already uses a precompiled kernel

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -677,6 +660,19 @@ def _assert_all_activations_consumed(self):
...
+
+
+@torch.compile()
```
kylesayrs, line 668: I wonder if torch.compile performs better without the asserts?

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: good call, it's pretty much the same runtime-wise but a lot cleaner code-wise, we don't need a custom function. updated

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I tried just now with asserts commented out, seems to shave off a few seconds. i think the if statements might not make it not all that beneficial? I think i'll leave the asserts as is, will be helpful for debugging. we can always add in future

kylesayrs: Tracing has asserts turned off by default
https://github.com/pytorch/pytorch/blob/main/torch/fx/proxy.py#L131

I wasn't sure about torch.compile

kylesayrs (dismissed):

brian-dellabetta:
Closing this as it is pretty stale, and as of v0.9.0 the batched calibration runs provide a significant speedup to AWQ far surpassing minor improvements of this branch

PR #1565, opened 2025-06-18
[BugFix] Fix quantizaiton_2of4_sparse_w4a16 example 

shanjiaz:
SUMMARY:
Pass saved model directly to next stage since now `train` and `oneshot` can properly initialize models from path.

TEST PLAN:
Testing `test_quantization_2of4_sparse_w4a16` locally

```
collected 2 items                                                                                                                                    

tests/examples/test_quantization_2of4_sparse_w4a16.py::TestQuantization24SparseW4A16::test_doc_example_commPASSED
tests/examples/test_quantization_2of4_sparse_w4a16.py::TestQuantization24SparseW4A16::test_alternative_recipe PASSED

=========================================================== 2 passed in 6123.28s (1:42:03) ===========================================================
```


tests/examples/test_quantization_2of4_sparse_w4a16.py
```diff
@@ -49,7 +49,7 @@ def test_doc_example_command(self, example_dir: str, tmp_path: Path):
         stages = {
             "quantization": {
                 "path": Path("quantization_stage"),
-                "format": "marlin-24",
+                "format": "pack-quantized",
```
dsikka: This should be marlin-24

dsikka (changes requested):

dsikka (changes requested):
You need to rebase your PR

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
```diff
@@ -63,32 +67,52 @@
...
     stage="finetuning_stage",
 )
 
+finetune_applied_model = AutoModelForCausalLM.from_pretrained(
+    output_dir + "/finetuning_stage"
+)
+
+compressor = get_model_compressor(
+    model=finetune_applied_model,
+    save_compressed=False,
```
dsikka: save_compressed should be True?

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
```diff
@@ -63,32 +67,52 @@
...
+    compressor.decompress_model(finetune_applied_model)
+
 # Oneshot quantization
 model.to("cpu")
 quantized_model = oneshot(
     model=finetune_applied_model,
     **oneshot_kwargs,
     stage="quantization_stage",
 )
-quantized_model.save_pretrained(
```
dsikka: Why don’t we save here?

tests/examples/test_quantization_2of4_sparse_w4a16.py
shanjiaz: Huh, I'm looking into this!

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
shanjiaz: It's already saved within the `oneshot` call.

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
shanjiaz: Will update!

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
dsikka: I think we want to be saving explicitly, not use the output_dir arg

shanjiaz: I see! Will update!

brian-dellabetta (dismissed):
it looks like this keeps reference to model, oneshot_applied_model, and finetune_applied_model. so essentially 3 models are living in memory. same behavior as original example, but just wanted to call it out

shanjiaz:
> it looks like this keeps reference to model, oneshot_applied_model, and finetune_applied_model. so essentially 3 models are living in memory. same behavior as original example, but just wanted to call it out

Good point! Not sure if there's anything I can do with saving three different models. I'm testing to see if I can skip the manual decompression step at least

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
```diff
@@ -63,21 +66,47 @@
...
+compressor = get_model_compressor(
+    model=finetune_applied_model,
+    save_compressed=True,
+    skip_sparsity_compression_stats=False,
+)
+if compressor is not None:
+    compressor.decompress_model(finetune_applied_model)
+
 # Oneshot quantization
 model.to("cpu")
```
dsikka: Do we need this anymore?

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
```diff
@@ -63,21 +66,47 @@
...
+    save_compressed=True,
+    skip_sparsity_compression_stats=False,
+)
+if compressor is not None:
+    compressor.decompress_model(oneshot_applied_model)
+
 # Sparse finetune
 dispatch_for_generation(model)
-finetune_applied_model = train(
+train(
```
dsikka, line 77: Train calls dispatch_for_generation, so I dont think we need this anymore.

Since we're loading the model and decompressing from disk, we can just pass in the path to train:

train(model=Path(output_dir)/ "sparsity_stage", ...) - this will load the model and handle the decompression for you:

https://github.com/vllm-project/llm-compressor/blob/f6010ce1fba76232dfea586bd03e44b4482c348b/src/llmcompressor/entrypoints/utils.py#L133

dsikka (changes requested):

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
shanjiaz: Nice! Will do 🫡

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
shanjiaz: Looks like it was added as part of [this PR](https://github.com/vllm-project/llm-compressor/pull/1263), addressing an issue [here](https://github.com/vllm-project/llm-compressor/issues/1383). I think it might be moving the wrong model to cpu anyways, will remove.

dsikka (dismissed):

examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
```diff
@@ -63,25 +64,25 @@
...
+    output_dir=output_dir,
     stage="sparsity_stage",
 )
 
 # Sparse finetune
-dispatch_for_generation(model)
-finetune_applied_model = train(
-    model=oneshot_applied_model,
+train(
+    model=Path(output_dir + "/sparsity_stage"),
```
kylesayrs: Is this the correct use of `pathlib.Path`? I think you might be looking for

```python3
output_dir = Path("output_llama7b_2of4_w4a16_channel")

train(
    model=(output_dir / "sparsity_stage"),
    ...
)
```

See examples
https://github.com/vllm-project/llm-compressor/blob/a144d8b44c5541a2a87dc87a41ed1f436b04f1ce/tests/llmcompressor/transformers/obcq/test_consecutive_runs.py#L87

kylesayrs:
Looks good, just make sure you're using pathlib in a way you expect

kylesayrs (approved):

brian-dellabetta (approved):
nice

jessiewiswjc:
@shanjiaz Good job! Have you tested the accuracy of the model produced by this script? I am sad to find that my 14B model's mmlu score dropped a lot (from 72 to 60) after using this script.

PR #1568, opened 2025-06-18
Remove `oneshot_device`

dsikka:
Summary
- Not used and annoying to see everywhere


kylesayrs:
I agree with this philosophically. Device handling will only become more complicated as of https://github.com/vllm-project/llm-compressor/pull/1558. The user can control which device use to via environment variables, but beyond that we should control all other device dispatching.

kylesayrs:
Will approve on behalf of @dsikka

PR #1572, opened 2025-06-19
[Pipelines] infer model device with optional override

kylesayrs:
## Purpose ##
* Fix support for deepseekv2.5
  * Add more robustness inference for model devices when calibrating

## Prerequisites ##
* https://github.com/neuralmagic/compressed-tensors/pull/363

## Background ##
Normally, starting model inputs on the cpu is not an issue for the sequential pipeline, since the sequential pipeline offloads models and offloaded models automatically place inputs on the proper devices. However, the deepseekv2.5 model is an exception, as this model [performs an add operation](https://huggingface.co/deepseek-ai/DeepSeek-V2.5/blob/main/modeling_deepseek.py#L886) between a module output (`attn_weights` and a model input `attention_mask`) before the model input has a chance to be placed on the proper device.

## Changes ##
* Use `model_device` when deciding the onload device for model inputs

## Testing ##
* Ran deepseekv2.5 example to completion
* TODO: run nightly to confirm other models work with new input device placement


dsikka (dismissed):

shanjiaz (commented):
Looks good to me! Thanks for testing

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -38,7 +38,9 @@ def __call__(
         :param dataset_args: dataset arguments relevant to pipelines
         """
         dispatch_for_generation(model)  # basic dispatch is identical to generation
-        model_device = get_execution_device(model)
+        model_device = getattr(dataset_args, "model_input_device")
+        if model_device is None:
+            model_device = infer_model_device(model)
```
brian-dellabetta: are we deprecating `get_execution_device` in favor of `infer_model_device` now? not clear to me when would be used over the other

shanjiaz (dismissed):

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -14,6 +14,7 @@
     get_sequential_targets,
     trace_subgraphs,
 )
+from llmcompressor.utils.dev import infer_model_device
```
brian-dellabetta: same ask here, do we expect users to be using this anywhere? if so importing from a `.dev` might be confusing

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -196,3 +197,10 @@ class DatasetArguments(CustomDatasetArguments):
             "definition"
         },
     )
+    model_input_device: Optional[str] = field(
```
brian-dellabetta: does inferring not work for deepseek? wondering if we can avoid another field that might confuse users

src/llmcompressor/pipelines/basic/pipeline.py
kylesayrs: In the future we could fold `infer_model_device` into `get_execution_device`. For now, `get_execution_device` is for modules, whereas `infer_model_device` is used for models

src/llmcompressor/args/dataset_arguments.py
kylesayrs: Inferring works for all the models I've tested, but isn't strictly guaranteed to work, hence why there's a fallback option.

I agree it's probably not necessary though, we can remove it

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs: In the future it's most robust to have this be imported in the example scripts, so I can raise this import path

src/llmcompressor/pipelines/basic/pipeline.py
brian-dellabetta: cool, good to know

src/llmcompressor/args/dataset_arguments.py
brian-dellabetta: cool, yeah if it's not needed for deepseek i think it's better to exclude for now, given the input arg bloat of `oneshot`. thanks!

kylesayrs:
@brian-dellabetta Thanks for your review!

dsikka (approved):
So we’re just using get_execution_device?
Is there a scenario where we infer it incorrectly/as meta for what reason? Seems like we removed the fallback if I read it correctly

kylesayrs:
@dsikka Yes, we removed the user-defined fallback (if it becomes necessary, we can add it back).

The `get_execution_device` is now expanded when https://github.com/neuralmagic/compressed-tensors/pull/363 lands

rahul-tuli (approved):

kylesayrs:
@dsikka 
> Is there a scenario where we infer it incorrectly/as meta for what reason

One of the rules of accelerate offloading (with AlignDevicesHook) is that a parameter is on the meta device iff its direct parent module is offloaded.

```python3
if has_offloaded_params(module):
   return module._hf_hook.execution_device
```

Since `get_execution_device` checks for this case, we should be safe from ever returning the device of a parameter on the meta device.

PR #1574, opened 2025-06-20
[Bugfix] Fix Whisper compatibility with CPU offloading

### LINKED ISSUE #1573: Problem faced with quantizing whisper
**Describe the bug**
RuntimeError: Tensor on device meta is not on the expected device cuda:0!

**Expected behavior**
No error happens and quantized weight could be obtained 

**Environment**
1. OS [Ubuntu 20.04]:Ubuntu 20.04
2. Python version [3.10.12]: 3.10.12
3. LLM Compressor version or commit hash [e.g. 0.1.0, `f7245c8`]:
4. ML framework version(s) [torch 2.7.1]: torch 2.7.1
5. Other Python package versions [e.g. vLLM, compressed-tensors, numpy, ONNX]:
6. Other relevant environment information [CUDA version]: 12.6

**To Reproduce**
cd /llm-compressor/examples/multimodal_audio 
python3 whisper_example.py

**Errors**

Image

**Additional context**

kylesayrs:
## Purpose ##
* Fixes #1573

## Background ##
The whisper model includes these lines where the embedding weight is used as a bare addend. The model can do this because the inputs_embeds are outputs of conv layers, which always output the same shape. `self.embed_positions` acts essentially as an unlearned bias buffer.
```python3
embed_pos = self.embed_positions.weight
hidden_states = inputs_embeds + embed_pos
```

However, this naked usage of `torch.nn.Embedding` means that the forward function is never called, and the weight is never onloaded, breaking CPU offloading compatibility with the whisper model. This is likely not a concern for most users, since the whisper models are quite small and don't usually require CPU offloading.

## Changes ##
* Support CPU offloading (which is required by sequential onloading) by modifying the whisper model definition
  * An upstream PR will be requested for future transformers versions

## Testing ##
* Ran whisper example to completion


kylesayrs:
Upstream fix: https://github.com/huggingface/transformers/pull/38994

src/llmcompressor/modeling/whisper.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(
+        self, config, conv1, conv2, embed_positions, layers, layer_norm
+    ):
+        super().__init__(config)
+        self.dropout = config.dropout
+        self.layerdrop = config.encoder_layerdrop
+
+        embed_dim = config.d_model
+        self.num_mel_bins = config.num_mel_bins
+        self.padding_idx = config.pad_token_id
```
brian-dellabetta, line 43: Is this correct?

src/llmcompressor/modeling/whisper.py
```diff
@@ -0,0 +1,143 @@
...
+
+import math
+
+import torch
+from torch import nn
+from transformers.modeling_outputs import BaseModelOutput
+from transformers.models.whisper.modeling_whisper import WhisperEncoder
+
+
+class WhisperEncoderPatched(WhisperEncoder):
```
brian-dellabetta, line 27: might be good to call out what has been patched from `WhisperEncoder` here, the whole thing or certain lines?

src/llmcompressor/modeling/whisper.py
kylesayrs: Yes, here is the original implementation: https://github.com/huggingface/transformers/blob/main/src/transformers/models/whisper/modeling_whisper.py#L611-L612

src/llmcompressor/modeling/whisper.py
kylesayrs: Yeah thanks for the callout

brian-dellabetta (approved):

kylesayrs:
The upstream fix has been merged, closing this but this can be used as reference for fixing the issue in the future prior to the upstream fix being released

PR #1578, opened 2025-06-23
Add fallback for sparsegpt inversion failure

kylesayrs:
## Purpose ##
* Add fallback for spgt inversion failure, which may occur randomly


shanjiaz (approved):

brian-dellabetta (approved):
cool, i think will resolve one of the issues reported in #1410 (part 2 of their issue)

PR #1593, opened 2025-06-25
[MoE] Add MoE calibration options

kylesayrs:
## Purpose ##
* Add more options for configuring how an MoE model is calibrated via `prepare_for_calibration`

## Changes ##
* Add moe calibration arguments to `prepare_for_calibration`
```
:param moe_calibrate_all_experts: send all tokens to all experts for calibration
:param moe_calibrate_gated_acts: use moe gating mechanism when computing
    expert input and output activations. If this is True, the model computes
    activations similar to those found during inference. If this is False, the
    model computes activations similar to those found during training.
```


dsikka:
Putting this as draft as I make some changes

kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/1596

PR #1595, opened 2025-06-25
Drop torch v1 support, allow numpy v2 support

brian-dellabetta:
SUMMARY:
Drop torch v1 support and extraneous warnings, including warning about use of torch.compile on torch v2.

- vllm pins torch==2.7.0 and has no restrictions on numpy
- compressed-tensors has torch>=1.7.0 and no numpy dependency

TODOs if we want to merge:
- [ ] Drop torch v1 support in [compressed-tensors](https://github.com/neuralmagic/compressed-tensors/blob/main/setup.py#L91)

TEST PLAN:
no net new src code


setup.py
```diff
@@ -112,10 +112,10 @@ def localversion_func(version: ScmVersion) -> str:
     install_requires=[
         "loguru",
         "pyyaml>=5.0.0",
-        "numpy>=1.17.0,<2.0",
+        "numpy>=1.17.0",
         "requests>=2.0.0",
         "tqdm>=4.0.0",
-        "torch>=1.7.0",
+        "torch>=2.7.0",
```
brian-dellabetta: Alternatively, if we want to allow torch v1 support, this can be
```suggestion
        #torch 1.10 and 1.11 do not support quantized onnx export
        "torch>=1.7.0,!=1.10,!=1.11",
```

kylesayrs (approved):
Looks like the numpy restriction was related to some random tensorboard conflict

https://github.com/neuralmagic/sparseml/pull/2333

shanjiaz (approved):
Looks good to me! Thanks for cleaning this up : )

shanjiaz (dismissed):

shanjiaz (approved):

kylesayrs (approved):

PR #1596, opened 2025-06-25
[Calibration] Add MoE Calibration Context

dsikka:
# Summary:
- Introduce an `moe_calibration_context` which during calibration, replaces MoE blocks with custom modules which are needed to properly calibrate all experts requiring data
- The context can be optionally enabled through a new `calibrate_moe_context` argument which if set to True, will enable the context
- Modules are replaced with new definitions defined in the `prepare` folder (shared with `replace_modules_for_calibration`)
- This enables a second pathway for calibrating MoEs and other models that require updates to their modules to be compatible with llm-compressor:

1. Replacing modules during calibration
2. Replacing modules permanently (as done by `replace_modules_for_calibration`, previously called `prepare_for_calibration`). 

- Similar to `replace_modules_for_calibration`, a dictionary defining the replacement has been added: `moe_context`

# Testing
- Tested with a `Qwen/Qwen3-30B-A3B` NVFP4 example and added the example to the folder as well

# Next Steps:
- Definitions for updated the MoE modules are hardcoded atm - we want to expand and add additional parameters to have more control over the MoE forward pass, such as through parameters defined here: https://github.com/vllm-project/llm-compressor/pull/1593  - this is especially important if we find a certain configuration results in optimal calibration
- We may find it easier to refactor out calibration args into their own pydantic model and not put everything under datraset args


src/llmcompressor/modeling/prepare.py
```diff
@@ -2,19 +2,58 @@
...
+# ------------------- module replacements; during calibration --------------------
+
+
+def update_qwen3_moe(model, stack):
+    for _, module in model.named_modules():
+        cls_name = module.__class__.__name__
+        if cls_name == "Qwen3MoeDecoderLayer":
+            stack.enter_context(
+                patch_attr(module, "mlp", replace_Qwen3MoE(model.config, module.mlp))
+            )
```
kylesayrs: Can you add a quick comment about how exiting the context still preserves parameter updates upon exit?

src/llmcompressor/modeling/prepare.py
```diff
@@ -2,19 +2,58 @@
...
             replace_module(model, name, new_module)
 
     return model
+
+
+# ------------------- module replacements; during calibration --------------------
+
+
+def update_qwen3_moe(model, stack):
+    for _, module in model.named_modules():
```
kylesayrs:
```suggestion
    for module in model.modules():
```

src/llmcompressor/modeling/prepare.py
```diff
@@ -2,19 +2,58 @@
...
+        cls_name = module.__class__.__name__
+        if cls_name == "Qwen3MoeDecoderLayer":
```
kylesayrs, line 34: Could you use something like this pattern for matching? This way things don't break if the parent's structure changes, and we can also share matching logic between replacements
```
        cls_name = module.__class__.__name__
        if cls_name in replacements:
            new_module = replacements[cls_name](module)
            replace_module(model, name, new_module)
```

src/llmcompressor/modeling/prepare.py
```diff
@@ -2,19 +2,58 @@
...
+            )
+
+
+moe_context = {
+    "Qwen3MoeForCausalLM": update_qwen3_moe,
+    # "DeepseekV3ForCausalLM": update_deepseek3_moe, TODO: uncomment when tested
+}
+
+
+def moe_calibration_context(model: PreTrainedModel, stack):
```
kylesayrs, line 50: Can you show what it's like to pass additional calibration options (`moe_calibrate_all_experts` `moe_calibrate_gated_acts`), if these are still options we want to supply research/users

src/llmcompressor/modeling/qwen3_moe.py
```diff
@@ -0,0 +1,69 @@
+import torch
```
kylesayrs, line 17: May want to add the HF copyright and a note about the amendments

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -42,7 +44,12 @@ def __call__(
 
         LifecycleCallbacks.calibration_epoch_start()
 
-        with calibration_forward_context(model):
+        with contextlib.ExitStack() as stack:
+            stack.enter_context(calibration_forward_context(model))
+
+            if dataset_args is not None and dataset_args.calibrate_moe_context:
```
kylesayrs, line 50: Maybe we can just use the calibration options moe_calibrate_all_experts and moe_calibrate_gated_acts directly, and always enter this context?

kylesayrs (commented):
Is the plan to use this for llama4 as well, or will that be a separate function?

src/llmcompressor/modeling/deepseek_v3.py
```diff
@@ -46,7 +45,7 @@ def forward(self, hidden_states):
         return hidden_states
 
 
-def replace(module: DeepseekV3MoE) -> DeepseekV3MoECalibrate:
-    return DeepseekV3MoECalibrate(
+def replace(module: "DeepseekV3MoE") -> DeepseekV3MoE:
```
brian-dellabetta: Aren't you replacing module of type `transformers.models.deepseek_v3.modeling_deepseek_v3.DeepseekV3MoE` with one of type `llmcompressor.modeling.deepseek_v3.DeepseekV3MoE` here? Why the change in signature?

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -228,6 +232,7 @@ def oneshot(
     preprocessing_num_workers: Optional[int] = None,
     min_tokens_per_module: Optional[float] = None,
     trust_remote_code_data: bool = False,
+    calibrate_moe_context: bool = False,
```
brian-dellabetta, line 234: Should this always be true if it's a MoE model passed into oneshot? Wondering if this can be inferred instead of requiring user setting

src/llmcompressor/modeling/deepseek_v3.py
dsikka: Just a name change.

src/llmcompressor/entrypoints/oneshot.py
dsikka: We can start inferring this once this pathway has been hardened but for now I'd like to make it easy to turn off/on

src/llmcompressor/pipelines/basic/pipeline.py
dsikka: Yeah we could. I was hoping to add these option in a follow-up

src/llmcompressor/modeling/deepseek_v3.py
brian-dellabetta: the type of input `module` is different in your changes, it's now indicating it's of type `llmcompressor.modeling.deepseek_v3.DeepseekV3MoE` rather than the one in transformers. Is that what we want?

dsikka:
> Is the plan to use this for llama4 as well, or will that be a separate function?

I think for Llama4, we may want to change the structure permanently, in which case we'd want to use the `replace_modules_for_calibration` so that we can also compress is correctly post calibration as well.

src/llmcompressor/modeling/qwen3_moe.py
dsikka: I added the HF copyright - do you hava reference for what type of note should be made about amendments?

src/llmcompressor/modeling/prepare.py
dsikka: I think if we want to use patch_attr in order to follow the other context set-up functionality, we need both the parent and the child so it would still require setting "mlp" - I think replace_modules finds the parent for you when replacing the module. 

We could expand patch_attr I guess to follow that potentiallly

src/llmcompressor/modeling/deepseek_v3.py
dsikka: Oh I see what you're saying

src/llmcompressor/modeling/prepare.py
dsikka: I left a small comment but see other comment below.

examples/quantization_w4a4_fp4/qwen_30b_a2b.py
brian-dellabetta, line 1: should this be "_a3b.py" to match `"Qwen/Qwen3-30B-A3B"`?

examples/quantizing_moe/deepseek_r1_example.py
```diff
@@ -20,7 +20,7 @@
     model_id, torch_dtype="auto", config=config
 )
 tokenizer = AutoTokenizer.from_pretrained(model_id)
-model = prepare_for_calibration(model)
+model = replace_modules_for_calibration(model)
```
brian-dellabetta, line 23: i think this is a better name change, but might hit merge conflicts with the changes in #1608

brian-dellabetta (approved):
couple nits

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -117,6 +117,16 @@ class DatasetArguments(CustomDatasetArguments):
         default=512,
         metadata={"help": "Number of samples to use for one-shot calibration"},
     )
+    calibrate_moe_context: bool = field(
```
kylesayrs, line 120: Shouldn't this basically always be on? Is there ever a case where a user shouldn't use this?

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -189,7 +189,11 @@ def apply_recipe_modifiers(
         user_pipeline = self.dataset_args.pipeline
         modifiers = session.lifecycle.recipe.modifiers
         pipeline = CalibrationPipeline.from_modifiers(modifiers, user=user_pipeline)
-        pipeline(self.model, calibration_dataloader, self.dataset_args)
+        pipeline(
+            self.model,
+            calibration_dataloader,
+            self.dataset_args,
+        )
```
kylesayrs, line 196: nit: undo expansion

examples/quantization_w4a4_fp4/qwen_30b_a2b.py
dsikka: good catch

examples/quantizing_moe/deepseek_r1_example.py
dsikka: yes will resolve

src/llmcompressor/args/dataset_arguments.py
dsikka: when you want to call `prepare_for_calibration` and want to permanently change the module definition, as opposed to only the duration of calibration

examples/quantization_w4a4_fp4/qwen_30b_a3b.py
```diff
@@ -0,0 +1,86 @@
...
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+from llmcompressor.utils import dispatch_for_generation
+
+MODEL_ID = "Qwen/Qwen3-30B-A3B"
+
+# Load model.
+model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
+
```
brian-dellabetta, line 13: Why don't we need a 
```python
model = replace_modules_for_calibration(model)
```
line in this example?

src/llmcompressor/modeling/prepare.py
```diff
@@ -3,20 +3,53 @@
...
+                    module,
+                    "mlp",
+                    replace_Qwen3MoE(config=model.config, module=module.mlp),
+                )
+            )
+
+
+moe_context = {
+    "Qwen3MoeForCausalLM": update_qwen3_moe,
+}
```
brian-dellabetta, line 47: this is split up from the `replacements` above because we only want it to be enabled if `calibrate_moe_context=True`?

examples/quantization_w4a4_fp4/qwen_30b_a3b.py
dsikka: replace_modules_for_calibration targets permanent changes, as needed in the cases of llama4
The whole point of this pathway is to allow temporary changes during calibration

src/llmcompressor/modeling/prepare.py
dsikka: For the time being yes. We can combine once hardened

src/llmcompressor/args/dataset_arguments.py
kylesayrs: Hm, but there's no conflict between `prepare_for_calibration` and `calibrate_moe_context`, right? I think it'd also look a little confusing to calibrate an MoE model, but explicitly call
```python3
prepare_for_calibration(my_moe_model)
oneshot(my_moe_model, calibrate_moe_context=False)
```

dsikka: Would it be better to just remove it and always run with `calibrate_moe_context=True`?

We require `prepare_for_calibration` to be explicitly applied, the idea was to be the same with this moe calibration context being enabled

But yes, there is no conflict between the two. I think we can technically run deepseek with both but I haven't tested it with the context

kylesayrs: I feel that requiring this argument to be explicit is an unnecessary foot gun and adds unnecessary complication. There's nothing dangerous about always calibrating within the `calibrate_moe_context`, so I don't see why it shouldn't be always applied.

PR #1606, opened 2025-06-30
[Bugfix][Examples] drop data_collator causing failures in trl init

brian-dellabetta:
SUMMARY:
[A recent update to trl](https://github.com/huggingface/trl/pull/3076/files) no longer allows data_collator to be set when padding_free=True, causing `examples/trl_mixin/ex_trl_distillation.py` to fail. This drops the default collator, which is not needed, to allow for performant packing.

Resolves INFERENG-959

TEST PLAN:
Resolved error in `examples/trl_mixin/ex_trl_distillation.py` in local check after this update.

### DIFF
--- a/examples/trl_mixin/ex_trl_distillation.py
+++ b/examples/trl_mixin/ex_trl_distillation.py
@@ -1,5 +1,5 @@
 from sft_trainer import SFTTrainer
-from transformers import AutoModelForCausalLM, AutoTokenizer, DefaultDataCollator
+from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor.args import DatasetArguments, ModelArguments
 from llmcompressor.transformers import TextGenerationDataset


shanjiaz (approved):
Looks good to me!

kylesayrs (approved):
Kind of interesting that this is only showing up now, given that this drop in support was released in March.
https://github.com/huggingface/trl/pull/3076
https://github.com/huggingface/trl/releases

It looks like it may have been triggered by some new computation of padding_free by https://github.com/huggingface/trl/pull/3526

brian-dellabetta:
> Kind of interesting that this is only showing up now, given that this drop in support was released in March. [huggingface/trl#3076](https://github.com/huggingface/trl/pull/3076) https://github.com/huggingface/trl/releases
> 
> It looks like it may have been triggered by some new computation of padding_free by [huggingface/trl#3526](https://github.com/huggingface/trl/pull/3526)

Yeah good point, it looks like the resolution of padding_free is what has changed more than anything

PR #1622, opened 2025-07-04
[Transforms][NVFP4] Transforms + NVFP4

dsikka:
SUMMARY:
- Branch to test out NVFP4 with transforms


brian-dellabetta:
Results look good! Cut out ~50% of the degradation moving from fp16 to nvfp4

PR #1633, opened 2025-07-09
Add opt-in code coverage to test workflows

dbarbuzzi:
In order to enable the code coverage, the workflow(s) must be executed manually with the `code_coverage` input set to `true`. This will trigger two additional steps per workflow/job:

1. One step uses a new action to prepare/configure the code coverage
2. The other step uploads the code coverage results to the run as an artifact

Runs with the added conditionals removed to force-enable code coverage (since you cannot manually execute a run with `workflow_dispatch` until the workflow is on the primary branch with `workflow_dispatch` defined):

https://github.com/vllm-project/llm-compressor/actions/runs/16178848704
https://github.com/vllm-project/llm-compressor/actions/runs/16178848721

One of the transformers job test failures seems like it's related to the datasets changes around remote code, but the others I’m not sure about – though they shouldn’t be related to this PR.


shanjiaz (approved):
Thanks for adding this!

PR #1636, opened 2025-07-10
Minor speedup for `infer_quantization_format` when `save_compressed=False`

kylesayrs:
## Purpose ##
* Avoid calling `_get_unique_quant_args` in non-save compressed case, which can be expensive

## Changes ##
* Check for `not save_compressed` earlier within `infer_quantization_format` function


brian-dellabetta (approved):

kylesayrs:
Closing in favor of https://github.com/vllm-project/llm-compressor/pull/1786

PR #1637, opened 2025-07-10
[Transform] Norm fusing utilities

kylesayrs:
## Purpose ##
* Provide utilities for fusing norms and embeddings for SpinQuantModifier

## Changes ##
* Implement `center_embeddings` and `fuse_norm_linears`
  * `center_embeddings` doesn't seem to have an effect (and theoretically shouldn't have an effect, and makes the implementation less resilient), but is implemented by the SpinQuant paper. We can implement the utility here and decide to not use it later

 ## Testing ##
* Add `test_center_embeddings` and `test_fuse_norm_linears`


shanjiaz (dismissed):
Looks great!

tests/llmcompressor/modeling/test_fuse.py
```diff
@@ -0,0 +1,32 @@
...
+    center_embeddings(embedding)
+
+    assert torch.allclose(
+        embedding.weight.mean(dim=1), torch.zeros(embedding.num_embeddings), atol=1e-5
+    )
+
+
+@pytest.mark.unit
+def test_fuse_norm_linears():
+    norm = torch.nn.LayerNorm((5,))
```
brian-dellabetta, line 19: `torch.nn.LayerNorm` defaults weights to identity unless updated:
```
In [2]: a=torch.nn.LayerNorm((5,))

In [3]: a
Out[3]: LayerNorm((5,), eps=1e-05, elementwise_affine=True)

In [4]: a.weight
Out[4]: 
Parameter containing:
tensor([1., 1., 1., 1., 1.], requires_grad=True)
```
So i think this would be a more meaningful test:
```suggestion
    norm = torch.nn.LayerNorm((5,))
    norm.weight.data = torch.rand(norm.weight.shape)
```

brian-dellabetta (changes requested):
one request for unit test based on my findings [here](https://github.com/vllm-project/llm-compressor/blob/3c216dd685fdac9172213b1200be5a5ee91be532/examples/transform/spinquant_dummy.py#L85-L89)

tests/llmcompressor/modeling/test_fuse.py
kylesayrs: I totally missed this, thanks for the catch

brian-dellabetta (dismissed):

shanjiaz (approved):
Looks good to me!

PR #1651, opened 2025-07-16
[Transform] Online Rotations

kylesayrs:


kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/17246609866

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/18477528039

src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -0,0 +1,221 @@
...
+        exclude=True,
+    )
+
+    # optional override for more fine-grained control
+    # also included in recipe serialization
+    transform_config: Optional[TransformConfig] = Field(default=None, repr=False)
+
+    @field_validator("randomize", "learnable", mode="before")
+    def validate_not_implemented(cls, value, info: ValidationInfo):
+        raise NotImplementedError(f"{info.field_name} is not supported right now")
```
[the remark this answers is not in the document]
kylesayrs: Not true

kylesayrs:
Last nightly worked, but e2e failed due to model storage issues
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/18483826999

src/llmcompressor/observers/helpers.py
```diff
@@ -107,14 +114,18 @@ def _flatten_activation(value: torch.Tensor, args: QuantizationArgs):
...
 
 
 def _flatten_attention(value: torch.Tensor, args: QuantizationArgs):
+    # value.shape = (batch_size, num_heads, seq_len, head_dim)
+
     if args.strategy == QuantizationStrategy.TENSOR:
-        # (batch_size, seq_len, num_heads, head_dim)
         # (batch_size * seq_len, 1, num_heads * head_dim)
-        return value.flatten(0, 1).flatten(-2, -1).unsqueeze(-2)
+        return value.transpose(1, 2).flatten(0, 1).flatten(-2, -1).unsqueeze(-2)
```
HDCharles, line 128: nice

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -1,6 +1,13 @@
 from typing import Any, Dict, List, Optional, Set, Union
 
 import torch
+from compressed_tensors.modeling import (
+    IMPL_ATTR,
```
rahul-tuli, line 5: nit: could we use a better name for this, it's sort of unclear from the name what it is

kylesayrs:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/18663440870

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -145,6 +152,10 @@ def resolved_targets(self) -> Set[str]:
         for config_group in self.resolved_config.config_groups.values():
             for target in config_group.targets:
                 targets.add(target)
+
+        if self.resolved_config.kv_cache_scheme is not None:
+            targets.add("re:.*self_attn$")
```
brian-dellabetta, line 155: I thought we didn't want to do this to find target attention modules, instead looping over `model.modules()` and checking `is_attention(module)` as the true check?

kylesayrs: I decided to leave decoupling `"re:.*self_attn$"` and `is_attention_module` for a later PR.

brian-dellabetta: i think it'd be worth leaving a todo or note in that case

experimental/llama3_attention.py
```diff
@@ -0,0 +1,87 @@
+from datasets import load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
```
dsikka, line 4: why experimental?

kylesayrs: Because vLLM doesn't have integration support for this yet.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -260,60 +269,51 @@ def _initialize_observers(self, module: torch.nn.Module):
...
             initialize_observer(module, base_name="output")
 
-    def _initialize_hooks(self, model: torch.nn.Module) -> Set[RemovableHandle]:
+    def _initialize_hooks(self, module: torch.nn.Module) -> Set[RemovableHandle]:
         hooks = set()
-        for _, module in match_named_modules(model, self.resolved_targets, self.ignore):
-            if not hasattr(module, "quantization_scheme"):
-                continue
+        if not hasattr(module, "quantization_scheme"):
+            hooks
```
brian-dellabetta: missing a return here?

experimental/llama3_attention.py
```diff
@@ -0,0 +1,87 @@
...
+model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+# Select calibration dataset.
+DATASET_ID = "HuggingFaceH4/ultrachat_200k"
+DATASET_SPLIT = "train_sft"
+
+# Select number of samples. 512 samples is a good place to start.
+# Increasing the number of samples can improve accuracy.
+NUM_CALIBRATION_SAMPLES = 512
```
dsikka, line 20: do we know how much data is actually needed for getting decent results?

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -93,7 +88,8 @@ def call_observer(
         if should_calculate_qparams:
             scale, zero_point = observer(value)
             update_offload_parameter(module, f"{base_name}_scale", scale)
-            update_offload_parameter(module, f"{base_name}_zero_point", zero_point)
+            if hasattr(module, f"{base_name}_zero_point"):
+                update_offload_parameter(module, f"{base_name}_zero_point", zero_point)
```
dsikka, line 92: we usually dont run asym quant - why wasn't this a problem before?

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -202,60 +199,16 @@ def calibrate_output_hook(module: Module, _args: Any, output: torch.Tensor):
...
-    if (
-        scheme is None
-        or not is_kv_cache_quant_scheme(scheme)
-        or isinstance(existing_kv_cache, QuantizedKVParameterCache)
-    ):
-        return
-
-    quantized_kv_cache = QuantizedKVParameterCache(scheme.output_activations)
-    setattr(module, "kv_cache", quantized_kv_cache)
+def calibrate_value_hook(module: Module, value_states: torch.Tensor):
```
dsikka, line 210: so clean

src/llmcompressor/modifiers/quantization/quantization/mixin.py
kylesayrs: Ah! Not sure how that sneaked through

experimental/llama3_attention.py
kylesayrs: 512 is enough for the basic GSM8K evals

src/llmcompressor/modifiers/quantization/calibration.py
kylesayrs: This is added because KV cache quantization is weird: it's the only scheme which does not have a compressor. For that reason, it's the only scheme where we cannot force zero points (vllm throws an error if zero points are present)

### What should happen ###
1. KV cache quant is initialized with forced zero points
2. Calibration happens and zero points are updated (and stay zero if symmetric)
3. Model does not have compressed weights and is saved in frozen state
4. vLLM loads and throws away zero points if symmetric (only symmetric is implemented atm)

(alternatively, in step (3) we write an "attention" compressor which throws away zero points)

### What this refactor does to avoid this ###
1. KV cache quant is initialized without forced zero points
2. Calibration happens and zero points are updated, only if they're present (they're not unless asymmetric)
3. Model does not have compressed weights and is saved in frozen state (without zero points)
4. vLLM loads scales only (since only symmetric kv cache quantization is supported atm)

src/llmcompressor/modifiers/quantization/calibration.py
kylesayrs: This comes from aligning with the patterns and abstractions we've already created, not creating new ones which don't integrate and therefore don't provide as many features.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
kylesayrs: Thanks for your comment, this is addressed now.

PR #1658, opened 2025-07-18
[Examples] Fix ignore layers for Qwen2.5-VL

SorenDreano:
SUMMARY:
Oneshot() prints the following warning:
Some layers that were to be ignored were not found in the model: {'re:visual.*'} as the vision module in Qwen2.5-vl is in model.visual

TEST PLAN:
Removing re:visual.* from the list of modules to be ignored did not result in any warning or error

### DIFF
--- a/examples/multimodal_vision/qwen_2_5_vl_example.py
+++ b/examples/multimodal_vision/qwen_2_5_vl_example.py
@@ -73,7 +73,7 @@ recipe = [
     GPTQModifier(
         targets="Linear",
         scheme="W4A16",
-        ignore=["lm_head", "re:visual.*", "re:model.visual.*"],
+        ignore=["lm_head", "re:model.visual.*"],
     ),
 ]
 


kylesayrs:
Hi @SorenDreano,

Thank you for your contribution! However, in this case, these ignore names are kept in place in order to support older checkpoints created prior to https://github.com/huggingface/transformers/pull/37033.

Perhaps instead of removing them, we can add a comment in the example explaining why they exist.

EDIT: See @brian-dellabetta's review

examples/multimodal_vision/qwen_2_5_vl_example.py
```diff
@@ -73,7 +73,7 @@ def data_collator(batch):
     GPTQModifier(
         targets="Linear",
         scheme="W4A16",
-        ignore=["lm_head", "re:visual.*", "re:model.visual.*"],
+        ignore=["lm_head", "re:model.visual.*"],
```
brian-dellabetta, line 76: Since we want to catch both `visual.` and `model.visual.`, we can use an OR condition in our regex like
```suggestion
        # NOTE: catch anything beginning with `model.visual.` or `visual.`
        # to account for different behavior across transformers versions
        ignore=["lm_head", "re:(model\.|)visual\..*"],
```
This will prevent the warning from appearing

SorenDreano:
Sorry for the delay. I would go with readability, even if it costs a Warning, compared to correctness.

PR #1674, opened 2025-07-23
Add code coverage combination and reporting

dbarbuzzi:
SUMMARY:
Update code coverage support to include results in the job summaries, as well as combine the results from the multiple jobs in the "test-checks.yaml" workflow.

Some open questions:
* Should the reporting in the summary skip empty files (files with no Python code/statements)?
  * Current: yes
* Should the reporting use different types of sorting (e.g., by coverage)?
  * Current: file path/name (tool default)
* Should the GitHub jobs have coverage summaries for intermediate jobs or just the combined results?
  * Current: all (though it does seem more verbose than is useful)
  * Note: The coverage result artifacts will still be attached to the jobs even without summaries, they are currently required for the job that combines the results.
* Should coverage be enabled by default for eithe of the workflows (here or another PR)?
  * Current: no (adds ~1 minute to base test run wall time and ~6-7 minutes to transformers run time)

TEST PLAN:
For examples, check these runs and the summary for each job in the run:

* base: https://github.com/vllm-project/llm-compressor/actions/runs/16471546227 (49% combined coverage)
* transformers: https://github.com/vllm-project/llm-compressor/actions/runs/16472609090 (65% coverage)

### DIFF
--- a/examples/multimodal_vision/llama4_example.py
+++ b/examples/multimodal_vision/llama4_example.py
@@ -3,7 +3,7 @@ from datasets import load_dataset
 from transformers import Llama4ForConditionalGeneration, Llama4Processor
 
 from llmcompressor import oneshot
-from llmcompressor.modeling import prepare_for_calibration
+from llmcompressor.modeling import replace_modules_for_calibration
 from llmcompressor.modifiers.quantization import GPTQModifier
 
 # Select model and load it.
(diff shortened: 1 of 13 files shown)


brian-dellabetta (approved):
thank you!

derekk-nm:
Nice that the coverage markdown is compatible w/ GitHub markdown.

PR #1683, opened 2025-07-25
Fix `site_url` (and therefore `sitemap.xml`)

hmellor:
Same as in https://github.com/vllm-project/vllm/pull/21564


dsikka:
@aireilly

PR #1692, opened 2025-07-30
[Tests] Fix GPTQ Tests

dsikka:
SUMMARY:
- The current tests are failing because when loading the tinystories model, the lm_head is ending up with device type "meta"
- This model is generally problematic so we swap to use TinyLlama 
- With the size of the model being large, we target just one layer for quantization to contain runtime, while updating the asserts to be reflective of the just one layer being quantized 

TESTING:
- All tests pass with these changes

### DIFF
--- a/tests/llmcompressor/transformers/gptq/test_oneshot.py
+++ b/tests/llmcompressor/transformers/gptq/test_oneshot.py
@@ -20,14 +20,15 @@ quant_stage:
                         type: "int"
                         symmetric: true
                         strategy: "channel"
-                    targets: ["Linear"]
+                    targets: ["re:.*model.layers.2.self_attn.q_proj$"]
 """
 
 recipe_modifier_full = GPTQModifier(
     ignore=["lm_head"],
     config_groups={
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):

kylesayrs (changes requested):
It'd be really nice if we could maintain our own version of this model, or use the smaller version. This test is less complete now

https://huggingface.co/nm-testing/llama2.c-stories15M

I can make a copy later today

dsikka:
> It'd be really nice if we could maintain our own version of this model, or use the smaller version. This test is less complete now
> 
> https://huggingface.co/nm-testing/llama2.c-stories15M
> 
> I can make a copy later today

Not really. The point of this test is to check if gptq was applied correctly. I would argue this is a better test as it checks for gptq being applied but using a case which is usually more error-prone.  We have plenty of cases where we check for quantization in the entire model.

We can maintain a separate model definition but that is separate from the usefulness of this test.

kylesayrs (approved):
Fair enough!

Issue #1688, opened 2025-07-29
Add an AWQ lm-eval test


dsikka:
**Is your feature request related to a problem? Please describe.**
- Similar to the existing test cases, add an AWQ test case that evaluates  the model through lm-eval
- The test case should apply an AWQ recipe and then evaluate the mode
- The model should be a meta-llama/Meta-Llama-3-8B-Instruct

**Describe the solution you'd like**
- Solution should follow a similar example: 
https://github.com/vllm-project/llm-compressor/tree/main/tests/lmeval/configs

**Additional context**
- This will require first running lm-eval using hf to first asses the model, before adding the example

aladerran:
I am on it~

dsikka:
Thank you!

dsikka:
Completed by: https://github.com/vllm-project/llm-compressor/pull/1702

PR #1702, opened 2025-08-01
[WIP] Add an awq_asym lm-eval test

aladerran:
SUMMARY:
In response to https://github.com/vllm-project/llm-compressor/issues/1688, add an awq_asym lm-eval test config.

TEST PLAN:
```
export CADENCE=weekly
export TEST_DATA_FILE=tests/lmeval/configs/w4a16_awq_asym.yaml
python -m pytest tests/lmeval/test_lmeval.py::TestLMEval::test_lm_eval -v
```

MISC:
As Llama is a gated model, I'm using a local model path instead of meta-llama/Meta-Llama-3-8B-Instruct during local testing. I'm not sure if this might affect the test flow / accuracy. It would be great if you could help me verify it. That's also why I've marked it as WIP.

### DIFF
--- a/tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_asym.yaml
+++ b/tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_asym.yaml
@@ -1,0 +1,22 @@
+quant_stage:
+  quant_modifiers:
+    AWQModifier:
+      mappings:
+        - smooth_layer: "re:.*input_layernorm$"
+          balance_layers: ["re:.*q_proj$", "re:.*k_proj$", "re:.*v_proj$"]
+        - smooth_layer: "re:.*v_proj$"
+          balance_layers: ["re:.*o_proj$"]
+        - smooth_layer: "re:.*post_attention_layernorm$"
+          balance_layers: ["re:.*gate_proj$", "re:.*up_proj$"]
+        - smooth_layer: "re:.*up_proj$"
+          balance_layers: ["re:.*down_proj$"]
+      ignore: ["lm_head"]
+      config_groups:
+        group_0:
+          weights:
+            num_bits: 4
+            type: "int"
+            symmetric: false
+            strategy: "group"
+            group_size: 128
+          targets: ["Linear"]
--- a/tests/lmeval/configs/w4a16_awq_asym.yaml
+++ b/tests/lmeval/configs/w4a16_awq_asym.yaml
@@ -1,0 +1,13 @@
+cadence: "weekly"
+model: meta-llama/Meta-Llama-3-8B-Instruct
+scheme: W4A16_ASYM
+recipe: tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_asym.yaml
+dataset_id: HuggingFaceH4/ultrachat_200k
+dataset_split: train_sft
+lmeval:
+  model: "vllm"
+  model_args:
+    add_bos_token: true
+  metrics:
+    exact_match,flexible-extract: 0.73
+    exact_match,strict-match: 0.73


aladerran:
I guess Gemini's comments are right, will check.

tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_asym.yaml
```diff
@@ -0,0 +1,22 @@
...
+    AWQModifier:
+      mappings:
+        - smooth_layer: "re:.*input_layernorm$"
+          balance_layers: ["re:.*q_proj$", "re:.*k_proj$", "re:.*v_proj$"]
+        - smooth_layer: "re:.*v_proj$"
+          balance_layers: ["re:.*o_proj$"]
+        - smooth_layer: "re:.*post_attention_layernorm$"
+          balance_layers: ["re:.*gate_proj$", "re:.*up_proj$"]
+        - smooth_layer: "re:.*up_proj$"
+          balance_layers: ["re:.*down_proj$"]
```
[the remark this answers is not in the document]
brian-dellabetta: yeah we can probably remove mappings from the yaml

tests/lmeval/configs/w4a16_awq_asym.yaml
```diff
@@ -0,0 +1,13 @@
+cadence: "weekly"
+model: meta-llama/Meta-Llama-3-8B-Instruct
+scheme: W4A16_ASYM
+recipe: tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_asym.yaml
+dataset_id: HuggingFaceH4/ultrachat_200k
+dataset_split: train_sft
+lmeval:
+  model: "vllm"
+  model_args:
+    add_bos_token: true
```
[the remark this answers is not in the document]
brian-dellabetta: we use the HF one, not the vllm one. so this should be ok

tests/lmeval/configs/w4a16_awq_asym.yaml
```diff
@@ -0,0 +1,13 @@
+cadence: "weekly"
+model: meta-llama/Meta-Llama-3-8B-Instruct
+scheme: W4A16_ASYM
+recipe: tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_asym.yaml
+dataset_id: HuggingFaceH4/ultrachat_200k
+dataset_split: train_sft
+lmeval:
+  model: "vllm"
```
brian-dellabetta: we use the HF model inside lm_eval for these tests
```suggestion
```

tests/lmeval/configs/w4a16_awq_asym.yaml
```diff
@@ -0,0 +1,13 @@
...
+  model_args:
+    add_bos_token: true
+  metrics:
+    exact_match,flexible-extract: 0.73
+    exact_match,strict-match: 0.73
```
brian-dellabetta: just confirming, did you get these as the end result of running the test?

tests/lmeval/configs/w4a16_awq_asym.yaml
aladerran: When using HF, it raises ValueError with ```Decompression of packed zero points is currently not supported``` for awq-asym model.

Could you try to reproduce? Here is my command:
```
lm_eval --model hf \     
  --model_args pretrained="./Meta-Llama-3-8B-Instruct-awq-asym",add_bos_token=true \
  --tasks gsm8k \
  --num_fewshot 5 \
  --batch_size 'auto'
  ```

File: tests/lmeval/configs/w4a16_awq_asym.yaml (hunk 4 above)
Comment (aladerran) [reply tests/lmeval/configs/w4a16_awq_asym.yaml]: Yes, with:
```
import os
import lm_eval

model_path = os.path.expanduser("Meta-Llama-3-8B-Instruct-awq-asym")

    model="vllm",
    model_args={
        "pretrained": model_path,
        "enforce_eager": True,
        "add_bos_token": True,
    },
    tasks=["gsm8k"],
    num_fewshot=5,
    batch_size="auto",
    log_samples=False,
)

print(results)
 
{'gsm8k': {'exact_match,strict-match': 0.733131159969674, 'exact_match_stderr,strict-match': 0.012183780551887954, 'exact_match,flexible-extract': 0.7354056103108415, 'exact_match_stderr,flexible-extract': 0.012150554001563226, 'alias': 'gsm8k'}}

...
 ```

tests/lmeval/configs/w4a16_awq_asym.yaml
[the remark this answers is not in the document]
aladerran: See below. If we have to use ```vllm```, I am considering add ```enforce_eager: true```, which may affect test result.

tests/lmeval/configs/w4a16_awq_asym.yaml
dsikka: Can we test no asym? This is expected as asym decompression is not yet supported

dsikka (changes requested):
Please use hf for the lmeval

tests/lmeval/configs/w4a16_awq_asym.yaml
aladerran: Np. I found awq-sym model has lower accuracy though.

aladerran:
@dsikka Thanks for the info. Updated.

Accuracy test for hf awq-sym:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7058|±  |0.0126|
|     |       |strict-match    |     5|exact_match|↑  |0.7058|±  |0.0126|

tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_sym.yaml
brian-dellabetta, line 1: rather than in a new folder, this can live in `e2e/vLLM/recipes/WNA16`

brian-dellabetta (commented):
Thanks for updating, one comment on file location and then I think this is good to go!

tests/e2e/vLLM/recipes/AWQ/recipe_w4a16_awq_sym.yaml
aladerran: np

tests/lmeval/configs/w4a16_awq_sym.yaml
```diff
@@ -0,0 +1,10 @@
+cadence: "weekly"
+model: meta-llama/Meta-Llama-3-8B-Instruct
+scheme: W4A16
```
dsikka, line 3: We dont need a scheme here as we're using a recipe in this case

PR #1713, opened 2025-08-07
[Quantization] Support mixed-precision compression

dsikka:
SUMMARY:

TEST PLAN:

### DIFF
--- a/src/llmcompressor/transformers/compression/quantization_format.py
+++ b/src/llmcompressor/transformers/compression/quantization_format.py
@@ -5,7 +5,7 @@ from compressed_tensors.config import SparsityStructure
 from compressed_tensors.quantization import QuantizationStrategy, QuantizationType
 from compressed_tensors.quantization.utils import is_module_quantized
 
-__all__ = ["infer_quantization_format"]
+__all__ = ["infer_quantization_format", "infer_per_module_quantization"]
 
 
 def infer_quantization_format(
@@ -95,6 +95,60 @@ def infer_quantization_format(
     else:
         # format will be inferred from config
         return None
+
+
+def _get_quant_method(input_args, weight_args, sparsity_structure):
+    is_24_structure = (
+        SparsityStructure(sparsity_structure) == SparsityStructure.TWO_FOUR
+    )
+    is_weight_only = weight_args is not None and input_args is None
+
+    if weight_args.num_bits == 4 and weight_args.type == QuantizationType.FLOAT.value:
+        return CompressionFormat.nvfp4_pack_quantized
+
+    if is_weight_only:  # w4a16 and w8a16
+        is_valid_pack = (
+            weight_args.num_bits in [4, 8]
+            and weight_args.type == QuantizationType.INT.value
+        )
+        if not is_valid_pack:  # packing only valid for int4 and int 8
+            return CompressionFormat.naive_quantized
+        if is_24_structure:
+            if (
+                weight_args.strategy is not QuantizationStrategy.CHANNEL.value
+                and weight_args.strategy is not QuantizationStrategy.GROUP.value
+            ):
+                # marlin24 kernel only applicable for channel/group quantization
+                return CompressionFormat.pack_quantized
+            return CompressionFormat.marlin_24
+        return CompressionFormat.pack_quantized
+
+    else:  # w8a8 float and int
+        if (
+            weight_args.type == QuantizationType.FLOAT.value
+            and weight_args.num_bits == 8
+        ):
+            return CompressionFormat.float_quantized
+        if weight_args.type == QuantizationType.INT.value:
+            return CompressionFormat.int_quantized
+
+        return CompressionFormat.naive_quantized
+
+
+def infer_per_module_quantization(model, sparsity_structure):
+    unique_formats = []
+    for submodule in model.modules():
+        if is_module_quantized(submodule):
+            weight_scheme = submodule.quantization_scheme.weights
+            input_scheme = submodule.quantization_scheme.input_activations
+            compression_format = _get_quant_method(
+                input_scheme, weight_scheme, sparsity_structure
+            )
+            print(compression_format)
+            submodule.quantization_scheme.format = compression_format.value
+            if compression_format not in unique_formats:
+                unique_formats.append(compression_format)
+    return unique_formats
 
 
 def _get_unique_quant_args(model):
--- a/src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
+++ b/src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
@@ -19,6 +19,7 @@ from transformers import PreTrainedModel
 from llmcompressor.core import active_session
 from llmcompressor.pytorch.model_load.helpers import copy_python_files_from_model_cache
 from llmcompressor.transformers.compression.quantization_format import (
+    infer_per_module_quantization,
     infer_quantization_format,
 )
 from llmcompressor.transformers.compression.sparsity_metadata_config import (
@@ -228,10 +229,19 @@ def get_model_compressor(
                 SparsityConfigMetadata.infer_sparsity_structure(model)
             )
 
+    """
     quantization_format: Optional[CompressionFormat] = infer_quantization_format(
         model=model,
         quantization_format=quantization_format,
         save_compressed=save_compressed,
+        sparsity_structure=None
+        if sparsity_config is None
+        else sparsity_config.sparsity_structure,
+    )
+    """
+
+    quantization_format = infer_per_module_quantization(
+        model,
         sparsity_structure=None
         if sparsity_config is None
         else sparsity_config.sparsity_structure,


src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -228,13 +228,15 @@ def get_model_compressor(
...
             )
 
-    quantization_format: Optional[CompressionFormat] = infer_quantization_format(
-        model=model,
-        quantization_format=quantization_format,
-        save_compressed=save_compressed,
-        sparsity_structure=None
-        if sparsity_config is None
-        else sparsity_config.sparsity_structure,
+    quantization_format: Optional[CompressionFormat] = (
```
kylesayrs: Type hint does not match return type of function

src/llmcompressor/transformers/compression/quantization_format.py
```diff
@@ -1,22 +1,70 @@
...
+    QuantizationType,
+)
 from compressed_tensors.quantization.utils import is_module_quantized
 
-__all__ = ["infer_quantization_format"]
+__all__ = ["infer_per_module_quantization_format"]
 
 
-def infer_quantization_format(
+def _get_quant_method(
```
kylesayrs: This function is called `_get_quant_method`, but returns a format

src/llmcompressor/transformers/compression/quantization_format.py
```diff
@@ -44,74 +94,26 @@ def infer_quantization_format(
...
-                if weight_scheme not in quant_info_weight:
-                    quant_info_weight.append(weight_scheme)
-            if input_scheme is not None:
-                if input_scheme not in quant_info_inputs:
-                    quant_info_inputs.append(input_scheme)
-
-    return quant_info_weight, quant_info_inputs
+            if weight_scheme is None:
+                continue  # no weight quant - nothing to compress
+            compression_format = _get_quant_method(
```
kylesayrs: Why are we not inferring from `scheme.format`? Doesn't this override anything the user might have passed?

kylesayrs: Rather than waiting until the end to do inference, why not infer a format at the source?

```python3
class QuantizationScheme:
    format: Optional[CompressionFormat] = None

    @validate_field("format")
    def validate_format(self, value):
        if self.weights and self.input_activations ...
        
        inferred_format = ...
        if value is not None and value != inferred_format:
            logger.warn_once("Consider using inferred scheme")
            
        return value or inferred_format
```

kylesayrs (changes requested):

src/llmcompressor/transformers/compression/quantization_format.py
dsikka: LLM Compressor interprets the compression format at the time of the compression based on the quant args. This is our current lifecycle.  We can still override using a global format.

We can support overriding on a per module basis in a follow-up. The goal of this PR isn't to support per-module format overriding. The goal is to support the most common pathway to start, which is where we determine the format during compresssion time.

kylesayrs: Okay. I'll also mention that this way of doing it also adds unnecessary runtime and redundant work. `submodule.quantization_scheme` is a pointer to a scheme, and this function modifies each submodule.

In effect, this function overwrites `submodule.quantization_scheme.format` for each module that shares that scheme.

src/llmcompressor/transformers/compression/quantization_format.py
```diff
@@ -44,74 +94,26 @@ def infer_quantization_format(
...
+                input_scheme, weight_scheme, sparsity_structure
+            )
+            submodule.quantization_scheme.format = compression_format.value
+            if compression_format not in unique_formats:
+                unique_formats.append(compression_format)
+    if len(unique_formats) > 0:
+        return unique_formats
+    return None
```
kylesayrs, line 114: Is it possible to stop supporting nullable formats? Supporting `CompresssionFormat.dense` and `None` leads to a  lot of redundant parsing logic and loose typing

src/llmcompressor/transformers/compression/quantization_format.py
```diff
@@ -44,74 +94,26 @@ def infer_quantization_format(
...
-                    return CompressionFormat.int_quantized
-
-            return CompressionFormat.naive_quantized
-    else:
-        # format will be inferred from config
+    if not save_compressed:
         return None
 
+    if quantization_format:
+        return [quantization_format]
```
kylesayrs, line 84: It might be nice to warn a user if their chosen format does not match the recommended formats

kylesayrs: Also, if the user passes mixed, subsequent steps to initialize quantization compressors will fail

src/llmcompressor/transformers/compression/quantization_format.py
```diff
@@ -1,22 +1,70 @@
...
+            and weight_args.num_bits == 8
+        ):
+            return CompressionFormat.float_quantized
+        if weight_args.type == QuantizationType.INT.value:
+            return CompressionFormat.int_quantized
+
+        return CompressionFormat.naive_quantized
+
+
+def infer_per_module_quantization_format(
```
kylesayrs: This function does more than just infer format, it also `sets` the format of the modules. Can you update the function name to reflect this?

src/llmcompressor/transformers/compression/quantization_format.py
dsikka: This is the same runtime that exists currently. We are not adding anything extra.
What you are suggesting are lifecycle changes, which I am not targeting at the moment.

src/llmcompressor/transformers/compression/quantization_format.py
dsikka: We really only support Dense iirc. None just gets mapped to Dense. I dont totally understand why we need a dense compressor ... but a format as "dense" make sense as a designation for no compression

src/llmcompressor/transformers/compression/quantization_format.py
```diff
@@ -1,22 +1,71 @@
...
+            and weight_args.num_bits == 8
+        ):
+            return CompressionFormat.float_quantized
+        if weight_args.type == QuantizationType.INT.value:
+            return CompressionFormat.int_quantized
+
+        return CompressionFormat.naive_quantized
+
+
+def infer_and_set_per_module_quantization_format(
```
kylesayrs, line 58: It seems like this function only sets the scheme on the modules if a `quantization_format` is not passed

kylesayrs (approved):

src/llmcompressor/transformers/compression/quantization_format.py
kylesayrs: It's be nice if we could stop using None and always return Dense instead.

src/llmcompressor/transformers/compression/quantization_format.py
dsikka: Yeah in the current lifecycle, this `quantization_format` is how we've been overriding the inferred format if given by the user.  It's kinda silly because it is passed only to be returned unchanged by get_compressor when really, you probably dont need a call to `infer_and_set_per_module_quantization_format` at all if you're passing in a quantization_format to begin with.
 
We can keep it so that the global format would also override any per-module defined format.  Which I think would make sense.

src/llmcompressor/transformers/compression/quantization_format.py
dsikka: Yeah I'll verify this is how it's being mapped when I move this function to compressed-tensors and clean some of this up.

yiliu30:
Hi @dsikka  do we have an e2e example for mixed precision?

dsikka:
> Hi @dsikka do we have an e2e example for mixed precision?

Not yet, I will put up a follow-up PR with examples!

dsikka:
> Hi @dsikka do we have an e2e example for mixed precision?

Hey! You can refer to some examples here: https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_non_uniform


To run the model produced from the nvfp4 + fp8 example, you will need vllm main

yiliu30:
> > Hi @dsikka do we have an e2e example for mixed precision?
> 
> Hey! You can refer to some examples here: https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_non_uniform
> 
> To run the model produced from the nvfp4 + fp8 example, you will need vllm main

Hi @dsikka, for the [quantization_nvfp4_fp8.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_non_uniform/quantization_nvfp4_fp8.py),
 I guess we need to add `CompressionFormat.mixed_precision.value` to `_ACTIVATION_QUANTIZATION_FORMATS` to make it work?

https://github.com/vllm-project/vllm/blob/8ea0c2753a273e24957ab4587c200a3254ebe970/vllm/model_executor/layers/quantization/compressed_tensors/utils.py#L13-L20

dsikka:
> > > Hi @dsikka do we have an e2e example for mixed precision?
> > 
> > 
> > Hey! You can refer to some examples here: https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_non_uniform
> > To run the model produced from the nvfp4 + fp8 example, you will need vllm main
> 
> Hi @dsikka, for the [quantization_nvfp4_fp8.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_non_uniform/quantization_nvfp4_fp8.py), I guess we need to add `CompressionFormat.mixed_precision.value` to `_ACTIVATION_QUANTIZATION_FORMATS` to make it work?
> 
> https://github.com/vllm-project/vllm/blob/8ea0c2753a273e24957ab4587c200a3254ebe970/vllm/model_executor/layers/quantization/compressed_tensors/utils.py#L13-L20

You should be able to do it without it!
MoE support is still being worked on however.

yiliu30:
> > > > Hi @dsikka do we have an e2e example for mixed precision?
> > > 
> > > 
> > > Hey! You can refer to some examples here: https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_non_uniform
> > > To run the model produced from the nvfp4 + fp8 example, you will need vllm main
> > 
> > 
> > Hi @dsikka, for the [quantization_nvfp4_fp8.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_non_uniform/quantization_nvfp4_fp8.py), I guess we need to add `CompressionFormat.mixed_precision.value` to `_ACTIVATION_QUANTIZATION_FORMATS` to make it work?
> > https://github.com/vllm-project/vllm/blob/8ea0c2753a273e24957ab4587c200a3254ebe970/vllm/model_executor/layers/quantization/compressed_tensors/utils.py#L13-L20
> 
> You should be able to do it without it! MoE support is still being worked on however.

It works, thanks!

PR #1719, opened 2025-08-11
Update dependency lower bounds

dhuangnm:
SUMMARY:
To update dependency lower bounds to proper versions. Some of the versions were too old and can cause issues to install or run tests.

Will have another PR to set the upper bounds once llmcompressor is ready for next release.

TEST PLAN:
Ran nightly tests using the lower bounds: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/16884522733

Corresponding PR to update test workflow is here: https://github.com/neuralmagic/llm-compressor-testing/pull/116


setup.py
```diff
@@ -111,17 +111,17 @@ def localversion_func(version: ScmVersion) -> str:
...
+        "tqdm>=4.66.3",
         # torch 1.10 and 1.11 do not support quantized onnx export
-        "torch>=1.7.0,!=1.10,!=1.11",
-        "transformers>4.0",
-        "datasets>=3.0.0",
-        "accelerate>=0.20.3,!=1.1.0",
+        "torch>=2.7.0",
+        "transformers>=4.53.0",
+        "datasets>=4.0.0",
+        "accelerate>=1.6.0",
```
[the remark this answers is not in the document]
brian-dellabetta: Gemini's comment here looks out of date, checked all the versions and confirmed

kylesayrs (approved):

setup.py
```diff
@@ -111,17 +111,17 @@ def localversion_func(version: ScmVersion) -> str:
...
-        "tqdm>=4.0.0",
+        "numpy>=2.0.0,<2.3.0",
+        "requests>=2.32.2",
+        "tqdm>=4.66.3",
         # torch 1.10 and 1.11 do not support quantized onnx export
-        "torch>=1.7.0,!=1.10,!=1.11",
-        "transformers>4.0",
-        "datasets>=3.0.0",
-        "accelerate>=0.20.3,!=1.1.0",
+        "torch>=2.7.0",
```
kylesayrs, line 120: Is this as low as we can go? We should try to support 2.6 if we can

kylesayrs (dismissed):

brian-dellabetta (approved):
This drops support of numpy<2, torch<2, and adds tighter bounds to core dependencies. I am in favor of this, but I thought this was a design choice at least for transformers dependency version.

setup.py
dhuangnm: Kicked off a run with torch 2.6: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/16887688359, will see if the tests can run through. If so, I'll update torch to >=2.6.0.

dhuangnm:
> This drops support of numpy<2, torch<2, and adds tighter bounds to core dependencies. I am in favor of this, but I thought this was a design choice at least for transformers dependency version.

There were some install issues when using versions lower than numpy 2 and torch 2. Those versions were too old (>1 year) too. I think it made sense to use >2 for them.

setup.py
dhuangnm: It looks 2.6 also ran through. I'll update the required version to 2.6.

brian-dellabetta (dismissed):

setup.py
dhuangnm: Ran some more testing and it looks that lm eval tests failed with 2.6 with "RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::BFloat16 != c10::Float8_e4m3fn". We probably have to revert back to 2.7. Running tests again to confirm.

dhuangnm: Actually the error seems being introduced this couple of days and not related to torch. Can reproduce using the latest main from ct and lc. Need more investigating.

brian-dellabetta: I saw this dtype error recently as well, do you know which test? I can try to reproduce

dhuangnm: Yep I filed https://issues.redhat.com/browse/INFERENG-1536 to track it. The ticket has the details.

dsikka (changes requested):
Please run e2e and lm-eval as well

dhuangnm:
> Please run e2e and lm-eval as well

After discussions we'll run the lm-eval/nightly tests using the lower bounds. The e2e tests cannot use the lower bounds directly as they're conflicting with the vllm nightly.

setup.py
dsikka: 2.6 is missing many fp8 operations. So we would have to keep it at 2.7

brian-dellabetta (approved):

setup.py
```diff
@@ -111,17 +111,16 @@ def localversion_func(version: ScmVersion) -> str:
...
-        "torch>=1.7.0,!=1.10,!=1.11",
-        "transformers>4.0",
-        "datasets>=3.0.0",
-        "accelerate>=0.20.3,!=1.1.0",
+        "numpy>=2.0.0,<2.3.0",
+        "requests>=2.32.2",
+        "tqdm>=4.66.3",
+        "torch>=2.7.0",
+        "transformers>=4.53.0",
+        "datasets>=4.0.0",
```
dsikka, line 122: With your upper bound, we could only support one version of datasets?

dhuangnm: Yeah there were test failures with "datasets<=3.6.0".

PR #1726, opened 2025-08-13
Make calibration quantization configurable

shanjiaz:
SUMMARY:
Allow activation quantization to be configurable in sequential/layer_squential pipeline. Usage:

``` python
oneshot(
        model=model,
        dataset=ds,
        recipe=recipe,
        quantization_aware_calibratio=False,
    )
```

TEST PLAN:
Tested on w8a8 and nvfp4 schemes, using both quantizationModifier and GPTQModifier(when available), there are no major differences.  See this [notion page](https://www.notion.so/QDQ-24830c7e73f3809a9942c0f5442d9924?source=copy_link) for more information. 

Scheme | Modifier | Activation Quant. | Winogrande Acc (±stderr) | MMLU Acc (±stderr)
-- | -- | -- | -- | --
w8a8 | gptq | disabled | 0.7609 ± 0.0121 | 0.6629 ± 0.0038
w8a8 | gptq | enabled | 0.7609 ± 0.0121 | 0.6629 ± 0.0038
w8a8 | quantization | disabled | 0.7609 ± 0.0120 | 0.6629 ± 0.0038
w8a8 | quantization | enabled | 0.7609 ± 0.0120 | 0.6629 ± 0.0038
nvfp4 | quantization | disabled | 0.7490 ± 0.0122 | 0.6514 ± 0.0038
nvfp4 | quantization | enabled | 0.7514 ± 0.0121 | 0.6518 ± 0.0038
w4a16 | awqmodifier | disabled | 0.7490 ± 0.0122 | 0.6481 ± 0.0038
w4a16 | awqmodifier | enabled | 0.7506 ± 0.0122 | 0.6488 ± 0.0038


src/llmcompressor/args/dataset_arguments.py
```diff
@@ -208,3 +208,11 @@ class DatasetArguments(CustomDatasetArguments):
             "definition"
         },
     )
+    disable_quantization: bool = field(
+        default=True,
+        metadata={
+            "help": "Whether to disable activation quantization during calibration in "
```
brian-dellabetta: this would disable quantization just for activations, while leaving weights quantized? Is that the previous behavior or is this a breaking change?

shanjiaz: Good question! This would disabled quantization just for activation. Weights are quantized separately before this. This was the behavior before, not a breaking change. 🫡

brian-dellabetta: ok cool! i misread the diff, thought it was a breaking change

dsikka (commented):
General default behaviour we should support:
- If running activation quantization, should run QDQ (enabled through RTN/QuantModifier)
- If running GPTQ, should disable QDQ
- For AWQ, I believe we want the same behaviour as gptq?

src/llmcompressor/args/dataset_arguments.py
shanjiaz: Haha no worries!

shanjiaz:
> General default behaviour we should support:
> 
> * If running activation quantization, should run QDQ (enabled through RTN/QuantModifier)
> * If running GPTQ, should disable QDQ
> * For AWQ, I believe we want the same behaviour as gptq?

Looks like it's already disabled for gptq [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/gptq/base.py#L184) at least? We can talk about this more tomorrow during deep dive.

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -75,7 +75,9 @@ def __call__(
 
         with contextlib.ExitStack() as stack:
             stack.enter_context(calibration_forward_context(model))
-            stack.enter_context(DisableQuantization(model))
+            # Optionally disable quantization
+            if not dataset_args.quantization_aware_calibration:
+                stack.enter_context(DisableQuantization(model))
```
brian-dellabetta: if `quantization_aware_calibration` defaults to True, doesn't this mean the default behavior is now different?

shanjiaz: yeah! I think we had calibration disabled in the past.

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -181,7 +177,6 @@ def on_start(self, state: State, event: Event, **kwargs):
         QuantizationMixin.start_calibration(self, state.model)
         # Unlike qmod, do not quantize as we calibrate
         # This choice does not seem to have a meaningful impact on accuracy
-        state.model.apply(disable_quantization)
```
kylesayrs, line 184: We've done evals with this removed, right?

shanjiaz: yeah the evals I did last night with gptqModifier was with this removed.

shanjiaz:
Failed test is unrelated. It's also failing on main.

dsikka (approved):
lgtm, thanks!

I dont believe the failed tests are failing for our other PRs - do you mind taking another look?

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -181,7 +177,6 @@ def on_start(self, state: State, event: Event, **kwargs):
         QuantizationMixin.start_calibration(self, state.model)
         # Unlike qmod, do not quantize as we calibrate
         # This choice does not seem to have a meaningful impact on accuracy
```
brian-dellabetta: I think you want to remove these comments as well? they no longer apply

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -237,6 +237,10 @@ def on_start(self, state: State, event: Event, **kwargs):
         # Unlike qmod, do not quantize as we calibrate
         # This choice does not seem to have a meaningful impact on accuracy
         state.model.apply(disable_quantization)
+        logger.info(
+            "quantization aware calibration is currently not supported for AWQ, "
+            "disabling quantization during calibration"
+        )
```
brian-dellabetta: can we only log this only if `model.quantization_enabled` is set to true? Or just remove?

kylesayrs: +1

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -182,6 +182,10 @@ def on_start(self, state: State, event: Event, **kwargs):
...
         state.model.apply(disable_quantization)
+        logger.info(
```
kylesayrs: Similarly, only log if option is set. You might find it easier to write this logic in the pipeline.

```
If GPTQModifier in modifiers
```

src/llmcompressor/modifiers/awq/base.py
shanjiaz: Yes!

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -234,9 +231,6 @@ def on_start(self, state: State, event: Event, **kwargs):
         # register quantization calibration hooks
         # assume quantization has been initialized by this modifier or one before it
         QuantizationMixin.start_calibration(self, state.model)
-        # Unlike qmod, do not quantize as we calibrate
-        # This choice does not seem to have a meaningful impact on accuracy
-        state.model.apply(disable_quantization)
```
brian-dellabetta, line 239: i think we should still keep this. Once AWQ is complete, quantization can be re-enabled for any subsequent modifier. So I think keeping the disable in the modifier lifecycles is preferred to the `disable_qac` checks in the pipelines

dsikka: I don’t think I follow the logic as to why this is preferred compared to having the pipeline handle it?

dsikka: I actually think the correct way would be a per modifier arg since in one oneshot call, you can have modifiers with different preferred behaviour but that isn’t supported atm

shanjiaz: We can remove awq from the disable_qac logic once it's fixed?

dsikka (approved):

kylesayrs (approved):
Looks good for now. Consider updating the argument help text to point out the caveat with GPTQ and AWQ

PR #1734, opened 2025-08-14
[WIP] Add upper bounds to dependencies for release

dhuangnm:
SUMMARY:

- added upper bounds to dependencies when build type is release
- still allow latest dependency versions when build type is nightly

The upper bounds are based on the current latest version of each dependency. May need to update again when we're ready to release 0.7.0.

TEST PLAN:
Run all tests.

Kicked off nightly tests: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/16976533641/job/48126802385


setup.py
```diff
@@ -110,19 +110,63 @@ def localversion_func(version: ScmVersion) -> str:
...
+        ),
+        (
+            "datasets>=4.0.0,<=4.0.0"
+            if BUILD_TYPE == "release"
+            else "datasets>=4.0.0"
+        ),
+        (
+            "accelerate>=1.6.0,<=1.10.0"
+            if BUILD_TYPE == "release"
+            else "datasets>=1.6.0"
```
brian-dellabetta: looks to be a typo
```suggestion
            else "accelerate>=1.6.0"
```

setup.py
```diff
@@ -110,19 +110,63 @@ def localversion_func(version: ScmVersion) -> str:
...
-        "requests>=2.32.2",
-        "tqdm>=4.66.3",
-        "torch>=2.7.0",
-        "transformers>=4.53.0",
-        "datasets>=4.0.0",
-        "accelerate>=1.6.0",
-        "pynvml>=11.5.3",
-        "pillow>=10.4.0",
+        (
+            "numpy>=2.0.0,<=2.3.2"
```
brian-dellabetta, line 126: We'll want to validate `numpy>=2.3`, hopefully librosa has resolved its issue

setup.py
dhuangnm: Thanks for catching this! Updated.

setup.py
dhuangnm: Yep, kicked off a run to test with the upper bounds: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/17047907241/job/48328293155#step:11:803

brian-dellabetta: If it works we can remove the lines 123-124 above it, but it doesn't look like they've updated it to be compatible with `numpy>2.3` here -- https://numba.readthedocs.io/en/stable/user/installing.html#version-support-information

dhuangnm: The job reached timeout so didn't run through completely, but for the tests it ran they looked fine except for some know failures. 

The numba 0.61.2 was released on Apr 2025 which is earlier than the numpy 2.3.0 release on Jun 2025 so I think the numba release page to limit numpy<2.3 is due to the lack of testing at the time of release.

brian-dellabetta: thanks for confirming!

brian-dellabetta (approved):

PR #1744, opened 2025-08-17
[Tests] Fix failing nightly quantization tests

dsikka:
## Purpose ##
* Fix failing tests introduced by adding `dispatch_for_generation` to the data free pipeline

## Changes ##
* After oneshot, oneshot should remove any dispatches from the model. However, there is a bug fixed [here](https://github.com/neuralmagic/compressed-tensors/pull/427) where models which fit entirely on one GPU do not have their dispatches removed (since they do not have hooks)
  * As a result, we need to move weights to the same device before comparing them for `test_quantization_reload`
* The `test_perplexity` test was implicitly relying on the model being dispatched to GPUs. Now explicitly `dispatch_for_generation`, similar to how we do in our examples

## Testing ##
* Nightly and commit tests passed locally

### DIFF
--- a/tests/llmcompressor/transformers/compression/test_quantization.py
+++ b/tests/llmcompressor/transformers/compression/test_quantization.py
@@ -108,7 +108,7 @@ class TestQuantizationMatches(unittest.TestCase):
             assert o_scale.dtype == n_scale.dtype == self.weight_dtype
             assert torch.equal(o_scale, n_scale.to(o_scale.device))
             assert o_zp.dtype == n_zp.dtype
-            assert torch.equal(o_zp, n_zp.to(o_scale.device))
+            assert torch.equal(o_zp, n_zp.to(o_zp.device))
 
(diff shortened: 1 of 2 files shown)


tests/llmcompressor/transformers/compression/test_quantization.py
```diff
@@ -150,6 +152,7 @@ def test_perplexity(self):
             max_seq_length=self.max_seq_length,
         )
         dataloader = self._get_dataloader(dataset_args, tokenizer)
+        dispatch_for_generation(self.model)
```
dsikka, line 155: Seems like we dispatch --> remove dispatch / move to cpu --> dispatch again.

Is the additional dispatch only required because we're removing the device map from the model and moving to cpu in remove_dispatch? Without the explicit move to cpu in `remove_dispatch` what was this doing incorrectly?

kylesayrs (approved):

PR #1746, opened 2025-08-18
[Transform] SpinQuant R4

kylesayrs:
## Purpose ##
* Support R4 transforms before R3. R3 requires hooking into the attention module, where as R4 does not

## Prerequisites ##
* https://github.com/vllm-project/vllm/pull/22486

## Testing ##
* Performed sanity checks with HF and vLLM


shanjiaz (approved):
Looks good to me. Niiice!

PR #1750, opened 2025-08-18
[Tests] Add recovery-based validation to LM-Eval tests

rahul-tuli:
# Recovery-Based Testing for LM-Eval

This PR implements **recovery-based testing** as the default validation mechanism for all lm-eval tests. Tests now compare compressed model performance against base model performance, making them robust to upstream changes while ensuring quantization quality.

**Current Problem:**
- Tests fail when base models regress due to external changes (e.g., transformers updates, lm-eval changes)
- False positives block CI even when quantization maintains expected recovery
- Absolute thresholds become stale as models/libraries evolve

**Example:**
Qwen2.5-VL tests fail with transformers ≥ 4.54.0 due to ~10% base model accuracy drop, despite quantization maintaining the same relative performance.

**Solution:**
Recovery testing validates that compressed models retain ≥95% (configurable) of base model performance, regardless of absolute score changes.

---

### Default Behavior (Zero Config Required)

All lm-eval tests now **automatically**:
1. ✅ Evaluate the base (uncompressed) model
2. ✅ Quantize the model using configured scheme
3. ✅ Evaluate the compressed model
4. ✅ Validate recovery ≥ 95% (default threshold)
5. ✅ Show optional warnings for absolute metrics

**Recovery Formula:**
```python
# For "higher is better" metrics (accuracy, F1, etc.)
recovery = compressed_score / base_score

# For "lower is better" metrics (perplexity, loss)
recovery = base_score / compressed_score  # Inverted!

# Validation
assert recovery >= threshold  # Default: 0.95
```

**Recovery Interpretation:**
- `1.00` = Perfect (0% degradation)
- `0.96` = 96% retained (4% degradation) ✅
- `0.93` = 93% retained (7% degradation) ❌ (with default threshold)

---

### Option 1: Use Default (Recommended)

No configuration needed - uses 95% recovery threshold:

```yaml
cadence: "weekly"
model: meta-llama/Meta-Llama-3-8B-Instruct
scheme: FP8_DYNAMIC
lmeval:
  # That's it! Uses recovery_threshold: 0.95 by default
```

### Option 2: Override Global Threshold

Set a different threshold for all metrics:

```yaml
lmeval:
  recovery_threshold: 0.93  # All metrics need ≥93% recovery
```

### Option 3: Per-Metric Thresholds

Set different thresholds for different metrics:

```yaml
lmeval:
  recovery_threshold:
    exact_match,flexible-extract: 0.95  # Strict threshold
    exact_match,strict-match: 0.90      # Relaxed threshold
```

### Option 4: With Absolute Metric Warnings

Keep absolute metrics for informational warnings (not failures):

```yaml
lmeval:
  recovery_threshold: 0.95  # Required - TEST FAILS if not met
  metrics:                   # Optional - warnings only, no failures
    exact_match,flexible-extract: 0.75
    exact_match,strict-match: 0.72
```

---

### ✅ Recovery Validation (Always Shown)

```
================================================================================
RECOVERY TESTING COMPARISON
================================================================================
✓ exact_match,flexible-extract   | Base: 0.7890 | Compressed: 0.7601 |
  Recovery: 96.34% ↑ | Threshold: ≥95.00%

✓ exact_match,strict-match       | Base: 0.7564 | Compressed: 0.7262 |
  Recovery: 96.01% ↑ | Threshold: ≥95.00%
================================================================================
✓ ALL METRICS PASSED RECOVERY THRESHOLDS
================================================================================
```

### Absolute Metric Warnings (If Configured)

```
================================================================================
ABSOLUTE METRICS CHECK (warnings only, not failures)
================================================================================
✓ exact_match,flexible-extract   | Expected: 0.7500 (±5%) | Got: 0.7601 |
  Within expected range

⚠ exact_match,strict-match       | Expected: 0.8000 (±5%) | Got: 0.7262 |
  Below expected range
================================================================================
```

**Note:** The warning above doesn't fail the test - recovery validation already passed!

---

### Existing Configs with Absolute Metrics

**Before (absolute thresholds cause failures):**
```yaml
lmeval:
  metrics:
    exact_match: 0.75  # TEST FAILS if not met
```

**After (minimal - uses recovery testing):**
```yaml
lmeval:
  # Uses default recovery_threshold: 0.95
  # No other config needed!
```

**After (keep warnings):**
```yaml
lmeval:
  # recovery_threshold: 0.95 is implicit (default)
  metrics:  # Now just warnings, won't fail tests
    exact_match: 0.75
```

### No Breaking Changes

- ✅ All existing configs continue to work
- ✅ `metrics` dict now shows warnings instead of failing
- ✅ Recovery testing automatically enabled with sensible default
- ✅ Backward compatible with all test infrastructure

---

### Files Changed

- **`tests/lmeval/test_lmeval.py`** (+151/-31 lines)
  - Added `recovery_threshold` config field (default: 0.95)
  - Made `metrics` field optional
  - Added `_eval_base_model()` method
  - Added `_validate_recovery()` method
  - Modified `_check_absolute_warnings()` to only warn, not fail
  - Updated test flow to always evaluate base model first

### Key Features

1. **Direction-Aware Recovery**
   - Automatically detects "higher is better" vs "lower is better" metrics
   - Inverts ratio for perplexity-style metrics

2. **Edge Case Handling**
   - Zero base values: `recovery = 1.0 if compressed == 0 else 0.0`
   - Missing metrics: Skipped gracefully
   - Metadata filtering: Skips stderr and alias keys

3. **Flexible Thresholds**
   - Global float: `recovery_threshold: 0.93`
   - Per-metric dict: `recovery_threshold: {metric1: 0.95, metric2: 0.90}`
   - Fallback to 0.95 for unlisted metrics when using dict

4. **Comprehensive Logging**
   - Recovery threshold displayed at test start
   - Detailed comparison table with base/compressed/recovery values
   - Clear pass/fail indicators with direction arrows (↑/↓)
   - Separate section for optional absolute warnings

---

## Performance Impact

**Additional Runtime:**
- Base model evaluation: ~2-10 minutes
- Compressed model evaluation: ~2-10 minutes (unchanged)
- **Total: ~2x single evaluation time**

**Trade-off:** Doubled evaluation time for robust, meaningful metrics that don't break from upstream changes.

**Mitigation:** Tests run on weekly cadence, making the additional time acceptable.

---

## ✅ Benefits

| Benefit | Description |
|---------|-------------|
| 🛡️ **Robustness** | Tests never break from lm-eval or transformers updates |
| 📊 **Meaningful** | Measures actual compression degradation, not arbitrary thresholds |
| 🎯 **Automatic** | Works out of box, no config needed |
| 🔧 **Flexible** | Override threshold globally or per-metric |
| ↔️ **Compatible** | Zero breaking changes, existing configs work |
| 🧹 **Simple** | ~150 lines in single file, no new dependencies |
---

## Testing

To test recovery-based validation:

```bash
# Uses default recovery threshold (0.95)
CADENCE=weekly TEST_DATA_FILE=tests/lmeval/configs/fp8_dynamic_per_token.yaml \
  pytest tests/lmeval/test_lmeval.py -v
```

---


rahul-tuli:
@gemini-code-assist review

tests/lmeval/test_lmeval.py
```diff
@@ -23,8 +24,12 @@ class LmEvalConfig(BaseModel):
     task: str = "gsm8k"
     num_fewshot: int = 5
     limit: int = 1000
-    metrics: dict
     batch_size: int = 100
+    # Recovery testing (default): compare against base model performance
+    # Default threshold is 0.95 (retain ≥95% of base), can be overridden
+    recovery_threshold: Union[float, dict] = 0.95
+    # Optional absolute metrics for warnings (not failures)
+    metrics: Optional[dict] = None
```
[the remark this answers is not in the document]
rahul-tuli: Too much of a nit

dsikka (approved):

tests/lmeval/test_lmeval.py
```diff
@@ -119,11 +143,29 @@ def test_lm_eval(self, test_data_file: str):
...
+        logger.info("================= Running LM Eval on COMPRESSED model ==========")
         self._run_lm_eval()
 
         self.tear_down()
 
+    @log_time
+    def _eval_base_model(self):
+        """Evaluate the base (uncompressed) model."""
+        model_args = {"pretrained": self.model}
+        model_args.update(self.lmeval.model_args)
```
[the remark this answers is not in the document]
dsikka: @rahul-tuli what do we think of this comment?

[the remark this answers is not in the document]
rahul-tuli: Let's accept it it's a good change!

tests/lmeval/test_lmeval.py
```diff
@@ -152,46 +194,124 @@ def _run_lm_eval(self):
...
-                # If higher is better, assert actual val >= expected val * (1 - stderr)
-                if higher_is_better:
-                    assert actual_val >= expected_val * (0.95)
-                # If higher is worse, assert actual val <= expected val * (1 + stderr)
-                else:
-                    assert actual_val <= expected_val * (1.05)
 
+            # Check if within ±5% relative tolerance
+            lower_bound = expected_val * 0.95
+            upper_bound = expected_val * 1.05
```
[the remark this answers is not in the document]
brian-dellabetta: Do we want these to be more strict now?

tests/lmeval/test_lmeval.py
```diff
@@ -97,6 +117,10 @@ def test_lm_eval(self, test_data_file: str):
         # Run vLLM with saved model
         self.set_up(test_data_file)
 
+        # Always evaluate base model for recovery testing
+        logger.info("================= Evaluating BASE model ======================")
+        self.base_results = self._eval_base_model()
```
brian-dellabetta, line 122: won't this increase runtime quite a bit? is this to prevent our ci/cd from breaking if regressions are introduced in transformers?

tests/lmeval/test_lmeval.py
```diff
@@ -152,46 +193,124 @@ def _run_lm_eval(self):
...
+
+    def _validate_recovery(self, compressed_results):
+        """Validate using recovery testing - compare against base model."""
+        base_metrics = self.base_results["results"][self.lmeval.task]
+        compressed_metrics = compressed_results["results"][self.lmeval.task]
+        higher_is_better_map = compressed_results["higher_is_better"][self.lmeval.task]
+
+        # Get recovery threshold(s)
+        recovery_threshold = self.lmeval.recovery_threshold
+        is_dict = isinstance(recovery_threshold, dict)
```
kylesayrs: As a nit, remove the use of this variable and instead put the condition in the if statement.

tests/lmeval/test_lmeval.py
```diff
@@ -152,46 +193,124 @@ def _run_lm_eval(self):
...
+
             actual_val = metrics.get(metric_key)
+            if actual_val is None:
+                continue
+
             higher_is_better = results["higher_is_better"][self.lmeval.task].get(
                 metric_key.split(",")[0], True
             )
-            stderr_key = metric_key.replace(",", "_stderr,")
-            std_err = self.lmeval.metrics.get(stderr_key)
```
brian-dellabetta, line 166: we should remove stderr values from the yaml files in `configs/` if we're no longer using them

tests/lmeval/test_lmeval.py
```diff
@@ -152,46 +193,124 @@ def _run_lm_eval(self):
...
+
+        # If absolute metrics provided, show warnings (not failures)
+        if self.lmeval.metrics:
+            self._check_absolute_warnings(results)
+
+    def _validate_recovery(self, compressed_results):
+        """Validate using recovery testing - compare against base model."""
+        base_metrics = self.base_results["results"][self.lmeval.task]
+        compressed_metrics = compressed_results["results"][self.lmeval.task]
+        higher_is_better_map = compressed_results["higher_is_better"][self.lmeval.task]
```
kylesayrs: Is higher_is_better guaranteed to exist?

tests/lmeval/test_lmeval.py
rahul-tuli: Yes, it will at the moment, see the PR description, as a follow up, we'll be caching the results for our base models

tests/lmeval/test_lmeval.py
```diff
@@ -152,46 +193,124 @@ def _run_lm_eval(self):
...
+            if "stderr" in metric_key or metric_key.startswith("alias"):
+                continue
+
+            compressed_val = compressed_metrics.get(metric_key)
+            if compressed_val is None:
+                continue
+
+            # Get threshold for this metric
+            if is_dict:
+                threshold = recovery_threshold.get(metric_key, 0.95)
```
kylesayrs: I agree with gemini here, two sources of truth is more to maintain

tests/lmeval/test_lmeval.py
rahul-tuli: both stderr keys, along with metrics will be pruned in a follow up PR

tests/lmeval/test_lmeval.py
```diff
@@ -152,46 +193,124 @@ def _run_lm_eval(self):
...
+
+        logger.info("=" * 80)
+        logger.info("RECOVERY TESTING COMPARISON")
+        logger.info("=" * 80)
+
+        failures = []
+        for metric_key, base_val in base_metrics.items():
+            # Skip stderr and other metadata
+            if "stderr" in metric_key or metric_key.startswith("alias"):
+                continue
```
kylesayrs, line 223: Shouldn't we be matching metrics to results, rather than results to metrics? That way we don't have to hard code against stuff like this.

```
for metric_key in compressed_metrics:
     ...

for metric_key in recovery_threshold:
     ....
```

brian-dellabetta (dismissed):

tests/lmeval/test_lmeval.py
```diff
@@ -152,46 +193,124 @@ def _run_lm_eval(self):
...
         for metric_key, expected_val in self.lmeval.metrics.items():
-            # stderr metrics are only used as absolute tolerance
-            # checks for actual values
+            # Skip stderr metrics
             if "stderr" in metric_key:
                 continue
+
             actual_val = metrics.get(metric_key)
+            if actual_val is None:
+                continue
```
kylesayrs, line 309: Should we error here if the config has a key that the results don't?

kylesayrs (approved):

kylesayrs:
Just nits, can address later if need be

tests/lmeval/test_lmeval.py
rahul-tuli: No because, the absolute error are just used for warnings, we will completely remove them in a follow up

brian-dellabetta (approved):
Good stuff! Looks like we can lower the threshold for the failing AWQ tests, after release

dsikka (approved):

PR #1759, opened 2025-08-19
[AWQ] fixes to matching logic and #1742 bugfix

brian-dellabetta:
SUMMARY:
- [x] I introduced a bug in #1742 that caused `lm_eval` awq test to fail. This reverts that change, re-setting the original state dict of the parent module in the grid search for best scales. 
- [x] This also updates to the new module matching API, excluding from the resolved mappings any modules that match the list in `ignore`. This should resolve a user issue with command-a-vision, which has k_proj etc. layers in the vision_encoder that we want to exclude in our resolved mappings.

TEST PLAN:
awq lm_eval test is passing now. Running command-a-vision check

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -2,13 +2,12 @@ import inspect
 from typing import Dict, List, Optional, Tuple, Union
 
 import torch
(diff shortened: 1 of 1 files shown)


shanjiaz (dismissed):
woohoo!

PR #1760, opened 2025-08-19
[MoE] MoE Calibration with `calibrate_all_experts`

kylesayrs:
Coauthored with @dichn!

## Purpose ##
* Add support for `calibrate_all_experts` option, which sends all tokens to all experts, but still produces the same outputs as if tokens had been gated

## Changes ##
* Modify model definitions such that, in the case of `calibrate_all_experts=True` token gating occurs after passing tokens to experts, rather than before

```python3
# `calibrate_all_experts=True` by default
model = replace_modules_for_calibration(model, calibrate_all_experts=True)
```

## Testing ##
* Added correctness tests for new model definitions which checks that outputs are exactly the same
* Added hook tests to make sure all experts are being sent tokens


src/llmcompressor/modeling/llama4.py
```diff
@@ -15,33 +15,41 @@
...
-
-        router_top_value, router_indices = torch.topk(router_logits, self.top_k, dim=1)
-
-        router_scores = (
-            torch.full_like(router_logits, float("-inf"))
-            .scatter_(1, router_indices, router_top_value)
-            .transpose(0, 1)
-        )
-        router_scores = torch.sigmoid(router_scores.float()).to(hidden_states.dtype)
+        router_scores, router_logits = self.router(hidden_states)
```
dsikka: fyi - will break for transformers < 4.54.

kylesayrs: Added a note

src/llmcompressor/modeling/llama4.py
```diff
@@ -17,39 +15,57 @@
...
+        else:
+            router_top_value, router_indices = torch.topk(
+                router_logits, self.top_k, dim=1
+            )
+            router_logits = router_outputs
+            router_scores = (
+                torch.full_like(router_logits, float("-inf"))
+                .scatter_(1, router_indices, router_top_value)
+                .transpose(0, 1)
+            )
```
brian-dellabetta, line 50: rather than not using topk for calibrate_all_experts, why not set `self.top_k=num_experts` during calibration, and back to original value after calibration?

brian-dellabetta: I think it would reduce the amount of code you'd have to change, particularly useful when we're copy-pasting model def code from `transformers`

src/llmcompressor/modeling/deepseek_v3.py
```diff
@@ -30,24 +36,40 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
         # End MoE
 
         hidden_states = final_hidden_states.type(hidden_states.dtype).view(*orig_shape)
         hidden_states = hidden_states + self.shared_experts(residuals)
         return hidden_states
 
 
-def replace(config: DeepseekV3Config, module: OriginalDeepseekV3MoE):
-    return DeepseekV3MoECalibrate(config=config, original=module)
+def replace(
```
fynnsu, line 68: I think it would be cleaner to drop these `replace` functions and just use the class directly.

Since we define the calibration classes, wed can ensure their signatures are consistent.  

For `SequentialLlama4TextMoe` we would just have to update the init so that it takes the config value directly and then calls `config.get_text_config()` inside.

fynnsu (commented):
Left a couple comments. I also agree with @brian-dellabetta's point that this could maybe be simplified by patching `self.top_k` temporarily.

src/llmcompressor/modeling/deepseek_v3.py
kylesayrs: I agree, I'd prefer to have the class define `from` and `to` class and instance methods, but this (along with consolidating `prepare_for_calibration` and the moe context) is out of scope imho.

src/llmcompressor/modeling/llama4.py
kylesayrs: `self.top_k=num_experts` would produce ungated outputs, which are not representative of how the model activates (neither during training, nor inference). We should not be calibrating with values that aren't representative of how the model activates during inference.

src/llmcompressor/modeling/deepseek_v3.py
kylesayrs: This is PR is focused on enabling proper MoE calibration for now, I'm trying to keep the diff small to focus on the routing logic for now.

src/llmcompressor/modeling/llama4.py
dsikka: @brian-dellabetta fyi - not setting top_k to num_experts is also consistent with how we're currently calibrating on main. 

You may be able to get away with this for some models / on certain tasks but we want to reflect the expected behaviour

I would say just setting the top_k is more or less a hack

brian-dellabetta: got it. i thought they were equivalent

kylesayrs:
Running those examples now

src/llmcompressor/utils/helpers.py
```diff
@@ -974,7 +974,8 @@ def getattr_chain(obj: Any, chain_str: str, *args, **kwargs) -> Any:
     return res
 
 
-class DisableKVCache:
+@contextlib.contextmanager
+def disable_cache(module: torch.nn.Module):
```
brian-dellabetta, line 978: definitely agree with these changes, but might be better in a separate PR or added in the PR summary. Seems orthogonal to `calibrate_all_experts`

brian-dellabetta (approved):

dsikka:
FYI - Produced the following Qwen Model: `nm-testing/Qwen3-30B-A3B-NVFP4-0912`

With the following set-up:
```bash
lm_eval \
  --model vllm \
  --model_args pretrained="Qwen/Qwen3-30B-A3B",dtype=auto,max_model_len=4096,add_bos_token=True\
  --tasks gsm8k \
  --batch_size auto
    
```

NVFP4:
```bash
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8734|±  |0.0092|
|     |       |strict-match    |     5|exact_match|↑  |0.8719|±  |0.0092|
```

Dense:

```bash
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8544|±  |0.0097|
|     |       |strict-match    |     5|exact_match|↑  |0.8916|±  |0.0086|
```


The only thing I noticed was speed but that may also be because of beaker just being slow

dsikka (commented):
Similarly, for Llama4 NVFP4 

```bash

lm_eval \
  --model vllm \
  --model_args pretrained="nm-testing/Llama-4-Scout-17B-16E-Instruct-NVFP4-0913",dtype=auto,max_model_len=4096,tensor_parallel_size=2,enable_chunked_prefill=True,enforce_eager=True \
  --tasks gsm8k_llama \
  --apply_chat_template \
  --fewshot_as_multiturn \
  --batch_size auto
```

tests/testing_utils.py
```diff
@@ -252,3 +253,12 @@ def preprocess(example):
     ds = ds.map(process, remove_columns=ds.column_names)
 
     return ds
+
+
+def requires_cadence(cadence: Union[str, List[str]]) -> Callable:
```
dsikka, line 258: Would be good if we could use this here as well: https://github.com/vllm-project/llm-compressor/blob/4caf5406c42b5882d78660645adc9b2d1724ca0e/tests/testing_utils.py#L93

to minimize the number of similar functions we have

src/llmcompressor/utils/helpers.py
```diff
@@ -61,7 +61,7 @@
     "is_package_available",
     "import_from_path",
     "getattr_chain",
-    "DisableKVCache",
```
dsikka, line 64: why did we make this change?

dsikka (commented):
LGTM. 
Two small comments

kylesayrs:
Thanks a ton @dsikka

tests/testing_utils.py
kylesayrs: Yeah I considered this, but it was hard to integrate and only reuses about 2 lines

src/llmcompressor/utils/helpers.py
kylesayrs: This was prompted by a type-hint change. I think this should be land in this PR

PR #1761, opened 2025-08-19
Fix install issue due to librosa

dhuangnm:
SUMMARY:
Had to pin librosa version to 0.11.0 in order for installation with [dev] to work.

TEST PLAN:
A run here has passed the installation across the jobs: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/17083270223/job/48442139981

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -187,7 +187,7 @@ setup(
             "trl>=0.10.1",
             "pandas<2.3.0",
             "torchvision",
-            "librosa>=0.11.0",
+            "librosa==0.11.0",
             "soundfile",
             "torchcodec",
             # linting, formatting, and type checking
--- a/setup.py
+++ b/setup.py
@@ -187,7 +187,7 @@ setup(
             "trl>=0.10.1",
             "pandas<2.3.0",
             "torchvision",
-            "librosa",
+            "librosa>=0.11.0",
             "soundfile",
             "torchcodec",
             # linting, formatting, and type checking


dsikka (commented):
Can you explain in the PR description why this is now required?

dhuangnm:
> Can you explain in the PR description why this is now required?

Yep updated the summary of the PR.

setup.py
```diff
@@ -187,7 +187,7 @@ def localversion_func(version: ScmVersion) -> str:
             "trl>=0.10.1",
             "pandas<2.3.0",
             "torchvision",
-            "librosa",
+            "librosa==0.11.0",
```
brian-dellabetta, line 190: any reason not to use `librosa>=0.11.0`?

dsikka: Yeah to help with flexibility. Although this package is currently on a pace of like one release per year

dsikka (changes requested):
I don’t really understand why we need this at all.

For the nightly, one job failed but one did not, which makes me think it’s a runner issue? https://github.com/neuralmagic/llm-compressor-testing/actions/runs/17086003244/job/48450172364

setup.py
dhuangnm: Tried librosa>=0.11.0 initially but it failed too unfortunately.

dhuangnm:
> I don’t really understand why we need this at all.
> 
> For the nightly, one job failed but one did not, which makes me think it’s a runner issue? https://github.com/neuralmagic/llm-compressor-testing/actions/runs/17086003244/job/48450172364

This is due to different python versions. The issue happens in 3.12 and the uv pip command looks picking up different versions for librosa if there is no constraint.

setup.py
dsikka: But 0.11.0 is latest - why would >= fail?
https://pypi.org/project/librosa/#history

dsikka (approved):

PR #1772, opened 2025-08-21
[Multi-modifier] Support scoped application of quantization config/status

brian-dellabetta:
SUMMARY:
Prerequisites:
* https://github.com/neuralmagic/compressed-tensors/pull/432

This allows for multi-modifier support by scoping the application of quantization config/status to only the modules in the model that match the given targets/ignore configuration, rather than all modules. Initialization of observers is moved to on_start (instead of on_initialize) to match their removal on_end (and not on_finalize). This prevents collision during the multi-modifier lifecycle

- [x] Update AWQ
- [x] Update QuantizationModifier
- [x] Update QuantizationMixin
- [x] Update GPTQ
- [x] No other quantization modifiers exist

TEST PLAN:
- Tests were added to https://github.com/neuralmagic/compressed-tensors/pull/432 to confirm correct application of multiple modifiers. 
- Added an example in this PR to show how AWQ and GPTQ can be applied heterogeneously to a model, along with a small README. Logs show alternating AWQ and GPTQ messages for `"sequential"`, and correct behavior for `"independent"` pipelines. [Model checkpoint](https://huggingface.co/nm-testing/Meta-Llama-3-8B-Instruct-selfattn-w8a8-mlp-w4a16-sequential/tree/main) for the sequential pipeline shows correct application of W8A8 to self_attn layers and W4A16 to mlp layers. config.json and safetensors weights all look as expected


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -178,7 +182,7 @@ def on_start(self, state: State, event: Event, **kwargs):
 
         # register gptq hooks
         added_hook = False
-        for module in state.model.modules():
+        for _, module in match_named_modules(state.model, self.targets, self.ignore):
             if getattr_chain(module, "quantization_scheme.weights", None) is not None:
```
kylesayrs, line 182: Should this be changed into an assert rather than an if?

brian-dellabetta: yeah i think that makes sense, I can update

brian-dellabetta: hmm on second thought, it doesn't look like there's anything in the validation layer confirming each quantization args instance has a weights field. so if a user sets an invalid config where weight quantization isn't configured, it would error out here. Is that what we want?

kylesayrs: I'd prefer explicit error rather than silent skip here

examples/multi_modifier/llama3_example.py
```diff
@@ -0,0 +1,101 @@
+from datasets import load_dataset
```
kylesayrs, line 1: Can we maybe generalize this folder name to `mixed-precision` so that people associate this with enabling mixed precision workloads?

kylesayrs: Even though multi-modifier recipes don't necessarily need to be mixed precision? I just think mixed-precision is a stronger message than multi-modifier

examples/multi_modifier/llama3_example.py
```diff
@@ -0,0 +1,101 @@
...
+
+# Apply algorithms.
+oneshot(
+    model=model,
+    dataset=ds,
+    recipe=recipe,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+    # Option 1) run both modifiers in a single calibrated run
+    pipeline="sequential",
```
kylesayrs, line 83: Is the pipeline not already sequential by default?

examples/multi_modifier/llama3_example.py
```diff
@@ -0,0 +1,101 @@
...
+print("========== SAMPLE GENERATION ==============")
+dispatch_for_generation(model)
+sample = tokenizer("Hello my name is", return_tensors="pt")
+sample = {key: value.to(model.device) for key, value in sample.items()}
+output = model.generate(**sample, max_new_tokens=100)
+print(tokenizer.decode(output[0]))
+print("==========================================\n\n")
+
+# Save to disk compressed.
+SAVE_DIR = model_id.rstrip("/").split("/")[-1] + "-W4A16-G128"
```
kylesayrs: Add GPTQ and AWQ to the names?

examples/multi_modifier/llama3_example.py
brian-dellabetta: it actually defaults to independent. Without it set, it infers to use `SequentialPipeline` for GPTQ, running just GPTQ independently. It then infers to use `SequentialPipeline` for AWQ, running just AWQ independently. Not sure what we want for default behavior though, this just makes it explicit

examples/multi_modifier/llama3_example.py
brian-dellabetta: I can rename to `examples/mixed_precision` if we find that to be more appropriate. Let's discuss in standup

src/llmcompressor/modifiers/quantization/gptq/base.py
brian-dellabetta: Shouldn't we do this in the validation layer though? I can add a check to model validate, and switch to assert?

examples/multi_modifier/llama3_example.py
brian-dellabetta: changed

examples/multi_modifier/llama3_example.py
brian-dellabetta: updated

src/llmcompressor/modifiers/quantization/gptq/base.py
brian-dellabetta: Updated to assert, and included an assertion in the validation layer so this is caught at the beginning, rather than GPTQ on_start (which could occur later in the lifecycle if other modifiers are included in the recipe)

brian-dellabetta: The assert on this line could possibly be dropped, but good to have another check to have a nice error in event of another bug, so i'll leave in. It's only checked once for each targeted model, so shouldn't be a heavy performance hit

kylesayrs (dismissed):
Nice job

tests/llmcompressor/modifiers/quantization/test_base.py
```diff
@@ -95,12 +95,11 @@ def test_block_strategy_parsing(block_q_config_kwargs):
 def test_actorder_resolution(
     has_actorder, actorder, q_config_kwargs, expected_0, expected_1
 ):
-    if has_actorder:
-        modifier = GPTQModifier(**q_config_kwargs, actorder=actorder)
-    else:
-        modifier = GPTQModifier(**q_config_kwargs)
-
     with pytest.raises(ValueError) if expected_0 == "error" else nullcontext():
+        if has_actorder:
```
brian-dellabetta, line 99: because GPTQ now calls `.resolve_quantization_config` on init/validation, this error occurs during instantiation instead of at line `resolved = modifier.resolve_quantization_config()`

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -174,14 +192,17 @@ def on_start(self, state: State, event: Event, **kwargs):
...
-                # HACK: previously, embeddings were not quantized because they were not
-                # accessible by the layer compressor. For now, we manually ignore it,
-                # but in the FUTURE this should be ignored by the user
-                if not isinstance(module, torch.nn.Embedding):
-                    self.register_hook(module, self.calibrate_module, "forward")
-                    added_hook = True
+        for name, module in match_named_modules(state.model, self.targets, self.ignore):
+            assert (
+                getattr_chain(module, "quantization_scheme.weights", None) is not None
+            ), f"Matched target {name} does not have associated weight quantization"
```
fynnsu: Can we use if/raise instead of assert?

Generally I think that's preferred for production code because asserts can be removed / ignored by the end user.

fynnsu (approved):
Looks good! Added 1 comment below.

src/llmcompressor/modifiers/quantization/gptq/base.py
brian-dellabetta: Yeah good point. Since this is blocking Ci/Cd, i'd like to get it in as is if transformers tests pass. I can make those changes in a follow-up

brian-dellabetta: i ended up removing completely. `match_named_modules` is returning non-quantizable layers like `self_attn` (in addition to `self_attn.q_proj) and we need to skip them here

brian-dellabetta: this wasn't in the original implementation, kyle suggested it in a comment, but in retrospect this won't work with an assert/raise

tests/llmcompressor/transformers/compression/recipes/new_quant_weight.yaml
```diff
@@ -14,4 +14,3 @@ test_stage:
                     targets: ["Linear", "Embedding"]
         GPTQModifier:
             block_size: 128
-            targets: ["re:model.layers.\\d+$"]
```
brian-dellabetta, line 17: Pretty sure this was failing because the change to use `match_named_targets` isn't applying quantization config to non-quantizable layers, so having `targets` that aren't the leaf Linear layers will fail out now. Do we want to support this target to mean "all modules within target module" as well?

tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py
```diff
@@ -187,7 +187,6 @@ def test_quant_model_reload(format, dtype, tmp_path):
     for _, module in model.named_modules():
         if hasattr(module, "quantization_scheme"):
             assert module.weight.dtype == dtype
-            assert module.quantization_status == QuantizationStatus.FROZEN
```
brian-dellabetta, line 190: this seemed appropriate because we are deprecating the meaning of quantization status on a module.

examples/mixed_precision/llama3_example.py
```diff
@@ -0,0 +1,101 @@
...
+ds = ds.map(tokenize, remove_columns=ds.column_names)
+
+# Configure the quantization algorithm to run.
+#   * quantize self_attn layers to W8A8 with GPTQ
+#   * quantize mlp layers to W4A16 with AWQ
+#       only include mappings pertaining to target layers
+recipe = [
+    GPTQModifier(targets=r"re:.*self_attn\.(k|q|o|v)_proj$", scheme="W8A8"),
+    AWQModifier(
+        targets=r"re:.*mlp\.(down|gate|up)_proj$",
```
dsikka, line 60: this is real regex?? amazing

examples/mixed_precision/llama3_example.py
```diff
@@ -0,0 +1,101 @@
+from datasets import load_dataset
```
dsikka, line 1: FYI - mixed_precision is what we typically use for something like W4A16.
We should add this example / readme to the non-uniform quantization folder. There is already a readme in that folder that we should extend

tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py
dsikka: have we cleaned up module level statuses throughout both here and ct? just worried about some cases still  floating around

examples/mixed_precision/llama3_example.py
brian-dellabetta: haha yeah, regex is super flexible, saw some crazy stuff in my information security days lol

examples/mixed_precision/llama3_example.py
brian-dellabetta: Moved script to `examples/quantization_non_uniform` and merged README

tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py
brian-dellabetta: I found the root cause and updated. It works now with the quantization_status check

shanjiaz (approved):
Woohoo!

tests/llmcompressor/modifiers/quantization/test_base.py
```diff
@@ -155,8 +154,8 @@ def test_config_resolution(strategies, actorder):
 )
 def test_serialize_actorder(has_actorder, actorder, exp_actorder):
     if has_actorder:
-        modifier = GPTQModifier(targets=["Linear"], actorder=actorder)
+        modifier = GPTQModifier(targets=["Linear"], scheme="W8A8", actorder=actorder)
```
dsikka, line 157: How was this targeting before you added the scheme?

fynnsu (approved):
Looks good!

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -114,43 +116,71 @@ def validate_scheme(
...
-        enable quantization as we calibrate
+        Attach observers, register activation calibration hooks (including
+        kv_cache quantization) and enable quantization as we calibrate
 
         :param model: model to prepare for calibration
         """
         self._calibration_hooks = self._initialize_hooks(model)
-        model.apply(apply_calibration_status)
+        for _, module in match_named_modules(model, self.targets, self.ignore):
+            self._initialize_observers(module)
```
dsikka, line 168: Why can't we keep this iniitialize_quantization?

tests/llmcompressor/modifiers/quantization/test_base.py
brian-dellabetta: I think it just passed init/validation but was never used. It was never applied to a model, so it would've never worked, i just added it to make sure improper configuration wasn't the reason the test was failing (it was ultimately something else causing the test to fail)

src/llmcompressor/modifiers/quantization/quantization/mixin.py
brian-dellabetta: observers should be initialized on start to align with them being removed on_end. so this was moved into on_start instead. without this change the lifecycle with multiple quant modifiers will trigger observer hooks before the modifier starts (before it sees any data), which can now happen in a previous modifier lifecycle

PR #1773, opened 2025-08-21
[Bugfix] [Example] Fix QuIP example

kylesayrs:
## Purpose ##
* In order to support LC which did not have [QuIP default arguments fix](#1770), update example to explicitly set targets


rahul-tuli (approved):
Thanks Kyle!

PR #1774, opened 2025-08-22
Fix wonky dependency range on datasets

timkpaine:
SUMMARY:
Fixes dependencies

- [`datasets>=4.0.0,<=4.0.0`](https://github.com/vllm-project/llm-compressor/pull/1734/files#r2277726469)
- [`pillow>=10.4.0,<=10.4.0`](https://github.com/vllm-project/llm-compressor/pull/1734/files#r2277726473)

TEST PLAN:
None

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,8 +147,10 @@ setup(
             if BUILD_TYPE == "release"
             else "transformers>=4.53.0"
         ),
+        # `datasets==4.0.0` is latest release, so
+        # pin directly. NOTE: add range after new release
         (
-            "datasets>=4.0.0,<=4.0.0"
+            "datasets==4.0.0"
             if BUILD_TYPE == "release"
             else "datasets>=4.0.0"
         ),
@@ -163,7 +165,7 @@ setup(
             else "pynvml>=11.5.3"
         ),
         (
-            "pillow>=10.4.0,<=10.4.0"
+            "pillow>=10.4.0,<11.0.0"
             if BUILD_TYPE == "release"
             else "pillow>=10.4.0"
         ),


setup.py
```diff
@@ -148,7 +148,7 @@ def localversion_func(version: ScmVersion) -> str:
             else "transformers>=4.53.0"
         ),
         (
-            "datasets>=4.0.0,<=4.0.0"
+            "datasets>=4.0.0,<=4.1.0"
```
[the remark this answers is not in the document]
timkpaine: mentioned in description, question is posed to maintainers

kylesayrs:
Hi @timkpaine, thanks for the catch!

These bounds were determined empirically by https://github.com/vllm-project/llm-compressor/pull/1734. Since 4.0.0 is the latest release, it remains unbounded outside of release builds and bounded during release builds for future-proofing.

I agree that the upper bounding is equivalent to `==`, so it can be changed, but it may be changed back to an explicit upper bound when the next `datasets` version is released.

kylesayrs:
Adding a comment might also clarify that this notation was not a mistake, and should be changed for the next datasets release

setup.py
```diff
@@ -163,7 +165,7 @@ def localversion_func(version: ScmVersion) -> str:
             else "pynvml>=11.5.3"
         ),
         (
-            "pillow>=10.4.0,<=10.4.0"
+            "pillow>=10.4.0,<11.0.0"
```
dhuangnm, line 168: The "<=10.4.0" was used initially to allow the upper bound testing to have a version to install from otherwise it doesn't know which one is the highest for "<11.0.0". We can make calls to the PyPi server to get that info but it'll add extra round-trip and I'd like to avoid that if possible. Also if there is newer version of 10.x.0 coming out, using "<11.0.0" will pick it up automatically which seems against the use of upper bound here since we want to be specific about the upper bound?

setup.py
```diff
@@ -147,8 +147,10 @@ def localversion_func(version: ScmVersion) -> str:
             if BUILD_TYPE == "release"
             else "transformers>=4.53.0"
         ),
+        # `datasets==4.0.0` is latest release, so
+        # pin directly. NOTE: add range after new release
         (
-            "datasets>=4.0.0,<=4.0.0"
+            "datasets==4.0.0"
```
dhuangnm, line 153: I'll update the workflow to use this version for both lower and upper bounds otherwise it'll fail sine it's trying to use the ">=" for the lower bound testing and the "<=" for the upper bound testing.

dhuangnm: PR to update the test workflow to handle pinned dependencies: https://github.com/neuralmagic/llm-compressor-testing/pull/121

setup.py
dhuangnm: Or maybe we can just pin pillow here, i.e. "pillow==10.4.0"?

setup.py
dhuangnm: PR landed.

kylesayrs:
Closing for inactivity, feel free to reopen if you think the changes should still be made

timkpaine:
> Closing for inactivity, feel free to reopen if you think the changes should still be made

all good, i can set the dependencies as i like in the conda recipe.

PR #1782, opened 2025-08-27
[Transform] Support separating v and u transforms of quip

kylesayrs:
## Purpose ##
* Support `rotation` argument with `QuIPModifier` which allows users to select V and U rotations separately.
  * According to research, the default can be changed to V after a basic recovery evaluation is done

## Changes ##
* Add `rotations` argument
* Updated docstrings and comments to remove references to spinquant

## Testing ##
* Tested model coherence with V and U separately


src/llmcompressor/modifiers/transform/quip/base.py
```diff
@@ -58,6 +58,7 @@ class QuIPModifier(Modifier):
     :param transform_config: Optional transform config for overriding provided arguments
     """  # noqa: E501
 
+    rotations: List[Literal["v", "u"]] = Field(default_factory=lambda: ["v", "u"])
```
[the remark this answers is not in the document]
kylesayrs: Pass, this is consistent with spinquant

src/llmcompressor/modifiers/transform/quip/base.py
```diff
@@ -32,18 +32,18 @@ class QuIPModifier(Modifier):
 
     Lifecycle:
         - on_initialize
-            - infer SpinQuantMappings & NormMappings
-            - as needed, create transform schemes for R1, R2, R3, & R4
```
brian-dellabetta, line 36: 😅

examples/transform/quip_example.py
```diff
@@ -14,14 +14,14 @@
 # NOTE: because the datafree pipeline is being used in this
 # example, you can use additional GPUs to support larger models
 MODEL_ID = "meta-llama/Llama-3.1-8B-Instruct"
-model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
+model = AutoModelForCausalLM.from_pretrained(MODEL_ID, dtype="auto")
```
rahul-tuli: Is this an intentional change? Did `AutoModelForCausalLM.from_pretrained` api change?

examples/transform/quip_example.py
```diff
@@ -35,7 +35,7 @@
 input_ids = tokenizer("Hello my name is", return_tensors="pt").input_ids.to(
     model.device
 )
-output = model.generate(input_ids, max_new_tokens=100)
+output = model.generate(input_ids, max_new_tokens=50)
```
rahul-tuli, line 38: nit: Why change this?

src/llmcompressor/modifiers/transform/quip/base.py
```diff
@@ -77,6 +78,12 @@ def validate_not_implemented(cls, value, info: ValidationInfo):
             raise NotImplementedError(f"{info.field_name} is not supported right now")
         return value
 
+    @field_validator("rotations", mode="before")
+    def validate_lowercase_list(cls, value):
+        if isinstance(value, list):
+            value = [v.lower() if isinstance(v, str) else v for v in value]
+        return value
```
rahul-tuli, line 85: Question: Pydantic will automatically throw an error if user specifies anything other than v or u right? And what about a single string input like "v" we are not accepting that right? if the latter case is True should throw an error maybe?

examples/transform/quip_example.py
kylesayrs: Yep, new transformers release deprecates `torch_dtype`

kylesayrs: Still need to support old

rahul-tuli (approved):

src/llmcompressor/modifiers/transform/quip/base.py
kylesayrs: Yes, anything except v or u should throw an error. I don't plan on supporting single "v" or "u" for now

examples/transform/quip_example.py
kylesayrs: Running quip with transformers takes especially long (because of the extra transform layers). 50 tokens is sufficient to check for coherence.

brian-dellabetta (approved):

PR #1785, opened 2025-08-29
Recovered skipped w8a8 compression related tests

shanjiaz:
SUMMARY:
Recovered skipped w8a8 compression/decompression tests now that transformer side of code is merged. (Should we wait till next release?)

TEST PLAN:
tested locally on transformers 4.56.0.dev0 they passed.


dsikka:
These will fail until the next transformers release?

shanjiaz:
> These will fail until the next transformers release?

Yeah they won't work till our changes are applied

shanjiaz:
> Is moving to the CPU required to fix the memory leak? If so, that would be vey strange.
> 
> Perhaps `torch.cuda.synchronize` is having an effect? I never tested with this when I last looked at the memory leak issue.

yeah the same code doesn't work if I don't move the models to cpu first.

PR #1786, opened 2025-08-29
[QuantizationFormat] Remove code inferring format

dsikka:
Summary
- Removed and moved to compressed-tensors - requires: https://github.com/neuralmagic/compressed-tensors/pull/470
- Fix a bad test - we were setting `resume_from_checkpoint` as True when we're already loading in a model for training

### DIFF
--- a/tests/llmcompressor/transformers/finetune/test_oneshot_then_finetune.py
+++ b/tests/llmcompressor/transformers/finetune/test_oneshot_then_finetune.py
@@ -92,7 +92,6 @@ def test_oneshot_sparsification_then_finetune(tmp_path):
             concatenate_data=concatenate_data,
             splits=splits,
             recipe=recipe,
-            resume_from_checkpoint=True,  # use last checkpoint
         )
 
 
@@ -158,5 +157,4 @@ def test_oneshot_quantization_then_finetune(tmp_path):
             concatenate_data=concatenate_data,
             splits=splits,
             num_train_epochs=0.05,
-            resume_from_checkpoint=True,  # use last checkpoint
         )


tests/llmcompressor/transformers/compression/test_infer_quant_format.py
```diff
@@ -1,31 +0,0 @@
...
-    [
-        ["W8A8", "unstructured", "int-quantized"],
-        ["W8A16", "unstructured", "pack-quantized"],
-        ["W8A16", "2:4", "marlin-24"],
-        ["W4A16", "unstructured", "pack-quantized"],
-        ["W4A16", "2:4", "marlin-24"],
-        ["FP8", "unstructured", "float-quantized"],
-    ],
-)
-def test_infer_quant_format(preset, sparsity_structure, expected_format):
```
kylesayrs, line 21: Should we move this test?

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -227,21 +228,21 @@ def get_model_compressor(
...
-            model=model,
-            quantization_format=quantization_format,
-            save_compressed=save_compressed,
-            sparsity_structure=None
-            if sparsity_config is None
-            else sparsity_config.sparsity_structure,
+    if not save_compressed:
+        quantization_format = CompressionFormat.dense.value
+
+    if quantization_format is None and save_compressed:
```
kylesayrs: Is this condition now redundant given the above `if not save_compressed` check?

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -227,21 +228,21 @@ def get_model_compressor(
...
+        quantization_format: Optional[List[str]] = (
+            infer_and_set_per_module_quantization_format(
```
kylesayrs: FYI this function is still coupled with `ModelCompressor.from_pretrained_model`. I'd ideally like to see this function called inside `ModelCompressor.from_pretrained_model` at some point

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
```diff
@@ -227,21 +228,21 @@ def get_model_compressor(
...
-    quantization_format: Optional[List[str]] = (
-        infer_and_set_per_module_quantization_format(
-            model=model,
-            quantization_format=quantization_format,
-            save_compressed=save_compressed,
-            sparsity_structure=None
-            if sparsity_config is None
-            else sparsity_config.sparsity_structure,
+    if not save_compressed:
+        quantization_format = CompressionFormat.dense.value
```
kylesayrs: Should there not be a warning if a user specifies a format but also specifies `save_compressed=False`? This seems like an invalid configuration that they should be aware of

tests/llmcompressor/transformers/compression/test_infer_quant_format.py
dsikka: It’s in ct now as the method it is testing is in ct

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
kylesayrs: I thought some more about it, if you swap to
```
if save_compressed and quantization_format is None:
```

It's a lot clearer that the two statements are disjoint.

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
brian-dellabetta: is there ever a time we'd want to infer quant format outside of instantiating a ModelCompressor? If not, tying the logic into the call makes sense to me

src/llmcompressor/transformers/sparsification/compressed_tensors_utils.py
brian-dellabetta: could also use `elif quantization_format is None` instead.

brian-dellabetta (approved):

PR #1787, opened 2025-08-29
Updating API docs

aireilly:
1. Updated `mkdocs.yml` with the following changes:

    * Added `pymdownx.magiclink` extension - automatic URL linking

    * Excluded utility modules (`llmcompressor.version`, `llmcompressor.typing`) from auto-generated docs

    * Updated mkdocstrings config with new options:
        - `docstring_section_style`: list - Added list section formatting
        - `filters`: ["!^_"] - Hide private/internal methods from docs
        - `merge_init_into_class`: true - Merge class documentation
        - `separate_signature`: true - Function signature display
        - `show_root_heading`: true - Clear module headers
        - `show_signature_annotations`: true - Type hint visibility
        - `show_symbol_type_heading`: true - Clear type indicators
        - `summary`: true - Include summary sections

        https://mkdocstrings.github.io/python/reference/api/

2. Added high level doc strings to most modules.

    **Note**: Used Claude to summarize and create new docstrings.
    
Preview:
https://vllm--1787.org.readthedocs.build/projects/llm-compressor/en/1787/

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -70,6 +70,7 @@ markdown_extensions:
   - pymdownx.mark
   - pymdownx.smartsymbols
   - pymdownx.snippets
+  - pymdownx.magiclink
   - pymdownx.superfences:
       custom_fences:
         - name: mermaid
@@ -85,6 +86,9 @@ markdown_extensions:
 plugins:
   - api-autonav:
       modules: ['src/llmcompressor']
+      exclude:
+        - "llmcompressor.version"
+        - "llmcompressor.typing"
   - gen-files:
       scripts:
         - docs/scripts/gen_files.py
@@ -96,7 +100,21 @@ plugins:
       handlers:
         python:
           options:
-            docstring_style: sphinx
+            docstring_section_style: list
+            show_if_no_docstring: false
+            docstring_style: "sphinx"
+            filters:
+              - "!docstring"
+              - "!^_"
+            heading_level: 1
+            merge_init_into_class: true
+            parameter_headings: true
+            separate_signature: true
+            show_root_heading: true
+            show_signature_annotations: true
+            show_symbol_type_heading: true
+            show_symbol_type_toc: true
+            summary: true
   - search
   - section-index
   - social
--- a/src/llmcompressor/args/dataset_arguments.py
+++ b/src/llmcompressor/args/dataset_arguments.py
@@ -1,3 +1,12 @@
+"""
+Dataset argument classes for LLM compression workflows.
+
+This module defines dataclass-based argument containers for configuring dataset
+loading, preprocessing, and calibration parameters across different dataset
+sources and processing pipelines. Supports various input formats including
+HuggingFace datasets, custom JSON/CSV files, and DVC-managed datasets.
+"""
+
 from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, List, Optional, Union
 
--- a/src/llmcompressor/core/helpers.py
+++ b/src/llmcompressor/core/helpers.py
@@ -1,3 +1,11 @@
+"""
+Helper functions for core compression operations.
+
+Provides utility functions for logging model information and state
+management during compression workflows. Includes functionality for
+conditional logging and parameter tracking.
+"""
+
 from typing import Any, Generator, Optional, Tuple, Union
 
 from llmcompressor.core.state import State
--- a/src/llmcompressor/core/session.py
+++ b/src/llmcompressor/core/session.py
@@ -1,3 +1,11 @@
+"""
+Compression session management for LLM compression workflows.
+
+Provides the main CompressionSession class for managing compression
+workflows, including lifecycle management, event handling, callback
+registration, and state tracking.
+"""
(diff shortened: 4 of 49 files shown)


rahul-tuli:
Hi @aireilly Thanks for doing this, I have a ask though, since we don't have Module level docstrings for most of our modules, would it be possible to ignore them for the time being?

Screenshot 2025-09-02 at 6 14 22 PM

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -1,3 +1,12 @@
+""" Dataset argument classes for LLM compression workflows.
+
+This module defines dataclass-based argument containers for configuring dataset
+    loading, preprocessing, and calibration parameters across different dataset
+    sources and processing pipelines. Support various input formats including
```
brian-dellabetta:
```suggestion
    sources and processing pipelines. Supports various input formats including
```

src/llmcompressor/core/helpers.py
```diff
@@ -1,3 +1,10 @@
+""" Helper functions for core compression operations.
+
+This module provides utility functions for logging model information and state
+    management during compression workflows. Includes functionality for
+    conditional logging, model introspection, and parameter tracking.
```
brian-dellabetta: Not sure what is meant by model introspection
```suggestion
    conditional logging and parameter tracking.
```

src/llmcompressor/transformers/finetune/text_generation.py
```diff
@@ -14,6 +14,15 @@
...
 # limitations under the License.
 
+""" Text generation fine-tuning script with compression support.
+
+This module provides a command-line script for fine-tuning language models
+    on text generation tasks with integrated compression capabilities.
+    Supports various datasets, compression recipes, and training configurations
+    for optimizing large language models.
+
+"""
```
brian-dellabetta: this file is entirely deprecated, should we just exclude from the docs in that case?

brian-dellabetta (approved):
few comments, but otherwise lgtm

src/llmcompressor/transformers/finetune/text_generation.py
dsikka: I am going to get rid of this so it wont be in the next release. 
I think we can just update the docs to state that this no longer exists, and that they should be using the following paths (or something more elegant than the statements below):

```
 from llmcompressor.transformers import oneshot` is deprecated,  please use `from llmcompressor import oneshot.

from llmcompressor import train` is deprecated,  please use `from llmcompressor import train`.

 from llmcompressor.transformers import apply, compress` is deprecated, please use oneshot and train.

```

dsikka: https://github.com/vllm-project/llm-compressor/pull/1793

aireilly:
> Hi @aireilly Thanks for doing this, I have a ask though, since we don't have Module level docstrings for most of our modules, would it be possible to ignore them for the time being?


I tried with this in the the `mkdocs.yml` but it didn't work. Perhaps we could just add the missing docstrings? This PR is open let's strike while the iron is hot WDYT @rahul-tuli @brian-dellabetta 

```yaml
filters:
- "!docstring"
- "!^_"
``` 

edit: hmm. maybe we can use https://mkdocstrings.github.io/python/usage/configuration/docstrings/#show_if_no_docstring

aireilly:
OK. Not going to add more docstrings than are already reviewed. I do have a git stash of more maybe can add them in later PR.

I did add `show_if_no_docstring` to `mkdocs.yml`. This hides modules from the API docs if they have no docstrings, and provided they don't have child classes or functions that have docs. Not quite what you are looking for @rahul-tuli. We could manually hide modules in the `mkdocs.yml` but that wouldn't be great either. WDYT?

mkdocs.yml
```diff
@@ -85,6 +86,9 @@ markdown_extensions:
 plugins:
   - api-autonav:
       modules: ['src/llmcompressor']
+      exclude:
```
aireilly, line 89: @rahul-tuli We can exclude more modules like this.

brian-dellabetta (approved):
sorry @aireilly for the late reply, was OOO the past few working days. I agree, it would help with reviewing if we merge in what we have and add docs for the remainder, WDYT?

aireilly:
@dsikka think this one is ready PTAL

src/llmcompressor/core/session.py
```diff
@@ -47,7 +55,7 @@ def __init__(self):
     def lifecycle(self) -> CompressionLifecycle:
         """
         Lifecycle is used to keep track of where we are in the compression
-        process and what modifiers are active. It also provides the ability
+        process and what modifiers are active. It also Provides the ability
```
brian-dellabetta: nit
```suggestion
        process and what modifiers are active. It also provides the ability
```

brian-dellabetta (dismissed):
one nit, but otherwise LGTM

rahul-tuli (dismissed):
Thank you @aireilly

PR #1791, opened 2025-09-02
add support for per-head attention quantization

eldarkurtic:
This PR adds stuff needed for per-attention-head attention/KV-cache quantization.
Should be reviewed along with the corresponding PR in compressed-tensors: https://github.com/neuralmagic/compressed-tensors/pull/442


dsikka (changes requested):
See CT PR comments

kylesayrs:
https://github.com/vllm-project/compressed-tensors/pull/481

PR #1792, opened 2025-09-02
fix variable names to align with the new transformer version

shanjiaz:
SUMMARY:
Update past_key_value to past_key_values. 

TEST PLAN:
This should resolve the first three kv cache failures. I'm looking into the last one related to meta device.


dsikka (approved):
Thank you!

PR #1801, opened 2025-09-05
Consolidate example script tests into single parametrized test

fynnsu:
SUMMARY:

Combines the logic from `tests/examples/test_*.py` into `tests/examples/test_example_scripts.py` which has a parametrized test fn that runs them instead.

Notes:
- Removed README parsing: some tests previously parsed a README file and then ran a code block from that file. However, these code blocks where always just `python3 some_script.py`, so I replaced these by just calling the script instead.
- A handful of tests had additional extra handling. I maintained the behavior of these by adding additional options to the `TestCase` namedtuple / making the test function more flexible. (i.e. pre-processing, flags, and post-processing verification)

TEST PLAN:

All of the example tests essentially boil down to "run the example script and check if it crashes". A few also have additional checks or preprocessing steps but that is the main idea. To run the scripts, the tests all (both before and after changes) call `run_cli_command` in `tests/testing_utils.py` which does the actual `python ...` call. 

Therefore, to test this change I replaced that function with a dummy function that just prints the command and returns a success. Then I verified that the printed commands matched before and after changes (excluding reorderings of the calls). 

```python

def run_cli_command(cmd: List[str], cwd: Optional[Union[str, Path]] = None):
    print()
    print(" ".join(cmd), "in", str(cwd).split("examples/")[-1])

    class DummyReturn:
        returncode = 0

    return DummyReturn

    # return run(cmd, stdout=PIPE, stderr=STDOUT, check=False, encoding="utf-8", cwd=cwd)
```

In addition, to verify that the special handling works correctly I ran the full examples test suite and confirmed that the tests with special handling still passed.

### DIFF
--- a/tests/llmcompressor/transformers/compression/configs/channelwise_smoke.yaml
+++ b/tests/llmcompressor/transformers/compression/configs/channelwise_smoke.yaml
@@ -1,4 +1,4 @@
 cadence: "commit"
 test_type: "regression"
-model_stub: "nm-testing/llama2.c-stories15M"
+model_stub: "nm-testing/tinysmokellama-3.2"
 new_recipe: "tests/llmcompressor/transformers/compression/recipes/new_quant_channel.yaml"
--- a/tests/llmcompressor/transformers/compression/configs/fp8_smoke.yaml
+++ b/tests/llmcompressor/transformers/compression/configs/fp8_smoke.yaml
@@ -1,4 +1,4 @@
 cadence: "commit"
 test_type: "regression"
-model_stub: "nm-testing/llama2.c-stories15M"
+model_stub: "nm-testing/tinysmokellama-3.2"
 new_recipe: "tests/llmcompressor/transformers/compression/recipes/new_quant_fp8.yaml"
--- a/tests/llmcompressor/transformers/compression/configs/inputs_smoke.yaml
+++ b/tests/llmcompressor/transformers/compression/configs/inputs_smoke.yaml
@@ -1,4 +1,4 @@
 cadence: "commit"
 test_type: "regression"
-model_stub: "nm-testing/llama2.c-stories15M"
+model_stub: "nm-testing/tinysmokellama-3.2"
 new_recipe: "tests/llmcompressor/transformers/compression/recipes/new_quant_full.yaml"
--- a/tests/llmcompressor/transformers/compression/configs/weights_only_smoke.yaml
+++ b/tests/llmcompressor/transformers/compression/configs/weights_only_smoke.yaml
@@ -1,4 +1,4 @@
 cadence: "commit"
 test_type: "regression"
-model_stub: "nm-testing/llama2.c-stories15M"
+model_stub: "nm-testing/tinysmokellama-3.2"
 new_recipe: "tests/llmcompressor/transformers/compression/recipes/new_quant_weight.yaml"
--- a/tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py
+++ b/tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py
@@ -46,7 +46,7 @@ from tests.testing_utils import requires_gpu
 def test_sparse_model_reload(compressed, config, dtype, tmp_path):
     recipe_str = "tests/llmcompressor/transformers/obcq/recipes/test_tiny2.yaml"
     expected_sparsity = 0.5
-    model_path = "nm-testing/llama2.c-stories15M"
+    model_path = "nm-testing/tinysmokellama-3.2"
     dataset = "open_platypus"
     concatenate_data = False
     num_calibration_samples = 64
@@ -124,7 +124,7 @@ def test_sparse_model_reload(compressed, config, dtype, tmp_path):
 def test_dense_model_save(tmp_path, skip_compression_stats, save_compressed):
     reset_session()
 
-    model_path = "nm-testing/llama2.c-stories15M"
+    model_path = "nm-testing/tinysmokellama-3.2"
     model = AutoModelForCausalLM.from_pretrained(model_path)
 
     inferred_global_sparsity = SparsityConfigMetadata.infer_global_sparsity(model)
@@ -161,7 +161,7 @@ def test_quant_model_reload(format, dtype, tmp_path):
     recipe_str = (
         "tests/llmcompressor/transformers/compression/recipes/new_quant_simple.yaml"
     )
-    model_path = "nm-testing/llama2.c-stories15M"
+    model_path = "nm-testing/tinysmokellama-3.2"
     device = "cuda:0" if not torch.cuda.is_available() else "cpu"
     dataset = "open_platypus"
     concatenate_data = False
@@ -244,7 +244,7 @@ def test_quant_model_reload(format, dtype, tmp_path):
     ],
 )
 def test_model_reload(offload, torch_dtype, tie_word_embeddings, device, tmp_path):
-    model_path = "nm-testing/llama2.c-stories15M"
+    model_path = "nm-testing/tinysmokellama-3.2"
     save_path = tmp_path / "save_path"
 
     model = AutoModelForCausalLM.from_pretrained(model_path, torch_dtype=torch_dtype)
@@ -296,7 +296,7 @@ def test_model_reload_gpu(offload, torch_dtype, tie_word_embeddings, device, tmp
 )
 def test_model_shared_tensors(offload, torch_dtype, tie_word_embeddings, device):
     # load model
-    model_path = "nm-testing/llama2.c-stories15M"
+    model_path = "nm-testing/tinysmokellama-3.2"
     model = AutoModelForCausalLM.from_pretrained(model_path, torch_dtype=torch_dtype)
     if offload:
         model = dispatch_model(model, {"": device}, force_hooks=True)
@@ -337,7 +337,7 @@ def test_model_shared_tensors_gpu(offload, torch_dtype, tie_word_embeddings, dev
     "model_stub, recipe, sparse_format, quant_format",
     [
         (
-            "nm-testing/llama2.c-stories15M",
+            "nm-testing/tinysmokellama-3.2",
             "tests/llmcompressor/transformers/compression/recipes/sparse_24_fp8.yaml",
             CompressionFormat.sparse_24_bitmask.value,
             CompressionFormat.float_quantized.value,
@@ -418,7 +418,7 @@ def test_compressor_stacking(model_stub, recipe, sparse_format, quant_format, tm
     "model_stub, recipe, sparse_format",
     [
         (
-            "nm-testing/llama2.c-stories15M",
+            "nm-testing/tinysmokellama-3.2",
             "tests/llmcompressor/transformers/compression/recipes/sparse_24.yaml",
             CompressionFormat.sparse_24_bitmask.value,
         ),
--- a/tests/llmcompressor/transformers/compression/test_recipe_parsing.py
+++ b/tests/llmcompressor/transformers/compression/test_recipe_parsing.py
@@ -17,7 +17,7 @@ def setup_model_and_config(tmp_path):
     Loads a test model and returns common arguments used in oneshot runs.
     """
     model = AutoModelForCausalLM.from_pretrained(
-        "nm-testing/llama2.c-stories15M",
+        "nm-testing/tinysmokellama-3.2",
         torch_dtype="auto",
     )
 
--- a/tests/llmcompressor/transformers/finetune/data/conftest.py
+++ b/tests/llmcompressor/transformers/finetune/data/conftest.py
@@ -6,7 +6,7 @@ from llmcompressor.args import ModelArguments
 
 @pytest.fixture
 def tiny_llama_path():
-    return "nm-testing/llama2.c-stories15M"
+    return "nm-testing/tinysmokellama-3.2"
 
 
 @pytest.fixture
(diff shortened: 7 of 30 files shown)


tests/examples/utils.py
```diff
@@ -51,101 +42,60 @@ def requires_gpu_mem(required_amount: Union[int, float]) -> pytest.MarkDecorator
...
-        that language (happens prior to indexing).
-
-        :param position: position of code block to get (starting at 1)
-        :param lang: language of code block to filter by
-        :return: content of the code block
-        """
-        code_blocks = self.get_code_blocks(lang=lang)
-        code = code_blocks[position - 1].text.strip()
-        return code
+def replace_2of4_w4a16_recipe(content: str) -> str:
```
kylesayrs: Maybe separate these into `verify.py`?

tests/examples/test_example_scripts.py
```diff
@@ -0,0 +1,130 @@
...
+        ),
+        pytest.param(
+            "trl_mixin/ex_trl_constant.py",
+            marks=pytest.mark.skip("disabled until further updates"),
+        ),
+        pytest.param(
+            "trl_mixin/ex_trl_distillation.py",
+            marks=(
+                pytest.mark.skip("disabled until further updates"),
+                pytest.mark.multi_gpu,
```
kylesayrs, line 156: Can you explain how you're handling examples which require multiple GPUs?

tests/examples/utils.py
fynnsu: I'm a little resistant to doing this because they're not just verifiers but also preprocessors. I could rename the file `helpers.py` but that feels kind of the same as `utils.py`.

Also in some other pr, I will probably remove the `requires_gpu_mem` (currently unused) and `requries_gpu_count` (sort of has a duplicate in `testing_utils.py`) functions, which will further simplify this file.

tests/examples/test_example_scripts.py
fynnsu: Yes, so all but two of the original tests had `requires_gpu_count(1)`. The two exceptions are
1. `trl_mixin/ex_trl_distillation.py`: This test is also currently skipped. The "handling" is essentially just the `pytest.mark.multi_gpu`, which I copied over from the old implementation. I'm not sure if that gets used anywhere to be honest, but I maintained the existing behavior by adding it here. This is also one of the tests/examples I briefly advocated for removing in yesterday's meeting because it seems outdated and is skipped anyways. I copied it over here so that this pr is a refactor not a real code change, but I would also be good with removing it.
2. `quantizing_moe/mixtral_example.py`: This test was actually being called twice previously (once with `requires_gpu_count(2)` + `pytest.mark.multi_gpu` and once without). Because it was already completing successfully on the single gpu runs, I removed the second test.

That being said, this `pytest.param(..., marks=...)` structure does seem to work well, so it would be easy to mark any individual test with `requires_gpu_count(2)` which would then cause it only be run in multi-gpu environments. If you like me to restore the second mixtral test, or if we want in the future to add multi-gpu tests, that should be easy to do.

dsikka:
Please don’t merge without @dbarbuzzi’s approval

tests/examples/test_example_scripts.py
kylesayrs: Thanks. All that matters is that we have a system for adding multi-gpu tests moving foward.

I think something like this would be good moving forward.
```suggestion
                requires_gpu_count(2),
```

tests/examples/utils.py
kylesayrs: This code is specific to helping `tests/examples/test_example_scripts.py` and won't be used by other tests. Perhaps we should put it in the `tests/examples/test_example_scripts.py` file?

kylesayrs (approved):

tests/examples/utils.py
fynnsu: Done

tests/examples/test_example_scripts.py
dbarbuzzi: Just to chime in briefly, these two marks serve different purposes.

`pytest.mark.multi_gpu` is a custom mark that is used primarily for test collection. For example, example tests are run across multiple jobs, some using multiple GPUs and some using a single GPU, and thus they use the markers to know which tests to run.

`requires_gpu_count(num)` is more of a guard to ensure that, if the tests are run without specificity on which to execute, then it will cause tests that _cannot_ succeed due to insufficient hardware to be skipped as to not waste time executing until they eventually fail. Similarly, I had added a more appropriate one based on required VRAM since that is a key benefit of using multiple GPUs, but only one test I think ended up using that because it was hard to pin down what appropriate VRAM requirements were for different models/tests/etc.

tests/examples/test_example_scripts.py
```diff
@@ -0,0 +1,188 @@
...
+        content = test_case.preprocess_fn(content)
+        path.write_text(content, encoding="utf-8")
+
+    result = run_cli_command(command, cwd=script_working_dir)
+
+    if result.returncode != 0:
+        raise RuntimeError(
+            f"command failed with exit code {result.returncode}:\n"
+            f"Command:\n{shlex.join(command)}\nOutput:\n{result.stdout}"
+        )
```
dbarbuzzi: Any reason to use the `if` → `raise` structure here instead of `assert` with an error message?

fynnsu:
> Historically, one of the intents of the examples testing has been to test code blocks from the README as they were considered part of the example. We wanted to ensure that end users who were utilizing the rendered README (not just raw Markdown) and potentially copying code from it were not going to hit errors in the copied code.

> If we are not as concerned with this aspect any longer, it’s fine to remove these tests. However, if we do want to cover that angle, they shouldn’t be removed, though perhaps they can be modified with a different approach to validate the command without actually executing it (though it should still be extracted from rendered Markdown if this is the case, not the raw source).

I agree with the idea of testing the README codeblocks, however all of the existing README parsing tests were just running the "Quickstart" codeblocks which just call `python some_script.py`. 

- https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_2of4_sparse_w4a16#quickstart
- https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_kv_cache/README.md#quickstart
- https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/README.md#quickstart
- https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_w8a8_fp8#quickstart
- https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_w8a8_int8#quickstart
- https://github.com/vllm-project/llm-compressor/blob/main/examples/quantizing_moe/README.md#quickstart

You can also see this because a lot of the README tests have
```python
assert command.startswith("python")
```

I think it could be valuable to test the other walkthrough codeblocks in some way and we could use README parsing for that. But the current system is basically just testing that the script name in the readme matches, while adding a lot of complexity. If we think that's worthwhile confirming then I can add a test which just checks?

fynnsu:
> If interested, one refactor I had on my to-do list was to update how the script execution output was handled. Currently, it is inserted as part of the failure message, which was originally done so that it would only be written to output if the test failed. However, even if we are to keep that condition, I wanted to move it out of the error message; instead, just logging it separately (potentially only if the test failed if we wanted to keep similar behavior). I only point this out in case we wanted to try and incorporate that into this refactor or make a new task for future work.

Yeah I noticed the script outputs are currently suppressed while running. I think it would be useful to record the logs for all of the tests. Although maybe writing these to log files and uploading them as artifacts might be one way to do this without cluttering the console output.

I think this could be a separate pr, but I can help implement it if you'd like.

tests/examples/test_example_scripts.py
fynnsu: No not really, I can update it

dbarbuzzi:
> I agree with the idea of testing the README codeblocks, however all of the existing README parsing tests were just running the "Quickstart" codeblocks which just call `python some_script.py`.

As trivial as it sounds, that is actually part of the original intent (because there were occasionally things like typos in file names or the file gets renamed without the command being updated, etc.). Therefore, a lighter weight validation is still warranted if we are concerned with README contents (maybe checks that the command is basically comprised of two tokens – `python`/`python3` and a filename that matches one of the scripts in the current example’s folder).

fynnsu:
Okay that's fair. I think I would still rather test that explicitly in a separate test and keep the script running tests clean. I will add the test to this pr.

tests/examples/test_example_scripts.py
fynnsu: That makes sense. Perhaps we could update `requires_gpu_count` to automatically add the `pytest.mark.multi_gpu` marker if `num > 1`?

fynnsu: Okay, I ended up just adding the `mark.multi_gpu` manually. 

And restored the second `mixtral.py` test I had previously removed because that is the only test that is currently running on `mark.multi_gpu`

fynnsu:
@dbarbuzzi Please take a look when you get a chance. I added explicit readme tests to confirm that the script exists and is called using `python` or `python3`. I think I have also resolved the `multi_gpu` marking now. 

I was initially concerned that the readme tests would interfere with sharding (if all 6 lightweight readme were on one gpu and displaced other jobs), but
1. I tested this manually (e.g. `pytest tests/examples/ --collect-only --shard-id 0 --num-shards 2`) and the tests were split evenly
2. I looked at the source code for `pytest-shard` and I think the distribution of tests on shards doesn't depend on other tests but rather just the hash of each tests name

tests/examples/test_readmes.py
```diff
@@ -0,0 +1,31 @@
...
+        "quantization_w8a8_int8",
+        "quantizing_moe",
+    ],
+)
+def test_readmes(subdir):
+    path = Path("examples") / subdir / "README.md"
+
+    readme = ReadMe(path)
+
+    cmd = readme.get_code_block_content(position=1, lang="bash").split()
```
brian-dellabetta: shouldn't this be position=0 now that position has been changed to 0-indexed [here](https://github.com/vllm-project/llm-compressor/pull/1801/files#diff-0c00f7518dc4f4ab0db41870780e6541be28e81aaad6cb77d8c51cd403e779e4R81)

tests/examples/test_readmes.py
```diff
@@ -0,0 +1,31 @@
...
+        "quantization_2of4_sparse_w4a16",
+        "quantization_kv_cache",
+        "quantization_w4a16",
+        "quantization_w8a8_fp8",
+        "quantization_w8a8_int8",
+        "quantizing_moe",
+    ],
+)
+def test_readmes(subdir):
+    path = Path("examples") / subdir / "README.md"
```
brian-dellabetta, line 47: any way to do a search for `README.md` files to avoid having to update this file every time we add a new readme?

tests/examples/test_readmes.py
fynnsu: No already updated! The tests passing also confirms that it is set correctly.

tests/examples/test_readmes.py
fynnsu: Yeah I could implement this. I didn't yet because there are a few `README.md` files that don't fit this pattern (and were never tested previously). I could do auto collection + an ignore list to exclude these?

brian-dellabetta: eh, if there are places we don't want to test, we can leave this off for now. allow list or exclude list, we'd still have to maintain something so it's not a huge value-add

tests/examples/test_example_scripts.py
```diff
@@ -0,0 +1,190 @@
...
+from pathlib import Path
+from typing import Callable, List, NamedTuple
+
+import pytest
+from transformers import AutoConfig
+
+from tests.testing_utils import requires_gpu, run_cli_command
+
+
+def replace_2of4_w4a16_recipe(content: str) -> str:
```
brian-dellabetta, line 16: can you explain why this is needed? Can't we just rename the recipe directly?

tests/examples/test_readmes.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+
+    def __init__(self, path: Path) -> None:
+        self.path = path
+        self.content = self.path.expanduser().read_text(encoding="utf-8")
+        self.__normalize_code_fence_lang()
+        self.html = gfm_to_html(self.content)
+        self.soup = BeautifulSoup(self.html, "html.parser")
+
+    def __normalize_code_fence_lang(self):
```
brian-dellabetta: why double underscore here?

tests/examples/test_example_scripts.py
fynnsu: This is maintain the existing test behavior from `tests/examples/test_quantization_2of4_sparse_w4a16.py`. One of the original tests `test_doc_example_command` calls `llama7b_sparse_w4a16.py` and verifies the outputs. 

The other `test_alternative_recipe`, patches `llama7b_sparse_w4a16.py` to replace the recipe and then calls it.

So we can't rename the recipe directly because the original value is also used in a different test.

tests/examples/test_readmes.py
fynnsu: This class was just moved from `tests/examples/utils.py` as is (with the exception of switching to 0-based indexing). I didn't really do any clean up on it but I can do a little bit now.

fynnsu: Fixed in latest commit + other clean up of this class.

brian-dellabetta (approved):
thanks for updating, looks good!

tests/examples/test_readmes.py
```diff
@@ -0,0 +1,53 @@
...
+        "quantization_w4a16",
+        "quantization_w8a8_fp8",
+        "quantization_w8a8_int8",
+        "quantizing_moe",
+    ],
+)
+def test_readmes(subdir):
+    path = Path("examples") / subdir / "README.md"
+    readme = ReadMe(path)
+    cmd = readme.get_code_block_content(position=1, lang="shell").split()
```
dbarbuzzi, line 49: nit: It shouldn’t matter for this use-case since the commands being tested are very trivial, but [`shlex.split`](https://docs.python.org/3/library/shlex.html#shlex.split) should be used here instead of `str.split`.

PR #1805, opened 2025-09-10
Replace uses of deprecated `find_name_or_class_matches` fn

fynnsu:
SUMMARY:
`find_name_or_class_matches` was deprecated in https://github.com/neuralmagic/compressed-tensors/pull/406 and replaced by `match_targets`. 

This pr update llm-compressor to use the new fn. 

TEST PLAN:
`match_targets` should be a direct replacement. Testing is left to CI


brian-dellabetta (approved):
thanks for updating!

PR #1806, opened 2025-09-10
[transforms] SpinQuantModifier & QuIPModifier `transform_block_size` field

brian-dellabetta:
SUMMARY:
Resolves `INFERENG-1882`

The research community [has pointed out](https://github.com/IST-DASLab/FP-Quant?tab=readme-ov-file#fp-format-quantization-harness) that the rotation/transform block size is important when performing transforms:

> Key to efficiency is that the Hadamard block size matches the microscaling format group size (16 or 32)

This exposes a new field on SpinQuantModifier and QuIPModifier to allow the user to set it to an arbitrary value, as long as the model's hidden_size and head_dim are both evenly divisible by it.

- [x] Add to SpinQuant Modifier. Option to allow for different `transform_block_size`s for R1 vs. R2 can be added at a future time.
- [x] Add to QuIPModifier. Option to allow for different `transform_block_size`s for U vs. V can be added at a future time.

Merge in conjunction with:
 * https://github.com/neuralmagic/compressed-tensors/pull/466

TEST PLAN:
`transform_block_size` added to parameterized `tests/llmcompressor/modifiers/transform/(test_correctness.py|test_serialization.py)`


src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -219,6 +227,13 @@ def _create_r2_scheme(self, model: PreTrainedModel) -> TransformScheme:
         else:
             raise NotImplementedError()
 
+        if self.hadamard_size:
+            assert (
```
kylesayrs: Can you change this to a ValueError? Imho asserts should never trigger in practice, while ValueErrors are used to signal something wrong about the user's configuration.

kylesayrs (commented):
Looks good!

src/llmcompressor/modifiers/transform/spinquant/base.py
brian-dellabetta: updated

kylesayrs (approved):

brian-dellabetta:
> Sorry, could we use `transform_size`? Transform size truly describes the size of the transform and makes it clear the the transform is actually smaller.

I prefer `block` to be somewhere in the name, to distinguish the block size from the size of the entire transform matrix. WDYT?

src/llmcompressor/modifiers/transform/quip/base.py
```diff
@@ -115,6 +121,7 @@ def _create_config(self) -> TransformConfig:
             config_groups={
                 "v": TransformScheme(
                     type=self.transform_type,
+                    head_dim=self.transform_block_size,
```
kylesayrs:
```suggestion
                    block_size=self.transform_block_size,
```

src/llmcompressor/modifiers/transform/quip/base.py
```diff
@@ -134,6 +141,7 @@ def _create_config(self) -> TransformConfig:
                 ),
                 "u": TransformScheme(
                     type=self.transform_type,
+                    head_dim=self.transform_block_size,
```
kylesayrs:
```suggestion
                    block_size=self.transform_block_size,
```

src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -186,6 +193,7 @@ def _create_r1_scheme(self) -> TransformScheme:
             randomize=self.randomize,
             requires_grad=self.learnable,
             precision=self.precision,
+            head_dim=self.transform_block_size,
```
kylesayrs:
```suggestion
            block_size=self.transform_block_size,
```

kylesayrs:
@brian-dellabetta All good! `transform_block_size` is good

src/llmcompressor/modifiers/transform/quip/base.py
brian-dellabetta: Thanks, forgot to commit these changes, these should all be updated now

src/llmcompressor/modifiers/transform/spinquant/base.py
brian-dellabetta: updated

src/llmcompressor/modifiers/transform/quip/base.py
brian-dellabetta: updated

rahul-tuli (approved):
LGTM!

PR #1808, opened 2025-09-10
Remove `isort`, `black` and `flake8`, replace with `ruff`

fynnsu:
SUMMARY:
Ruff supports import sorting, formatting, and linting. We already use if for formatting (`black` is listed as a dev dependency but isn't used) and linting.

In this pr:
- 6af05b191ba171a0cc66c92088a768a2d5a8226b
  - Cleans up `pyproject.toml`, removing old configs/updating ruff configs to also check import sorting
  - Updates `Makefile` to just call `ruff` in `make quality` and `make style`
  - Removes `black`, `flake8`, and `isort` from dev dependencies
  - Updates dev docs description of `make quality` command
- 796a5b8632e88ca0d33e14737fba684ef97aa863
  - Replaces `flake8: noqa` file wide markers with `ruff: noqa` (not strictly necessary since `ruff` will respect the `flake8: noqa` markers but for cleanliness)
  - Removes unnecessary `isort: skip_file` markers

Note: this is builds off changes from #1807

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,27 +2,17 @@
 requires = ["setuptools", "wheel", "setuptools_scm==8.2.0"]
 build-backend = "setuptools.build_meta"
 
-[tool.black]
-line-length = 88
-target-version = ['py38']
-
-[tool.isort]
-profile = "black"
-skip = ["src/llmcompressor/transformers/tracing/", "src/llmcompressor/version.py"]
-
 [tool.mypy]
 files = "src/guidellm"
 
 [tool.ruff]
-exclude = ["build", "dist", "env", ".venv", "src/llmcompressor/transformers/tracing/", "src/llmcompressor/version.py"]
-lint.select = ["E", "F", "W"]
+extend-exclude = ["env", "src/llmcompressor/transformers/tracing/", "src/llmcompressor/version.py"]
+line-length = 88
+lint.select = ["E", "F", "W", "I"]
 lint.extend-ignore = ["E203", "W605"]
 
-# flake8 configuration is non-functional, only exists for vscode extensions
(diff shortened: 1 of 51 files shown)


fynnsu:
> isort is moving `transformers` below `llmcompressor` in the imports ordering now. We should update so that llmcompressor always come last. This can be done by adding to the pyproject.toml:
> 
> ```
> [tool.isort]
> known_first_party = ["compressed_tensors", "llmcompressor"]
> ```

Yeah this is a good point. I've added 
```toml
[tool.ruff.lint.isort]
known-first-party = ["llmcompressor", "compressed_tensors"]
```

Note: that this is still a change because compressed_tensors was previously not considered first party. So a lot of the `compressed_tensors` imports moved down but I think that's probably what we want.

pyproject.toml
```diff
@@ -2,27 +2,17 @@
...
+lint.select = ["E", "F", "W", "I"]
 lint.extend-ignore = ["E203", "W605"]
 
-# flake8 configuration is non-functional, only exists for vscode extensions
-# for the true flake8 configuration, see `Makefile`
-[tool.flake8]
-max-line-length = 88
-extend-ignore = ["E203", "W605"]
+[tool.ruff.lint.isort]
+known-first-party = ["llmcompressor", "compressed_tensors"]
```
brian-dellabetta: Sorry fynn, I told you wrong. Looks like compressed_tensors is imported in third party rather than first party. So i think we should do this instead to maintain what the ordering was before these changes:
```suggestion
known-first-party = ["llmcompressor"]
```

fynnsu: Yeah I think there's an argument to be made for treating it as first-party but I've removed it now and that does make the output the same as isort's.

PR #1815, opened 2025-09-12
[Actorder] Fix GPTQ actorder logic, only apply actorder to weight group args

dsikka:
SUMMARY:
- Don't set actorder to static if running channel quantization
- The modifier level actoder value is also getting serialized incorrectly - we set it to None if the strategy is not GROUP but still missing a step to serialize it correctlt when it is None
- FIx test case which is using an incorrect activation quantization strategy
-  Update compress / decompress test case 

Testing
- Fixes failing tests
- There is still a bug where `Sentinel` values are not serialized correctly

### DIFF
--- a/tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
+++ b/tests/llmcompressor/transformers/sparsification/test_compress_tensor_utils.py
@@ -348,7 +348,6 @@ def test_compressor_stacking(model_stub, recipe, sparse_format, quant_format, tm
     concatenate_data = False
     num_calibration_samples = 64
     splits = {"calibration": "train[:10%]"}
-    empty_model = AutoModelForCausalLM.from_pretrained(model_stub, torch_dtype="auto")
 
     oneshot(
         model=model_stub,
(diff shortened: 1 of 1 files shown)


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -149,9 +154,12 @@ def resolve_actorder(existing):
 
         for scheme in config.config_groups.values():
             assert isinstance(scheme, QuantizationScheme)
+            # actorder only valid for group quantization
             if scheme.weights is not None:
-                scheme.weights.actorder = resolve_actorder(scheme.weights.actorder)
-
+                if scheme.weights.strategy != QuantizationStrategy.GROUP.value:
+                    self.actorder = None
```
kylesayrs: This will override any subsequent grouped schemes.

For example, if the first scheme is channel but the second scheme is group, the second scheme will no longer have activation ordering (even if the user specified it/ the default is weight)

We should just pass in this case

brian-dellabetta (approved):

brian-dellabetta (approved):

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -149,9 +154,11 @@ def resolve_actorder(existing):
 
         for scheme in config.config_groups.values():
             assert isinstance(scheme, QuantizationScheme)
-            if scheme.weights is not None:
+            if (
+                getattr_chain(scheme, "weights.strategy", None)
```
rahul-tuli, line 158: 💯

PR #1822, opened 2025-09-16
Update beginner documentation

rayan-syed:
SUMMARY:
While exploring the LLM-Compressor project, I noticed that several beginner-level examples in the documentation were out of date and no longer run as written. This PR aims to fix these small issues, making the docs use non-deprecated code. A summary of the changes is below:
- Utilize `SamplingParams` as an input to `model.generate()` since old code no longer worked
- Align CLI + `curl` examples: use "TinyLlama-1.1B-Chat-v1.0-INT8" consistently (removes `./` prefix but keeps model key consistent between `vllm serve` and `curl`)
- Update import paths as needed
These changes affect only documentation, not runtime code.

TEST PLAN:
All changes here **only affect documentation**. All changes to the example code blocks were tested locally on a blank Python 3.9 conda environment with `llmcompressor` and `vllm` installed.

### DIFF
--- a/examples/quantization_w8a8_fp8/README_granite4.md
+++ b/examples/quantization_w8a8_fp8/README_granite4.md
@@ -1,0 +1,159 @@
+# `fp8` Weight and Activation Quantization for Granite 4
+
+`llmcompressor` supports quantizing weights and activations to `fp8` for memory savings and inference acceleration with `vllm`
+
+For Granite 4, in addition to typical `nn.Linear` layers in `mamba` or `mlp` modules, there are three "Linear-like" layers in `GraniteMoeHybridMoe` (`moe` module) that could be quantized as well. Among the three layers, usually `router` should be kept in high precision for accuracy reason. Therefore, users could choose to quantize the other two layers, `input_linear` and `output_linear`, for better model compression.
+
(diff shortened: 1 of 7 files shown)


docs/guides/saving_a_model.md
```diff
@@ -69,7 +69,7 @@ If you need more control, you can wrap `save_pretrained` manually:
 
 ```python
-from llmcompressor.transformers.sparsification import modify_save_pretrained
+from llmcompressor.transformers.sparsification.compressed_tensors_utils import modify_save_pretrained
```
brian-dellabetta, line 72: Does this get used anywhere? I think we can remove it entirely
```suggestion
```

rayan-syed: `modify_save_pretrained` is used in line 78:
```
# Manually wrap save_pretrained
modify_save_pretrained(model)
```

Unless you mean remove the entire **Manual Approach (Without oneshot)** section

brian-dellabetta: My bad, I missed a patch of hidden code. Will resolve and merge this in shortly, thanks @rayan-syed !

PR #1823, opened 2025-09-16
regex raw string when backslash is used

### LINKED ISSUE #1776: Escape backslash in ignore regex pattern
[Here](https://github.com/vllm-project/llm-compressor/blob/b26703f8835f6a3c81abdd5b41279d4367013932/examples/multimodal_vision/gemma3_example.py#L29) the string describing the regex pattern contains backslashes.
Python raises a `SyntaxWarning: invalid escape sequence '\.'` as a result.
Should these backslashes be escaped?

I don't think it's a bug but clarification would be nice.

brian-dellabetta:
SUMMARY:
Resolves #1776 

The `gemma3_example.py` script was not correctly escaping the `\` character in its regexes. This adds the `r"` prefix to the string, so that python knows it is a raw string that needs to be escaped, avoiding warnings.

It appears to be isolated to this example, I couldn't find other regexes with `\` characters that weren't being properly escaped.

TEST PLAN:
Example runs successfully after changes.

### DIFF
--- a/examples/multimodal_vision/gemma3_example.py
+++ b/examples/multimodal_vision/gemma3_example.py
@@ -32,8 +32,8 @@ recipe = [
         scheme="W4A16",
         ignore=[
             "lm_head",
-            "re:model\.vision_tower.*",
-            "re:model\.multi_modal_projector.*",
+            r"re:model\.vision_tower.*",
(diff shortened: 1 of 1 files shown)


shanjiaz (approved):
Nice

PR #1825, opened 2025-09-16
[tests] move recipe parsing test so it is run nightly

brian-dellabetta:
SUMMARY:
`text_recipe_parsing.py` was previously in `tests/llmcompressor/recipe`, but since it is decorated with `@requires_gpu` it was never being run. `llm-compressor-testing` only runs these tests if marked as `integration` tests and they are inside `tests/llmcompressor/transformers/(compression|finetune|obcq)` (see [here](https://github.com/neuralmagic/llm-compressor-testing/blob/main/.github/actions/test/action.yml#L28)).

This moves the test into compression and marks it as an integration test so it will be included in tests. This is valid since it is calling into oneshot and running a compression flow on the model.

TEST PLAN:
- [x] Confirm with @dbarbuzzi that these tests are now being run in nightly. Tests should succeed now that #1810 is merged in


dbarbuzzi (dismissed):
I can confirm that these combined changes should result in the test getting picked up in our nightly runs.

PR #1830, opened 2025-09-17
[Observers] Small observers cleanup, add e2e quantization tests

kylesayrs:
## Purpose ##
* Add tests to quantization in order to add more confidence to future quantization refactors
  * Each of the tested values has been checked and look correct

## Changes ##
* Observers changes
  *  Update `calculate_updated_min_max` to store running values, even if the function computation shortcuts (helpful testing)
  * Update `get_qparams_along_dim` to support multiple dims and negative dims
    * This actually results in a silent typing bug with token quantization, and is fixed on the base class implementation
    * This change essentially duplicates the base class implementation. Future work could involve cleaning up the inheritance structure here

## Testing ##
* Added tests pass

### DIFF
--- a/src/llmcompressor/observers/base.py
+++ b/src/llmcompressor/observers/base.py
@@ -171,7 +171,7 @@ class Observer(InternalModule, RegistryMixin):
                     group_indices, group_sizes = torch.unique(g_idx, return_counts=True)
                     group_sizes = group_sizes[torch.argsort(group_indices)]
 
-                    observed = observed.index_select(g_idx, -1)
+                    observed = observed.index_select(-1, g_idx)
 
                 # TODO: experiment with vectorizing for loop for performance
                 end = 0


tests/llmcompressor/modifiers/calibration/test_observers.py
```diff
@@ -59,3 +64,320 @@ def test_observers_update(shape, group_size, actorder):
...
+        #         1: torch.tensor([[3]]),
+        #     },
+        #     {
+        #         "default": torch.tensor([[2]]),
+        #         1: torch.tensor([[5]]),
+        #     },
+        #     torch.tensor([[0.0000, 0.9766, 1.9531, 3.3125, 3.3125, 4.9688]]),
+        #     0.1,
+        # ),
+        # block is not supported, but is in principle similar to group
```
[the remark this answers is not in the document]
kylesayrs: I can't believe I'm listening to a clanker

shanjiaz (commented):
Do you think we should update the mse observer as well? Since reset was removed from calibration.

kylesayrs:
@shanjiaz global parameters are not supported by MSE observer anyways, so it's not relevant without that support

shanjiaz (approved):
Looks good to me!

src/llmcompressor/observers/min_max.py
```diff
@@ -126,14 +130,23 @@ def calculate_qparams(
...
-        reduce_dims = tuple(idx for idx in range(observed.ndim) if idx != dim)
+        # cast to set
+        if isinstance(dim, int):
+            dim = [dim]
+        dim = set(dim)
+
+        # convert negative dims
+        dim = [d if d >= 0 else observed.ndim + d for d in dim]
+
+        # reduce all dimensions except the the one passed as argument to this function
```
brian-dellabetta:
```suggestion
        # reduce all dimensions except the ones passed as argument to this function
```

src/llmcompressor/observers/min_max.py
```diff
@@ -126,14 +130,23 @@ def calculate_qparams(
...
         Calculate quantization parameters along the specified dimension
         """
-        reduce_dims = tuple(idx for idx in range(observed.ndim) if idx != dim)
+        # cast to set
+        if isinstance(dim, int):
+            dim = [dim]
+        dim = set(dim)
+
+        # convert negative dims
+        dim = [d if d >= 0 else observed.ndim + d for d in dim]
```
brian-dellabetta, line 145: Shouldn't the cast to set happen after this line?

src/llmcompressor/observers/min_max.py
```diff
@@ -86,9 +88,11 @@ def calculate_gparam(self, observed: torch.Tensor) -> torch.Tensor:
         :return: updated global scale derived from the observed tensor
         """
 
-        updated_min_val, updated_max_val = self.calculate_updated_min_max(
-            observed=observed
-        )
+        # patch to avoid affecting running means
```
brian-dellabetta: this is because we are calculating global scale right? we don't want the `calculate_qparams` result to change based on this calculation?

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -147,7 +147,6 @@ def update_weight_global_scale(module: Module):
         should_calculate_gparam=True,
         should_calculate_qparams=False,
     )
-    module.weight_observer.reset()
```
dsikka: Because we only attach one observer, I’m fairly sure we’re resetting to prevent global scale  metrics from impacting quant scale metrics

kylesayrs: > Update calculate_gparam to restore original running values, rather than relying on resetting after calculation

Reset is replaced by patching and restoring the metrics

src/llmcompressor/observers/min_max.py
kylesayrs: > Update calculate_gparam to restore original running values, rather than relying on resetting after calculation

Yes

src/llmcompressor/observers/min_max.py
kylesayrs: Technically either is fine, since the argument type just needs to be an iterable. I'm purely matching the implementation on the base model for now

> Update get_qparams_along_dim to support multiple dims and negative dims
This actually results in a silent typing bug with token quantization, and is fixed on the base class implementation
This change essentially duplicates the base class implementation. Future work could involve cleaning up the inheritance structure here

src/llmcompressor/observers/min_max.py
dsikka: Why is this preferable? If anything, this now seems more confusing

kylesayrs: From a programming standpoint, this decouples `calculate_gparam` and `Observer.reset` (there's no way to footgun yourself by calling `calculate_gparam` and forgetting to call `Observer.reset`.

From a functionality standpoint, I think this fixes a bug where metrics would be updated twice (which has implications for running values), specifically when called from `calibrate_activations`. In the case of activations, we don't want to reset after each gparam calculation, since we still need those metrics to compute running values.

dsikka: I think you're right about the 2nd point.

I don't know if I agree with the first point. This feels like a hack.

kylesayrs: ## Case 1 ##

Consider the case of `strategy="tensor_group"`, `dynamic=False` and `averaging_constant != 1`.

On activation hook, `calibrate_activations` calls `call_observer` with `should_calculate_gparam=True`\*. This causes `calculate_updated_min_max` to be called twice, which **causes the running min/max to move faster than if no global param was calculated**.

## Case 2 ##

Consider the case of `strategy="tensor_group"`, `dynamic="local"` and `averaging_constant != 1`.

Originally, `calculate_gparam` would call `calculate_updated_min_max` would be called and the running values would update (twice\*). Now, the running values will not update.

\* Note that running values are updated, even if `should_calculate_qparams=False`

## TLDR ##
So it seems that this change fixes a bug where running values are updated twice, but changes the behavior of `dynamic="local"` to calculate global parameters based on true values, not running values. I assumed that global parameters should be the true min/max of all values, not running values, but maybe @dsikka you think this shouldn't be the case?

I've reverted the change since it's not necessary for group quant, but we should definitely look into exactly the behavior we want for global scales (and scales in general. Runnings means are slightly strange anyways and seem to be a vestige of QAT).

src/llmcompressor/observers/min_max.py
brian-dellabetta: I mean more that you might end up with duplicates in `dim` if you create this list and don't cast back to a set.

e.g. if there are 3 dims and dim=`{1,2,-1}`, then `dim=[1,2,2]` after this line.

kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/1903

PR #1840, opened 2025-09-17
MSE observer for NVFP4 

shubhra:
SUMMARY:
"MSE Observer implemented for NVFP4 for both weights and activations. Originally, I had added a new observer `mseact` for input activations global scale calculation but moved it under the same observer to keep it clean."

### DIFF
--- a/src/llmcompressor/observers/mse.py
+++ b/src/llmcompressor/observers/mse.py
@@ -89,7 +89,7 @@ class MovingAverageMSEObserver(Observer):
 
             if (is_fp4(self.quantization_args)) and global_scale is None:
                 # If the quantization scheme is fp4 and global_scale is still None
-                # i.e it has not yet been optimized, then we are should first get
+                # i.e it has not yet been optimized, then we should first get
                 # the global scale and then optimize the local scales.
                 # Local scales are set to by the absolute min and max.
                 iteration_global_scale = generate_gparam(


src/llmcompressor/observers/mse.py
```diff
@@ -53,6 +53,13 @@ def calculate_mse_min_max(
         :param global_scale: optional scale to further scale local quantization scales
         :return: tuple of min and max values derived from the observed tensor
         """
+        print(f"[MSE DEBUG] calculate_mse_min_max called")
+        print(f"[MSE DEBUG]   observed.shape: {observed.shape}")
+        print(f"[MSE DEBUG]   observed.dtype: {observed.dtype}")
+        print(f"[MSE DEBUG]   reduce_dims: {reduce_dims}")
+        print(f"[MSE DEBUG]   global_scale: {global_scale}")
+        print(f"[MSE DEBUG]   maxshrink: {self.maxshrink}, grid: {self.grid}, norm: {self.norm}, patience: {self.patience}")
```
dsikka: no prints

src/llmcompressor/observers/mse.py
```diff
@@ -61,6 +68,9 @@ def calculate_mse_min_max(
             absolute_min_val = torch.amin(observed, dim=reduce_dims, keepdims=True)
             absolute_max_val = torch.amax(observed, dim=reduce_dims, keepdims=True)
 
+        print(f"[MSE DEBUG]   absolute_min_val: {absolute_min_val}")
+        print(f"[MSE DEBUG]   absolute_max_val: {absolute_max_val}")
```
dsikka: remove

src/llmcompressor/observers/mseact.py
```diff
@@ -0,0 +1,303 @@
...
+from compressed_tensors.quantization.utils import calculate_qparams
+from torch import FloatTensor, IntTensor, Tensor
+
+from llmcompressor.observers.base import Observer
+
+__all__ = ["MovingAverageMSEObserverActivations"]
+
+
+@Observer.register("mseact")
+class MovingAverageMSEObserverActivations(Observer):
```
dsikka: Please add these changes to the original MSE observer. We shouldn't be duplicating

dsikka (changes requested):

src/llmcompressor/observers/mseact.py
shubhra: Yes, as soon as I finalize the core implementation I'll move it under there. Just wanted to keep it separate for now to test cleanly.

src/llmcompressor/observers/mse.py
```diff
@@ -36,6 +42,7 @@ def __init__(
         self.averaging_constant = averaging_constant
         self.grid = grid
         self.norm = norm
+        self.is_activation = base_name != "weight"
```
brian-dellabetta, line 45: if this is the only place we use `base_name`, it might just be better to expose `is_activation: bool = False` to the constructor instead of base_name

src/llmcompressor/observers/mse.py
```diff
@@ -25,6 +30,7 @@ def __init__(
         averaging_constant: float = 0.01,
         grid: float = 100.0,
         norm: float = 2.4,
+        base_name: str = "weight",
```
brian-dellabetta, line 33: will have to confirm with @shanjiaz if this needs to be added elsewhere, since there are a couple different places Observers are instantiated

src/llmcompressor/observers/mse.py
```diff
@@ -75,18 +85,38 @@ def calculate_mse_min_max(
             shrinked_min_val = p * absolute_min_val
             shrinked_max_val = p * absolute_max_val
 
+            from compressed_tensors.quantization.utils import generate_gparam
+
+            if (is_fp4(self.quantization_args)) and global_scale is None:
+                # If the quantization scheme is fp4 and global_scale is still None
+                # i.e it has not yet been optimized, then we are should first get
```
brian-dellabetta:
```suggestion
                # i.e it has not yet been optimized, then we should first get
```

kylesayrs:
I think this is going to be a simpler approach

https://github.com/vllm-project/llm-compressor/pull/1903

dsikka:
@mergifyio refresh

PR #1844, opened 2025-09-18
Refactor to standardize `pytest` usage

fynnsu:
SUMMARY:
- Replaces all instances of `unittest.TestCase` classes with pytest test functions. `setUp` methods were generally replaced with fixtures or folded into the test functions.

- Tests were updated to use `tmp_path` fixture instead of a local output dir. Test tmp dir cleanup logic was removed as this will be handled automatically by pytest. 
- Tests that used `parameterized` classes were replaced with uses of `pytest.mark.parameterized` or parameterized `pytest.fixtures`
- Cleaned up/combined some testing utils.
- The `parameterized` package was removed as a dependency

TEST PLAN:
I've thoroughly reviewed the pr and as far as I can tell this should be a refactor. There were minimal changes to the code logic made.

I also ran most of the tests while updating them. And will verify that CI runs correctly.


tests/llmcompressor/transformers/compression/test_decompress.py
```diff
@@ -31,106 +27,81 @@ class TestDecompression(unittest.TestCase):
...
-            decompressed_model_hf_quantizer_out = (
-                self.decompressed_model_hf_quantizer.generate(**inputs, max_length=50)
-            )
-
-            assert torch.equal(
-                decompressed_model_hf_quantizer_out, decompressed_model_manual_output
-            )
-
-    @classmethod
-    def tearDownClass(self):
```
brian-dellabetta, line 131: does this teardown still occur?

tests/llmcompressor/transformers/finetune/test_finetune_no_recipe_custom_dataset.py
```diff
@@ -1,154 +1,137 @@
...
-
-        mock_filename = f"{filename}.{extension}"
-        mock_filepath = os.path.join(path, mock_filename)
-
-        with open(mock_filepath, "w") as mock_file:
-            mock_file.write(mock_content)
-
-        return mock_filepath  # Return the file path
-
-    def tearDown(self):
```
brian-dellabetta, line 108: same here

tests/llmcompressor/transformers/finetune/test_finetune_without_recipe.py
```diff
@@ -1,44 +1,31 @@
...
-            model=self.model,
-            dataset=self.dataset,
-            output_dir=self.output,
-            recipe=recipe_str,
-            max_steps=max_steps,
-            concatenate_data=concatenate_data,
-            splits=splits,
-        )
-
-    def tearDown(self):
```
brian-dellabetta, line 42: this looks to be missing

brian-dellabetta (approved):
LGTM! Super glad to be getting rid of `unittest`

tests/llmcompressor/transformers/compression/test_decompress.py
fynnsu: As discussed offline:
1. tmp dir cleanup (previously handled in teardowns) was removed and is now managed automatically by `pytest` through the usage of the `tmp_path` fixture.
2. Teardown logic removing models / objects stored as class attributes is no longer required because we no longer store those values as attributes. 
3. Clear cuda cache and similar logic is now being handled by `pytest` fixtures which can contain cleanup code after `yield`ing to the test function.

tests/llmcompressor/transformers/finetune/test_finetune_no_recipe_custom_dataset.py
fynnsu: Same as above.

tests/llmcompressor/transformers/finetune/test_finetune_without_recipe.py
fynnsu: Same as above

dsikka:
Very nice, will review soon

tests/llmcompressor/modifiers/logarithmic_equalization/test_base.py
```diff
@@ -11,28 +9,22 @@
...
-
-    def test_log_equalization_is_registered(self):
-        modifier = ModifierFactory.create(
-            type_="LogarithmicEqualizationModifier",
-            allow_experimental=False,
-            allow_registered=True,
-            **self.kwargs,
-        )
+def test_logarithmic_equalization_is_registered():
+    setup_modifier_factory()
```
dbarbuzzi, line 13: nit (maybe better as a separate/future pass) – this function seems ripe for converting to a pytest fixture, and applied case via the `pytest.mark.usefixtures` decorator:

```python
@pytest.mark.usefixtures("setup_modifier_factory")
def test_logarithmic_equalization_is_registered():
    ...
```

tests/llmcompressor/pytorch/modifiers/smoothquant/test_pytorch.py
```diff
@@ -8,22 +6,22 @@
 from tests.llmcompressor.pytorch.helpers import LinearNet
 
 
-@pytest.mark.unit
-class TestSmoothQuantMapping(unittest.TestCase):
-    def setUp(self):
-        self.model = LinearNet()
-        self.state = State(model=self.model)
+@pytest.fixture
```
dbarbuzzi, line 9: nit: this seems to be identically defined in at least two places, if there were more than two then it might be better to hoist this to somewhere it can be imported or to `conftest` where it because automatically available without needing import.

dbarbuzzi (approved):
This looks great! Love to see the embrace of the `pytest` built-ins to reduce lots of code to let things be handled automatically.

Only left a couple of very tiny nitpicks as specific comments, and in general it would be nice for fixtures to defined in `conftest.py` files where possible if they’re used in more than one file – there can be a `conftest.py` in each directory to use as a scoping mechanism, and this allows their use without needing to be imported.

tests/llmcompressor/modifiers/logarithmic_equalization/test_base.py
fynnsu: Fixed!

tests/llmcompressor/pytorch/modifiers/smoothquant/test_pytorch.py
fynnsu: Also updated!

tests/data.py
```diff
@@ -19,8 +19,3 @@ class Cadence(Enum):
 class TestConfig:
     test_type: TestType
     cadence: Cadence
-
```
dsikka, line 22: maybe also move TestConfig to the uils and remove this file?

dsikka (approved):

tests/data.py
fynnsu: Yeah make sense!

dsikka (approved):

PR #1847, opened 2025-09-19
Fix logic to identify MLP blocks that need to be fused

anmarques:
SUMMARY:
This PR fixes the logic used to determine if up_proj and gate_proj should be merged during quantization. It checked whether up_proj OR gate_proj were present whereas in fact you need BOTH to be present for the fusion to occur.

This bug was blocking the quantization of the Apertus models since they don't have the gate_proj modules. The broken logic was flagging the MLP layers for fusion but then the code broke because it couldn't find gate_proj.

TEST PLAN:
Successfully quantized the Apertus models:

https://huggingface.co/RedHatAI/Apertus-8B-Instruct-2509-FP8-dynamic
https://huggingface.co/RedHatAI/Apertus-70B-Instruct-2509-FP8-dynamic

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -114,7 +114,7 @@ setup(
         ("pyyaml>=6.0.1,<=6.0.2" if BUILD_TYPE == "release" else "pyyaml>=6.0.1"),
         # librosa dependency numba is currently not compatible with numpy>=2.3
         # https://numba.readthedocs.io/en/stable/user/installing.html#version-support-information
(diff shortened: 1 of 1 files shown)


rahul-tuli (approved):
Thank you Alex!

brian-dellabetta (approved):

brian-dellabetta (approved):

dsikka (commented):
I lied. We check for the presence of the gate_proj before checking for tensor_group.
This is correct change. Thank you!

PR #1851, opened 2025-09-22
Moe calibration refactor

### LINKED ISSUE #1829: [Help Wanted] Refactor/ Clean up MoE calibration logic
## Background ##
MoE models require special logic in order to calibrate because the majority of expert activations are gated to behind the MoE gating mechanism. Attempting to naively calibrate experts with the base huggingface model definition means that some activations will not receive enough samples in order to properly calibrate.

Another reason why MoEs are difficult to calibrate is that sometimes the huggingface model definition fuses all experts together into one weight. This one weight might be too large to fit into one GPU's memory, which is a limiting factor for memory-limited use cases.

The solution is to write specialized logic to replace MoE experts modules into individual expert modules which can be individually calibrated and offloaded.

At a high level, the routing logic is converted as follows:

```python3
if calibrate_all_experts:
   output += expert(x)[top_k_tokens] * weights[expert_index]
else:
   output += expert(x[top_k_tokens]) * weights[expert_index]
```

Right now, the logic for determining when and how module replacements happen is written in [modeling/prepare.py](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/prepare.py). However, this logic is split between `replace_modules_for_calibration` and `moe_calibration_context`. This confusing logic makes it difficult for people to contribute new MoE model replacements.

> **_NOTE:_** Some models such as Llama4 can be loaded in vllm in their replaced form, and therefore do not need to be restored after `moe_calibration_context` exit.

## Goals ##
* Make MoE model contribution as easy and standardized as possible
* Standardize on `moe_calibration_context` (remove/deprecate `replace_modules_for_calibration`

## Suggested task list ##
- [ ] Remove/deprecate `replace_modules_for_calibration`
- [ ] Remove/deprecate [DatasetArgs.calibrate_moe_context](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/args/dataset_arguments.py#L120-L129) (it should always be on)
- [ ] Refactor `moe_calibration_context` to not require that the context stack is passed as argument
- [ ] Create a standardized and simple interface that `moe_calibration_context` and the files in the [modeling](https://github.com/vllm-project/llm-compressor/tree/main/src/llmcompressor/modeling) folder can use to easy contribute moe contexts
- [ ] Add tests for changing and restoring model structure in and outside of the `moe_calibration_context` (you can skip downloading model weights for the tests, see [these examples](https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/modeling/test_calib_deepseek_v3.py#L21-L22))

sairampillai:
SUMMARY:
Fixes #1829
- Introduce new MoeContextCalibration class to easily add Contextual and Permanent MoE expert calibration
- Depreate `replace_modules_for_calibration`

TEST PLAN:
WIP

### DIFF
--- a/src/llmcompressor/modeling/moe_context.py
+++ b/src/llmcompressor/modeling/moe_context.py
@@ -1,0 +1,155 @@
+"""
+Standardized interface for MoE model calibration.
+MoE calibration context is used to apply MoE calibration modifications to the model.
+There are two types of MoE calibration contexts:
+1. ContextualMoECalibration: uses context managers for temporary modifications
+2. PermanentMoECalibration: permanently modifies the model
+"""
+
+import contextlib
+from abc import ABC, abstractmethod
+from typing import Dict, TypeVar, Union
+
+from transformers import PreTrainedModel
+
+T = TypeVar("T", bound="MoECalibrationContext")
+
+
+class MoECalibrationContext(ABC):
+    """
+    Abstract base class for MoE calibration.
+    This provides a standardized interface for MoE model calibration.
+    """
+
+    @abstractmethod
+    def apply(self, model: PreTrainedModel, calibrate_all_experts: bool = True) -> None:
+        """
+        Apply MoE calibration modifications to the model.
+        :param model: The model to modify
+        :param calibrate_all_experts: Whether to calibrate all
+        experts or only routed ones
+        """
+        pass
+
+    @abstractmethod
+    def restore(self, model: PreTrainedModel) -> None:
+        """
+        Restore the model to its original state.
+        :param model: The model to restore
+        """
+        pass
+
+
+class ContextualMoECalibration(MoECalibrationContext):
+    """
+    MoE calibration that uses context managers for temporary modifications.
+    This is suitable for models that need to be restored after calibration.
+    """
+
+    def __init__(self, model_class_name: str, update_function):
+        """
+        Initialize the context manager-based MoE calibration.
+        :param model_class_name: The class name of the model this context applies to
+        :param update_function: Function that applies the MoE modifications
+        """
+        self.model_class_name = model_class_name
+        self.update_function = update_function
+        self._stack = None
+
+    def apply(self, model: PreTrainedModel, calibrate_all_experts: bool = True) -> None:
+        """Apply MoE calibration modifications using context managers."""
+        if self._stack is None:
+            self._stack = contextlib.ExitStack()
+
+        self.update_function(model, self._stack, calibrate_all_experts)
+
+    def restore(self, model: PreTrainedModel) -> None:
+        """Restore the model by exiting the context stack."""
+        if self._stack is not None:
+            self._stack.close()
+            self._stack = None
+
+
+class PermanentMoECalibration(MoECalibrationContext):
+    """
+    MoE calibration context that permanently modifies the model.
+    This is suitable for models that can be loaded in their modified form
+    (e.g., Llama4 in vLLM).
+    """
+
+    def __init__(self, model_class_name: str, replacement_function):
+        """
+        Initialize the permanent MoE calibration.
+        :param model_class_name: The class name of the model this context applies to
+        :param replacement_function: Function that permanently replaces MoE modules
+        """
+        self.model_class_name = model_class_name
+        self.replacement_function = replacement_function
+        self._original_modules = {}
+
+    def apply(self, model: PreTrainedModel, calibrate_all_experts: bool = True) -> None:
+        """Apply permanent MoE calibration modifications."""
+        # Store original modules for potential restoration
+        for name, module in model.named_modules():
+            if module.__class__.__name__ == self.model_class_name:
+                self._original_modules[name] = module
+
+        # Apply the replacement
+        self.replacement_function(model, calibrate_all_experts)
+
+    def restore(self, model: PreTrainedModel) -> None:
+        """Restore original modules (if needed)."""
+        # For permanent MoE calibrations, restoration is typically not needed
+        # as the model is meant to stay in its modified form
+        pass
+
+
+# Registry for MoE calibrations
+_MOE_CONTEXTS: Dict[str, MoECalibrationContext] = {}
+
+
+def register_moe_context(model_class_name: str, context: MoECalibrationContext) -> None:
+    """
+    Register a MoE calibration context for a model class.
+    :param model_class_name: The class name of the model
+    :param context: The MoE calibration context to register
+    """
+    _MOE_CONTEXTS[model_class_name] = context
+
+
+def get_moe_context(model_class_name: str) -> Union[MoECalibrationContext, None]:
+    """
+    Get the registered MoE calibration context for a model class.
+    :param model_class_name: The class name of the model
+    :return: The MoE calibration context or None if not found
+    """
+    return _MOE_CONTEXTS.get(model_class_name)
+
+
+def list_supported_models() -> list:
+    """
+    List all model classes that have registered MoE calibration contexts.
+    :return: List of supported model class names
+    """
+    return list(_MOE_CONTEXTS.keys())
+
+
+# Convenience function for backward compatibility
+def create_context_manager_context(model_class_name: str, update_function):
+    """
+    Create a context manager-based MoE calibration.
+    :param model_class_name: The class name of the model
+    :param update_function: Function that applies the MoE modifications
+    :return: A ContextualMoECalibration instance
+    """
+    return ContextualMoECalibration(model_class_name, update_function)
+
+
+def create_permanent_context(model_class_name: str, replacement_function):
+    """
+    Create a permanent MoE calibration.
+    :param model_class_name: The class name of the model
+    :param replacement_function: Function that permanently replaces MoE modules
+    :return: A PermanentMoECalibration instance
+    """
+    return PermanentMoECalibration(model_class_name, replacement_function)
--- a/src/llmcompressor/modeling/prepare.py
+++ b/src/llmcompressor/modeling/prepare.py
@@ -1,3 +1,5 @@
+import warnings
+
 import tqdm
 from compressed_tensors.utils import replace_module
 from transformers import PreTrainedModel
@@ -20,6 +22,14 @@ def replace_modules_for_calibration(
     model: PreTrainedModel,
     calibrate_all_experts: bool = True,
 ) -> PreTrainedModel:
+    # This function is deprecated. Use moe_calibration_context instead.
+    warnings.warn(
+        "replace_modules_for_calibration is deprecated. \
+    Use moe_calibration_context instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+
     for name, module in tqdm.tqdm(list(model.named_modules())):
         cls_name = module.__class__.__name__
         if cls_name in replacements:


sairampillai:
@kylesayrs @dsikka Few clarifications:
- I pushed a couple of commits without signing, how do you suggest I fix that?
- I have deprecated the [`calibrate_moe_context`](src/llmcompressor/args/dataset_arguments.py) parameter, do we want to plan how to phase it out?
- I have tested using unit tests but without GPU (gpu poor), can you point me to best ways to test this change end-to-end?

brian-dellabetta:
@sairampillai , regarding DCO, you can ignore that. We can sign it via github once reviewed/approved

examples/multimodal_vision/llama4_example.py
```diff
@@ -3,18 +3,15 @@
...
-from llmcompressor.modeling import replace_modules_for_calibration
 from llmcompressor.modifiers.quantization import GPTQModifier
 
 # Select model and load it.
 model_id = "meta-llama/Llama-4-Scout-17B-16E-Instruct"
 model = Llama4ForConditionalGeneration.from_pretrained(model_id, torch_dtype="auto")
 processor = Llama4Processor.from_pretrained(model_id)
-# We update `Llama4TextMoe` modules with custom `SequentialLlama4TextMoe`.
-# This change allows compatibility with vllm.
-# To apply your own custom module for experimentation, consider updating
```
dsikka, line 15: I think we want to keep the note about the requirement of restructuring for vLLM and allowing the user to update the calibrate as they need to

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -46,9 +46,7 @@ def __call__(
 
         with contextlib.ExitStack() as stack:
             stack.enter_context(calibration_forward_context(model))
-
-            if dataset_args is not None and dataset_args.calibrate_moe_context:
-                moe_calibration_context(model, stack)
+            stack.enter_context(moe_calibration_context(model))
```
dsikka: I dont think we want to do this for every case as not every model will be an MoE

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -85,8 +85,7 @@ def __call__(
             if not dataset_args.quantization_aware_calibration or disable_qac:
                 stack.enter_context(DisableQuantization(model))
 
-            if dataset_args.calibrate_moe_context:
-                moe_calibration_context(model, stack)
+            stack.enter_context(moe_calibration_context(model))
```
dsikka: same as above

src/llmcompressor/pipelines/basic/pipeline.py
sairampillai: This is implemented along with a quick check to see if a particular model is added to the experts replacement list in [`modeling/prepare.py`](https://github.com/sairampillai/llm-compressor/blob/ba42881c3367a2b04be889ba122a71fd806803ec/src/llmcompressor/modeling/prepare.py#L52-L57) as such here in [`modeling/prepare.py`](https://github.com/sairampillai/llm-compressor/blob/ba42881c3367a2b04be889ba122a71fd806803ec/src/llmcompressor/modeling/prepare.py#L102C5-L111C52):

This would end the need for any parameters in DatasetArgs, simplifying MoE calibration. Do you think there would be overhead when we do stack.enter_context()? Do you recommend a better way to implement this?

examples/multimodal_vision/llama4_example.py
sairampillai: Done

dsikka:
@kylesayrs

src/llmcompressor/pipelines/basic/pipeline.py
kylesayrs: I think it's fair to assume that a model is an MoE if its architecture is listed in the dictionary `MOE_EXPERTS_REPLACEMENTS`, therefore entering the context in all cases is fine.

kylesayrs: As a small nit, consider entering the moe context [here](https://github.com/vllm-project/llm-compressor/blob/b4a99d4cd802d0cfca0becf30117ac5a47aa9ef0/src/llmcompressor/entrypoints/oneshot.py#L201). Entering the context before the pipeline call comes with some benefits

1. We no longer need to enter the context for each pipeline explicitly
2. We no longer need to enter, exit, and reenter in cases where multiple pipelines are composed (independent pipeline)

```python3
with moe_calibration_context(self.model):
    pipeline(...)
```

src/llmcompressor/modeling/prepare.py
```diff
@@ -20,6 +28,14 @@ def replace_modules_for_calibration(
     model: PreTrainedModel,
     calibrate_all_experts: bool = True,
 ) -> PreTrainedModel:
+    # This function is deprecated. Use moe_calibration_context instead.
+    warnings.warn(
```
kylesayrs: nit: Use [compressed_tensors/deprecated](https://github.com/neuralmagic/compressed-tensors/blob/0d8c7c36d3eb2c6a00ee5fbcf496fb15a489d9e2/src/compressed_tensors/utils/helpers.py#L173)

src/llmcompressor/modeling/prepare.py
```diff
@@ -35,37 +51,67 @@ def replace_modules_for_calibration(
...
+    if moe_context is None:
+        # No MoE context registered for this model, yield unchanged
+        yield model
+        return
+
+    # Apply MoE calibration
+    moe_context.apply(model, calibrate_all_experts)
+
+    try:
+        yield model
```
kylesayrs: It seems like this yield value is never used. Do you still want to include it?

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -126,16 +127,6 @@ class DatasetArguments(CustomDatasetArguments):
         default=512,
         metadata={"help": "Number of samples to use for one-shot calibration"},
     )
-    calibrate_moe_context: bool = field(
```
kylesayrs, line 129: Can we add a dataset argument called `moe_calibrate_all_experts` which defaults to True?

src/llmcompressor/modeling/prepare.py
```diff
@@ -35,37 +51,67 @@ def replace_modules_for_calibration(
...
-                )
-            )
+# MoE model configurations - centralized registry
+# Adding a new MoE model is now as simple as adding an entry here!
+# This follows the same pattern as MAPPINGS_REGISTRY in SmoothQuant and AWQ
+MOE_EXPERTS_REPLACEMENTS = {
+    "Qwen3MoeForCausalLM": MoEModelConfig(
+        calibration_type=MoECalibrationType.CONTEXTUAL,
+        target_class_name="Qwen3MoeDecoderLayer",
+        target_attribute="mlp",
```
kylesayrs: Ideally we shouldn't need to ever target attributes, only target parent modules. For example, only targeting `Qwen3MoeMLP`

src/llmcompressor/modeling/moe_context.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def restore(self, model: PreTrainedModel) -> None:
+        """Restore original modules (if needed)."""
+        # For permanent MoE calibrations, restoration is typically not needed
+        # as the model is meant to stay in its modified form
+        pass
+
+
+# Registry for MoE calibrations
+_MOE_CONTEXTS: Dict[str, MoECalibrationContext] = {}
```
kylesayrs: I think having `_MOE_CONTEXTS`, `replacements`, and `MOE_EXPERTS_REPLACEMENTS` is more than we need.

Additionally, mapping from model class names to replacement modules will run into issues for nested modules (for example, if an MoE model is inside of a nested model architecture, we will not be able to replace its modules).

Ideally we could simplify this to just one dictionary which maps module class names to their contexts. We can also use ABC to define `from_original` and `restore` interfaces. For example:

```python3
T = TypeVar("T")

class MoECalibrationModule(ABC):
     is_permanent = False

     @classmethod
     def from_original(self, original: T) -> Self:
         # converts from original module to moe calibration module
         ...

     def restore(self) -> T:
         # might include repacking weights/qparams into 3d structure, or not
         ...
```

```python3
from llmcompressor.modeling.deepseek_v3 import CalibrationDeepseekV3MoE


moe_modules: Dict[str, MoECalibrationModule] = {
    "DeepseekV3MoE": CalibrationDeepseekV3MoE,
    ...
}

def moe_calibration_context(model, dataset_args):
        if module.__class__.__name__ in moe_modules:
            replacement = moe_modules[module.__class__.__name__].from_original(module, dataset_args.calibrate_all_experts)
            model.set_submodule(name, replacement)

    # ... maybe some logging about if/which modules were replaced
    # maybe some logging about if the structure will stay `is_permanent`

    yield

        if isinstance(module, MoECalibrationModule):
            original = module.restore()
            model.set_submodule(name, original)
```

While keying by *module class names* rather than by *model class names* requires iterating through all of the modules in the model in order to check, I think this is minimal overhead and acceptable in order to support nested models. What do you think @dsikka?

src/llmcompressor/modeling/moe_context.py
```diff
@@ -0,0 +1,304 @@
...
+    target_attribute: Optional[str] = None
+    description: Optional[str] = None
+
+    def __post_init__(self):
+        """Validate configuration after initialization."""
+        if (
+            self.calibration_type == MoECalibrationType.CONTEXTUAL
+            and self.target_attribute is None
+        ):
+            raise ValueError("target_attribute is required for contextual calibration")
```
kylesayrs: Is this coupling necessary? Why do we need to specify an attribute at all?

src/llmcompressor/modeling/moe_context.py
```diff
@@ -0,0 +1,304 @@
...
+                            module=getattr(module, target_attr),
+                            calibrate_all_experts=calibrate_all_experts,
+                        ),
+                    )
+                )
+
+    return update_function
+
+
+def register_moe_model(model_class_name: str, config: MoEModelConfig):
```
kylesayrs: This might be more abstraction than is necessary

src/llmcompressor/pipelines/basic/pipeline.py
sairampillai: Good idea! I will try this change and test it out

src/llmcompressor/modeling/prepare.py
kylesayrs: I'm actually unfamiliar, do we need to specify `target_attribute`s?

kylesayrs:
Hey @sairampillai! Are you still interested in contributing to this PR? If not, please let me know and I can assign someone to pick up where you left off!

sairampillai:
@kylesayrs I am working on the updates, I will push an update soon for review!

src/llmcompressor/modeling/prepare.py
sairampillai: We can remove it

src/llmcompressor/pipelines/sequential/pipeline.py
sairampillai: Addressing this with kylesayrs's suggestion

src/llmcompressor/modeling/prepare.py
sairampillai: That's my oversight, we don't need to specify it for either contextual or permanent replacements.

src/llmcompressor/args/dataset_arguments.py
sairampillai: I will remove the deprecated field and replace it with `moe_calibrate_all_experts` which will be used in entrypoint and anywhere required.

src/llmcompressor/modeling/moe_context.py
```diff
@@ -0,0 +1,177 @@
...
+        with moe_calibration_context(model):
+            # Run calibration - all experts will see data
+            for batch in dataloader:
+                model(**batch)
+        # Model is now restored (unless permanent)
+    """
+    replaced = {}
+
+    # Step 1: Find and replace MoE modules
+    for name, module in model.named_modules():
```
kylesayrs, line 123: Iterating through this might take some time, especially if there are modules to replace. You'll want to make sure the user has some feedback, rather than hanging right before calibration starts without explanation.

Consider doing one pass to collect all the `(name, module)`s to replace, then a tqdm iteration to replace those modules, along with any logging.

kylesayrs (commented):
Looks great so far, thanks for following up!

src/llmcompressor/modeling/moe_context.py
sairampillai: Got it. Is there any performance related tests you recommend to run?

src/llmcompressor/pipelines/sequential/pipeline.py
sairampillai: Resolved with context manager invoked in entrypoint/oneshot.py

src/llmcompressor/pipelines/basic/pipeline.py
sairampillai: Added the context manager in `entrypoint/oneshot.py`.

src/llmcompressor/modeling/prepare.py
sairampillai: Done

src/llmcompressor/modeling/moe_context.py
sairampillai: Done!

src/llmcompressor/modeling/moe_context.py
sairampillai: Done

src/llmcompressor/modeling/moe_context.py
```diff
@@ -0,0 +1,188 @@
...
+            f"{self.__class__.__name__} has is_permanent=False but doesn't "
+            "implement restore()"
+        )
+
+
+# Registry: module class name -> calibration module class
+MOE_CALIBRATION_MODULES: Dict[str, Type[MoECalibrationModule]] = {}
+
+
+def register_moe_calibration(module_class_name: str):
```
kylesayrs, line 69: Something like this is also implemented via the [RegistryMixin](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/registry/registry.py#L68), but we can standardize that in a follow up as well.

kylesayrs: Your registry is slightly different, let's leave this for a follow up

examples/quantization_w8a8_fp8/llama4_fp8_block_example.py
```diff
@@ -10,7 +9,9 @@
 # Load model.
 model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
 tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
-model = replace_modules_for_calibration(model)
+# MoE calibration is now handled automatically by the pipeline.
+# The `SequentialLlama4TextMoe` modules will be applied during calibration
+# to enable proper expert calibration and vLLM compatibility.
```
brian-dellabetta: do we want to keep this note for all examples? Might be cleaner without them, what do people think?

src/llmcompressor/modeling/deepseek_v3.py
```diff
@@ -65,11 +87,18 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
-    return DeepseekV3MoECalibrate(
-        config=config, original=module, calibrate_all_experts=calibrate_all_experts
+    """
+    Legacy replacement function.
+    Use CalibrationDeepseekV3MoE.from_original() instead.
+    """
+    return CalibrationDeepseekV3MoE.from_original(
+        original=module,
+        config=config,
+        calibrate_all_experts=calibrate_all_experts,
```
brian-dellabetta: Why not just use the constructor? We can probably remove the `from_original` class method
```suggestion
    return CalibrationDeepseekV3MoE(
        original=module,
        config=config,
        calibrate_all_experts=calibrate_all_experts,
```

src/llmcompressor/modeling/llama4.py
```diff
@@ -74,9 +108,14 @@ def __init__(self, config: Llama4TextConfig, original: Llama4TextExperts):
...
 
+# Legacy function for backward compatibility
 def replace(config: Llama4Config, module: Llama4TextMoe, calibrate_all_experts: bool):
-    return SequentialLlama4TextMoe(
-        config=config.get_text_config(),
+    """
+    Legacy replacement function.
+    Use SequentialLlama4TextMoe.from_original() instead.
+    """
+    return SequentialLlama4TextMoe.from_original(
```
brian-dellabetta: Same here, should just be able to use the constructor directly, no?

examples/quantization_w8a8_fp8/llama4_fp8_block_example.py
sairampillai: I felt it helpful to have the note in varied examples since this would be a breaking change once we deprecate older methods. Open to recommendations.

src/llmcompressor/modeling/deepseek_v3.py
sairampillai: neater this way! updated

src/llmcompressor/modeling/moe_context.py
sairampillai: Right, I saw and didn't want to delay this any further. We can work on the follow up.

kylesayrs (commented):
Hi @sairampillai!

I think the new interface looks great! It was easy for me to contribute qwen3_vl_moe. I think there's a small import here which I've fixed in this patch (along with adding qwen3_vl_moe, rearranging some imports). I think once the import issue is fixed, we should land this ASAP.

[import_qwen3vl.patch](https://github.com/user-attachments/files/23168501/import_qwen3vl.patch)

I agree with @brian-dellabetta that there's some overlap between `__init__` and `from_original`, but we can revisit that later.

src/llmcompressor/modeling/deepseek_v3.py
```diff
@@ -23,6 +31,20 @@ def __init__(
...
+        config: DeepseekV3Config,
+        calibrate_all_experts: bool = True,
+    ) -> "CalibrationDeepseekV3MoE":
+        """Create calibration module from original DeepseekV3MoE."""
+        return cls(
+            config=config,
+            original=original,
+            calibrate_all_experts=calibrate_all_experts,
+        )
+
```
brian-dellabetta: We should be able to remove all of these, since it mimics the constructor we can just the constructor instead
```suggestion
```

examples/quantizing_moe/deepseek_r1_example.py
```diff
@@ -20,7 +19,9 @@
     model_id, torch_dtype="auto", config=config
 )
 tokenizer = AutoTokenizer.from_pretrained(model_id)
-model = replace_modules_for_calibration(model)
+# MoE calibration is now handled automatically by the pipeline.
+# The `CalibrationDeepseekV3MoE` modules will be applied during calibration
```
dsikka: If we're going to make a note of this in every example, it is probably helpful indicating where these custom modules are being defined as they are not in the original model definition

tests/llmcompressor/modeling/test_calib_llama4.py
```diff
@@ -28,39 +27,43 @@ def test_calib_replace_llama4_moe_all_experts(model_stub):
             model_stub, torch_dtype="auto"
         )
 
-    replace_modules_for_calibration(model, calibrate_all_experts=True)
+    with contextlib.ExitStack() as stack:
```
dsikka, line 30: Is your linter up-to-date with llmcompressor? Seems to just be adding spaces here?

examples/quantizing_moe/deepseek_r1_example.py
sairampillai: Makes sense, will add to the note

tests/llmcompressor/modeling/test_calib_llama4.py
sairampillai: Let me recheck, I did use `make style` to check for linting errors

sairampillai:
@dsikka, I will create the issues `Qwen3-VL-MoE` and `Qwen3-Next` but it looks like @kylesayrs already has a patch ready for `Qwen3-VL-MoE`.

dsikka:
Hi @sairampillai - do you mind doing the following so that we can get this over the line?
1. Resolve the conflicts with `dataset_arguments.py`
3. Add in @kylesayrs's patch for qwen3_vl_moe (I believe this will fix imports as well)

We can land this once these two things are done

dsikka:
@sairampillai Seems like there are some test failures.
Can you also make sure you're using the proper linter, by running the following commands in a clean env in the root directory?

```python
uv pip install -e .[dev]
make style
```

sairampillai:
@dsikka @kylesayrs 
- Created #1978 and #1979 
- Resolved merge conflict
- Can't access Kyle's [Qwen3-VL-MoE patch](https://github.com/user-attachments/files/23168501/import_qwen3vl.patch) - Let me know how I can get access to that attachment

examples/quantizing_moe/deepseek_r1_example.py
sairampillai: Done

src/llmcompressor/modeling/deepseek_v3.py
sairampillai: Done for all relevant files and removed usage of `from_original()` from the abstract base class too.

src/llmcompressor/modeling/moe_context.py
sairampillai: Can you add an issue for this and assign to me once this is merged?

kylesayrs:
@sairampillai No worries, we can land this in a follow up

kylesayrs (approved):
This is a much needed change! Thank you

sairampillai:
> @sairampillai Seems like there are some test failures. Can you also make sure you're using the proper linter, by running the following commands in a clean env in the root directory?
> 
> ```python
> uv pip install -e .[dev]
> make style
> ```

Linting seems to not catch the test files issue, I will manually verify and fix:
```
(.new-env) ➜  llm-compressor git:(moe_calibration_refactor) ✗ make style
Running python styling
ruff format src tests examples setup.py; 
356 files left unchanged
ruff check --fix src tests examples setup.py;
All checks passed!
ruff format --silent src tests examples setup.py; 
```

Yes, post rebase I am seeing import errors in my tests too:
```
(.new-env) ➜  llm-compressor git:(moe_calibration_refactor) ✗ CADENCE=weekly HF_TOKEN=token pytest tests/llmcompressor/modeling/test_calib_deepseek_v3.py tests/llmcompressor/modeling/test_calib_qwen3.py tests/llmcompressor/modeling/test_calib_llama4.py -v
ImportError while loading conftest 'llm-compressor/tests/llmcompressor/conftest.py'.
tests/llmcompressor/conftest.py:8: in <module>
    from llmcompressor.modifiers.factory import ModifierFactory
E   ModuleNotFoundError: No module named 'llmcompressor'
```

examples/quantization_w4a4_fp4/llama4_example.py
```diff
@@ -3,18 +3,18 @@
...
 model_id = "meta-llama/Llama-4-Scout-17B-16E-Instruct"
 model = Llama4ForConditionalGeneration.from_pretrained(model_id, torch_dtype="auto")
 processor = Llama4Processor.from_pretrained(model_id)
-# We update `Llama4TextMoe` modules with custom `SequentialLlama4TextMoe`.
-# This change allows compatibility with vllm.
-# To apply your own custom module for experimentation, consider updating
-# `SequentialLlama4TextMoe` under llmcompressor/modeling/llama4.py
-model = replace_modules_for_calibration(model)
+# MoE calibration is now handled automatically by the pipeline.
+# The `SequentialLlama4TextMoe` modules (from `llmcompressor.modeling.llama4`)
```
brian-dellabetta, line 13: nit -- Any reason this one is prefixed Sequential while the qwen and deepseek are prefixed Calibration?

sairampillai: I have tried to mimic the existing class names but with prefix and I think I didn't update the class name for Llama4. The prefix `Sequential` was preexisting for llama4.

sairampillai:
> Yes, post rebase I am seeing import errors in my tests too:
> 
> ```
> (.new-env) ➜  llm-compressor git:(moe_calibration_refactor) ✗ CADENCE=weekly HF_TOKEN=token pytest tests/llmcompressor/modeling/test_calib_deepseek_v3.py tests/llmcompressor/modeling/test_calib_qwen3.py tests/llmcompressor/modeling/test_calib_llama4.py -v
> ImportError while loading conftest 'llm-compressor/tests/llmcompressor/conftest.py'.
> tests/llmcompressor/conftest.py:8: in <module>
>     from llmcompressor.modifiers.factory import ModifierFactory
> E   ModuleNotFoundError: No module named 'llmcompressor'
> ```

Tests work when installing default version `uv pip install -e .` without dev dependencies.
```
2025-10-29T20:46:22.528503+0530 | moe_calibration_context | INFO - Entering MoE calibration context
2025-10-29T20:46:22.566975+0530 | moe_calibration_context | INFO - Found 58 MoE modules to replace
Replacing MoE modules for calibration: 100%|███████████████████████████████████████████████████████████████████████████████████████████| 58/58 [00:00<00:00, 47972.71it/s]
2025-10-29T20:46:22.568660+0530 | moe_calibration_context | INFO - Replaced 58 MoE modules for calibration
2025-10-29T20:46:22.568703+0530 | moe_calibration_context | INFO - 58/58 modules will remain in calibration form (permanent)
PASSED2025-10-29T20:46:30.325090+0530 | reset | INFO - Compression lifecycle reset
```

examples/quantization_w4a4_fp4/llama4_example.py
brian-dellabetta: Ok, we can close off for now, thanks for the info

kylesayrs:
Hi @sairampillai, looks like CI was able to run and pass the tests. I think this may have something to do with your local install? Glad to see it's resolved now.

Issue #1850, opened 2025-09-22
[Help Wanted] Update `oneshot` args to match `*Arguments` config classes


fynnsu:
`src/llmcompressor/entrypoints/oneshot.py:oneshot` relists all the arguments from the `DatasetArguments, ModelArguments, RecipeArguments, TrainingArguments` class in `src/llmcompressor/args`.

However, it seem like a few arguments are missing or have different type hints or defaults. These should be updated to match. Missing arguments include `preprocessing_func, data_collator, raw_kwargs, max_train_samples, pipeline, tracing_ignore, sequential_targets` from the `DatasetArguments` class.

ArkaSanka:
Please assign this to me, I am interested in looking into this.

dsikka:
Thank you!

Etelis:
If @ArkaSanka is not working on this, could you assign me? @dsikka

ArkaSanka:
Hi @Etelis, I'm working on it, submitting the PR soon

dsikka:
Hi @ArkaSanka are you interested in working on this?

ArkaSanka:
Yes @dsikka, if you could getting the PR #1957, I can rebase it

PR #1854, opened 2025-09-22
Small param fixes

fynnsu:
SUMMARY:
Two small fixes:
1. `splits` isn't included in the `oneshot` function signature even though it is used that way in multiple tests. There are also a few other args that are missing (which aren't used by tests currently), opened #1850 to track.
2. `sequential_update` is deprecated but was still listed in a test and a couple examples. Removed.


src/llmcompressor/modifiers/obcq/sgpt_base.py
```diff
@@ -34,7 +34,7 @@ class SparsityModifierBase(Modifier):
     owl_lmbda: Optional[float] = None
 
     # data pipeline arguments
-    sequential_update: Optional[bool] = False  # deprecated
+    sequential_update: Optional[bool] = True  # deprecated
```
fynnsu: I changed this default because in the `field_validator` on line 49-58 (below), we
1. overwrite this value to `True`
2. raise a warning if it was originally `False`

That means we were by default raising the warning.

dsikka: why can't we just remove

fynnsu: True. Added a commit to remove.

PR #1860, opened 2025-09-23
[Deprecations] clear usage of deprecated package, function, and field

brian-dellabetta:
SUMMARY:
Clear out some deprecation warnings in logs:

- [x] Replace `pynvml` with `nvidia-ml-py` (which has the same API):
```
../../../../../opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/torch/cuda/__init__.py:63
  /opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/torch/cuda/__init__.py:63: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.
    import pynvml  # type: ignore[import]
```

- [x] Replace`safe_permute(x,y, dim=d)` with `x.index_select(dim=d, index=y)`:
```
tests/llmcompressor/observers/test_min_max.py::test_g_idx
  /opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/llmcompressor/observers/base.py:172: DeprecationWarning: safe_permute is deprecated and will be removed in a future release. Please use Tensor.index_select instead.
    observed = safe_permute(observed, perm, dim=1)
```

- [x] Only display warning for `MagnituePruningModifier.leave_enabled` if field has been set to something other than default value
```
  /opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/llmcompressor/modifiers/pruning/magnitude/base.py:43: DeprecationWarning: MagnitudePruningModifier.leave_enable has been deprecated
    warnings.warn(
```

TEST PLAN:
No net new src code, tests pass

PR #1861, opened 2025-09-23
[Logging] clean up CompressionLogger verbosity

brian-dellabetta:
SUMMARY:
Prerequisites:
* #1860 

This updates the `CompressionLogger`, used by GPTQ and other modifiers to log monitoring metrics during runtime. Previously, all GPUs were logged even if a subset were visible. This updates the logs to only log visible devices, and move helper functions into methods. If visible devices cannot be parsed, it will default to logging all of them like before

TEST PLAN:
logs for visible GPUS show correctly when `CUDA_VISIBLE_DEVICES=0,1,2` or `CUDA_VISIBLE_DEVICES=[0,1,2]`, and logs for all GPUs show when `CUDA_VISIBLE_DEVICES=all` or some invalid value.


dsikka:
@mergifyio refresh

dsikka:
@brian-dellabetta can we close this?

brian-dellabetta:
> @brian-dellabetta can we close this?

@dsikka this should be ready for review, I just resolved conflicts. Goal is to only show metrics for visible GPUs instead of all of them

rahul-tuli (approved):

PR #1863, opened 2025-09-24
Add Qwen3-Coder awq moe example

fynnsu:
SUMMARY:
Add a `Qwen3-Coder` awq example based on https://huggingface.co/nm-testing/Qwen3-Coder-30B-A3B-Instruct-W4A16-awq.

TEST PLAN:
Ran script locally and verified resulting model outputs were reasonable.

Prompt: "Write a binary search function"
Output:
```python
def binary_search(arr, target):
    left = 0
    right = len(arr) - 1

    while left <= right:
        mid = (left + right) // 2

        if arr[mid] == target:
            return mid
        elif arr[mid] < target:
            left = mid + 1
        else:
            right = mid - 1

    return -1
```

### DIFF
--- a/examples/awq/qwen3_coder_moe_example.py
+++ b/examples/awq/qwen3_coder_moe_example.py
@@ -1,0 +1,83 @@
+from datasets import load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.awq import AWQModifier
+from llmcompressor.utils import dispatch_for_generation
+
+MODEL_ID = "Qwen/Qwen3-Coder-30B-A3B-Instruct"
+SAVE_DIR = MODEL_ID.split("/")[-1] + "-W4A16-awq"
+
+# Configure the quantization algorithm to run.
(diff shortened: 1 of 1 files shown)


examples/awq/qwen3_coder_moe_example.py
```diff
@@ -0,0 +1,83 @@
...
+    ### Apply algorithms.
+    ###
+    oneshot(
+        model=model,
+        dataset=get_calib_dataset(tokenizer),
+        recipe=recipe,
+        max_seq_length=MAX_SEQUENCE_LENGTH,
+        num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+        log_dir=None,
+        trust_remote_code_model=True,
```
brian-dellabetta: is this needed for qwen 3?

fynnsu: Nope, seems like they're not needed. I've removed them.

brian-dellabetta (approved):
looks great! excited to have this in

shanjiaz (approved):
Nice

PR #1870, opened 2025-09-26
[Transforms] Update examples for R3 and `transform_block_size` option

brian-dellabetta:
SUMMARY:
Prerequisites:
- [ ] https://github.com/neuralmagic/compressed-tensors/pull/472

This PR updates the SpinQuant and Quip examples to include `transform_block_size` and the latest R4 feature in SpinQuant. It also reverts the `TransformScheme.block_size` changes previously introduced into CT, and updated in Pr linked above. While `block_size` is a more appropriate name, `head_dim` has already landed in vllm, and it would be too much of a pain to change. Users will rarely create their own `TransformScheme` anyway.

TEST PLAN:
Both examples run and the saved model can be run in vllm, defaulting to use the hadacore kernel

### DIFF
--- a/examples/transform/spinquant_example.py
+++ b/examples/transform/spinquant_example.py
@@ -12,7 +12,7 @@ model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
 tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 
 # NOTE: currently only rotations R1, R2, and R4 are available
-# Learned rotations and R3 will be added in a future release.
+# R3 and learned R1/R2 rotations will be added in a future release.
 # Configure the quantization algorithm to run.
 #   * apply spinquant transforms to model to reduce quantization loss
 #   * quantize the weights to 4 bit with group size 128
--- a/examples/transform/spinquant_example.py
+++ b/examples/transform/spinquant_example.py
@@ -11,14 +11,15 @@ MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
 model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
 tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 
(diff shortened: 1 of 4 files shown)


examples/transform/spinquant_example.py
```diff
@@ -11,14 +11,15 @@
...
 tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 
-# NOTE: currently only fused rotations (R1 & R2) are available
-# Learned rotations and online rotations (R3 & R4) will be added
-# in a future release.
+# NOTE: currently only rotations R1, R2, and R4 are available
+# R3 and learned R1/R2 rotations will be added in a future release.
 # Configure the quantization algorithm to run.
 #   * apply spinquant transforms to model to reduce quantization loss
 #   * quantize the weights to 4 bit with group size 128
```
fynnsu, line 18: Is this "group size 128" still correct? Or is the group size equal to the `transform_block_size` here?

brian-dellabetta: group size 128 is defined in the `W4A16` scheme, it's a different value than `transform_block_size` though i think research is showing it is good to set them to be equivalent or multiples of one another

src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -262,7 +262,7 @@ def _create_r4_scheme(self) -> TransformScheme:
             randomize=self.randomize,
             requires_grad=self.learnable,
             precision=self.precision,
-            block_size=self.transform_block_size,
+            head_dim=self.transform_block_size,
```
kylesayrs, line 265: I assume this is for backwards compatibility?

kylesayrs (approved):

src/llmcompressor/modifiers/transform/spinquant/base.py
brian-dellabetta: right, to revert the deprecation and just go back to using `head_dim`

PR #1871, opened 2025-09-26
[examples] fix vision_tower/multi_modal_projector regexes

### LINKED ISSUE #1652: Error when loading GPTQ-quantized Mistral-Small model with vLLM
Hi, 

Just tried to quantize Mistral-Small-3.1-24B with GPTQ using your [example script](https://github.com/vllm-project/llm-compressor/blob/main/examples/multimodal_vision/mistral3_example.py).

Everything went smoothly but when I tried to load the model with vLLM, I got the following error message 

> ValueError: There is no module or parameter named 'multi_modal_projector.patch_merger.merging_layer.weight_packed' in Mistral3ForConditionalGeneration

I'm using one of the latest commits on the `main` branch of vLLM (0.9.2rc2.dev269+gf29fd8a7f) and same for llm-compressor (llmcompressor-0.6.1.dev35+g53240c63-py3-none-any.whl).

Thanks!

brian-dellabetta:
SUMMARY:
Resolves #1652

Our multimodal examples all ignore `"re:vision_tower.*"`, but this misses cases where the name is prefixed with something else (e.g. `model.vision_tower`). This PR loosens the regexes to allow for anything to precede `vision_tower` or `multi_modal_projector` and still be caught by the ignore. Layers beginning with `vision_tower`, without a prefix, will still be caught.

Also some formatting fixes, which must not be included on `examples/` as part of ci/cd checks.

TEST PLAN:
Running `llm-compressor/examples/multimodal_vision/mistral3_example.py` on latest main shows we are quantizing layers we don't want to be:
```
2025-09-26T20:02:43.571160+0000 | compress_modules | INFO - Quantizing model.vision_tower.transformer.layers.4.feed_forward.gate_proj using 512 samples
```

After these changes, those don't appear in the logs


kylesayrs (approved):

examples/quantization_w4a4_fp4/llama4_example.py
```diff
@@ -67,8 +69,8 @@ def data_collator(batch):
         "re:.*lm_head",
         "re:.*self_attn",
         "re:.*router",
-        "re:vision_model.*",
-        "re:multi_modal_projector.*",
+        "re:.*vision_model.*.*",
```
fynnsu:
```suggestion
        "re:.*vision_model.*",
```

examples/quantization_w8a8_fp8/llama3.2_vision_example.py
```diff
@@ -17,7 +17,7 @@
 recipe = QuantizationModifier(
     targets="Linear",
     scheme="FP8_DYNAMIC",
-    ignore=["re:.*lm_head", "re:multi_modal_projector.*", "re:vision_model.*"],
+    ignore=["re:.*lm_head", "re:.*multi_modal_projector.*", "re:.*vision_model.*.*"],
```
fynnsu:
```suggestion
    ignore=["re:.*lm_head", "re:.*multi_modal_projector.*", "re:.*vision_model.*"],
```

examples/quantization_w8a8_fp8/llama4_fp8_block_example.py
```diff
@@ -22,8 +22,8 @@
         "re:.*lm_head",
         "re:.*self_attn",
         "re:.*router",
-        "re:vision_model.*",
-        "re:multi_modal_projector.*",
+        "re:.*vision_model.*.*",
```
fynnsu:
```suggestion
        "re:.*vision_model.*",
```

tests/e2e/vLLM/recipes/FP8/recipe_fp8_dynamic.yaml
```diff
@@ -1,7 +1,7 @@
 quant_stage:
   quant_modifiers:
     QuantizationModifier:
-      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:vision_model.*", "re:model.visual.*"]
+      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:.*vision_model.*.*", "re:model.visual.*"]
```
fynnsu:
```suggestion
      ignore: ["lm_head", "re:vision_tower.*", "re:.*multi_modal_projector.*", "re:.*visual.*", "re:.*vision_model.*"]
```

tests/e2e/vLLM/recipes/INT8/recipe_int8_channel_weight_dynamic_per_token.yaml
```diff
@@ -3,7 +3,7 @@ quant_stage:
     SmoothQuantModifier:
       smoothing_strength: 0.8
     GPTQModifier:
-      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:vision_model.*", "re:model.visual.*"]
+      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:.*vision_model.*.*", "re:model.visual.*"]
```
fynnsu:
```suggestion
      ignore: ["lm_head", "re:.*vision_tower.*", "re:.*multi_modal_projector.*", "re:.*visual.*", "re:.*vision_model.*"]
```

tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_weight.yaml
```diff
@@ -1,7 +1,7 @@
 quant_stage:
   quant_modifiers:
     GPTQModifier:
-      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:vision_model.*", "re:model.visual.*"]
+      ignore: ["lm_head", "re:vision_tower.*", "re:multi_modal_projector.*", "re:visual.*", "re:.*vision_model.*.*", "re:model.visual.*"]
```
fynnsu:
```suggestion
      ignore: ["lm_head", "re:.*vision_tower.*", "re:.*multi_modal_projector.*", "re:.*visual.*", "re:.*vision_model.*"]
```

fynnsu (commented):
I added suggestions for some that weren't updated.

I'm also wondering if we should approach this another way, given this seems very common. For example we could:
1. Have this behavior by default (essentially auto-append `.*` to regexes). Slightly risky but probably 95+% of the time what the user wants.
2. Have some kind of flag (maybe on by default) that appends `.*` to the regexes.

Idk, maybe that doesn't make sense but it seems like we will almost always want to have the `.*` there and would help prevent silent failures.

brian-dellabetta:
> I added suggestions for some that weren't updated.
> 
> I'm also wondering if we should approach this another way, given this seems very common. For example we could:
> 
>     1. Have this behavior by default (essentially auto-append `.*` to regexes). Slightly risky but probably 95+% of the time what the user wants.
> 
>     2. Have some kind of flag (maybe on by default) that appends `.*` to the regexes.
> 
> 
> Idk, maybe that doesn't make sense but it seems like we will almost always want to have the `.*` there and would help prevent silent failures.

Thanks @fynnsu for the catches. We may want to in future, but that would be a change in behavior, we'd have to think about that

kylesayrs:
Re: @fynnsu this is philosophical take, but I usually prefer explicit over implicit behavior. This also avoids allows users to be more specific for matching paths, especially when it comes to nested models (which might have repeated path segment names)

Issue #1657, opened 2025-07-18
W4fp8 AWQ


LugerW-A:
W4fp8 achieves good quantization and acceleration effects. Can LLMCompressor implement this quantization method and adapt it to vLLM?

Ref:
https://nvidia.github.io/TensorRT-Model-Optimizer/guides/_choosing_quant_methods.html#:~:text=Ampere%20and%20later.-,INT4%2DFP8%20AWQ%20(W4A8),-High

brian-dellabetta:
Hi @LugerW-A , my understanding from the AWQ work is that W4A8 won't work well for AWQ, because AWQ moves the variance into the activations so weights can be quantized. But maybe A8 is still high enough precision to make this ok? Seems like `modelopt` provides it without much issue, and I don't believe we have to do anything other than loosening [these validation checks](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L191-L204) to allow for 8-bit as well.

Feel free to try it out. I have some other priorities but can possibly take a look next week.

LugerW-A:
@brian-dellabetta Thank you very much. My understanding is that Fp8 offers higher practical precision compared to Int8. W4Fp8 can effectively meet the production requirements for model accuracy, as well as prefill and decode inference performance.

brian-dellabetta:
Would you have time to experiment and validate? I can set up a PR later this week, making sure it can run AWQ with W4A8 (int8 or fp8), and pass it back to you to try? I am unsure of the state of W4A8 support in vllm, but I believe it could at least be activated in `llm-compressor` fairly easily

LugerW-A:
Okay, I will test the performance on both VLM and LLM

LugerW-A:
> I am unsure of the state of W4A8 support in vllm

I hope the process of integrating this algorithm with vLLM isn't too difficult.
Maybe this pr can help https://github.com/vllm-project/vllm/pull/9855

brian-dellabetta:
Hi @LugerW-A , had some time to dig in today. I've created a PR to allow for activation quantization:

* #1682

Can you check out branch `bdellabe/awq-w4a8` and try this out? A quick preliminary check with "meta-llama/Llama-3.2-3B-Instruct" seems promising 👀 :

```
Wikitext PPL (lower is better)
W4AFP8 AWQ+Quant --  13.0959
W4AFP8 Quant -- 13.2527
W4A16 AWQ+Quant --  13.1008
[...]

PR #1873, opened 2025-09-27
Add awq activation fp8 support in  loss compute

Bluedyson:
SUMMARY:
Introduced FP8 quantization error into loss calculation with new config options, validation checks, and tests. 
See https://github.com/vllm-project/llm-compressor/issues/1657#issuecomment-3339402117

TEST PLAN:

```
"Qwen/Qwen3-32B", On our internal dataset, the KL divergence

W4A16: 81
W4AFP8 : 89
TensorRT-Model-Optimizer NVFP4-FP8: 88
```
I'll quickly add general dataset tests result to this PR.

Keep tracing: 
- https://github.com/vllm-project/llm-compressor/issues/1657


brian-dellabetta:
@Bluedyson thanks for this! I will take a look after our release this week.

brian-dellabetta:
Thanks @Bluedyson  for the draft. We decided this can be handled more generically with some built-in utils from compressed tensors, so that any scheme can be used (or even different schemes for different parts of the model). You can track the progress here
* #1961 
Once merged and validated, we can close this PR

PR #1874, opened 2025-09-27
[Qwen3VLMoe] Add linearized definition and FP8 Quantization Example

dsikka:
SUMMARY:
- Updates the MoE layer to use a linearized definition such that we can quantize and run the model in vLLM
- Wraps the gate layer so that it is properly ignored - this is hack for now. We will need to do this properly in ct
- Not adding forward pass for now; will add a forward pass as a follow-up but would like it in for the release to enable FP8 quantization
- Note - requires latest transformers

TEST PLAN:
- Produces `/proving-grounds/engine/hub_cache/Qwen3-VL-235B-A22B-Instruct-FP8_DYNAMIC` which generates coherent generations: 

```python
if __name__ == '__main__':
    import torch

    from vllm import LLM, SamplingParams
    import torch

    prompts = [
        "The Swiss Alps are", 
        "Brad Marchand is",
        "The Toronto Maple Leafs are"
    ]

    # Create a sampling params object for greedy sampling
    sampling_params = SamplingParams(temperature=0.80, top_p=0.95, max_tokens=40, min_tokens=10)
    llm = LLM("/proving-grounds/engine/hub_cache/Qwen3-VL-235B-A22B-Instruct-FP8_DYNAMIC", tensor_parallel_size=2, max_model_len=4096, enforce_eager=True)
    output = llm.generate(prompts, sampling_params)
    for out in output:
        print(out.outputs[0].text)


```

Generations:
```bash
 a true paradise for nature lovers and outdoor enthusiasts. With their snow-capped peaks, lush green valleys, and crystal-clear lakes, the Alps offer a stunning backdrop for a wide range of activities. Whether
 a prominent figure in the NHL, known for his exceptional performance and leadership. He has won the Art Ross Trophy as the NHL's leading scorer, with 110 points (32 goals and
 a professional ice hockey team based in Toronto, Ontario, Canada. They are members of the Atlantic Division in the Eastern Conference of the National Hockey League (NHL). The team was established in 1

```


rahul-tuli (approved):

src/llmcompressor/modeling/prepare.py
```diff
@@ -1,10 +1,21 @@
...
+
+try:
+    from llmcompressor.modeling.qwen3_vl_moe import replace as replace_Qwen3VLMoE
+except ImportError:
+    logger.warning(
+        "Qwen3-VL-MoE support is not available. "
+        "Please ensure that you have the correct version of transformers installed."
+    )
+    replace_Qwen3VLMoE = None
+
```
fynnsu: This will raise the warning for any current user of llmcompressor that doesn't have an up-to-date version of transformers, whether they are using this model or not. I think we should avoid that.

Instead we could move the 
```python
from transformers.models.qwen3_vl_moe.modeling_qwen3_vl_moe import (
    Qwen3VLMoeTextMLP,
)
```
in `src/llmcompressor/modeling/qwen3_vl_moe.py` into the `replace` function or maybe the `SequentialQwen3VLMoeTextExperts.__init__` where it gets used and raise the error there when the function is called. 

Then in this file we can just import the replace function normally and add it to the `replacements` dict. 

That way only users that are trying to use this model will be notified.

brian-dellabetta: Yeah, agree with this, this would be pretty noisy for the logs as is

dsikka: The replacement logic all lives in prepare.py atm - how do you suggest we do this without having to rewrite the way the logic is currently handled?

brian-dellabetta: If you move the try import directly into `replace_Qwen3VLMoE` it should be fine, right?

brian-dellabetta: the warning will only appear if a user then tries to call the function at `replacements["Qwen3VLMoeTextSparseMoeBlock"]`

dsikka:
> Can you add a test in this style? https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/modeling/test_calib_qwen3.py

No because I haven't written the forward pass yet to validate all experts are calibrated.
So far only landing datafree support so that we can enable fp8 support

kylesayrs (approved):
Let's add test in follow up

shanjiaz (approved):

PR #1879, opened 2025-09-29
[tests] remove pile-val-backup dataset from tests

brian-dellabetta:
SUMMARY:
This PR
- [x] Removes `pile-val-dataset` from e2e tests, as it is no longer used in examples and the processing logic was flawed
- [x] Fixes a model validation error introduced in #1772 that was preventing AWQModifier from running one of the validations, causing it to be in an invalid state (`AWQModifier.validate_model_after` was preventing `QuantizationMixin.validate_model_after` from running). With these changes, tests pass and the compressed model generates meaningful responses. It was previously generating all 0s

TEST PLAN:
`CADENCE=nightly TEST_DATA_FILE=tests/e2e/vLLM/configs/w4a16_grouped_quant_sym_awq.yaml pytest -s tests/e2e/vLLM/test_vllm.py` and 
`CADENCE=nightly TEST_DATA_FILE=tests/e2e/vLLM/configs/w4a16_grouped_quant_asym_awq.yaml pytest -s tests/e2e/vLLM/test_vllm.py` 
both pass with output like
```
PROMPT:
The capital of France is
GENERATED TEXT:
 Paris, which is also the country's largest city.

PROMPT:
The president of the US is
GENERATED TEXT:
 named, but the name of the Vice President is not given. In the case

PROMPT:
My name is
GENERATED TEXT:
 Emily and I am from Canada. I have always been fascinated with
```

### DIFF
--- a/tests/e2e/vLLM/recipes/WNA16/recipe_w4a16_awq_sym.yaml
+++ b/tests/e2e/vLLM/recipes/WNA16/recipe_w4a16_awq_sym.yaml
@@ -11,4 +11,3 @@ quant_stage:
             strategy: "group"
             group_size: 128
           targets: ["Linear"]
-          


fynnsu (dismissed):
Looks good!

My only comment is maybe we should change the name of
`tests/e2e/vLLM/recipes/WNA16/recipe_w4a16_group_quant_asym_awq.yaml`
to match
`tests/e2e/vLLM/recipes/WNA16/recipe_w4a16_awq_sym.yaml` if we're just directly using the `W4A16_ASYM` scheme. But also fine as is.

shanjiaz (dismissed):
Looks good to me! Thanks for fixing!

fynnsu (approved):

PR #1884, opened 2025-09-30
Improve stability of flaky perplexity test

fynnsu:
SUMMARY:
`tests/llmcompressor/transformers/compression/test_quantization.py:test_perplexity` is currently flaky, with the test [occasionally failing](https://github.com/neuralmagic/llm-compressor-testing/actions/runs/17994161150/job/51234264145) due to the recorded `avg_ppl` exceeding the test threshold.

Through debugging, it seems like most of the high perplexity samples are samples where most of the target labels are not trained (i.e. set to `-100`). This makes the loss calculation averaging over the remaining tokens more volatile and can result in high perplexity values recorded.

To correct this, I added a check that filters out samples where less than `25%` of the tokens have training labels. This should make the perplexity calculation more consistent, while still testing the model's perplexities are reasonable.

TEST PLAN:
Ran the test locally and all cases passed. Although this is a flaky test, so that doesn't guarantee this has solved the problem.

### DIFF
--- a/tests/llmcompressor/transformers/compression/test_quantization.py
+++ b/tests/llmcompressor/transformers/compression/test_quantization.py
@@ -145,15 +145,19 @@ def test_perplexity(setup_model_and_config):
     dispatch_for_generation(model)
 
     total_ppl = 0.0
-    total_non_nan = 0
-    for idx, sample in enumerate(dataloader):
-        if idx >= config["num_eval"]:
+    total_samples = 0
+    for sample in dataloader:
+        if total_samples >= config["num_eval"]:
             break
         output = model(**tensors_to_device(sample, "cuda:0"))
-        if torch.isnan(output.loss):
+        # -100 in labels indicates that the token is not part of the loss calculation
+        pct_labels_in_sample = (sample['labels'] != -100).to(torch.float).mean().item()
+        if pct_labels_in_sample <= 0.25:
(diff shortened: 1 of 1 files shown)


tests/llmcompressor/transformers/compression/test_quantization.py
```diff
@@ -145,15 +145,19 @@ def test_perplexity(setup_model_and_config):
...
     total_ppl = 0.0
-    total_non_nan = 0
-    for idx, sample in enumerate(dataloader):
-        if idx >= config["num_eval"]:
+    total_samples = 0
+    for sample in dataloader:
+        if total_samples >= config["num_eval"]:
             break
         output = model(**tensors_to_device(sample, "cuda:0"))
-        if torch.isnan(output.loss):
```
fynnsu, line 153: Previously we needed this `torch.isnan` check because when all labels are equal to `-100` there are no labels to compute loss on and the result is `output.loss` is `nan`. Adding this check should handle that case as well.

kylesayrs (approved):
Do we need to adjust tolerances for this new threshold?

fynnsu:
> Do we need to adjust tolerances for this new threshold?

@kylesayrs I don't think so. We're just enforcing that the samples used in the test are longer. I suppose we could maybe reduce the ppl threshold, but I'd worry that would re-introduce flakiness if we lower it too far.

PR #1888, opened 2025-09-30
lmeval threshold p1

brian-dellabetta:
SUMMARY:
There are 2 options to change the threshold for lm_eval tests.

    Add stderr to each tests, as shown in diff for tests/lmeval/configs/fp8_dynamic_per_token.yaml. This would require re-running all tests and grabbing it
    Reduce threshold from default 5% to 3% (or lower?).

I am unsure how we want to handle this, so posing the two options for now

TEST PLAN:


dsikka (commented):
I think we just want 2% threshold for anything that ins't a vl model

brian-dellabetta:
Closing in favor of recovery-based testing, #1750

PR #1893, opened 2025-10-01
Run `ruff format` twice in `make style`

fynnsu:
There was an issue where `ruff check --fix` will fail on too long files before `ruff format` has a chance to fix them. To fix this, we now call `ruff format` before `ruff check --fix` but we also run formatting a second in case the lint fix introduces formatting issues.


shanjiaz (approved):
cool

PR #1898, opened 2025-10-03
[WIP] Add in-memory caching for lm-eval base model results

rahul-tuli:
Evaluate base model once across tests


rahul-tuli:
Not needed anymore in favor of #1900

PR #1903, opened 2025-10-07
[Observers] Refactor to fix gparam bug, support attention, readability

kylesayrs:


src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -100,36 +83,17 @@ def call_observer(
...
         if should_calculate_qparams:
-            updated_scale, updated_zero_point = observer(
-                value, g_idx=g_idx, global_scale=global_scale
-            )
-            # register or update scale & zero_point parameters (supports block shapes)
-            scale_name = f"{base_name}_scale"
-            zp_name = f"{base_name}_zero_point"
-            update_offload_parameter(module, scale_name, updated_scale)
-            update_offload_parameter(module, zp_name, updated_zero_point)
+            scale, zero_point = observer(value)
```
brian-dellabetta, line 94: isn't global_scale, if it exists, needed when calculating scale/zero_point?

src/llmcompressor/observers/base.py
```diff
@@ -25,287 +25,72 @@ class Observer(InternalModule, RegistryMixin):
...
-        :param tensor_id: optional id for tracking separate statistics when different
-            ranges of observed tensors are passed, useful for sharding tensors by
-            group_size or block quantization
-        :param global_scale: optional scale to further scale local quantization scales
-        :return: tuple of scale and zero point derived from the observed tensor
-        """
-        raise NotImplementedError(f"{self.__class__} must implement calculate_qparams")
+        # populate observer kwargs
+        self.args.observer_kwargs = self.args.observer_kwargs or {}
+        self.args.observer_kwargs.update(observer_kwargs)
```
brian-dellabetta, line 54: does it matter that the passed in `observer_kwargs` take precedent over `quantization_args.observer_kwargs`? Will only one or the other ever be set?

src/llmcompressor/observers/min_max.py
```diff
@@ -20,142 +18,39 @@ class MinMaxObserver(Observer):
...
-            "The class name `MovingAverageMinMaxObserver` has been deprecated, please "
-            "initialize with `MinMaxObserver` in the future"
-        )
-    )
-    def __new__(cls, *args, **kwargs):
-        return super().__new__(MinMaxObserver, *args, **kwargs)
+    def _lerp(
+        self, input: torch.Tensor, end: torch.Tensor, weight: float
+    ) -> torch.Tensor:
+        """torch lerp_kernel is not implemeneted for all data types"""
```
brian-dellabetta: how did lerp become the shorthand for linear interpolation? 😅 had to look this up

src/llmcompressor/observers/mse.py
```diff
@@ -110,104 +119,8 @@ def calculate_mse_min_max(
...
-        )
-
-    def reset(self):
-        """
-        Reset the state of the observer, including min and maximum values
-        """
-        super().reset()
-        self.min_val = {}
-        self.max_val = {}
+    def _lerp(
```
brian-dellabetta: you have this duplicated in minmax, can we move to the base class instead?

src/llmcompressor/modifiers/quantization/calibration.py
kylesayrs: Yes, global scale and g_idx are now retrieved from the module attributes directly [here](https://github.com/vllm-project/llm-compressor/pull/1903/files/a0b83b4ad2c6f14cb2fd2cc305331c9ad360b5a5#diff-0ad67c7cc9d87ea2a1aa7be179560a16d51c1bf491e6bce6d880893a0e444e4dR66-R67).

src/llmcompressor/observers/base.py
kylesayrs: Although there are some cases where we manually create observers with args (see [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/gptq/gptq_quantize.py#L97-L101)), we don't actively have any cases where passed args overwrite qargs.

In cases like the one above though, if we did allow users to pass kwargs there, it would make sense for the instance kwargs to override the qargs

src/llmcompressor/observers/min_max.py
kylesayrs: [Wikipedia](https://en.wikipedia.org/wiki/Linear_interpolation#History_and_applications) implies that this term became common in the field of computer graphics first, which is where I first learned it (see [Bezier Curves](https://en.wikipedia.org/wiki/B%C3%A9zier_curve)).

src/llmcompressor/observers/mse.py
kylesayrs: Trying to simplify indirection 🙃. But yeah, although that would be the only function that is shared (without having to create another "get_min_max" interface.

kylesayrs:
@brian-dellabetta Yes, this adds support for FP4 + MSE in a more direct way, and has a test to validate this.

src/llmcompressor/observers/min_max.py
brian-dellabetta: til

src/llmcompressor/observers/mse.py
brian-dellabetta: base classes are intended for housing attributes and methods shared across subclasses. since these both inherit `Observer`, i think it is preferred to repeating

src/llmcompressor/observers/base.py
```diff
@@ -25,287 +25,72 @@ class Observer(InternalModule, RegistryMixin):
...
-            layers in a MOE model. If the input tensor does not have
-            two dimensions, the _num_observed_tokens attribute will be set
-            to None.
+        :param observed: value being observed
+        :return: calibrated global parameter
         """
-        if not isinstance(batch_tensor, Tensor):
-            raise ValueError(f"Expected value to be a tensor, got {type(batch_tensor)}")
+        # avoid updating running min/max for global scales
+        with patch_attr(self, "min_vals", None), patch_attr(self, "max_vals", None):
```
dsikka: Wouldn't this mean that for activation quantization, we are ignoring all previous inputs when calculating the global scale? So the global scale value is only ever based on a single data value at a time?

kylesayrs: You are correct. I was under the impression that global scales should be calculated as absolute min/maxes, but it seems that we've been using the QAT-style moving min/maxes up until now.

I've reverted to the previous behavior of using moving min/maxes, we should explore with using static min/maxes for activation quantization in the future.

src/llmcompressor/observers/mse.py
kylesayrs: Sounds good, I've broken up into the following hierarchy.

```
Observer
StaticObserverBase, MovingAverageObserverBase
StaticMinMaxObserver, MinMaxObserver, StaticMSEObserver, MovingAverageMSEObserver
```

tests/llmcompressor/observers/test_min_max.py
brian-dellabetta, line 1: it doesn't look like we have tests for the new static observers in `test_min_max.py` and `test_mse.py`, if they're easy to add.

kylesayrs: Actually `test_observer_moving_static` does test values for both

src/llmcompressor/observers/min_max.py
```diff
@@ -1,161 +1,39 @@
...
-        self.min_val = {}
-        self.max_val = {}
 
+@Observer.register("minmax")
+class MinMaxObserver(MovingAverageObserverBase):
+    """
+    Implements a quantization observer that calculates scale and zero point based on the
+    minimum and maximum values of the tensor being observed. If averaging_constant is
+    specified, then the scales are updated using a moving average
+    """
```
HDCharles, line 82: i think the docstring is a bit confusing, the min/max are updated using a moving average, not the scales as far as i can tell

src/llmcompressor/observers/mse.py
```diff
@@ -1,71 +1,53 @@
...
+from llmcompressor.observers.static_base import StaticObserverBase
 
-__all__ = ["MovingAverageMSEObserver"]
+__all__ = ["StaticMSEObserver", "MovingAverageMSEObserver"]
 
 
-@Observer.register("mse")
-class MovingAverageMSEObserver(Observer):
+@Observer.register("static_mse")
+class StaticMSEObserver(StaticObserverBase):
```
HDCharles: This doesn't seem to be doing what I would expect it to do. It seems like for MinMaxObservers, Static[] means that it calculates Qparams as though it takes into account all seen data, whereas MovingAverage[] means it does the static algorithm for a single observation and slowly updates the min/max (using a moving average)

But for []MSEObserver, instead it looks like MovingAverageMSEObserver operates as expected but StaticMSEObserver finds the MSE optimized min/max and then if it ever finds a bigger one, overwrites it. The overwriting would seem to pretty much invalidate the whole shrinkage optimization process (Unsure if this actually works well in practice and my intuition is just broken). 

Whereas I would expect StaticMSEObserver to accumulate some high level stats and use that to calculate the best min/max over all seen data. This might be pytorch bias, the closest thing there uses a [histogram](https://github.com/pytorch/pytorch/blob/98b53961b9c917b6594e7322c16586c594e967e7/torch/ao/quantization/observer.py#L1126) which works well but is geared toward integer quantization. 

Maybe keeping track of unshrunkmin/max in the same way as StaticMinMaxObserver but then also calculating a shrinkage factor and gradually updating the error at different shrinkage factors across multiple observations could work?

src/llmcompressor/observers/mse.py
```diff
@@ -1,71 +1,53 @@
...
-        if not reduce_dims:
-            absolute_min_val, absolute_max_val = torch.aminmax(observed)
-        else:
-            absolute_min_val = torch.amin(observed, dim=reduce_dims, keepdims=True)
-            absolute_max_val = torch.amax(observed, dim=reduce_dims, keepdims=True)
+    def get_current_min_max(
+        self, observed: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        Grid search for MSE-optimal min and max values
```
HDCharles: Not really related to this PR but does this method work well in practice? Normally i'd expect this to mostly be about clipping outliers, but if you had something like [-10,...., 10, 100] trying to clip the 100 down to 10 would also result in clipping the -10 to -1. Whereas if you handle them separately you can do much better.

src/llmcompressor/observers/min_max.py
kylesayrs: While this is semantically correct, I agree that it's misleading. I'll change this

src/llmcompressor/observers/mse.py
kylesayrs: I totally agree, this is what I was referencing in our discussion offline. This is a hard case to reference elsewhere, as static activation quantization is still relatively unexplored.

I agree, finding a way to calculate an approximation of min(MSE across all observations) would be best.

src/llmcompressor/observers/mse.py
kylesayrs: We do observe significantly improved accuracy recovery when using the MSE observer. We can sync more later to figure out what's the best way to grid search this (and potentially accelerate it)

dsikka:
Just an FYI -  the Qwen3 VL MoE NVFP4 example has OOO errors on this branch, for both the static_minmax and minmax observer.

kylesayrs:
@dsikka The OOO errors were due to gradients being calculated. Now, gradient calculations are disabled for observers.

The bad calibrations you saw with the MSE observer were due to the MSE observer using the uninitialized global_scale when computing MSE loss. I was able to replicate and fix the issue by recomputing the global scale for each iteration (note that this is only done when computing global_scales, not scales/zps).

src/llmcompressor/observers/mse.py
kylesayrs: I've removed this implementation and replaced it with `memoryless_mse` for now

src/llmcompressor/observers/mse.py
kylesayrs: I've added a comment to explore this idea later

src/llmcompressor/observers/mse.py
```diff
@@ -1,213 +1,236 @@
...
+    :param norm: exponent used when computing the error. norm = 2 approximates MSE
+    :param global_scale: precomputed global scale to use for quantization. Ignored if
+        `optimize_global_scale` is True
+    :param optimize_global_scale: If True, recompute ``global_scale`` from the
+        candidate min/max during each step of the search
+    """
+    min_val = torch.amin(observed, dim=(0, -1))
+    max_val = torch.amax(observed, dim=(0, -1))
+    best_error = torch.full_like(min_val, torch.finfo(min_val.dtype).max)
+    best_min_val = min_val.clone()
```
HDCharles, line 185: does this introduce issues where you have nans in the data? Not sure if this is a concern.

PR #1907, opened 2025-10-08
[MoE] Fix llama4 calibration, bump transformers

kylesayrs:
## Purpose ##
It was noticed that llama4 calibration tests failed after weights were initialized with a normal distribution. This revealed that outputs from the calibrated module were slightly different than those of the original model.

The issue was that llama4 applies router scores **before** experts, whereas most MoEs apply router scores after experts.

## Changes ##
* Fix llama4 model definition to scale by router scores before experts are applied
* Drop support for `transformers<4.53` (current version is 4.57)

## Testing ##
* Updated llama4 test to use normally initialized weights and to check router logits as well as normal outputs.

### DIFF
--- a/src/llmcompressor/modeling/llama4.py
+++ b/src/llmcompressor/modeling/llama4.py
@@ -33,7 +33,7 @@ class SequentialLlama4TextMoe(torch.nn.Module):
 
     def forward(self, hidden_states: torch.Tensor) -> Tuple[torch.Tensor, torch.tensor]:
         hidden_states = hidden_states.reshape(-1, self.hidden_dim)
-        router_scores, router_logits = self.router(hidden_states)  # transformers>=4.53
+        router_scores, router_logits = self.router(hidden_states)  # transformers>=4.54
 
         out = self.shared_expert(hidden_states)
 
--- a/setup.py
+++ b/setup.py
@@ -123,9 +123,9 @@ setup(
         ("tqdm>=4.66.3,<=4.67.1" if BUILD_TYPE == "release" else "tqdm>=4.66.3"),
         ("torch>=2.7.0,<=2.8.0" if BUILD_TYPE == "release" else "torch>=2.7.0"),
         (
-            "transformers>=4.53.0,<=4.56.2"
+            "transformers>=4.54.0,<=4.56.2"
             if BUILD_TYPE == "release"
-            else "transformers>=4.53.0"
+            else "transformers>=4.54.0"
         ),
         ("datasets>=4.0.0,<=4.1.1" if BUILD_TYPE == "release" else "datasets>=4.0.0"),
         (


src/llmcompressor/modeling/llama4.py
```diff
@@ -33,39 +33,28 @@ def __init__(
...
-                router_logits, self.top_k, dim=1
-            )
-            router_logits = router_outputs
-            router_scores = (
-                torch.full_like(router_logits, float("-inf"))
-                .scatter_(1, router_indices, router_top_value)
-                .transpose(0, 1)
-            )
-            router_scores = torch.sigmoid(router_scores.float()).to(hidden_states.dtype)
+        router_scores, router_logits = self.router(hidden_states)  # transformers>=4.53
```
dsikka: We either have to bump our transformers version support in setup.py or make sure we're supporting < 4.53.

tests/llmcompressor/modeling/test_calib_llama4.py
```diff
@@ -68,21 +68,25 @@ def test_calib_llama4_module():
...
+        assert torch.nn.functional.mse_loss(true_out, out) < 1e-10
+        assert torch.nn.functional.mse_loss(true_router_logits, router_logits) < 1e-10
 
     module = SequentialLlama4TextMoe(config, original, calibrate_all_experts=False)
     with calibration_forward_context(module):
-        output = module(sample)[0]
-        assert torch.allclose(true_output, output, atol=1e-6)
+        out, router_logits = module(sample)
+        assert torch.nn.functional.mse_loss(true_out, out) < 1e-10
+        assert torch.nn.functional.mse_loss(true_router_logits, router_logits) < 1e-10
```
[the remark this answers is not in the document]
rahul-tuli: nit: Makes sense?

src/llmcompressor/modeling/llama4.py
rahul-tuli: Why <4.53? as of now we support `transformers>=4.53.0,<=4.56.2` we should try to bump up

kylesayrs: Why would we need to increase the support version when 4.53.0 is already the minimum.

dsikka: Sorry, initial comment was unclear 
Iirc, our lower bound transformers version uses a slightly different llama4 definition. It returns router_scores, not logits. 

I am ok with bumping up the lower bound but we will need to do that in setup.py if we’re not longer supporting the older definition

kylesayrs: My mistake, the original comment was wrong, we require `transformers>=4.54`

dsikka: yeah sorry, I should have said that initially.

PR #1912, opened 2025-10-09
[Bugfix] Refactor QuantizationMixin to use resolved config

### LINKED ISSUE #1906: [Bug]: Serialized QuantizationMixin subclasses fail to load

I introduced a bug in #1772 

### 🐛 Describe the bug

When a recipe with modifiers extending the `QuantizationMixin` class is serialized, it fails to load back up from the yaml. If `config_groups` field is set, the `targets` field is set during validation to contain the full list of targets nested in `config_groups`. When saved, both targets and config_groups are saved, and when re-loaded [this error](https://github.com/vllm-project/llm-compressor/commit/27303c494ba23be6e7e6b2208d981a454c35d797#diff-33303ae48e185b2fbb14dc45c2052805837deb5723248367b9579321c4c4e974R127) is hit, because both targets and config_groups cannot be set. 

### 🛠️ Steps to reproduce

```python
from compressed_tensors.quantization import (
    QuantizationScheme,
    QuantizationStrategy,
    QuantizationType,
    QuantizationArgs,
)
from llmcompressor.modifiers.quantization import (
    QuantizationModifier,
)
from llmcompressor.recipe import Recipe

recipe = Recipe.from_modifiers(
    QuantizationModifier(
        config_groups={
            "group_0": QuantizationScheme(
                targets=["Linear"],
                weights=QuantizationArgs(
                    num_bits=4,
                    type=QuantizationType.INT,
                    group_size=128,
                    symmetric=True,
                    dynamic=False,
                ),
            )
        },
        ignore=["lm_head"],
    )
)

yaml_str = recipe.yaml()

new_recipe = Recipe.create_instance(path_or_modifiers=yaml_str)
```
results in error
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for QuantizationModifier
  Value error, Please specify either `targets` or `config_groups` [type=value_error, input_value={'group': 'default', 'con..., 'ignore': ['lm_head']}, input_type=dict]
```

brian-dellabetta:
SUMMARY:
Fixes #1906 

This refactors QuantizationMixin to not update any pydantic fields during validation. Rather than modifying them in order to make them the source of truth, this adds properties `resolved_config` & `resolved_targets` that all modifiers should instead use as source of truth. These are resolved once, when needed, and not serialized, which should fix the bug in #1906 

TEST PLAN:
Added a `test_resolved_targets` unit test

### DIFF
--- a/src/llmcompressor/core/lifecycle.py
+++ b/src/llmcompressor/core/lifecycle.py
@@ -6,7 +6,7 @@ including initialization, finalization, and event handling.
 """
 
 from dataclasses import dataclass, field
-from typing import Any, List, Optional
+from typing import Any
 
 from loguru import logger
 
@@ -23,11 +23,11 @@ class CompressionLifecycle:
     A class for managing the lifecycle of compression events in the LLM Compressor.
 
     :param state: The current state of the compression process
-    :type state: Optional[State]
+    :type state: State
     :param recipe: The compression recipe
     :type recipe: Recipe
     :param modifiers: The list of stage modifiers
-    :type modifiers: List[StageModifiers]
+    :type modifiers: list[StageModifiers]
     """
 
     state: State = field(default_factory=State)
@@ -37,8 +37,8 @@ class CompressionLifecycle:
     finalized: bool = False
 
     # event order validation
-    _last_event_type: Optional[EventType] = EventType.BATCH_END
-    _event_order: List[EventType] = field(
+    _last_event_type: EventType | None = EventType.BATCH_END
+    _event_order: list[EventType] = field(
         default_factory=lambda: [
             EventType.BATCH_START,
             EventType.LOSS_CALCULATED,
@@ -72,11 +72,11 @@ class CompressionLifecycle:
 
     def initialize(
         self,
-        recipe: Optional[RecipeInput] = None,
-        recipe_stage: Optional[RecipeStageInput] = None,
-        recipe_args: Optional[RecipeArgsInput] = None,
+        recipe: RecipeInput | None = None,
+        recipe_stage: RecipeStageInput | None = None,
+        recipe_args: RecipeArgsInput | None = None,
         **kwargs,
-    ) -> List[Any]:
+    ) -> list[Any]:
         """
         Initialize the compression lifecycle.
 
@@ -114,7 +114,7 @@ class CompressionLifecycle:
 
         return mod_data
 
-    def finalize(self, **kwargs) -> List[Any]:
+    def finalize(self, **kwargs) -> list[Any]:
         """
         Finalize the compression lifecycle.
 
@@ -149,8 +149,8 @@ class CompressionLifecycle:
         return mod_data
 
     def event(
-        self, event_type: EventType, global_step: Optional[int] = 0, **kwargs
-    ) -> List[Any]:
+        self, event_type: EventType, global_step: int | None = 0, **kwargs
+    ) -> list[Any]:
         """
         Handle a compression event.
 
--- a/tests/e2e/vLLM/configs/fp8_block.yaml
+++ b/tests/e2e/vLLM/configs/fp8_block.yaml
@@ -1,0 +1,4 @@
+cadence: "nightly"
+test_type: "regression"
+model: TinyLlama/TinyLlama-1.1B-Chat-v1.0
+scheme: FP8_BLOCK


src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -83,12 +85,14 @@ class QuantizationMixin(HooksMixin):
     """
 
     config_groups: Optional[Dict[str, QuantizationScheme]] = None
-    targets: Union[str, List[str]] = Field(default_factory=list)
+    targets: Union[str, List[str]] = Field(default_factory=lambda: ["Linear"])
```
brian-dellabetta, line 92: this returns to the default behavior. the setting to `["Linear"]` was part of the validation layer that is now removed

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -116,27 +120,33 @@ def validate_scheme(
...
-            for config_group in model.config_groups.values():
+    @property
+    def resolved_targets(self):
+        """
+        List of all resolved targets, i.e. all unique targets listed
+        in resolved quantization config.
+        Use this property instead of the targets field, as targets can
+        also come from config_groups depending on how recipe is configured.
+        """
+        if self._resolved_targets is None:
```
kylesayrs: Caching this value comes with the risk that if the config changes, that won't be reflected in this value

I think getting targets should be infrequent enough that it can be done at runtime? What do you think?

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -116,27 +120,33 @@ def validate_scheme(
...
+            targets = []
+            for config_group in self.resolved_config.config_groups.values():
```
kylesayrs: Consider using a set

```python3
return (
    set(cg.targets for cg in self.resolved_config.config_groups.values())
)
```

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -116,27 +120,33 @@ def validate_scheme(
 
         return value
 
-    @model_validator(mode="after")
-    def validate_model_after(model: "QuantizationMixin") -> "QuantizationMixin":
+    @property
+    def resolved_config(self):
```
kylesayrs:
```suggestion
    def resolved_config(self) -> QuantizationConfig:
```

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -116,27 +120,33 @@ def validate_scheme(
...
+            self._resolved_config = self.resolve_quantization_config()
+        return self._resolved_config
 
-        if len(model.targets) > 0 and model.config_groups is not None:
-            raise ValueError("Please specify either `targets` or `config_groups`")
-
-        if len(model.targets) == 0 and model.config_groups is not None:
-            for config_group in model.config_groups.values():
+    @property
+    def resolved_targets(self):
```
kylesayrs:
```suggestion
    def resolved_targets(self) -> Set[str]:
```

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -116,27 +120,33 @@ def validate_scheme(
...
+        """
+        List of all resolved targets, i.e. all unique targets listed
```
kylesayrs: If you choose to use a set, update here

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -116,27 +120,33 @@ def validate_scheme(
...
+    @property
+    def resolved_config(self):
         """
-        - If targets have not been set, aggregate targets from config_groups
-          into a single unique list
-        - If targets have still not been found, default to targets=["Linear"]
+        Quantization config needs to be resolved just once based on
+        scheme and config_groups inputs.
         """
+        if self._resolved_config is None:
```
kylesayrs, line 132: I think caching makes sense here, since resolving the qconfig is quite intense.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
kylesayrs: Is this going to conflict with plans to use AWQ without quantization targets? You could also potentially set the AWQ subclass default to be None

kylesayrs: Consider adding a comment encouraging use of resolved_targets

src/llmcompressor/modifiers/quantization/quantization/mixin.py
brian-dellabetta: yeah, i wasn't sure about order, but it looks like `match_named_modules` doesn't care about order. will switch to set

src/llmcompressor/modifiers/quantization/quantization/mixin.py
brian-dellabetta: Regarding AWQ, I think we can get by with this default, but allowing the type to be Optional so that a user can set it to None. I wanted to do that with an example in a follow-up PR.
Regarding a comment, I did so in the docstring. If changing the field name to `unresolved_targets` is preferable, I can do that at as well. WDYT? It can still be serialized as targets by setting `alias="targets"` in the pydantic Field inputs

src/llmcompressor/modifiers/quantization/quantization/mixin.py
brian-dellabetta: I can update to do it on the fly. what about `self._resolved_config`? That is also evaluated once, do we expect users to change a modifier's config_groups in the middle of the lifecycle?

brian-dellabetta: updated

src/llmcompressor/modifiers/quantization/quantization/mixin.py
brian-dellabetta: Switched to set, but that for comprehension doesn't work. `set([1,2,3,4])` is acceptable but `set([[1,2],[3,4]])` is not

src/llmcompressor/modifiers/quantization/quantization/mixin.py
brian-dellabetta: Added a comment, so it is clear not to use self.targets in the docstring and comment directly above the field definition. In a follow-up, I will add the option to do AWQ with `targets=None`

PR #1921, opened 2025-10-14
[Tests] Remove Qwen E2E Test Cases

dsikka:
SUMMARY:
- The purpose of the e2e test cases is to act as smoke tests for the different quantization schemes supported by vllm (ensure proper loading)
- The current qwen model being used is dense / a normal decoder model ie not an MoE or VL model. It essentially has the same structure of the tiny llama model also already being tested  
- It is also used for the same schemes already tested by the Tiny Llama model
- As such, it provides no benefit over the Tiny Llama cases and should be removed as redundant in structure / format / vllm loading pathway 
- Will be replaced with MoE cases instead to test a different structure and loading pathway in vllm

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -13,10 +13,16 @@ jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


HDCharles:
looks good

Issue #1922, opened 2025-10-14
[Docs] W8A8-FP8_BLOCK Hardware Compatibility?


jessiewiswjc:
https://github.com/vllm-project/llm-compressor/blob/6c07ef35c39672de2190eb8590045b8d3b0c46c3/docs/guides/compression_schemes.md?plain=1#L14

Should the Hardware Compatibility of W8A8-FP8_BLOCK be the same as that of W8A8-FP8?

dsikka:
Good catch. Do you mind putting up a PR to update this?

jessiewiswjc:
> Good catch. Do you mind putting up a PR to update this?

Ok, i have putted up a [PR](https://github.com/vllm-project/llm-compressor/pull/1923) to update this.

dsikka:
Thank you for putting up a fix!

bash99:
Hello, May I ask anyone had made FP8_BLOCK works on capability ==8.9, for example 4090?

I've troubles to make FP8_BLOCK works on various vllm version, from 0.10.0 - 0.11.0, with Qwen3-30B-A3B model.
FP8-Dynamic always works for me, so what's the advantage of  FP8_BLOCK than FP8_Dynamic?

brian-dellabetta:
Hi @bash99 , I think vllm might not have FP8_BLOCK support for MoE models? You can confirm by trying one of the non-MoE models like https://huggingface.co/RedHatAI/Qwen3-8B-FP8-block

bash99:
> Hi [@bash99](https://github.com/bash99) , I think vllm might not have FP8_BLOCK support for MoE models? You can confirm by trying one of the non-MoE models like https://huggingface.co/RedHatAI/Qwen3-8B-FP8-block

I got it. So the follow-up question is, does Qwen3-8B-FP8-block have any advantage than Qwen3-8B-FP8-Dynamic? 
I still have some Qwen3-32B-FP8-Dynamic deployed, should I change to Qwen3-32B-FP8-Block?

brian-dellabetta:
Hi @bash99 , it really depends on your model and use case. Vllm docs have some information:
- [FP8_DYNAMIC](https://docs.vllm.ai/en/latest/features/quantization/fp8.html#online-dynamic-quantization)
- [W4A16](https://docs.vllm.ai/en/latest/features/quantization/int4.html) which are also provided in the HF hub [here](https://huggingface.co/models?search=redhatai/qwen3)
[...]

PR #1923, opened 2025-10-14
[Docs] Fix the hardware compatibility of W8A8-FP8_BLOCK

jessiewiswjc:
Motivated by https://github.com/vllm-project/llm-compressor/issues/1922#issuecomment-3399908179.


jessiewiswjc:
> Thanks for the catch! Technically [lovelace is also supported](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py#L83), could you update the text to match W8A8-FP8?

If lovelace is added, the compute capability needs to be changed from 9.0 to 8.9. I have done it.

PR #1930, opened 2025-10-14
Fix meta tensor materialization in SequentialPipeline

ronantakizawa:
SUMMARY:
## Problem
Sequential quantization (GPTQ/AWQ) fails on vision-language models with: `RuntimeError: Tensor.item() cannot be called on meta tensors`.

  This occurs in `SequentialPipeline` because PyTorch FX tracing creates meta-tensors
  during graph partitioning for offloaded modules. When the traced code executes
  operations like `.item()`, it fails on these meta tensors.

  ## Solution
  Added meta tensor materialization at two levels:
  1. **Model-level** (helpers.py): Materialize meta tensors in model parameters/buffers
     before subgraph execution
  2. **Input-level** (pipeline.py): Materialize meta tensors in inputs before passing
     to traced code

  This defense-in-depth approach ensures meta-tensors are materialized regardless of
  where they originate.

  ## Changes
  - `src/llmcompressor/pipelines/sequential/helpers.py`:
    - Added `_materialize_model_meta_tensors()` method to Subgraph class
    - Added `_materialized` flag to prevent duplicate materialization
    - Handles integer dtype tensors by converting to buffers instead of parameters
  - `src/llmcompressor/pipelines/sequential/pipeline.py`:
    - Added input materialization before both calibration and propagation passes

TEST PLAN:
Implemented unit test file `tests/llmcompressor/pipelines/test_sequential_vlm.py`
  Successfully quantized HuggingFaceTB/SmolVLM-Instruct (idefics3 architecture):
  - **GPTQ W4A16**: 168 layers quantized, 4.4 GB → 1.97 GB (55.1% reduction)
  - **AWQ W4A16**: 168 layers quantized, 4.4 GB → 1.97 GB (55.1% reduction)

### DIFF
--- a/src/llmcompressor/pipelines/sequential/pipeline.py
+++ b/src/llmcompressor/pipelines/sequential/pipeline.py
@@ -97,14 +97,9 @@ class SequentialPipeline(CalibrationPipeline):
                 if isinstance(obj, torch.Tensor) and obj.is_meta:
                     return torch.zeros_like(obj, device=model_device)
                 elif isinstance(obj, dict):
-                    return {
-                        k: _materialize_meta_tensors(v)
-                        for k, v in obj.items()
-                    }
+                    return {k: _materialize_meta_tensors(v) for k, v in obj.items()}
                 elif isinstance(obj, (list, tuple)):
-                    return type(obj)(
-                        [_materialize_meta_tensors(x) for x in obj]
-                    )
+                    return type(obj)([_materialize_meta_tensors(x) for x in obj])
                 return obj
 
             for subgraph_index, subgraph in enumerate(subgraphs):


kylesayrs:
Hi @ronantakizawa,

Do you have a model example which uses `Tensor.item`?

These kind of `Tensor.item` calls typically shouldn't be present in model definitions anyways. I think in this scenario, it might make more sense to add a patching layer to patch the model definition to avoid the item call.

ronantakizawa:
@kylesayrs here is the project I was working on that caused the meta tensor error. My patch fixed it. 

https://colab.research.google.com/drive/15L5F-p01M5mBv4UDueyw8C57KxkFaqjS?usp=sharing

The .item() call doesn't occur in the model definition itself - it happens inside the traced subgraph code that PyTorch FX generates during sequential pipeline partitioning.

When `SequentialPipeline` partitions the model using PyTorch FX tracing, device-offloaded modules get replaced with meta tensors during symbolic execution. The traced code then executes `rotary_emb()`, which internally calls .item() on tensor dimensions/position indices. When those are meta tensors, it crashes.

brian-dellabetta:
@kylesayrs does this look good to you? A user reported it resolved their issue with Qwen 3 VL and AWQ -- https://github.com/vllm-project/llm-compressor/issues/1939#issuecomment-3410229448

I plan to validate Qwen 3 VL + AWQ use case tomorrow (along with updating the e2e tests from qwen 2.5 vl to qwen 3 vl)

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -79,6 +86,51 @@ def forward(self, *args, **kwargs) -> Dict[str, Any]:
...
+    def _materialize_model_meta_tensors(self, model: Optional[Module]) -> None:
+        """Materialize meta tensors in model parameters and buffers"""
+        if model is None:
+            return
+
+        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+
+        for module in model.modules():
+            # Materialize parameters
+            for name, param in list(module.named_parameters(recurse=False)):
```
brian-dellabetta, line 98: why don't we need to recurse here?

tests/llmcompressor/pipelines/test_sequential_vlm.py
```diff
@@ -0,0 +1,86 @@
...
+    # Verify model can run inference (no meta tensors remain)
+    input_ids = torch.tensor([[1, 2, 3, 4, 5]]).to("cuda:0")
+    with torch.no_grad():
+        output = model_loaded(input_ids)
+        assert output.logits is not None
+        assert not output.logits.is_meta  # Ensure output is not meta tensor
+
+
+@pytest.mark.unit
+def test_meta_tensor_materialization():
```
brian-dellabetta, line 64: this test looks like it might have been vibe coded 😄

tests/llmcompressor/pipelines/test_sequential_vlm.py
```diff
@@ -0,0 +1,86 @@
...
+from transformers import AutoModelForCausalLM
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization.gptq import GPTQModifier
+from tests.testing_utils import requires_gpu
+
+
+@pytest.mark.integration
+@requires_gpu
+def test_sequential_pipeline_with_meta_tensors(tmp_path):
```
brian-dellabetta, line 17: rather than include calls to oneshot, a better unit test here would just be to make sure the model, after materializing, has parameters than aren't meta type?

brian-dellabetta (changes requested):
Thank you for the very impressive contribution! I tried this and it does in fact resolve the issue using Qwen 3 inside the sequential pipeline. 

The code looks really good, but I have a question and a couple change requests for the tests

dsikka:
Are you using latest transformers? @ronantakizawa

brian-dellabetta:
Hi @ronantakizawa , it seems like materialized tensors might be a useful thing in general when tracing subgraphs, but I tried running a sequential pipeline after upgrading my torch/transformers versions, and it works without your PR. Can you try and see if it works on your side without your changes?

ronantakizawa:
@brian-dellabetta it seems like it worked for me as well after upgrading. Will close the issue.

brian-dellabetta:
Thanks @ronantakizawa , appreciate the effort nonetheless

kylesayrs:
I don't like this solution because by the time a tensor is on the meta device, you don't know what is true value should be. Replacing it with zeros will produce gibberish.

HAOHAOXUEXIEN:
> Hi @ronantakizawa , it seems like materialized tensors might be a useful thing in general when tracing subgraphs, but I tried running a sequential pipeline after upgrading my torch/transformers versions, and it works without your PR. Can you try and see if it works on your side without your changes?

what version is your torch/transformers?   thanks versy much

Issue #1927, opened 2025-10-14
[Python 3.10+] Modernize codebase with Python 3.10+ idioms and features


rahul-tuli:
Now that we've dropped Python 3.9 support (#1910), we can modernize our codebase to use Python 3.10+ features for better readability and maintainability. This is a great opportunity for community contributions!

## Background

Python 3.10 introduced several powerful features that make code more readable and Pythonic:
- **PEP 604**: Union types using `|` operator instead of `Union[]`
- **PEP 585**: Built-in collection types for generics (e.g., `list` instead of `List`)
- **PEP 634**: Structural pattern matching with `match/case`

Our codebase has **500+ instances** across **70+ files** that can benefit from these modern idioms.

### 1. Type Hints with `|` Operator

Replace `Union[]` with the more readable `|` syntax.

**Example:**
```python
# Before (Python 3.9 style)
from typing import Union, Optional, List, Dict

def oneshot(
    model: Union[str, PreTrainedModel],
    recipe: Optional[Union[str, List[str]]] = None,
    dataset: Optional[Dict[str, Any]] = None,
) -> PreTrainedModel:
    ...

# After (Python 3.10+ style)
from typing import TYPE_CHECKING, Any

if TYPE_CHECKING:
    from transformers import PreTrainedModel

def oneshot(
    model: str | PreTrainedModel,
    recipe: Optional[str | list[str]] = None,
    dataset: Optional[dict[str, Any]] = None,
) -> PreTrainedModel:
    ...
```

**Benefits:**
- More concise and readable
- Aligns with modern Python style guides
- Reduces import overhead

**Files affected:** ~70 files, ~500 instances

---

### 2. Built-in Generic Types

Use built-in types (`list`, `dict`, `tuple`, `set`) instead of importing from `typing`.

**Example:**
```python
# Before
from typing import List, Dict, Tuple, Optional

def process_batch(
    items: List[str],
    config: Optional[Dict[str, int]] = None,
) -> Tuple[List[str], int]:
    ...

# After
def process_batch(
    items: list[str],
    config: Optional[dict[str, int]] = None,
) -> tuple[list[str], int]:
    ...
```

**Benefits:**
- Cleaner imports
- Standard library usage
- Better IDE support

---

### 3. Structural Pattern Matching

Replace `if/elif` isinstance chains with `match/case` for type-based dispatch.

**Example 1: Type dispatch**
```python
# Before
def log_value(log_tag: str, log_value: Any, epoch: float):
[...]

PR #1934, opened 2025-10-14
Modernization of typehints in llmcompressor/core/lifecycle.py

siddhaka:
This PR refactors llmcompressor/core/lifecycle.py for improved readability, consistency, and adherence to modern Python style conventions.

### DIFF
--- a/src/llmcompressor/core/lifecycle.py
+++ b/src/llmcompressor/core/lifecycle.py
@@ -206,9 +206,9 @@ class CompressionLifecycle:
             if data is not None:
                 mod_data.append(data)
 
-        assert event is not None, (
-            f"Event lifecycle did not return an event for {event_type}"
-        )
+        assert (
+            event is not None
+        ), f"Event lifecycle did not return an event for {event_type}"
 
         return mod_data
 


rahul-tuli:
Hi @siddhaka, thank you for your contribution, it seems like both #1933 and #1934 changed alot more files than scoped in the PR description, I'm guessing it's due to styling changes, However we would like to keep the PRs crisp, could you rebase the diff on latest main, and limit the changes to intended files, for example if you wanted to change llmcompressor/core/lifecycle please limit the changes to just that file. Kindly let us know if we can help in any way, the diff looks good otherwise

siddhaka:
I’ve now cleaned up both PRs to include only the intended updates. Please let me know if any further changes are required.

kylesayrs:
Thanks for the contribution! I have one small nit that I think this worth discussing

From https://docs.python.org/3/library/typing.html#typing.Optional:
> On the other hand, if an explicit value of None is allowed, the use of Optional is appropriate, whether the argument is optional or not. For example:

I understand that functionally, `Optional` and `X | None` are equivalent, but this comment in the docs (and my personal intuition on the word) suggest that `Optional` should refer to when the argument is not required to be provided, whereas `X | None` refers to when the argument must be provided, but could be `None`.

src/llmcompressor/core/lifecycle.py
```diff
@@ -37,8 +37,8 @@ class CompressionLifecycle:
     finalized: bool = False
 
     # event order validation
-    _last_event_type: Optional[EventType] = EventType.BATCH_END
-    _event_order: List[EventType] = field(
+    _last_event_type: EventType | None = EventType.BATCH_END
```
rahul-tuli, line 40: @kylesayrs do you mean places like these should use `Optional[EventType]`? I agree that would read more appropriately @siddhaka could you make this change?

Essentially we'd like to keep things the following:

```python
var: Optional[SomeType] = object<SomeType>  # keep these as is 
```

and,
```python
var : Optional[SomeType] = None # should be replace with var: SomeType | None
```

rahul-tuli:
> Thanks for the contribution! I have one small nit that I think this worth discussing
> 
> From https://docs.python.org/3/library/typing.html#typing.Optional:
> 
> > On the other hand, if an explicit value of None is allowed, the use of Optional is appropriate, whether the argument is optional or not. For example:
> 
> I understand that functionally, `Optional` and `X | None` are equivalent, but this comment in the docs (and my personal intuition on the word) suggest that `Optional` should refer to when the argument is not required to be provided, whereas `X | None` refers to when the argument must be provided, but could be `None`.

@kylesayrs would you be kind enough to update https://github.com/vllm-project/llm-compressor/issues/1927 to make this more explicit

src/llmcompressor/core/lifecycle.py
HDCharles: Does the mypy type checking work correctly for the first one? Because in that case your var is always a Some type and never a None. But iirc the type checking will think it could be both based on that typehint?

HDCharles: e.g. if later you have var.method() it's going to complain that None doesn't have .method, right?

kylesayrs: I agree with @HDCharles. Optional will fail mypy checking if the value cannot be None. Imho optional should only be used if the argument has a default, and the argument’s value could be None

kylesayrs: I believe that in this case (based on how it's being assigned [here](https://github.com/vllm-project/llm-compressor/blob/4cfc0e6217c263cb7450cbf95764de4a1fbffab8/src/llmcompressor/core/lifecycle.py#L229)), `_last_event_type` must be an `EventType` and cannot be `None`

kylesayrs (dismissed):
Looks good! Either `Optional` or `X | None` is fine.

rahul-tuli (dismissed):

src/llmcompressor/core/lifecycle.py
rahul-tuli: Synced offline with the team, we can land this PR as is for now

rahul-tuli (approved):

kylesayrs (approved):

brian-dellabetta (approved):

PR #1942, opened 2025-10-17
AWQ per-channel fix

zhanglei1172:
SUMMARY:
"Fix AWQ modifier to support per-channel quantization configuration"

I tested the metrics of llama3 w4-perchannel-aym under three quant-setting as follows:

rtn
image

Awq
image

awq-rescale
image

It can be seen that as @brian-dellabetta  mentioned, rescale weights need to be performed

TEST PLAN:
"Previously, AWQ modifier defaulted to using group quantization's group_size configuration, which caused errors when per-channel quantization configuration was passed in."

### DIFF
--- a/src/llmcompressor/args/dataset_arguments.py
+++ b/src/llmcompressor/args/dataset_arguments.py
@@ -8,7 +8,7 @@ HuggingFace datasets, custom JSON/CSV files, and DVC-managed datasets.
 """
 
 from dataclasses import dataclass, field
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable
 
 from transformers import DefaultDataCollator
 
@@ -19,7 +19,7 @@ class DVCDatasetArguments:
     Arguments for training using DVC
     """
 
-    dvc_data_repository: Optional[str] = field(
+    dvc_data_repository: str | None = field(
         default=None,
         metadata={"help": "Path to repository used for dvc_dataset_path"},
     )
@@ -31,7 +31,7 @@ class CustomDatasetArguments(DVCDatasetArguments):
     Arguments for training using custom datasets
     """
 
-    dataset_path: Optional[str] = field(
+    dataset_path: str | None = field(
         default=None,
         metadata={
             "help": (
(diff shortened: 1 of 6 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -460,14 +462,16 @@ def _apply_smoothing(self, model: Module) -> None:
...
+                if self._group_size > 0:
+                    weight = weight.view(-1, self._group_size)
+                    # Calculates the relative magnitude of the weights within
+                    # each of the quantization groups, and rescales each group
+                    # individually so that each group has weights on a 0-1 scale.
+                    weight.abs_()
+                    weight.div_(weight.amax(dim=1, keepdim=True) + 1e-6)
+                    # Resizes the rescaled weight matrix back up to
+                    # its original dimensions
+                    weight = weight.view(org_shape)
```
brian-dellabetta, line 475: the abs and div lines would still need to be run, I think you need to run these in both cases
```suggestion
                if self._group_size > 0:
                    weight = weight.view(-1, self._group_size)
                # Calculates the relative magnitude of the weights within
                # each of the quantization groups, and rescales each group
                # individually so that each group has weights on a 0-1 scale.
                weight.abs_()
                weight.div_(weight.amax(dim=1, keepdim=True) + 1e-6)
                if self._group_size >0:
                    # Resizes the rescaled weight matrix back up to
                    # its original dimensions
                    weight = weight.view(org_shape)
```

brian-dellabetta (changes requested):
Thanks for the contribution, but I don't think this will work correctly as is. See change request.

Have you been able to run AWQ with channel-wise quant on this branch and seen improvement over simple round-to-nearest channel-wise quant?

src/llmcompressor/modifiers/awq/base.py
zhanglei1172: I tested the metrics of llama3 w4-perchannel-aym under three quant-setting as follows: 

  rtn
image

  Awq
image

  awq-rescale
image

It can be seen that as you mentioned, rescale weights need to be performed

brian-dellabetta: Awesome! Can you update the PR summary with these results? would be good for posterity

zhanglei1172: done

zhanglei1172:
Hi~ @brian-dellabetta, I try to run make style with my enviroment, but so many files formated, Is this a normal phenomenon?
BTW, I only fixed some format issues in the newly added code and updated it. The original repository code with formatting issues has been retained.
image
image

brian-dellabetta:
Thanks @zhanglei1172 , quality checks are passing now. You might need to upgrade your versions of flake8/ruff/black/isort, they often change formatting style between versions unfortunately

kylesayrs (approved):
Thank you for your contribution! There's an implicit assumption that AWQ only supports group and channel quantization, but that's fine for now

brian-dellabetta:
Yeah, I can add a validation when we further enhance AWQ in the next month

PR #1947, opened 2025-10-18
[AWQ][Qwen3 VL] Add qwen3-vl-30b-a3b-Instruct-example

JartX:
You can follow the thread here:

https://github.com/vllm-project/llm-compressor/issues/1939#issuecomment-3418264625

After confirming with @dsikka that the issue when quantizing the model was a datasheet problem, and that the flickr30k dataset / processing you were trying also results in the same failure (which specifically comes from transformers, suggesting there may be a mismatch in the flickr30k processing and what is needed by the Qwen3 VL Model), I finished a functional model without a datafree pipeline, but a sequential one, as you requested. I proceed to publish, as promised, both the quantization script and the already uploaded model here: https://huggingface.co/jart25/Qwen3-VL-30B-A3B-Instruct-AWQ-8bit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,7 +126,7 @@ venv.bak/
 
 # mkdocs documentation
 /site
-docs/.cache/
+docs/.cache/*
 
 # mypy
 .mypy_cache/
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
 
 ## 🚀 What's New!
 
-Big updates have landed in LLM Compressor! To get a more in-depth look, check out the [deep-dive](https://x.com/RedHat_AI/status/1937865425687093554).
+Big updates have landed in LLM Compressor! To get a more in-depth look, check out the [LLM Compressor overview](https://docs.google.com/presentation/d/1WNkYBKv_CsrYs69lb7bJKjh2dWt8U1HXUw7Gr4Wn3gE/edit?usp=sharing).
 
 Some of the exciting new features include:
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -20,7 +20,7 @@ Review the [LLM Compressor v0.8.0 release notes](https://github.com/vllm-project
 !!! info "Support for multiple modifiers in oneshot compression runs"
     LLM Compressor now supports using multiple modifiers in oneshot compression runs such as applying both AWQ and GPTQ in a single model. 
 
-    Using multiple modifiers is an advanced usage of LLM Compressor and an active area of research. See [Non-uniform Quantization](examples/quantization_non_uniform/) for more detail and example usage.
+    Using multiple modifiers is an advanced usage of LLM Compressor and an active area of research. See [Non-uniform Quantization](/examples/quantization_non_uniform/) for more detail and example usage.
 
 !!! info "Quantization and calibration support for Qwen3 models"
     Quantization and calibration support for Qwen3 Next models has been added to LLM Compressor.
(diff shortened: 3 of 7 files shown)


examples/awq/qwen3-vl-30b-a3b-Instruct-example.py
```diff
@@ -0,0 +1,135 @@
...
+        {
+            "smooth_layer": "re:.*input_layernorm$",
+            "balance_layers": ["re:.*q_proj$", "re:.*k_proj$", "re:.*v_proj$"],
+        },
+        {"smooth_layer": "re:.*v_proj$", "balance_layers": ["re:.*o_proj$"]},
+        {
+            "smooth_layer": "re:.*post_attention_layernorm$",
+            "balance_layers": ["re:.*gate_proj$", "re:.*up_proj$"],
+        },
+        {"smooth_layer": "re:.*up_proj$", "balance_layers": ["re:.*down_proj$"]},
```
brian-dellabetta: these are equivalent to the default mappings, we should be fine to exclude them

examples/awq/qwen3-vl-30b-a3b-Instruct-example.py
```diff
@@ -0,0 +1,135 @@
...
+    ],
+    duo_scaling=True,
+    config_groups={
```
brian-dellabetta, line 82: any reason you chose this quantization config? We typically rely on the preset schemes in examples. This is W8A16, does W4A16 work as well?

brian-dellabetta (changes requested):
Thanks for the contribution! A couple change requests, but this is looking good:

examples/awq/qwen3-vl-30b-a3b-Instruct-example.py
JartX: removed :)

examples/awq/qwen3-vl-30b-a3b-Instruct-example.py
JartX: hi! @brian-dellabetta I use group_size 32 to force the exllamalinearkernel in vllm for my RDNA3, it's the only one that has support for CompressedTensors on my graphics card, I'll give you a link to the vllm PR:

https://github.com/vllm-project/vllm/pull/27187

Many thanks for your answer

JartX: Check it please,I'll update the example after the patch:

Please check it, https://github.com/vllm-project/llm-compressor/issues/1939

brian-dellabetta: It has to be W8A16 group_size 32 in order to work with the RDNA3 chip? Or is this something RocM specific?

JartX: That’s what I asked the vLLM team with ROCm to check — I can only test the ExllamaLinearKernel on my hardware. It also works for W4A16 with group_size 32.

brian-dellabetta: In that case, can you switch to w4a16 (the most common AWQ pathway) and leave a comment above on line 67 that this is using group size 32 instead of the `W4A16` [preset default of 128](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_scheme.py#L219-L228)? I think that should be sufficient

JartX: Do it

examples/awq/qwen3-vl-30b-a3b-Instruct-example.py
```diff
@@ -0,0 +1,125 @@
...
+                "strategy": "group",
+                "block_structure": None,
+                "dynamic": False,
+                "actorder": None,
+                "observer": "mse",
+                "observer_kwargs": {},
+            },
+            "input_activations": None,
+            "output_activations": None,
+            "format": None,
```
brian-dellabetta: we can clear some of these out
```suggestion
                "dynamic": False,
                "actorder": None,
                "observer": "mse",
            }
```

examples/awq/qwen3-vl-30b-a3b-Instruct-example.py
```diff
@@ -0,0 +1,125 @@
...
+            if key != "pixel_values"
+            else torch.tensor(value, dtype=torch.bfloat16).squeeze(0)
+        )
+        for key, value in batch[0].items()
+    }
+
+
+# Configure AWQ quantization with smoothing and balancing
+# NOTE: Using W4A16 quantization with group_size=32 
+# (default W4A16 preset uses 128)
```
brian-dellabetta:
```suggestion
# NOTE: This recipe uses W4A16 quantization with group_size=32
# rather than the default preset with group_size=128
```

brian-dellabetta (changes requested):
Sorry, 2 more nits and then this should be good to land. Thanks!

JartX:
@brian-dellabetta applied :)

Issue #1941, opened 2025-10-17
[Bug]: Qwen3-VL-4B 'Tensor' object has no attribute 'get'. Did you mean: 'det'?


justzhanghong:
### ⚙️ Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
### Environment Information ###
Operating System: `Linux-5.4.0-125-generic-x86_64-with-glibc2.35`
Python Version: `3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0]`
llm-compressor Version: `0.8.1`
compressed-tensors Version: `0.12.3a20251013`
transformers Version: `4.57.1`
torch Version: `2.8.0+cu128`
CUDA Devices: `['NVIDIA A100-SXM4-80GB']`
AMD Devices: `None`
```

</details>

### 🐛 Describe the bug

/usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py:63: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.
  import pynvml  # type: ignore[import]
`torch_dtype` is deprecated! Use `dtype` instead!
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:34<00:00, 17.45s/it]
Generating test split: 3532 examples [00:02, 1485.15 examples/s]
Map: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 12/12 [00:00<00:00, 12.18 examples/s]
2025-10-17T00:56:56.307532-0700 | reset | INFO - Compression lifecycle reset
2025-10-17T00:56:56.315648-0700 | _create_default_logger | INFO - Logging all LLM Compressor modifier-level logs to sparse_logs/17-10-2025_00.56.56.log
2025-10-17T00:56:56.317732-0700 | from_modifiers | INFO - Creating recipe from modifiers
2025-10-17T00:56:56.375590-0700 | initialize | INFO - Compression lifecycle initialized for 1 modifiers
2025-10-17T00:56:56.375803-0700 | IndependentPipeline | INFO - Inferred `SequentialPipeline` for `GPTQModifier`
huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks...
To disable this warning, you can either:
[...]

PR #1953, opened 2025-10-20
[tests] Update lm_eval VL tests to qwen 3

brian-dellabetta:
SUMMARY:
Upgrade the lm_eval vision languge tests from Qwen 2.5 to Qwen 3. After updating to include `apply_chat_template`, the scores closely align with what was achieved with Qwen 2.5

- [x] switch to `neuralmagic/calibration` dataset, based on suggestion [here](https://github.com/vllm-project/llm-compressor/issues/1941#issuecomment-3418252900), to avoid tracing issues related to VL dataset.
- [x] switch to `chartqa` task, to increase number of samples to 500 and reduce variance in accuracy.
- [x] pruned unused datasets (slimorca and llm_compression_calibration)

TEST PLAN:
The 3 lm_eval VL tests were run, and the accuracies were updated
- vl_fp8_dynamic_per_token.yaml runs in ~29m
- vl_int8_w8a8_dynamic_per_token.yaml runs in ~37m
- vl_w4a16_actorder_weight.yaml runs in ~34m

### DIFF
--- a/examples/multimodal_vision/llama4_example.py
+++ b/examples/multimodal_vision/llama4_example.py
@@ -3,18 +3,24 @@ from datasets import load_dataset
 from transformers import Llama4ForConditionalGeneration, Llama4Processor
 
 from llmcompressor import oneshot
-from llmcompressor.modeling import replace_modules_for_calibration
 from llmcompressor.modifiers.quantization import GPTQModifier
 
 # Select model and load it.
 model_id = "meta-llama/Llama-4-Scout-17B-16E-Instruct"
 model = Llama4ForConditionalGeneration.from_pretrained(model_id, torch_dtype="auto")
 processor = Llama4Processor.from_pretrained(model_id)
-# We update `Llama4TextMoe` modules with custom `SequentialLlama4TextMoe`.
-# This change allows compatibility with vllm.
-# To apply your own custom module for experimentation, consider updating
-# `SequentialLlama4TextMoe` under llmcompressor/modeling/llama4.py
-model = replace_modules_for_calibration(model)
+# MoE calibration is now handled automatically by the pipeline.
+# The `SequentialLlama4TextMoe` modules (from `llmcompressor.modeling.llama4`)
+# will be applied during calibration to enable proper
+# expert calibration and vLLM compatibility.
+# These replace the original `Llama4TextMoe` class from
+# `transformers.models.llama4.modeling_llama4`.
+#
+# NOTE: This restructuring is specifically required for vLLM compatibility.
+# To define custom calibration logic, create a new calibration module in
+# modeling/llama4.py that inherits from `MoECalibrationModule`, and register
+# it using the `@register_moe_calibration` decorator with the appropriate
+# module class name (e.g., "Llama4TextMoe").
 
 DATASET_ID = "neuralmagic/calibration"
 NUM_CALIBRATION_SAMPLES = 512
--- a/examples/quantization_w4a4_fp4/README.md
+++ b/examples/quantization_w4a4_fp4/README.md
@@ -84,11 +84,11 @@ We have successfully created an `nvfp4` model!
 
 # Quantizing MoEs
 
-To quantize MoEs, a few additional steps are required. An example quantizing Llama4 can be found under `llama4_example.py`. Here, we replace all `Llama4TextMoe` modules by calling `replace_modules_for_calibration`. This replacement allows us to:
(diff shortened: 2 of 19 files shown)


brian-dellabetta:
> Why not just use mmmu_val instead of the literature task? This gives us around 0.53 for the dense model?

`mmmu_val` is 900 evals total instead of 30. that would add probably ~40 minutes to each lm-eval run, and we run two for each config, so total test time would increase over 3 hours with that change

dsikka:
Why not use mmmu_val

> > Why not just use mmmu_val instead of the literature task? This gives us around 0.53 for the dense model?
> 
> `mmmu_val` is 900 evals total instead of 30. that would add probably ~40 minutes to each lm-eval run, and we run two for each config, so total test time would increase over 3 hours with that change

The 30 datapoints has proven to be very noisy historically. A happy medium might be better but we should also just validate the runtime for batch size of 100

tests/lmeval/test_lmeval.py
```diff
@@ -25,6 +25,7 @@ class LmEvalConfig(BaseModel):
     num_fewshot: int = 5
     limit: int = 1000
     batch_size: int = 100
+    apply_chat_template: bool = False
```
kylesayrs, line 28: Should we not always be applying a chat template?

brian-dellabetta: None of the others have, they are using `add_bos_token=True` instead. This is just to be backwards-compatible. Research does not always do so, the transforms benchmarks aren't using it -- https://github.com/neuralmagic/research/blob/bdellabe/transforms-benchmarks/examples/llm_compress_eval_example.py#L171-L176

kylesayrs: That's really strange to me, but okay to keep for consistency for now.

dsikka (commented):
LGTM. 3 questions:
1. Do we need to keep all the datasets in the testing_utils? Are thre some there that we can remove?
2. Is 100 enough?
3. Did we mention to MLR the variation we see without the chat_template?

brian-dellabetta:
> LGTM. 3 questions:
> 
>     1. Do we need to keep all the datasets in the testing_utils? Are thre some there that we can remove?
> 
>     2. Is 100 enough?
> 
>     3. Did we mention to MLR the variation we see without the chat_template?

@dsikka thanks, see responses below:

1. we could prune the code in `testing_utils`. should i get rid of gsm8k, open-platypus and slim-orca?
2. I can up this as well. I set it to 100 because tests were taking forever, but i think the cpu of the cluster was just under heavy load when i was trying. 500?
3. I can do so Monday

dsikka:
> > LGTM. 3 questions:
> > ```
> > 1. Do we need to keep all the datasets in the testing_utils? Are thre some there that we can remove?
> > 
> > 2. Is 100 enough?
> > 
> > 3. Did we mention to MLR the variation we see without the chat_template?
> > ```
> 
> @dsikka thanks, see responses below:
> 
> 1. we could prune the code in `testing_utils`. should i get rid of gsm8k, open-platypus and slim-orca?
> 2. I can up this as well. I set it to 100 because tests were taking forever, but i think the cpu of the cluster was just under heavy load when i was trying. 500?
> 3. I can do so Monday

Yeah I think we shoud up to 500 and remove any testing dataset that we're not using

tests/lmeval/configs/vl_w4a16_actorder_weight.yaml
```diff
@@ -1,20 +1,25 @@
...
+  batch_size: 100
+  limit: 500
+  # dense model achieves exact_match accuracy of 0.576
+  # dense model achieves relaxed_accuracy of 0.780
+  # dense model achieves anywhere_accuracy of 0.806
   metrics:
-    acc,none: 0.8333
-    acc_stderr,none: 0.0557
\ No newline at end of file
+    exact_match,none: 0.588
```
dsikka, line 23: Isn't this strange for w4a16?

brian-dellabetta: I talked to @kylesayrs about this, he said some improvement can be expected occasionally. There might be some stderr effects here too. This won't be used to validate results, they are tied to be within 5% of the base model.

brian-dellabetta: I will merge in for now and we can inspect weekly results to see how consistently the VL models perform

PR #1954, opened 2025-10-21
[Cache] Check cache before recipe fetch, for offline mode

ralphbean:
Previously, recipe_from_huggingface_model_id() would attempt network
requests to huggingface.co even in offline environments, causing
connection errors during model saving when trying to fetch recipe.yaml.

This change supports offline operation by:
- Checking local HuggingFace cache for recipe.yaml before attempting
  network requests using try_to_load_from_cache()
- Respecting HF_HUB_OFFLINE environment variable to skip network
  requests when explicitly in offline mode


kylesayrs (approved):
Awesome, great job

PR #1961, opened 2025-10-22
[AWQ] Generalize AWQ quantization

### LINKED ISSUE #1657: W4fp8 AWQ

kylesayrs:
### Summary
To allow for arbitrary heterogeneous quantization schemes, this PR switches several helpers from AutoAWQ to the observer and QDQ logic. AWQ no longer constrains that the quantization config needs to have the same settings for group_size, symmetric, and num_bits for each config_group.

Resolves #1657 

Prerequisites:
* https://github.com/vllm-project/compressed-tensors/pull/519

### Test plan
- [x] When running `llm-compressor/examples/awq/llama_example.py` with this (with `duo_scaling="both"`) and logging the best configuration of `(ratio, duo_scaling)`, I see a good mix of Falses and Trues. i.e. a good percentage of best_scales were found with duo_scaling=False and a good percentage were found with `duo_scaling=True`. Generated model output looks good.
- [x] When using `awq_one_shot.py` (pasted below), Wikitext PPL is consistent for w4a16 and w4a16_asym on this branch when compared to main, and better than what was reported in a [previous AWQ PR](https://github.com/vllm-project/llm-compressor/pull/1444#issue-3069873327), but those might have been differently configured. For W4A16_ASYM, the results are both 13.41 for main and this branch. This is what we've been historically using to test regressions.

|Scheme|Wikitext PPL RTN|AWQ main|AWQ this branch|
|-----------:|---------------------:|----------|-----:|
|W4A16|   13.784   |13.477| 13.426|
|W4A16_ASYM | 13.606 | 13.346 | 13.377|

- [x] I see a small regression in recovery when running `CADENCE=weekly TEST_DATA_FILE=~/projects/llm-compressor/tests/lmeval/configs/w4a16_awq_sym.yaml pytest -s ~/projects/llm-compressor/tests/lmeval/test_lmeval.py` on this branch, which causes the test to fail. This persists even when using `pseudo_quantize_tensor` instead of `call_observer`/`forward_quantize`, as shown in [this diff](https://github.com/vllm-project/llm-compressor/compare/kylesayrs/awq-generalize-quant...bdellabe/awq-generalize-quant?expand=1). I get the same result in this diff, so at least that means quantization logic in CT is consistent with AutoAWQ
Output:
```
<main>
2025-11-17T18:26:04.682699+0000 | _validate_recovery | INFO - ✓ exact_match,strict-match                 | Base: 0.7650 | Compressed: 0.7090 | Recovery: 92.68% ↑ | Threshold: ≥92.00%
2025-11-17T18:26:04.682811+0000 | _validate_recovery | INFO - ✓ exact_match,flexible-extract             | Base: 0.7630 | Compressed: 0.7100 | Recovery: 93.05% ↑ | Threshold: ≥93.00%
<this branch>
2025-11-17T17:55:00.648672+0000 | _validate_recovery | ERROR - ✗ exact_match,strict-match                 | Base: 0.7650 | Compressed: 0.6950 | Recovery: 90.85% ↑ | Threshold: ≥92.00%
2025-11-17T17:55:00.648967+0000 | _validate_recovery | ERROR - ✗ exact_match,flexible-extract             | Base: 0.7630 | Compressed: 0.6960 | Recovery: 91.22% ↑ | Threshold: ≥93.00%
```
This is already a pretty high drop in recovery, should we revisit this test?

- [x] Further regression testing against main was done in this [commit](https://github.com/vllm-project/llm-compressor/pull/1961/commits/8b6b0a5ae27084756df5d7e3fd0eca60cbe07b87)  see [run.sh](https://github.com/vllm-project/llm-compressor/blob/8b6b0a5ae27084756df5d7e3fd0eca60cbe07b87/run.sh) as of that commit which was removed in the final PR. Results look reasonable comparing branch and main, some up some down, within margin of error.

  Test Group Quantization (w4a16_awq_sym)
  
  | Branch    | Metric                       | Base   | Compressed | Recovery |
  |-----------|------------------------------|--------|------------|----------|
  | On Branch | exact_match,strict-match     | 0.7620 | 0.7170     | 94.09% ↑ |
  | On Branch | exact_match,flexible-extract | 0.7600 | 0.7130     | 93.82% ↑ |
  | On Main   | exact_match,strict-match     | 0.7620 | 0.7090     | 93.04%  |
  | On Main   | exact_match,flexible-extract | 0.7600 | 0.7060     | 92.89%  |

  Test Tensor Quantization (int8_tensor)

  | Branch    | Metric                       | Base   | Compressed | Recovery |
  |-----------|------------------------------|--------|------------|----------|
  | On Branch | exact_match,strict-match     | 0.7620 | 0.7220     | 94.75% ↓ |
  | On Branch | exact_match,flexible-extract | 0.7600 | 0.7240     | 95.26% ↓ |
  | On Main   | exact_match,strict-match     | 0.7620 | 0.7280     | 95.54%  |
  | On Main   | exact_match,flexible-extract | 0.7600 | 0.7310     | 96.18%  |

  Test Channel Quantization (fp8_dynamic)

  | Branch    | Metric                       | Base   | Compressed | Recovery |
  |-----------|------------------------------|--------|------------|----------|
  | On Branch | exact_match,strict-match     | 0.7650 | 0.7610     | 99.48%  |
  | On Branch | exact_match,flexible-extract | 0.7630 | 0.7580     | 99.34%  |

  Test Block Quantization (fp8_block)

  | Branch    | Metric                       | Base   | Compressed | Recovery  |
  |-----------|------------------------------|--------|------------|-----------|
  | On Branch | exact_match,strict-match     | 0.7650 | 0.7720     | 100.92%  |
  | On Branch | exact_match,flexible-extract | 0.7630 | 0.7690     | 100.79%  |
  
<details>
<summary>awq_oneshot.py script</summary>
```python
import os

os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"

import torch

from llmcompressor import oneshot, active_session
from llmcompressor.utils import dispatch_for_generation
from compressed_tensors.quantization import (
    QuantizationArgs,
    QuantizationScheme,
    QuantizationStrategy,
    QuantizationType,
)


MODEL_ID = "meta-llama/Llama-3.2-3B-Instruct"

SAVE_DIR = MODEL_ID.split("/")[-1] + "-awq-asym"


recipe = [
    AWQModifier(
        ignore=[
            "lm_head",
            "re:.*mlp.gate$",
            "re:.*mlp.shared_expert_gate$",
            "re:visual.*",
        ],
        scheme="W4A16_ASYM",
        duo_scaling="both",
        targets=["Linear"],
        # offload_device=torch.device("cpu"),
    ),
]

# Select calibration dataset.
DATASET_ID = "mit-han-lab/pile-val-backup"
DATASET_SPLIT = "validation"

# Select number of samples. 256 samples is a good place to start.
# Increasing the number of samples can improve accuracy.
NUM_CALIBRATION_SAMPLES = 256
MAX_SEQUENCE_LENGTH = 512


def get_calib_dataset(tokenizer):

    ds = load_dataset(
        DATASET_ID,
        split=f"{DATASET_SPLIT}[:{NUM_CALIBRATION_SAMPLES*10}]",
    )

    def preprocess(example):
        return {"input_ids": tokenizer.encode(example["text"].strip())}

    ds = (
        ds.shuffle(seed=42)
        .map(preprocess, remove_columns=ds.column_names)
        .select(range(NUM_CALIBRATION_SAMPLES))
    )

    return ds


if __name__ == "__main__":
        MODEL_ID, torch_dtype="auto", trust_remote_code=True
    )

    ###
    ### Apply algorithms.
    ###
    oneshot(
        model=model,
        dataset=get_calib_dataset(tokenizer),
        recipe=recipe,
        log_dir=None,
        trust_remote_code_model=True,
    )

    print("\n\n")

    # Save to disk compressed.

    ##
    ### Apply algorithms.
    ##

    ## LM EVAL

    active_session().reset()
    del model
    del tokenizer
    torch.cuda.empty_cache()

    import lm_eval
    from lm_eval.utils import make_table

        model="vllm",
        model_args={
            "pretrained": SAVE_DIR,
            "add_bos_token": True,
            "dtype": "bfloat16",
            "gpu_memory_utilization": 0.7,
            "max_model_len": 4096,
            # "max_num_batched_tokens": 128,
            # "max_num_seqs": 128,
        },
        tasks=["wikitext"],
        batch_size=128,
    )
    print(make_table(results))
```
</details>

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -358,7 +358,7 @@ class AWQModifier(Modifier, QuantizationMixin):
             ):
                 self._smooth_activation_means[smooth_name] = _accumulate_mean(
                     # Assume that first argument is the input
-                    args[0].cpu().abs().detach().squeeze(),
+                    args[0].cpu().abs().detach().flatten(0, -2),
                     self._smooth_activation_means.get(smooth_name, None),
                 )
 
--- a/examples/multimodal_audio/whisper_example.py
+++ b/examples/multimodal_audio/whisper_example.py
@@ -1,6 +1,10 @@
 import torch
 from datasets import load_dataset
-from transformers import WhisperForConditionalGeneration, WhisperProcessor
+from transformers import (
+    WhisperForConditionalGeneration,
+    WhisperProcessor,
+    default_data_collator,
+)
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.quantization import GPTQModifier
@@ -55,9 +59,12 @@ def process(sample):
         return_tensors="pt",
     )
 
-    inputs["input_features"] = inputs["input_features"].to(dtype=model.dtype)
+    # treat labels as calibration prefill
     inputs["decoder_input_ids"] = inputs["labels"]
     del inputs["labels"]
+
+    # strip extra dim added by multimodal processors
+    inputs = {key: value[0] for key, value in inputs.items()}
 
     return inputs
 
@@ -65,10 +72,14 @@ def process(sample):
 ds = ds.map(process, remove_columns=ds.column_names)
 
 
-# Define a oneshot data collator for multimodal inputs.
-def data_collator(batch):
-    assert len(batch) == 1
-    return {key: torch.tensor(value) for key, value in batch[0].items()}
+# Patch: mismatch between processor and model dtype
+def data_collator(features):
+    for feature in features:
+        feature["input_features"] = torch.tensor(
+            feature["input_features"], dtype=model.dtype
+        )
+
+    return default_data_collator(features, return_tensors="pt")
 
 
 # Recipe
--- a/examples/multimodal_vision/gemma3_example.py
+++ b/examples/multimodal_vision/gemma3_example.py
@@ -1,5 +1,4 @@
 import requests
-import torch
 from PIL import Image
 from transformers import AutoProcessor, Gemma3ForConditionalGeneration
 
@@ -13,17 +12,11 @@ model = Gemma3ForConditionalGeneration.from_pretrained(model_id, torch_dtype="au
 processor = AutoProcessor.from_pretrained(model_id, trust_remote_code=True)
 
 # Oneshot arguments
-DATASET_ID = "flickr30k"
-DATASET_SPLIT = {"calibration": "test[:512]"}
+BATCH_SIZE = 4
 NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQUENCE_LENGTH = 2048
-
-
-# Define a oneshot data collator for multimodal inputs.
-def data_collator(batch):
-    assert len(batch) == 1
-    return {key: torch.tensor(value) for key, value in batch[0].items()}
-
+DATASET_ID = "flickr30k"
+DATASET_SPLIT = {"calibration": f"test[:{NUM_CALIBRATION_SAMPLES}]"}
 
 # Recipe
 recipe = [
@@ -41,14 +34,13 @@ recipe = [
 # Perform oneshot
 oneshot(
     model=model,
-    tokenizer=model_id,
+    processor=processor,
     dataset=DATASET_ID,
     splits=DATASET_SPLIT,
     recipe=recipe,
+    batch_size=BATCH_SIZE,
     max_seq_length=MAX_SEQUENCE_LENGTH,
     num_calibration_samples=NUM_CALIBRATION_SAMPLES,
-    trust_remote_code_model=True,
-    data_collator=data_collator,
 )
 
 # Confirm generations of the quantized model look sane.
--- a/examples/multimodal_vision/internvl3_example.py
+++ b/examples/multimodal_vision/internvl3_example.py
@@ -37,20 +37,14 @@ def preprocess_and_tokenize(example):
         return_dict=True,
         return_tensors="pt",
     )
+
+    # remove extra dim added by multimodal processors
+    inputs = {key: value[0] for key, value in inputs.items()}
+
     return inputs
 
 
-ds = ds.map(preprocess_and_tokenize)
-
-
-def data_collator(batch):
-    assert len(batch) == 1
-    item = {key: value for key, value in batch[0].items()}
-    item["attention_mask"] = torch.tensor([item["attention_mask"]])
-    item["input_ids"] = torch.LongTensor([item["input_ids"]])
-
-    return item
-
+ds = ds.map(preprocess_and_tokenize, remove_columns=ds.column_names)
 
 # Recipe
 recipe = GPTQModifier(
@@ -68,7 +62,6 @@ oneshot(
     max_seq_length=MAX_SEQUENCE_LENGTH,
     num_calibration_samples=NUM_CALIBRATION_SAMPLES,
     trust_remote_code_model=True,
-    data_collator=data_collator,
 )
 
 # Save to disk compressed.
--- a/examples/multimodal_vision/llava_example.py
+++ b/examples/multimodal_vision/llava_example.py
@@ -1,5 +1,4 @@
 import requests
-import torch
 from PIL import Image
 from transformers import AutoProcessor, LlavaForConditionalGeneration
 
@@ -17,12 +16,6 @@ DATASET_ID = "flickr30k"
 DATASET_SPLIT = "test"
 NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQUENCE_LENGTH = 2048
-
-
-# Define a oneshot data collator for multimodal inputs.
-def data_collator(batch):
-    assert len(batch) == 1
-    return {key: torch.tensor(value) for key, value in batch[0].items()}
 
 
 # Recipe
@@ -44,7 +37,6 @@ oneshot(
     max_seq_length=MAX_SEQUENCE_LENGTH,
     num_calibration_samples=NUM_CALIBRATION_SAMPLES,
     trust_remote_code_model=True,
-    data_collator=data_collator,
     sequential_targets=["LlamaDecoderLayer"],
 )
 
--- a/examples/multimodal_vision/mistral3_example.py
+++ b/examples/multimodal_vision/mistral3_example.py
@@ -4,7 +4,11 @@ import os
 import requests
 import torch
 from PIL import Image
-from transformers import AutoProcessor, Mistral3ForConditionalGeneration
+from transformers import (
+    AutoProcessor,
+    Mistral3ForConditionalGeneration,
+    default_data_collator,
+)
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.quantization import GPTQModifier
@@ -27,17 +31,13 @@ NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQUENCE_LENGTH = 2048
 
 
-# Define a oneshot data collator for multimodal inputs.
-def data_collator(batch):
-    assert len(batch) == 1
-    return {
-        key: (
-            torch.tensor(value)
-            if key != "pixel_values"
-            else torch.tensor(value, dtype=model.dtype)
+# Patch: mismatch between processor and model dtype
+def data_collator(features):
+    for feature in features:
+        feature["pixel_values"] = torch.tensor(
+            feature["pixel_values"], dtype=model.dtype
         )
-        for key, value in batch[0].items()
-    }
+    return default_data_collator(features, return_tensors="pt")
 
 
 # Recipe
--- a/examples/multimodal_vision/mllama_example.py
+++ b/examples/multimodal_vision/mllama_example.py
@@ -1,5 +1,4 @@
 import requests
-import torch
 from PIL import Image
 from transformers import AutoProcessor, MllamaForConditionalGeneration
 
@@ -17,12 +16,6 @@ DATASET_ID = "flickr30k"
 DATASET_SPLIT = "test"
 NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQUENCE_LENGTH = 2048
-
-
-# Define a oneshot data collator for multimodal inputs.
-def data_collator(batch):
-    assert len(batch) == 1
-    return {key: torch.tensor(value) for key, value in batch[0].items()}
 
 
 # Recipe
@@ -44,7 +37,6 @@ oneshot(
     max_seq_length=MAX_SEQUENCE_LENGTH,
     num_calibration_samples=NUM_CALIBRATION_SAMPLES,
     trust_remote_code_model=True,
-    data_collator=data_collator,
     sequential_targets=["MllamaSelfAttentionDecoderLayer"],
 )
 
--- a/examples/multimodal_vision/pixtral_example.py
+++ b/examples/multimodal_vision/pixtral_example.py
@@ -1,7 +1,11 @@
 import requests
 import torch
 from PIL import Image
-from transformers import AutoProcessor, LlavaForConditionalGeneration
+from transformers import (
+    AutoProcessor,
+    LlavaForConditionalGeneration,
+    default_data_collator,
+)
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.quantization import GPTQModifier
@@ -19,16 +23,13 @@ NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQUENCE_LENGTH = 2048
 
 
-# Define a oneshot data collator for multimodal inputs.
-# NOTE: for transformers<4.48.0, please squeeze the first dimension of `pixel_values`
-# by appending `[0]` to the end of line 32
-def data_collator(batch):
-    assert len(batch) == 1
-    return {
-        "input_ids": torch.LongTensor(batch[0]["input_ids"]),
-        "attention_mask": torch.tensor(batch[0]["attention_mask"]),
-        "pixel_values": torch.tensor(batch[0]["pixel_values"]),
-    }
+# Patch: mismatch between processor and model dtype
+def data_collator(features):
+    for feature in features:
+        feature["pixel_values"] = torch.tensor(
+            feature["pixel_values"], dtype=model.dtype
+        )
+    return default_data_collator(features, return_tensors="pt")
 
 
 # Recipe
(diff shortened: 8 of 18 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -696,47 +600,32 @@ def _assert_all_activations_consumed(self):
...
+    def _compute_duo_scaling_means(self, mapping: ResolvedMapping):
+        balance_layers = mapping.balance_layers
+
+        # TODO: validate that all layers have the same quantization_scheme.weights
+        # either generalize this to compute means with different strategy shapes
+        # or throw error if strategy is not channel/group
+
+        # [STEP 1]: Compute per-channel mean of normalised weights
+        # All layer weights are concatted together
+        weight = torch.cat([bl.weight for bl in balance_layers], dim=0)
```
kylesayrs: Rather than concatting all these and using a ton of memory, we should just compute the mean/sum of each weight individually, then mean those.

brian-dellabetta: Updated

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -622,52 +539,76 @@ def _compute_best_scale(
...
+            "weight_observer",
+            [
+                Observer.load_from_registry(
+                    "memoryless_minmax",
+                    base_name="weight",
+                    args=balance_layer.quantization_scheme.weights,
+                    module=balance_layer,
                 )
+                for balance_layer in mapping.balance_layers
+                if hasattr(balance_layer, "quantization_scheme")
```
kylesayrs: Will this function error if balance layer doesn't have a quantization scheme?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -622,52 +539,76 @@ def _compute_best_scale(
...
+                        min=1e-4
+                    )
+                else:
+                    scales = x_mean.pow(ratio).clamp(min=1e-4).view(-1)
+                scales = scales / (scales.max() * scales.min()).sqrt()
+                _scalesview = scales.view(1, -1).to(device)
+
+                # avoid scaling values that overflow
+                scales[torch.isinf(scales)] = 1
+                scales[torch.isnan(scales)] = 1
```
kylesayrs, line 588: Is smoothing scale overflow newly introduced? Is overflow always expected to happen?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -720,48 +657,61 @@ def _assert_all_activations_consumed(self):
...
-
-    assert torch.isnan(scales).sum() == 0
-    assert torch.isnan(w).sum() == 0
-
-    scales = scales.view(org_w_shape[0], -1)
-    w = w.reshape(org_w_shape)
-
-    return w, scales, zeros
+        for layer in layers:
+            if not hasattr(layer, "weight"):
```
kylesayrs, line 691: I feel like a lot of this algorithm assumes that layers have weights? Should we be silently skipping here?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -720,48 +657,61 @@ def _assert_all_activations_consumed(self):
...
+            weight = layer.weight
+            org_shape = weight.shape
+
+            # If group-wise, calculate abs max based on group
+            # abs max, rather than channel
+            if (group_size := _infer_group_size(layer)) > 0:
+                weight = weight.view(-1, group_size)
+
+            weight.abs_()
+            weight.div_(weight.amax(dim=1, keepdim=True) + 1e-6)
```
kylesayrs: Infinity norm?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -720,48 +657,61 @@ def _assert_all_activations_consumed(self):
...
+
+            # Gets the average rescaled magnitude for each output channel
+            weight_total_count += weight.size(0)
+            weight_sum = weight.sum(0, dtype=torch.float64)
+            if weight_total_sum is None:
+                weight_total_sum = weight_sum
+            else:
+                weight_total_sum += weight_sum
+
+        return weight_total_sum / weight_total_count
```
kylesayrs: I think this is just `weight.mean(dim=0)`

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -720,48 +657,61 @@ def _assert_all_activations_consumed(self):
...
-
-    return w, scales, zeros
+        for layer in layers:
+            if not hasattr(layer, "weight"):
+                continue
+
+            weight = layer.weight
+            org_shape = weight.shape
+
+            # If group-wise, calculate abs max based on group
```
kylesayrs: Will this generalize to block or tensor quant? I think I personally need to get a better understanding of what "group normalization" is supposed to do and how it applies to other quant strategies.

Writing this function with torch native vectorized ops might help, I think it might be reducible to frobenius norm and mean/sum ops, but that doesn't have to be done now.

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: To date, all the balance_layers are linear layers. I could just change the AWQMapping type to linear and avoid this checking logic. WDYT?

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: I updated this to skip balance_layers that don't have a quant schema, if that does arise. So this should be robust enough to still work when someone wants to update a mapping like input_layernorm -> q/k/v proj, but does NOT want to quantize all q/k/v proj layers.

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: for tensor, i don't think using channel-wise means makes sense. We could validate that if need be, i.e. throw validation error if strategy==TENSOR and duo_scaling != False.

For block, we'd have to update this logic, yeah. I'll leave as a todo for now

src/llmcompressor/modifiers/awq/base.py
kylesayrs: Anything's good with me, I prefer loud errors if assumptions are violated in this case.

src/llmcompressor/modifiers/awq/base.py
kylesayrs: Nice, that seems like the most robust solution.

HDCharles (approved):

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -720,48 +664,64 @@ def _assert_all_activations_consumed(self):
...
+            # Reshape back to original dimensions
+            weight = weight.view(org_shape)
+
+            # Gets the average rescaled magnitude for each output channel
+            weight_total_count += weight.size(0)
+            weight_sum = weight.sum(0, dtype=torch.float64)
+            if weight_total_sum is None:
+                weight_total_sum = weight_sum
+            else:
+                weight_total_sum += weight_sum
```
fynnsu: This seems a little strange to me. Can't we just initialize `weight_total_sum = 0.0`?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -622,52 +539,81 @@ def _compute_best_scale(
...
-            if loss < best_error:
-                best_error = loss
-                best_ratio = ratio
-                best_scales = scales.clone()
+                history.append(loss)
+                if loss < best_error:
+                    best_error = loss
+                    best_duo_scaling = use_duo_scaling
+                    best_ratio = ratio
+                    best_scales = scales.clone()
```
fynnsu: It seems like `history` is currently just used for debugging when no `best_ratio` is found. I wonder if we could instead be recording saving the hyperparameter states and losses and printing them everytime (when in DEBUG logging mode.

e.g.

```python
for grid_idx, use_duo_scaling in product(range(n_grid), duo_scalings):
    ratio = grid_idx / n_grid
    ...
    history.append({"ratio": ratio, "duo_scaling": use_duo_scaling, "error": loss})
...
logger.debug(history)
```

This might be useful in the future as we look into improving the hyperparameter search / to get a sense of what parameters are most often selected. I think including the ratio/duo_scaling in some way is important now that we've switched from a simple linear search to a grid search, so that's easy to tell which arguments are being set.

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -622,52 +539,81 @@ def _compute_best_scale(
...
+                        min=1e-4
+                    )
+                else:
+                    scales = x_mean.pow(ratio).clamp(min=1e-4).view(-1)
+                scales = scales / (scales.max() * scales.min()).sqrt()
+                _scalesview = scales.view(1, -1).to(device)
+
+                # avoid scaling values that overflow
+                scales[torch.isinf(scales)] = 1
+                scales[torch.isnan(scales)] = 1
```
HDCharles, line 588: should the overflow handling happen before the 

```scales = scales / (scales.max() * scales.min()).sqrt()``` line?

I guess this line is the same as before

File: src/llmcompressor/modifiers/awq/base.py [hunk 11]
```diff
@@ -230,6 +165,24 @@ def on_initialize(self, state: State, **kwargs) -> bool:
         if QuantizationMixin.has_config(self):
             QuantizationMixin.initialize_quantization(self, state.model)
 
+        # Validate that duo_scaling is only used with per-channel quantization
+        if self.duo_scaling != False:
```
HDCharles: @kylesayrs added check for duo_scaling + per Tensor strategy

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -622,52 +559,81 @@ def _compute_best_scale(
...
+
+                # Q(W * s)
+                for balance_layer in balance_layers_to_patch:
+                    if not hasattr(balance_layer, "quantization_scheme") or not hasattr(
+                        balance_layer.quantization_scheme, "weights"
+                    ):
+                        continue
+
+                    balance_layer.weight.mul_(_scalesview)
+                    call_observer(balance_layer, "weight", balance_layer.weight)
```
kylesayrs: If you pass `should_calculate_gparam=scheme.strategy=="tensor_group"`, then I think this should be able to support nvfp4

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -682,6 +648,8 @@ def _compute_best_scale(
             torch.isnan(best_scales).sum() == 0
         ), f"Nan found in scales: {best_scales}"
 
+        print("BEST CONFIGURATION", best_duo_scaling, best_ratio)
```
kylesayrs: del

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -720,49 +684,110 @@ def _assert_all_activations_consumed(self):
...
+            raise ValueError(
+                f"duo_scaling must be True, False, or 'both', got {v!r}"
+            )
+        return v
 
-    scales = scales.view(org_w_shape[0], -1)
-    w = w.reshape(org_w_shape)
 
-    return w, scales, zeros
+def _orient_weight(weight: torch.Tensor, qscheme) -> torch.Tensor:
```
kylesayrs: Can we reuse `flatten_for_calibration`?

src/llmcompressor/modifiers/awq/base.py
HDCharles: will add that

src/llmcompressor/modifiers/awq/base.py
HDCharles: will do this in another PR

src/llmcompressor/modifiers/awq/base.py
HDCharles: actually added todo will do in another PR

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -127,17 +138,12 @@ class AWQModifier(Modifier, QuantizationMixin):
...
 
     # User-provided vars (in addition to QuantizationMixin args)
-    sequential_targets: str | list[str] | None = None
-    mappings: list[AWQMapping] | None = None
-    offload_device: torch.device | None = None
-    duo_scaling: bool | Literal["both"] = True
+    sequential_targets: Union[str, List[str], None] = None
+    mappings: Optional[List[AWQMapping]] = None
+    offload_device: Optional[torch.device] = None
+    duo_scaling: str | bool = True
```
kylesayrs: `Literal["both"]` doesn't work anymore?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -622,52 +558,88 @@ def _compute_best_scale(
...
+                        min=1e-4
+                    )
+                else:
+                    scales = x_mean.pow(ratio).clamp(min=1e-4).view(-1)
+                scales = scales / (scales.max() * scales.min()).sqrt()
+                _scalesview = scales.view(1, -1).to(device)
+
+                # avoid scaling values that overflow
+                scales[torch.isinf(scales)] = 1
+                scales[torch.isnan(scales)] = 1
```
kylesayrs, line 588: Definitely annoying that we still have to have this line.

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -690,20 +662,16 @@ def _compute_loss(
...
-            batch_loss = (
-                (fp16_batch.to(device) - int_w_batch.to(device))
-                .view(-1)
-                .float()
-                .pow(2)
-                .sum()
-                .item()
-            )
+            batch_loss = torch.nn.functional.mse_loss(
+                fp16_batch.to(device), int_w_batch.to(device)
```
kylesayrs: This algorithm has a lot of device movement

src/llmcompressor/modifiers/awq/base.py
HDCharles: (added todo) going to move this and simplified logic to a new PR

src/llmcompressor/modifiers/awq/base.py
HDCharles: will do in a different PR

src/llmcompressor/modifiers/awq/base.py
HDCharles: all the type hints got mangled in the rebase i think, fixing

src/llmcompressor/modifiers/awq/base.py
HDCharles: oh good catch, remove those .to(device) calls improved speed on the llama example from 10 minutes to 9 minutes on an A100

src/llmcompressor/modifiers/awq/base.py
HDCharles: i can do more testing on this section and see if there's a way around in the followup PR

HDCharles (approved):

PR #1963, opened 2025-10-24
Fixing untie to be used only as needed and automatic

HDCharles:
when models have shared in/out embedding, we had to call the untie function to
separate them before modifying those layers. This caused increased memory and was applied at all times
regardless of whether those layers were targeted for modification.

change: automatically detect when transform or mixin needs to untie
shared embeddings.

This also adds try except to the untieing code so that if it is invoked
on a model that can't be untied, it gives a warning rather than erroring

new tests are added to test this functionality, old tests are modified
to use the automatic untieing

the new tests were initially written using claude-code, I then rewrote them

TEST PLAN:

pytest tests/llmcompressor/modifiers/quantization/test_handling_shared_embeddings.py

### DIFF
--- a/src/llmcompressor/modifiers/transform/spinquant/base.py
+++ b/src/llmcompressor/modifiers/transform/spinquant/base.py
@@ -147,6 +147,7 @@ class SpinQuantModifier(Modifier, use_enum_values=True):
 
         return True
 
+    @torch.no_grad()
     def on_start(self, state: State, event: Event, **kwargs):
         self.started_ = True
 


src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -317,3 +321,38 @@ def _initialize_hooks(self, model: torch.nn.Module) -> Set[RemovableHandle]:
...
+            untie_word_embeddings,
+        )
+
+        try:
+            input_embeddings, output_embeddings = (
+                model.get_input_embeddings(),
+                model.get_output_embeddings(),
+            )
+        except NotImplementedError:
+            logger.warning(
```
kylesayrs: Can we only throw a warning if embeddings are being targeted? That way we don't need to warn the user unless it's relevant to their flow.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -317,3 +321,38 @@ def _initialize_hooks(self, model: torch.nn.Module) -> Set[RemovableHandle]:
...
+                f"{model.__class__} doesn't have get_input_embeddings and "
+                + "get_output_embeddings implmented.\n This can cause problems when "
+                + "trying to quantize layers with shared weights\n{e}"
```
kylesayrs: Is the {e} a typo?

```suggestion
               "get_output_embeddings implmented.\n This can cause problems when "
               "trying to quantize layers with shared weights\n{e}"
```

src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -121,6 +121,9 @@ def validate_rotations(cls, value):
         return value
 
     def on_initialize(self, state: State, **kwargs) -> bool:
+        from llmcompressor.transformers.compression.compressed_tensors_utils import (
+            untie_word_embeddings,
```
kylesayrs: I'm surprised that you need to dynamically import this. Maybe we can move this CT in the future.

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -126,8 +126,12 @@ def untie_word_embeddings(model: PreTrainedModel):
 
     :param model: model to fix
     """
-    input_embed = model.get_input_embeddings()
-    output_embed = model.get_output_embeddings()
+    try:
+        input_embed = model.get_input_embeddings()
+        output_embed = model.get_output_embeddings()
```
kylesayrs, line 132:
```suggestion
        input_embed = model.get_input_embeddings()
        output_embed = model.get_output_embeddings()
        assert input_embed is not None and output_embed is not None
```

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -126,8 +126,12 @@ def untie_word_embeddings(model: PreTrainedModel):
...
+    except NotImplementedError as e:
+        logger.warning(f"cannot untie model of type {model.__class__} which doesn't have get_input_embeddings and get_output_embeddings implmented\n{e}")
```
kylesayrs: Now that this is guarded, I think this should be a ValueError, not a warning. If we can't untie, the model output will be super weird.

tests/llmcompressor/modifiers/quantization/test_base.py
```diff
@@ -210,4 +210,4 @@ def test_resolved_targets(
             targets=targets, scheme=scheme, config_groups=config_groups
         )
 
-        assert modifier.resolved_targets == resolved_targets
+        assert modifier.resolved_targets == resolved_targets
```
kylesayrs: nit: keep line ending

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -317,3 +321,38 @@ def _initialize_hooks(self, model: torch.nn.Module) -> Set[RemovableHandle]:
...
+        from llmcompressor.transformers.compression.compressed_tensors_utils import (
+            untie_word_embeddings,
+        )
+
+        try:
+            input_embeddings, output_embeddings = (
+                model.get_input_embeddings(),
+                model.get_output_embeddings(),
+            )
+        except NotImplementedError:
```
kylesayrs: Isn't this warning already thrown by untie_word_embeddings?

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
```diff
@@ -0,0 +1,190 @@
...
+    handle shared input/output embeddings when they are targeted for quantization.
+    """
+    model_id = "meta-llama/Llama-3.2-1B-Instruct"
+    scheme = "W8A8"
+    targets = ["Linear"]  # This targets lm_head (Linear layer)
+
+    # Test 1: Apply quantization WITHOUT manually untieing first
+    # (relies on automatic untieing in start_calibration)
+    model_auto = AutoModelForCausalLM.from_pretrained(
+        model_id, device_map="cuda", torch_dtype=torch.float32
```
kylesayrs:
```suggestion
        model_id, device_map="cuda"
```

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+
+    # Verify embeddings are initially tied
+    input_embed = model_auto.get_input_embeddings()
+    output_embed = model_auto.get_output_embeddings()
+    initial_tied = input_embed.weight is output_embed.weight
```
kylesayrs: Do you mean to assert that they're initialized tied here?

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+
+    modifier.on_initialize(state)
+    modifier.on_start(state, None)
+
+    # Verify embeddings are still tied (since they weren't targeted)
+    still_tied = input_embed.weight is output_embed.weight
+
+    # If they were initially tied and not targeted, they should remain tied
+    if initial_tied:
```
kylesayrs: Do you ever expect this to be False?

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
```diff
@@ -0,0 +1,190 @@
...
+    model_auto = AutoModelForCausalLM.from_pretrained(
+        model_id, device_map="cuda", torch_dtype=torch.float32
+    )
+
+    # Verify embeddings are initially tied
+    input_embed = model_auto.get_input_embeddings()
+    output_embed = model_auto.get_output_embeddings()
+    initial_tied = input_embed.weight is output_embed.weight
+
+    state_auto = State(model=model_auto)
```
kylesayrs: Just curious, why the `_auto` suffix?

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
```diff
@@ -0,0 +1,190 @@
...
+        output_auto = model_auto(**input_data)
+
+    # Verify that embeddings were untied
+    auto_untied = input_embed.weight is not output_embed.weight
+
+    # Clean up
+    del model_auto, state_auto, modifier_auto
+    torch.cuda.empty_cache()
+
+    # Test 2: Apply quantization WITH manual untieing first
```
kylesayrs: Maybe make this a separate function?

kylesayrs: This way, you don't have to cleanup either

tests/llmcompressor/modifiers/transform/test_correctness.py
```diff
@@ -36,7 +36,7 @@ def test_apply_correctness(
     model = AutoModelForCausalLM.from_pretrained(
         "meta-llama/Llama-3.2-1B-Instruct", device_map="cuda", torch_dtype=model_dtype
     )
-    untie_word_embeddings(model)
+    # untie_word_embeddings(model)
```
kylesayrs: Delete

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
```diff
@@ -0,0 +1,190 @@
...
+    with torch.no_grad():
+        output_manual = model_manual(**input_data)
+
+    # Verify that automatic and manual untieing produce identical results
+    # The outputs should be identical (or very close due to numerical precision)
+    mse = torch.nn.MSELoss()(output_auto.logits, output_manual.logits)
+
+    # For quantization, we expect very low MSE between the two approaches
+    # (they should be essentially identical)
+    assert mse < 1e-6, (
```
kylesayrs: If you feel comfortable, consider removing the MSE tests and only verifying that the embeddings are now untied / can be modified independently.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
HDCharles: So e.g.

1) check if we can get embeddings, if not prepare a warning
2) go through normal module generation, if we see a module of the embedding class, then we give the warning?

or is there a different way to check if the embeddings are targeted?

src/llmcompressor/modifiers/quantization/quantization/mixin.py
HDCharles: just trying to pass the error up the stack in case its actually important/unsuspected.

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
HDCharles: i think its fine if they are None, right? it won't change any of the logic afaict

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
HDCharles: if we do that then we'll be back to where we started as far as erroring out when its not implemented but irrelevant.

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
HDCharles: edit: actually the MSE test is probably the better test to make sure that nothing is broken when doing these things.

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
HDCharles: the whole test was written by claude and i'm in the process of revising it

src/llmcompressor/modifiers/quantization/quantization/mixin.py
kylesayrs: Yeah makes sense. Doesn't seem like `e` is actually used though.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
kylesayrs: Ah totally, makes sense. We don't know if it's targeting embeddings without getting the embeddings 🙃. As you have it makes sense

src/llmcompressor/modifiers/quantization/quantization/mixin.py
HDCharles: oh i see

src/llmcompressor/modifiers/quantization/quantization/mixin.py
HDCharles: yes, if we add a check here, that won't trigger when using mixin, it would only trigger when doing spinquant

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -317,3 +321,38 @@ def _initialize_hooks(self, model: torch.nn.Module) -> Set[RemovableHandle]:
...
+            logger.warning(
+                f"{model.__class__} doesn't have get_input_embeddings and "
+                + "get_output_embeddings implmented.\n This can cause problems when "
+                + "trying to quantize layers with shared weights\n{e}"
+            )
```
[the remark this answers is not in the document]
brian-dellabetta: Yeah i think you need the f prefix here

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -317,3 +321,38 @@ def _initialize_hooks(self, model: torch.nn.Module) -> Set[RemovableHandle]:
...
+        if input_embeddings is None:
+            return
+        if input_embeddings.weight is not output_embeddings.weight: # if not shared
+            return
+
+        # if shared, check if either is targeted
+        matched_modules = list(
+            match_named_modules(model, self.resolved_targets, self.ignore)
+        )
+        if input_embeddings in matched_modules or output_embeddings in matched_modules:
```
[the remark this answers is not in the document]
brian-dellabetta: Your current changes should work, I think gemini's suggestion is outdated

tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py
```diff
@@ -229,21 +227,17 @@ def test_quant_model_reload(format, dtype, tmp_path):
...
         (False, torch.float32, False, "cpu"),
-        (False, torch.float32, True, "cpu"),
         # offloading
         (True, torch.float16, False, "cpu"),
         (True, torch.float32, False, "cpu"),
-        (True, torch.float16, True, "cpu"),
-        (True, torch.float32, True, "cpu"),
     ],
 )
-def test_model_reload(offload, torch_dtype, tie_word_embeddings, device, tmp_path):
```
brian-dellabetta, line 246: Any reason we want to remove these tests? Might still be good to test `untie_word_embeddings` here

src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -175,6 +179,7 @@ def on_finalize(self, state: State, **kwargs) -> bool:
         return True
 
     def _center_embeddings(self, model: PreTrainedModel):
+        untie_word_embeddings(model)
```
brian-dellabetta: if we are always calling this on SpinQuant modifier (i think `_center_embeddings` will always get called?), it might make more sense to move this up into the `on_initialize` code block (basically move it out of the `if SpinquantRotation.R1 in self.rotations:` check above), so it's clear to the user it is happening during the lifecycle.

brian-dellabetta:
This PR should resolve #1969

src/llmcompressor/modifiers/transform/spinquant/base.py
HDCharles: I wanted to make it so that if there was ever a flow without center embeddings, it would work automatically correctly.

How about I added it to on-start next to the center_embeddings call and remove it from the R1 call so it doesn't look like it only happens then?

tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py
HDCharles: ok will add them back.

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
HDCharles: outdated

tests/llmcompressor/modifiers/quantization/test_shared_embeddings.py
HDCharles: outdated

HDCharles:
> Make sure to add a test for nested models like llava/pixtral

@kylesayrs 

just tested "llava-hf/llava-1.5-7b-hf" and it doesn't  have tied embeddings. I can tie them manually but then we'll be using the same interfaces we already know we support.

src/llmcompressor/modifiers/transform/spinquant/base.py
brian-dellabetta: Yeah consolidating makes sense. agree, i think it would make most sense on-start just above center embeddings, since that's where we're doing all the preparation logic.

src/llmcompressor/modifiers/transform/quip/base.py
```diff
@@ -123,6 +130,46 @@ def on_finalize(self, state: State, **kwargs) -> bool:
 
         return True
 
+    def _untie_if_target_shared_embedding(self, model: torch.nn.Module):
```
brian-dellabetta: nit -- If this is the same as `QuantizationMixin._untie_if_target_shared_embedding`, might be worth moving to a shared util.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -317,3 +323,73 @@ def _initialize_hooks(self, module: torch.nn.Module) -> Set[RemovableHandle]:
...
+        def _get_embeddings_or_warn(
+            model: torch.nn.Module,
+        ) -> tuple[torch.nn.Module | None, torch.nn.Module | None]:
+            if not (
+                hasattr(model, "get_input_embeddings")
+                and hasattr(model, "get_output_embeddings")
+            ):
+                logger.warning(
+                    f"{model.__class__} doesn't have attribute get_input_embeddings and"
+                    + " get_output_embeddings implemented."
```
kylesayrs: nit:
```suggestion
                    " get_output_embeddings implemented."
```

src/llmcompressor/modifiers/transform/quip/base.py
kylesayrs: Maybe `def get_word_embeddings(model: PreTrainedModel) -> list[Module | None]:`

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
kylesayrs: Yeah that's fine

src/llmcompressor/modifiers/quantization/quantization/mixin.py
HDCharles: updated

src/llmcompressor/modifiers/transform/quip/base.py
HDCharles: ok i conslidated the logic

src/llmcompressor/args/model_arguments.py
```diff
@@ -64,10 +64,11 @@ class ModelArguments:
     )
 
     tie_word_embeddings: bool = field(
```
kylesayrs, line 66: We should consider renaming this to "untie_word_embeddings", since "untie" is really the active verb here. In the sense that if a model has untied weights already, we will not attempt to tie them.

HDCharles: Yeah i thought the same thing but I wasn't sure if it would be too weird to change the field to mean the opposite of what it was before.

PR #1964, opened 2025-10-24
[Misc] Remove NeuralMagic copyright

kylesayrs:


brian-dellabetta (approved):
so rule of thumb is copyright only in src files, right?

PR #1971, opened 2025-10-28
[`model_free_ptq`] Add pathway for day-zero weight quantization support

kylesayrs:
## Purpose ##
* Create a pathway which can quantize model weights without needing a model definition or the use of a calibration pipeline. Such a pathway provides fast and reliable support for models which:
  * Do not have a HF model definition yet
  * Have complications with sequential pipelines (very large vision towers, tracing failure, long calibration runtime)

## Usage ##
```python
model_free_ptq(
    model_stub="meta-llama/Llama-3.2-1B-Instruct",
    save_directory="Llama-3.2-1B-Instruct-FP8_block",
    scheme="FP8_BLOCK",
    ignore=["model.embed_tokens", "lm_head"],
    max_workers=15,
    device="cuda:0",
):
```

## Testing ##
* Added `test_model_free_ptq_matches_oneshot` which tests that saved tensors and configs exactly match between `model_free_ptq` and `oneshot` entrypoints for the same arguments. This test takes about 10 seconds to run.

## Future Extensions ##
* Mixed-precision quantization (multiple recipes/targets)
* Multi-GPU support (work is already parallelized by threads, but if GPU is the bottleneck we can split the work across GPUs)
* Multi-process support (is python processing is the bottleneck, we can replace multithreading with multiprocessing)


dsikka (commented):
I think keeping this as a separate entrypoint makes sense given the point is speed of execution / not wanting it getting bogged down from other flows in llmcomp but it seems like the lifecycle / steps we're applying are very similar to the datafree pipeline?

kylesayrs:
@dsikka Yes

src/llmcompressor/entrypoints/model_free/model_utils.py
```diff
@@ -0,0 +1,48 @@
...
+        for filename in filenames:
+            rel_path = os.path.relpath(os.path.join(dirpath, filename), root_dir)
+            if not (ignore and rel_path.startswith(ignore)):
+                all_files.append(rel_path)
+    return all_files
+
+
+# distinguish relative file paths from absolute/resolved file paths
+# relative file paths are used to find the save path
+# resolved file paths are what are used to load data
```
brian-dellabetta, line 48: consider moving to top of file or docstring

PR #1973, opened 2025-10-28
[AWQ] Allow users to disable quantization during AWQ

brian-dellabetta:
SUMMARY:
We want to provide users the ability to disable quantization in `AWQModifier`, so the best scales are found and applied to the weights without round-to-nearest quantization. This is useful when someone wants to run AWQ and GPTQ, before ultimately quantizing weights. This is a draft solution, see #1972 for discussion

TEST PLAN:

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -1,5 +1,4 @@
 import inspect
-from typing import Dict, List, Optional, Tuple, Union
 
 import torch
 from compressed_tensors.quantization import disable_quantization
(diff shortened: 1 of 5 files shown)


zhanglei1172:
Currently, if only want the scale of awq to take effect and do not want to perform the quantization process,  one can simply set save compressed= False in the model.save pretrained method (maybe the scale+zero point parameter needs to be manually deleted).

brian-dellabetta:
> Currently, if only want the scale of awq to take effect and do not want to perform the quantization process, one can simply set save compressed= False in the model.save pretrained method (maybe the scale+zero point parameter needs to be manually deleted).

Hi @zhanglei1172 , yes that's probably true. But it would require some manual overhead for a user who wants to do that, whereas this could just be a single boolean flag `disable_quantization`, based on discussion #1972. I think it would amount to the same thing, scales and zero points are just not created and quantization configs and statuses are pruned from the target modules, so the checkpoint at the end of the pipeline would just have modified weights

brian-dellabetta:
Closing in favor of #2206

PR #1976, opened 2025-10-29
[Transform] SpinQuant fix OOM

zhanglei1172:
SUMMARY:
"When using SpinQuantModifier for some fuse operations, it is necessary to add the torch.no_grad decorator. Otherwise, PyTorch will capture the grad graph by default, leading to a gradual increase in memory usage. I encountered a CUDA OOM issue when rotating the MOE model, and the OOM problem was resolved after fixing it."

TEST PLAN:
"Performed code quality evaluation locally"

### DIFF
--- a/src/llmcompressor/modifiers/transform/spinquant/base.py
+++ b/src/llmcompressor/modifiers/transform/spinquant/base.py
@@ -144,6 +144,7 @@ class SpinQuantModifier(Modifier, use_enum_values=True):
 
         return True
 
+    @torch.no_grad()
     def on_start(self, state: State, event: Event, **kwargs):
         self.started_ = True
 


kylesayrs (approved):
Awesome catch, thanks!

kylesayrs:
@brian-dellabetta I'm ambivalent towards moving this into a higher scope (calibration pipeline). There might be future modifier techniques which might require gradients (cayley optimization) with which a higher scope would conflict, but those don't exist atm.

kylesayrs:
Either scope would work. Even a lower scope (fuse norms and center embeddings) would work.

zhanglei1172:
@HDCharles I accidentally chose the rebase main; it may need to be approved again.

kylesayrs:
@zhanglei1172 The approvals persist through rebase if there's no conflicts

PR #1977, opened 2025-10-29
[InternVL3]Add internvl3 quantizing example

BigFaceBoy:
SUMMARY:
LLM Compressor doesn't currently have any examples of InternVL3. It truely took me a lot of time to quantize it successful. So I want to share the example of it.

TEST PLAN:

### DIFF
--- a/examples/multimodal_vision/internvl3_example.py
+++ b/examples/multimodal_vision/internvl3_example.py
@@ -143,8 +143,8 @@ def data_collator(batch):
     assert len(batch) == 1
     item = {key: value for key, value in batch[0].items()}
     item["pixel_values"] = load_image_from_PIL(item["image"])
-    item["labels"] = torch.LongTensor([item["input_ids"]])
     item["input_ids"] = torch.LongTensor([item["input_ids"]])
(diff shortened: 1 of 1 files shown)


examples/multimodal_vision/internvl3_example.py
```diff
@@ -0,0 +1,196 @@
...
+
+from llmcompressor import oneshot
+
+# IMPORTANT: Before running this script, you must manually modify the
+# `modeling_internvl_chat.py` file in your local copy of the `OpenGVLab/InternVL3-8B`
+# model directory 
+# Replace the original `forward` method of the `InternVLChatModel` class with the
+# version provided in `examples/multimodal_vision/internvl3_README.md`.
+# And put the `chat_template.jinja` in your local copy of the `OpenGVLab/InternVL3-8B`
+# model directory.
```
brian-dellabetta, line 16: Rather than manually doing this, we might be able to replace modules with custom wrappers to handle this automatically, once this PR is in we can follow its pattern:
* #1851

BigFaceBoy:
I find out that there is problem in this way, so I`ll close this PR

PR #1981, opened 2025-10-29
[MoE] Clean up imports, add qwen3_moe_vl, change logger level

kylesayrs:
## Purpose ##
* Change imports to make sure that modules are registered, even if the context is imported not from prepare.py
* Use `RegistryMixin` for registry functionality to make use of code reuse
* Do not log every every time moe context is entered (some models are not moes, some users may be confused)
* Add support for qwen3_moe_vl

## Changes ##
* Use `RegistryMixin` for registry functionality
* Change logger level of moe entering message
* Move registry init imports
* Add `CalibrateQwen3VLMoeTextSparseMoeBlock`

## Testing ##
* MoE tests pass, ran moe examples and see that modules are being replaced

### DIFF
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -55,7 +55,7 @@ jobs:
 
   transformers-tests:
     needs: [detect-changes]
-    runs-on: gcp-k8s-vllm-l4-solo
+    runs-on: ibm-wdc-k8s-vllm-h100-solo
     if: (contains(github.event.pull_request.labels.*.name, 'ready') || github.event_name == 'push') && needs.detect-changes.outputs.matched-changes != ''
     steps:
       - uses: actions/setup-python@v5
--- a/examples/transform/README.md
+++ b/examples/transform/README.md
@@ -41,7 +41,7 @@ recipe = [
 ]
 ```
 
-Note that `QuIPModifier` can be customized. For a full list of the available arguments, see the [docstring](/src/llmcompressor/modifiers/transform/spinquant/base.py) or documentation.
+Note that `QuIPModifier` can be customized. For a full list of the available arguments, see the [docstring](/src/llmcompressor/modifiers/transform/quip/base.py) or documentation.
 
 * `rotations` determines which of the input rotation (v) or output rotations (u) should be used.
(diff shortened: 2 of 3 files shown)


src/llmcompressor/modeling/moe_context.py
```diff
@@ -115,10 +115,17 @@ def moe_calibration_context(
                 model(**batch)
         # Model is now restored (unless permanent)
     """
+    # trigger registration
+    from .deepseek_v3 import CalibrationDeepseekV3MoE  # noqa: F401
+    from .llama4 import SequentialLlama4TextMoe  # noqa: F401
+    from .qwen3_moe import CalibrationQwen3MoeSparseMoeBlock  # noqa: F401
+    from .qwen3_vl_moe import CalibrateQwen3VLMoeTextSparseMoeBlock  # noqa: F401
+    # TODO: add granite4, Qwen3Next
```
brian-dellabetta: can't this be handled by including all of these in `__init__.py`?

kylesayrs: Great suggestion, done

kylesayrs:
FYI @sairampillai

sairampillai:
Looks awesome! @kylesayrs

dsikka:
@kylesayrs can you address conflicts?

src/llmcompressor/modeling/qwen3_vl_moe.py
```diff
@@ -64,6 +84,9 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         next_states = next_states.reshape(batch_size, sequence_length, hidden_dim)
         return next_states, router_logits
 
+    def restore(self, original: torch.nn.Module) -> torch.nn.Module:
+        return original
```
dsikka, line 88: We dont need to set a restore method if permanent is True, right? This wont ever be called?

fynnsu (dismissed):

src/llmcompressor/modeling/qwen3_vl_moe.py
kylesayrs: You're right, thanks for the catch. Will address in follow up.

PR #1984, opened 2025-10-30
[Qwen3 Next] Update `qwen3_next` to use `moe_calibration_context`

dsikka:
# SUMMARY:
- Update the qwen3_next_moe definition to use the moe context
- Update example to use the correct arguement
- Add tests
- Update qwen3_moe model definition to implement `restore` 

# Testing:
- All modeling tests pass

### DIFF
--- a/src/llmcompressor/args/model_arguments.py
+++ b/src/llmcompressor/args/model_arguments.py
@@ -64,10 +64,11 @@ class ModelArguments:
     )
 
     tie_word_embeddings: bool = field(
-        default=False,
+        default=True,
         metadata={
             "help": "Whether the model's input and output word embeddings "
-            "should be tied. Note that this is only relevant if the "
+            "should attempt to be left tied. False means always untie."
+            " Note that this is only relevant if the "
             "model has a output word embedding layer."
         },
     )
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -233,7 +233,7 @@ def oneshot(
     processor: Optional[Union[str, ProcessorMixin]] = None,
     use_auth_token: bool = False,
     precision: str = "auto",
-    tie_word_embeddings: bool = False,
+    tie_word_embeddings: bool = True,
     trust_remote_code_model: bool = False,
     save_compressed: bool = True,
     model_revision: str = "main",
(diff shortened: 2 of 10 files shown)


src/llmcompressor/modeling/qwen3_next_moe.py
```diff
@@ -16,13 +16,25 @@
...
+
+    is_permanent = False
 
-class Qwen3NextSparseMoeBlock(torch.nn.Module):
     def __init__(
         self,
-        config,
         original,
-        calibrate_all_experts: bool,
+        config,
```
kylesayrs: Add type hints

dsikka:
> Tests?

I'll add. Running basic evals on vLLM to start

tests/llmcompressor/modeling/test_calib_qwen3_next.py
```diff
@@ -0,0 +1,96 @@
...
+                break
+
+        assert moe_layer is not None
+
+        num_experts = len(moe_layer.experts)
+        expert_triggered = [False for _ in range(num_experts)]
+
+        # Define the hook function
+        def hook_fn(i, module, input, output):
+            expert_triggered[i] = True
```
HDCharles, line 41: should this check that all tokens were recieved?

expert_triggered[i] = input.shape[0] == seq_len 

or something liek that?

tests/llmcompressor/modeling/test_calib_qwen3_next.py
```diff
@@ -0,0 +1,96 @@
...
+            _ = moe_layer(sample)
+
+        # Assert all experts are used
+        assert all(
+            expert_triggered
+        ), f"Not all experts were triggered: {expert_triggered}"
+
+
+@requires_gpu
+def test_calib_qwen3_moe_module():
```
kylesayrs, line 63: We should consider standardizing these tests in the future.

src/llmcompressor/modeling/qwen3_next_moe.py
```diff
@@ -16,17 +16,34 @@
...
+
+    is_permanent = False
 
-class Qwen3NextSparseMoeBlock(torch.nn.Module):
     def __init__(
         self,
-        config,
-        original,
-        calibrate_all_experts: bool,
+        original: Qwen3NextSparseMoeBlock,
```
kylesayrs, line 40: You don't run into an import error here?

kylesayrs (approved):

tests/llmcompressor/modeling/test_calib_qwen3_next.py
dsikka: We can potentially add this in a follow-up. We’re using the same tests for all the models

src/llmcompressor/modeling/qwen3_next_moe.py
dsikka: It does not but it likely will if not in the init for older transformers Let me update in a follow up

PR #1985, opened 2025-10-30
[AWQ] allow for use of model-wide kwargs cache

brian-dellabetta:
SUMMARY:
It appears AutoAWQ makes an implicit assumption that, when caching inputs to the forward call of certain modules, [only the first input is needed](https://github.com/casper-hansen/AutoAWQ/blob/main/awq/quantize/quantizer.py#L627-L635), and [all other kwargs can be shared](https://github.com/casper-hansen/AutoAWQ/blob/main/awq/quantize/quantizer.py#L569-L602), so that they don't have to be redundantly stored in GPU VRAM. 

When we first ported AutoAWQ, we thought this seemed incorrect and could lead to poor behavior. Our implementation [cached all the args](https://github.com/vllm-project/llm-compressor/blob/0.8.1/src/llmcompressor/modifiers/awq/base.py#L388-L394) into a given module's forward call, so that it is guaranteed to be replicated correctly at the expense of GPU VRAM.

Now that we are revisiting performance improvements for AWQ, I wanted to add AutoAWQ's design choice as a toggleable field on AWQModifier.
- If `AWQModifier(..., use_auto_awq_mem_hack=True)`, we will use AutoAWQ's technique to cache to a field
  -  `_model_kwargs_cache: IntermediatesCache`
- Otherwise, we will cache to a field
  - `_parent_kwargs_cache: dict[Module, IntermediatesCache]`

I am pretty sure that's what this PR has, but when I compare the running of `examples/awq/qwen3_moe_example.py` side-by-side with the field False or True, I don't see any meaningful difference in the VRAM usage. I need to do some more debugging to make sure this is working as intended (and compare to VRAM usage when running AutoAWQ)

TEST PLAN:

### DIFF
--- a/examples/awq/qwen3_moe_example.py
+++ b/examples/awq/qwen3_moe_example.py
@@ -56,7 +56,7 @@ recipe = [
         ignore=["lm_head", "re:.*mlp.gate$", "re:.*mlp.shared_expert_gate$"],
         scheme="W4A16",
         targets=["Linear"],
-        # use_auto_awq_mem_hack=os.getenv("USE_HACK", "") == "yes",
+        use_auto_awq_mem_hack=os.getenv("USE_HACK", "") == "yes",
         # GPU VRAM consistently peakds at ~37784MiB regardless
     ),
 ]


brian-dellabetta:
After further investigation, even caching just the first batch's kwargs causes no change in the peak GPU VRAM used. This is surmised to be because even though there are fewer keys in all the caches, the values those keys point to include a lot of redundant `data_ptr`s to GPU VRAM. The CPU RAM usage decreases, maybe 10-20%, because of this change, but that hardly seems worth it make this a feature a user can enable.

I will close this for now, we can always revive in future but the value-add doesn't seem worth the additional complexity

To reproduce, on this branch run `USE_HACK=yes python examples/awq/qwen3_moe_example.py` and `USE_HACK=no python examples/awq/qwen3_moe_example.py`, and `watch nvidia-smi` to see GPU VRAM is equal for both

PR #1989, opened 2025-11-03
[Tracing] Copy `HFTracer` to support latest transformers

kylesayrs:
## Background ##
* https://github.com/huggingface/transformers/pull/41683 remove some fx tracing utilities used by LLM Compressor's sequential pipeline capabilities

## Purpose ##
* Allow LLM Compressor to support latest transformers

## Changes ##
* Copied `src/transformers/utils/fx.py` (with copyright)
* Use `trust_remote_code` for trace testing (the old deepseek model definitions are not supported with the latest transformers

## Testing ##
* Used `test_models` to test that all models can still trace with transformers main
* Quantized llama 8B e2e and confirmed sane outputs with transformers main


kylesayrs:
@dsikka I think we're good for now. Tracing is already tested e2e

PR #1991, opened 2025-11-04
Adding early warning for groupsize divisibility issues

HDCharles:
Summary:
Previously you would only run into issues when you trie dto save/run model, better to error out earlier and also give a helpful error rather than just saying layer is not divisible by groupsize

resolves https://github.com/issues/created?issue=vllm-project%7Cllm-compressor%7C1983

Tests and error message generation written by claude code, will revise tests once we confirm that this is the direction we want to go

SUMMARY:

TEST PLAN:

### DIFF
--- a/src/llmcompressor/modifiers/quantization/quantization/mixin.py
+++ b/src/llmcompressor/modifiers/quantization/quantization/mixin.py
@@ -186,7 +186,7 @@ class QuantizationMixin(HooksMixin):
             )
             validate_group_size_divisibility(
                 modules_to_validate,
-                current_ignore_list=self.ignore if self.ignore else [],
+                current_ignore_list=self.ignore,
             )
 
         # disable quantization until calibration


kylesayrs (changes requested):
Can we do this in Compressed Tensors? This code only checks weights quantization, but  we can achieve the same result and check inputs/outputs if we only change these lines

https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/lifecycle/initialize.py#L214-L240


```python3
elif strategy in (QuantizationStrategy.GROUP, QuantizationStrategy.TENSOR_GROUP):
    # no non-divisible kernels exist for group quant
    num_groups = strategy_cdiv(observed_shape[-1], group_size, strategy, strict=True)
elif strategy == QuantizationStrategy.BLOCK:
    # non-divisible kernels exist for block quant
    num_rows = strategy_cdiv(observed_shape[-2], block_structure[-2], strategy, strict=False)
    num_cols = strategy_cdiv(observed_shape[-1], block_structure[-1], strategy, strict=False)
```

PR #1992, opened 2025-11-04
[LM Eval] Update / fix testing

dsikka:
SUMMARY:
- Bump LM Eval to support chartqa - #1953  landed without updating lm-eval which is required to support chatqa for VL models
- Update recovery metrics for two cases to use different thresholds, not 95%


kylesayrs:
FYI changing lm-eval versions (could but shouldn't) change the results of other eval tests

Issue #1968, opened 2025-10-28
RFC: Add Intel AutoRound Quantization Algorithm Support


yiliu30:
Hi, here is the INC team from Intel.  Thank you for developing this amazing project!

Our team has developed **AutoRound**, a novel tuning-based quantization algorithm that delivers state-of-the-art accuracy with low tuning cost and no extra inference overhead. The key features are listed below:

- **Superior Accuracy:** Achieves leading performance across various scenarios, including very low bits (2-3 bits).
- **Multiple Data Types:** Effective for a variety of data types (W4A16, MXFP8, MXFP4, FP8, NVFP4, etc.).
- **Broad Model Support:** Validated on both Large Language Models (LLMs) and Vision-Language Models (VLMs).
- **Advanced Mixed-Precision:** Supports layer-wise mixed-bit quantization and automatic bit-width search.
- **Flexible Tuning:** Offers a configurable tuning space to balance tuning cost and accuracy goals.

 For more detailed information, please refer to our [paper](https://aclanthology.org/2024.findings-emnlp.662.pdf), and [GitHub repository](https://github.com/intel/auto-round/).

We propose integrating AutoRound into LLM Compressor to provide users with a simple, high-accuracy quantization option.

### The key Idea of AutoRound

AutoRound quantizes a given tensor by introducing three trainable parameters (`V`, &alpha; and &beta;) to adjust the rounding value and clipping range. For a given transformers model, AutoRound quantizes the decoder block one by one, using block-wise output reconstruction error as loss to train these parameters.

![autoround_overview](https://github.com/user-attachments/assets/bec1a573-f091-4b46-9dce-0302022746b3)

### Integration Proposal

AutoRound is already integrated into popular LLM ecosystems such as [Transformers](https://huggingface.co/docs/transformers/quantization/auto_round), [vLLM](https://docs.vllm.ai/en/latest/features/quantization/auto_round.html), [SGLang](https://github.com/sgl-project/sglang/pull/10153) and [TorchAO](https://github.com/pytorch/ao/tree/main/torchao/prototype/autoround).  For LLM Compressor, we proposed two integration options as below:

#### Option 1. A New `modifier` for AutoRound (Recommend)

This approach integrates AutoRound as a new modifier within LLM Compressor. Specifically, we propose creating an `AutoRoundModifier` that wraps the AutoRound algorithm, delegating the actual tuning logic to the [intel/auto-round](https://github.com/intel/auto-round) library. 
[...]

PR #1994, opened 2025-11-05
Add Intel AutoRound algorithm support

yiliu30:
Resolve https://github.com/vllm-project/llm-compressor/issues/1968

### Highlights
- Introduced `AutoRoundModifier` to enable AutoRound quantization for `wNa16`.
- Added an end-to-end example and unit tests.
- Verified functionality with local accuracy tests (GSM8K with a limit of 1000, the results may fluctuate due to non-determinism.)
```bash
- LLMC-AutoRound
vllm (pretrained=/storage/yiliu7/Meta-Llama-3-8B-Instruct-W4A16-G128-disbale-shuffule,tensor_parallel_size=1,max_model_len=8192,max_num_batched_tokens=32768,max_num_seqs=128,add_bos_token=True,gpu_memory_utilization=0.8,dtype=bfloat16,max_gen_toks=2048,enable_prefix_caching=False), gen_kwargs: (None), limit: 1000.0, num_fewshot: None, batch_size: 128
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.737|±  |0.0139|
|     |       |strict-match    |     5|exact_match|↑  |0.736|±  |0.0139|

- AutoRound result as ref
vllm (pretrained=/storage/yiliu7/meta-llama/Meta-Llama-3-8B-Instruct-ar/Meta-Llama-3-8B-Instruct-w4g128/,tensor_parallel_size=1,max_model_len=8192,max_num_batched_tokens=32768,max_num_seqs=128,add_bos_token=True,gpu_memory_utilization=0.8,dtype=bfloat16,max_gen_toks=2048,enable_prefix_caching=False), gen_kwargs: (None), limit: 1000.0, num_fewshot: None, batch_size: 128
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.739|±  |0.0139|
|     |       |strict-match    |     5|exact_match|↑  |0.740|±  |0.0139|

```
Attached [eval cmd](https://gist.github.com/yiliu30/a7881cd1cbf0d676e3ffac3e3833aa8e) FYI.

### Next stage (in later PRs)
- [ ] Extend support for additional data types.
- [ ] Add group-wise quantization recipes mapping between LLMC and AutoRound.
- [ ] Add end-to-end tests.

cc @hshen14 @thuang6 @wenhuach21

### DIFF
--- a/src/llmcompressor/modeling/moe_context.py
+++ b/src/llmcompressor/modeling/moe_context.py
@@ -105,8 +105,8 @@ def moe_calibration_context(
         ):
             replacement = MoECalibrationModule.load_from_registry(
                 class_name,
-                module,
-                model.config,
+                original=module,
+                config=model.config,
                 calibrate_all_experts=calibrate_all_experts,
             )
             model.set_submodule(name, replacement)


yiliu30:
Hi  @dsikka @kylesayrs, could you please review this when you get a chance? Thanks!

src/llmcompressor/modifiers/quantization/autoround/base.py
```diff
@@ -0,0 +1,311 @@
...
+            for _, module in decoding_layer.named_modules():
+                if hasattr(module, "weight_scale") and hasattr(
+                    module, "weight_zero_point"
+                ):
+                    # Note: The model's weight is already q-dq in-place by auto-round.
+                    weight_scale = module.scale
+                    del module.scale
+                    del module.zp
+                    # TODO: update zero_point as well if needed
+                    update_offload_parameter(module, "weight_scale", weight_scale)
```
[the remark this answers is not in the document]
yiliu30: Corrected the comments; will add `asym` support in the later PR.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,311 @@
...
+    def _mapping_config_to_autoround(self):
+        from auto_round.schemes import QuantizationScheme as ARQuantizationScheme
+
+        resolved_config = self.resolved_config
+        quant_scheme = None
+        for scheme in resolved_config.config_groups.values():
+            assert isinstance(
+                scheme, QuantizationScheme
+            ), f"Expected QuantizationScheme, got {type(scheme)}"
+            quant_scheme = scheme
```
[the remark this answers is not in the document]
yiliu30: Add a more strict check for now; will remove that in later PRs.

examples/autoround/llama3_example.py
```diff
@@ -0,0 +1,56 @@
+from auto_round.calib_dataset import get_dataset
```
dsikka, line 1: So that's it's easier to find, we can create a new autoround examples folder and add this in

src/llmcompressor/modifiers/autoround/__init__.py
```diff
@@ -0,0 +1,3 @@
+# ruff: noqa
+
```
dsikka, line 2: Can we move the autoround folder to be under modifiers? We will be removing the quantization folder soon

setup.py
```diff
@@ -144,6 +144,8 @@ def localversion_func(version: ScmVersion) -> str:
             if BUILD_TYPE == "release"
             else "compressed-tensors>=0.12.3a2"
         ),
+        # TODO: replace it with the release version
```
dsikka, line 147: We generally pin all our dependencies

src/llmcompressor/utils/helpers.py
```diff
@@ -1082,3 +1083,6 @@ def patch_attr(base: object, attr: str, value: Any):
             setattr(base, attr, original_value)
         else:
             delattr(base, attr)
+
+
+DISABLE_QAC_MODIFIERS = ["GPTQModifier", "AWQModifier", "AutoRoundModifier"]
```
dsikka, line 1088: just an fyi, this just wont run Quant-Dequant during any forward passes done through https://github.com/vllm-project/compressed-tensors/blob/52792be02ec09e59f3517104e755a02d0e003fbb/src/compressed_tensors/quantization/lifecycle/forward.py#L366

src/llmcompressor/modifiers/quantization/__init__.py
```diff
@@ -2,3 +2,4 @@
 
 from .gptq import *
 from .quantization import *
+from .autoround import *
```
dsikka: move to modifiers folder

yiliu30:
> Were you able to save a checkpoint in the compressed-tensors format that you can share?

Sure! Here’s the checkpoint for `Qwen/Qwen3-8B`, which comes with a more flexible license: 
[Yi30/Qwen3-8B-W4A16-G128-AutoRound](https://huggingface.co/Yi30/Qwen3-8B-W4A16-G128-AutoRound)

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,323 @@
...
+        For more details, please refer to the AutoRound repository:
+        https://github.com/intel/auto-round/
+        """
+        cur_layer_idx = self._cur_layer_idx
+        logger.info("Applying AutoRound to layer index: {}", cur_layer_idx)
+        self._cur_layer_idx += 1
+        if cur_layer_idx >= len(state.model.model.layers):
+            # skip the lm_head layer
+            return
+        decoding_layer = state.model.model.layers[cur_layer_idx]
```
kylesayrs: The sequential pipeline is not guaranteed to break a model into decoder layers. If a user specifies `sequential_targets="Linear"`, then each `SEQUENTIAL_EPOCH_END` will trigger on each linear layer of the model.

One way to generalize this would be to have the sequential pipeline return the modules that were in the sequential layer

```python3
class Subgraph:
    def get_modules(self, model: torch.nn.Module, recurse: bool = False) -> Set[torch.nn.Module]:
        nodes = self.graph.find_nodes(op="call_module")
        modules = set(model.get_submodule(node.target) for node in nodes)
        if recurse:
            modules = set(module.modules() for module in modules)

        return modules
```

```python3
class SequentialPipeline:
...
    for subgraph in subgraphs:
```

```python3
def apply_autoround(self, state, subgraph):
    decoding_layer = torch.nn.ModuleList(list(subgraph.modules()))
```

kylesayrs (commented):
Looks great. There's a few more things I'd like to point out, will give a full review soon.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,323 @@
...
+            param.requires_grad_(False)
+        return True
+
+    def start_calibration(self, model: torch.nn.Module):
+        """
+        Register activation calibration hooks and enable quantization as we calibrate
+
+        :param model: model to prepare for calibration
+        """
+
```
HDCharles, line 146: need to manually call the untie functionality since you aren't using the start_calibration from [QuantizationMixin](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/quantization/mixin.py#L189)

```suggestion
untie_if_target_shared_embedding(model, self._module_names.values())
```

src/llmcompressor/modifiers/quantization/autoround/base.py
```diff
@@ -0,0 +1,323 @@
...
+            mod._tmp_name = name
+
+    def _remove_temporary_names(self, model: torch.nn.Module):
+        for _, mod in model.named_modules():
+            if hasattr(mod, "_tmp_name"):
+                del mod._tmp_name
+
+    def on_initialize(self, state: State, **kwargs) -> bool:
+        """
+        Initialize and run the AutoRound algorithm on the current state
```
HDCharles: misleading docstring, it just does the initialization, right?

src/llmcompressor/modifiers/autoround/base.py
kylesayrs: FYI these changes are implemented [here](https://github.com/vllm-project/llm-compressor/pull/1998), and this PR can potentially rebase on them.

examples/autoround/llama3_example.py
yiliu30: Thanks, I’ve updated the examples!

src/llmcompressor/modifiers/autoround/__init__.py
yiliu30: Sure, updated.

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Good catch — I’ve added it!

src/llmcompressor/utils/helpers.py
yiliu30: I see — since the AutoRound tuning process already performs Q-DQ, we should disable it here and let AutoRound handle it. Otherwise, the loss calculation may be incorrect.

setup.py
yiliu30: Got it! We're preparing a patch release for LLMC integration.
Would it be okay to keep using the branch in this PR for now, and switch to the patch release once it's available?

cc @thuang6

src/llmcompressor/modifiers/quantization/autoround/base.py
yiliu30: Yes, I have updated it, thanks!

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Thanks for the detailed explanation — this approach is definitely more robust. I’ve tested it locally, and it works well.

Will https://github.com/vllm-project/llm-compressor/pull/1998 be merged soon? If so, I’d prefer to rebase on `main` and update my PR accordingly to avoid introducing too much code here.

src/llmcompressor/modifiers/quantization/__init__.py
yiliu30: I didn’t import `AutoRoundModifier` in the `modifiers/__init__.py` because it causes circular imports.
Instead, I import it like this:
```python
```
Let me know if you'd prefer a different structure to avoid this issue!

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,328 @@
...
+    :param ignore: optional list of module class names or submodule names to not
+        quantize even if they match a target in config_groups. Defaults to empty list.
+    :param scheme: a single quantization scheme to apply to the model. This is a
+        dictionary that supports all keys from QuantizationScheme except targets, which
+        will be set to the targets parameter set at the modifier level.
+    """
+
+    # AutoRound modifier arguments
+    iters: Optional[int] = 200
+    enable_torch_compile: Optional[bool] = True
```
brian-dellabetta: Is there any difference between `enable_torch_compile=False` and `None`? If not, we can simplify this to
```suggestion
    enable_torch_compile: bool = True
```

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # private variables
+    _module_names: Dict[torch.nn.Module, str] = PrivateAttr(default_factory=dict)
+    _cur_layer_idx = PrivateAttr(default=0)
+    _all_module_input: Dict[str, List[Tuple]] = PrivateAttr(default_factory=dict)
+
+    def resolve_quantization_config(self) -> QuantizationConfig:
+        config = super().resolve_quantization_config()
+        return config
```
brian-dellabetta: This should already be handled by the `resolved_config` property on QuantizationMixin

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,328 @@
...
+from llmcompressor.modifiers.quantization.calibration import apply_calibration_status
+from llmcompressor.modifiers.quantization.quantization import QuantizationMixin
+from llmcompressor.transformers.compression.compressed_tensors_utils import (
+    untie_if_target_shared_embedding,
+)
+
+__all__ = ["AutoRoundModifier"]
+
+
+def _is_decoding_layer(module, name):
```
brian-dellabetta: do we already have helpers for this in compressed-tensors?

brian-dellabetta (commented):
Thanks for the contribution! This is really cool. I have a few comments/questions in an initial review

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Good catch, updated!

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Thanks, updated!

src/llmcompressor/modifiers/autoround/base.py
yiliu30: I have rebased the `main` branch and updated it accordingly.

src/llmcompressor/modifiers/autoround/base.py
yiliu30: I made a small update to compare it with `sequential_targets`.
If you have a better implementation or suggestions, I’d love to hear them!

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,335 @@
...
+    :param ignore: optional list of module class names or submodule names to not
+        quantize even if they match a target in config_groups. Defaults to empty list.
+    :param scheme: a single quantization scheme to apply to the model. This is a
+        dictionary that supports all keys from QuantizationScheme except targets, which
+        will be set to the targets parameter set at the modifier level.
+    """
+
+    sequential_targets: Union[str, List[str], None] = None
+    # AutoRound modifier arguments
+    iters: Optional[int] = 200
```
brian-dellabetta: sorry, same here
```suggestion
    iters: int = 200
```

setup.py
brian-dellabetta: Hi @yiliu30 , do you have an estimate for when the next version of autoround will release? Does it have the appropriate licensing to avoid issues like [this](https://github.com/vllm-project/compressed-tensors/issues/468)?

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -0,0 +1,335 @@
...
+            f"found {len(decoding_layers)}."
+        )
+        decoding_layer = decoding_layers[0]
+
+        logger.info("Applying AutoRound on layer {}", decoding_layer._tmp_name)
+
+        wrapped_model = _wrap_decoding_layer(decoding_layer)
+
+        with torch.enable_grad(), align_module_device(decoding_layer):
+            import auto_round
```
brian-dellabetta: can we move this to the top of the file? would be good to show imports to users at top scope

yiliu30: Sure, I've updated all the imports to be at the top of the file.

yiliu30:
> Hi @yiliu30 , do you have an estimate for when the next version of autoround will release? Does it have the appropriate licensing to avoid issues like https://github.com/vllm-project/compressed-tensors/issues/468?

Hi @brian-dellabetta , We're planning to release the next version within the next 1–2 weeks—hope that works for you!
As for AutoRound, it's licensed under [Apache License 2.0](https://github.com/intel/auto-round/blob/main/LICENSE), so I guess there shouldn't be any licensing concerns.

src/llmcompressor/modifiers/quantization/__init__.py
kylesayrs: Yeah, this is a known circular import issue. Importing from the full path is fine for now, thanks!

tests/llmcompressor/transformers/autoround/test_oneshot.py
```diff
@@ -0,0 +1,94 @@
...
+)
+
+
+@pytest.mark.parametrize(
+    "recipe",
+    [
+        recipe_str,
+        recipe_modifier_full,
+    ],
+)
```
kylesayrs, line 50: This should have a `@requires_gpu` decorator

setup.py
brian-dellabetta: Next autoround release is planned in 1-2 weeks -- https://github.com/vllm-project/llm-compressor/pull/1994#issuecomment-3505837494

Should we merge in now or wait? We can always pin later before our next release

tests/llmcompressor/transformers/autoround/test_oneshot.py
yiliu30: Good catch, updated!

setup.py
yiliu30: We’d like to proceed with merging this first, and will update it to the release version (0.9) as soon as it becomes available.

PR #1997, opened 2025-11-05
Update untie embedding function and usage

HDCharles:
SUMMARY:
1) making this function a little more forgiving so you can optionally just pass in the normal name, module generator.
2) improve docstring for QuantizationMixin so it mentions the untieing

TEST PLAN:
see CI

### DIFF
--- a/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
+++ b/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
@@ -203,7 +203,8 @@ def _get_embeddings_or_warn(
 
 
 def untie_if_target_shared_embedding(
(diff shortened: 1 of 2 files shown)


src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -203,16 +203,18 @@ def _get_embeddings_or_warn(
 
 
 def untie_if_target_shared_embedding(
-    model: torch.nn.Module, matched_module_generator: Generator[torch.nn.Module]
+    model: torch.nn.Module,
+    matched_module_generator: Generator[torch.nn.Module | tuple[str, torch.nn.Module]],
```
brian-dellabetta, line 207: nit --  i typically prefer to leave these input types cleaner, but since we use `match_named_modules` in so many places I think this is fine

HDCharles:
i'm going to do something else

PR #1998, opened 2025-11-05
[Sequential Pipeline] Return subgraph on `sequential_epoch_end`, remove `layer_sequential` pipeline

kylesayrs:
## Purpose ##
* Enable better targeting of modules by modifiers such as [AutoRound](https://github.com/vllm-project/llm-compressor/pull/1994)
* Remove legacy pipeline (which is incompatible with this change)

## Changes ##
* Pass subgraph to `sequential_epoch_end`, allowing modifiers to view all of the module that were called in the subgraph
* Implement `submodules` method on `Subgraph` which returns all the modules called by this subgraph
* Remove `LayerSequentialPipeline`, which does not use the `Subgraph` API and has been superseded by the sequential pipeline

### DIFF
--- a/src/llmcompressor/modifiers/transform/spinquant/base.py
+++ b/src/llmcompressor/modifiers/transform/spinquant/base.py
@@ -9,7 +9,7 @@ from compressed_tensors.transform import (
     TransformScheme,
     apply_transform_config,
 )
-from compressed_tensors.utils import TorchDtype
+from compressed_tensors.utils import TorchDtype, get_head_dim
 from pydantic import Field, ValidationInfo, field_validator
 from transformers import PreTrainedModel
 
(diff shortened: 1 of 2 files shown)


HDCharles (approved):
Looks good

shanjiaz (approved):

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -79,6 +79,14 @@ def forward(self, *args, **kwargs) -> Dict[str, Any]:
 
         return outputs
 
+    def submodules(self, model: Module, recurse: bool = False) -> Set[Module]:
```
dsikka, line 82: docstring

PR #1999, opened 2025-11-06
[torch.compile] Use torch.compile for tracing subgraphs

kylesayrs:
## Purpose ##
* Support subgraph creation with reduced technical burden for supporting new models
* Support sequential onloading of the vision tower (which is required by very large models such as [Cohere Vision](https://github.com/vllm-project/llm-compressor/pull/1710))

## Technique ##
1. Torch compile doesn't like replacing tensors with tensors of a different device (cpu offloading). Instead, we can fake cpu offloading by moving everything to `cpu_params`, then getting and moving the params to the gpu as they are `get`ted.

2. With that fixed, we assume that most models are written such that they can be torch.compiled. If a graph break happens inside of a sequential target, it's not an issue. If a graph break happens outside of a sequential target, it is an issue.

3. Now, we only need to determine where a sequential target starts and ends in the graph. We can use `_export_tracepoint`, although it's not going to support weird args  types like Cache. If we ever needed it, we could  add support for Cache. There might also be another way of splitting this.

## Issues ##
1. This kind of CPU offloading works, and I believe a working implementation exists, but it's a little complicated. However, I don't see this as a major blocker.

2. I think the assumption that most models can be torch compiled is mostly correct.

3. I think the major issue is how to stitch together broken subgraphs. These subgraphs are not guaranteed to use the same names, and mapping input names to torch compile fx graph names is also very difficult (undefined?). There's also a high chance that graph breaks will occur outside of sequential targets, in which case there is no possible recovery.

Subgraph 0
```python3
class GraphModule(torch.nn.Module):
    def forward(self, L_self_modules_language_model_modules_embed_tokens_cpu_params_weight_: "bf16[151936, 1536]", L_input_ids_: "i64[1, 4096]"):
        l_self_modules_language_model_modules_embed_tokens_cpu_params_weight_ = L_self_modules_language_model_modules_embed_tokens_cpu_params_weight_
        l_input_ids_ = L_input_ids_
        
         # File: /home/kylesayrs/llm-compressor/src/llmcompressor/pipelines/sequential/helpers.py:307 in custom__getattr__, code: return cpu_params[name].to("cuda")
        to: "bf16[151936, 1536]" = l_self_modules_language_model_modules_embed_tokens_cpu_params_weight_.to('cuda');  l_self_modules_language_model_modules_embed_tokens_cpu_params_weight_ = None
        
         # File: /home/kylesayrs/llm-compressor/env/lib64/python3.12/site-packages/transformers/models/qwen2_vl/modeling_qwen2_vl.py:1184 in forward, code: inputs_embeds = self.get_input_embeddings()(input_ids)
        inputs_embeds: "bf16[1, 4096, 1536]" = torch.nn.functional.embedding(l_input_ids_, to, None, None, 2.0, False, False);  l_input_ids_ = to = None
        return (inputs_embeds,)
```

Subgraph 1
```
class GraphModule(torch.nn.Module):
    def forward(self, L_self_modules_patch_embed_modules_proj_cpu_params_weight_: "bf16[1280, 3, 2, 14, 14]", L_hidden_states_: "f32[864, 1176]"):
        l_self_modules_patch_embed_modules_proj_cpu_params_weight_ = L_self_modules_patch_embed_modules_proj_cpu_params_weight_
        l_hidden_states_ = L_hidden_states_
        
         # File: /home/kylesayrs/llm-compressor/src/llmcompressor/pipelines/sequential/helpers.py:307 in custom__getattr__, code: return cpu_params[name].to("cuda")
        to: "bf16[1280, 3, 2, 14, 14]" = l_self_modules_patch_embed_modules_proj_cpu_params_weight_.to('cuda');  to = None
        
         # File: /home/kylesayrs/llm-compressor/env/lib64/python3.12/site-packages/transformers/models/qwen2_vl/modeling_qwen2_vl.py:248 in forward, code: hidden_states = hidden_states.view(
        hidden_states: "f32[864, 3, 2, 14, 14]" = l_hidden_states_.view(-1, 3, 2, 14, 14);  l_hidden_states_ = None
        
         # File: /home/kylesayrs/llm-compressor/env/lib64/python3.12/site-packages/transformers/models/qwen2_vl/modeling_qwen2_vl.py:251 in forward, code: hidden_states = self.proj(hidden_states.to(dtype=target_dtype)).view(-1, self.embed_dim)
        to_1: "bf16[864, 3, 2, 14, 14]" = hidden_states.to(dtype = torch.bfloat16);  hidden_states = None
        
         # File: /home/kylesayrs/llm-compressor/src/llmcompressor/pipelines/sequential/helpers.py:307 in custom__getattr__, code: return cpu_params[name].to("cuda")
        to_2: "bf16[1280, 3, 2, 14, 14]" = l_self_modules_patch_embed_modules_proj_cpu_params_weight_.to('cuda');  l_self_modules_patch_embed_modules_proj_cpu_params_weight_ = None
        
         # File: /home/kylesayrs/llm-compressor/env/lib64/python3.12/site-packages/transformers/models/qwen2_vl/modeling_qwen2_vl.py:251 in forward, code: hidden_states = self.proj(hidden_states.to(dtype=target_dtype)).view(-1, self.embed_dim)
        conv3d: "bf16[864, 1280, 1, 1, 1]" = torch.conv3d(to_1, to_2, None, (2, 14, 14), (0, 0, 0), (1, 1, 1), 1);  to_1 = to_2 = None
        hidden_states_1: "bf16[864, 1280]" = conv3d.view(-1, 1280);  conv3d = None
        return (hidden_states_1,)
```

### DIFF
--- a/src/llmcompressor/pipelines/sequential/helpers.py
+++ b/src/llmcompressor/pipelines/sequential/helpers.py
@@ -175,12 +175,16 @@ def trace_subgraphs(
     # ----- run torch compile -----
     sample_input = {k: v.to("cuda") for k, v in sample_input.items()}
 
-    graph = None
-    def custom_backend(gm, example_inputs):
+    graphs = []
+    def custom_backend(graph, example_inputs):
         print("make graph")
-        nonlocal graph
-        graph = gm
-        # maybe return an empty callable
+
+        output = graph.print_readable(print_output=False)
+        with open(f"output_{len(graphs)}.py", "w") as file:
+            file.write(output)
+
+        graphs.append(graph)
+
         return graph.forward
     
     with contextlib.ExitStack() as stack:
@@ -188,15 +192,13 @@ def trace_subgraphs(
         stack.enter_context(calibration_forward_context(model))
         stack.enter_context(HooksMixin.disable_hooks())
 
-        torch.compile(fullgraph=True, backend=custom_backend)(model.forward)(**sample_input)
+        torch.compile(fullgraph=False, backend=custom_backend)(model.forward)(**sample_input)
         #program = torch.export.export(model, tuple(), kwargs=sample_input)
         #model.forward(**sample_input)
     # ----- run torch compile -----
 
-    output = graph.print_readable(print_output=False)
-    with open("output.py", "w") as file:
-        file.write(output)
-    exit(0)
+    assert len(graphs) > 0
+    return graphs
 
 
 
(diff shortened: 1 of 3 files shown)


kylesayrs:
Closing due to the issues described in PR description.

PR #2001, opened 2025-11-06
Granite4 FP8 Block Quantization

krishnateja95:
SUMMARY:
"Recipe for FP8 Block quantization of Granite4-small model (https://huggingface.co/ibm-granite/granite-4.0-h-small)"


src/llmcompressor/modeling/granite4.py
```diff
@@ -1,9 +1,289 @@
 import torch
+import json
+import os
+import shutil
```
dsikka: Please use this context: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/moe_context.py

dsikka (changes requested):
Couple of key issues:
1. We should be using the moe_context: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/moe_context.py. We have many examples you can use that apply the context
2. All saving logic should be applied by compressed-tensors. Seems like you're doing a lot of custom logic

src/llmcompressor/modeling/granite4.py
krishnateja95: @dsikka This saving logic is specific to granite4 small model to ensure the fp8 block quantized model is compatible with vLLM.

examples/quantization_w8a8_fp8/granite4_block_quantization_example.py
```diff
@@ -0,0 +1,47 @@
...
+
+model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
+
+model = replace_granite_moe_with_linear_experts(model)
+
+ignore_lay = ["lm_head"]
+ignore_lay += ["re:.*block_sparse_moe.router"]
+ignore_lay += ["re:.*mamba.in_proj"]
+ignore_lay += ["re:.*shared_mlp.input_linear"]
```
brian-dellabetta: nit
```suggestion
ignore_lay = ["lm_head", "re:.*block_sparse_moe.router", "re:.*mamba.in_proj", "re:.*shared_mlp.input_linear"]
```

brian-dellabetta (commented):
One comment, otherwise LGTM pending @dsikka 's comment

examples/quantization_w8a8_fp8/granite4_fp8_block_example.py
brian-dellabetta, line 1: Consider rename to `granite4_fp8_block_example.py` to match precedent in `examples/`

examples/quantization_w8a8_fp8/granite4_dynamic_quantization_example.py
brian-dellabetta, line 1: Outside of a different granite4 model id, how does this differ from the current `granite4_example.py`? Can we merge the two files?

brian-dellabetta (commented):
Looks like you've addressed the calibration context concerns. A couple nits, but this is looking good to me, thanks for the contribution!

examples/quantization_w8a8_fp8/granite4_dynamic_quantization_example.py
krishnateja95: @brian-dellabetta I initially tried running `granite4_example.py` and I got an error. So, I had to modify and submit this updated code. Now, this code works for both `FP8_BLOCK` and `FP8_DYNAMIC`. I modified the weight loading function in vLLM to be able to load this granite model produced by llm-compressor. The corresponding weight loading PR of vLLM is here: https://github.com/vllm-project/vllm/pull/28398

The models produced by llm-compressor are here:
https://huggingface.co/nm-testing/granite-4.0-h-small-FP8-block
https://huggingface.co/nm-testing/granite-4.0-h-small-FP8-dynamic

brian-dellabetta: Thanks @krishnateja95 , if the previous example doesn't work and yours does for the same `FP8_DYNAMIC` scheme, maybe we should just replace it with yours? It was as community user PR that may be outdated now. @dsikka  WDYT?

krishnateja95: @brian-dellabetta Yes, please replace the existing example with this PR. The vLLM PR supports FP8-DYNAMIC version as well.

examples/quantization_w8a8_fp8/granite4_fp8_block_example.py
brian-dellabetta: updated

examples/quantization_w8a8_fp8/granite4_dynamic_quantization_example.py
brian-dellabetta: I updated on this PR

dsikka:
I think we need to make sure the vLLM changes are accepted first before we make changes on our end.

brian-dellabetta (dismissed):

brian-dellabetta (approved):
Hi @krishnateja95 , sorry for the delay on this, I was on leave the past few weeks. Are we good to merge this in?

dsikka:
@mergifyio refresh

dsikka:
@krishnateja95 are you still working on this?

dsikka:
Closing as stale

PR #2003, opened 2025-11-07
[AWQ] modernized typehints, allow for n_grid/duo_scaling parameters

brian-dellabetta:
SUMMARY:
This updates the AWQModifier type hints to python 3.10+, exposes n_grid as a parameter rather than the hard-coded 20, and allows the value of duo_scaling to be None, in which case half of the grid search is performed with duo_scaling off and the other half with duo_scaling on. This stems from @fynnsu 's findings that the previous default value of True was actually leading to worse behavior in some circumstances

TEST PLAN:
Run benchmarks with different values of n_grid and duo_scaling to see how much this affects accuracy/runtime

### DIFF
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -55,7 +55,7 @@ jobs:
 
   transformers-tests:
     needs: [detect-changes]
-    runs-on: gcp-k8s-vllm-l4-solo
+    runs-on: ibm-wdc-k8s-vllm-h100-solo
     if: (contains(github.event.pull_request.labels.*.name, 'ready') || github.event_name == 'push') && needs.detect-changes.outputs.matched-changes != ''
     steps:
       - uses: actions/setup-python@v5
--- a/examples/model_free_ptq/README.md
+++ b/examples/model_free_ptq/README.md
@@ -1,0 +1,15 @@
+# Quantizing models without a model definition 
+
+`model_free_ptq` provides a PTQ pathway for data-free schemes (such for FP8 Dynamic Per Token or FP8 Block). Specifically, this pathway removes the requirement for a model definition or the need to load the model through transformers. If you are interested in applying a data-free scheme, there are two key scenarios in which applying this pathway may make sense for your model:
+
+1. The model does not have a model definition available through transformers. This may be the case for a brand new model which has not landed in transformers.
+2. The model is very large (such as Kimi K2 Thinking) and is running into issues with `oneshot`
+
+
+`model_free_ptq` works directly with the safetensors in the checkpoint to which observers are applied, thereby removing the requirement for a model definition or transformers.
+
+# Quantizing Kimi K2 Thinking to FP8 Block 
+
+In `kimi_k2_thinking_fp8_block.py`, we call `model_free_ptq` by providing a `scheme` and `ignore` list, similar to how we provide reicpes to `oneshot` calls. In the case of Kimi-K2 Thinking, we apply the `FP8_BLOCK` scheme and ignore layers that are incompatible with a block_size of 128 (specifically, `kv_a_proj_with_mqa` and `q_a_proj`).
+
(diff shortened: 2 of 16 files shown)


src/llmcompressor/modifiers/awq/mappings.py
```diff
@@ -186,13 +185,13 @@ class ResolvedMapping:
...
     smooth_name: str
     smooth_layer: Module
-    balance_layers: List[Module]
-    balance_names: Optional[List[str]] = None
-    parent: Optional[Module] = None
-    parent_name: Optional[str] = None
+    balance_layers: list[Module]
+    balance_names: list[str]
+    parent: Module
+    parent_name: str
```
fynnsu, line 191: Removed `None` default values?

brian-dellabetta: yeah, they were unnecessary. The resolved mapping should always have them

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -111,31 +111,40 @@ class AWQModifier(Modifier, QuantizationMixin):
         device. Defaults to None, so cached args are not offloaded. Consider setting
         to torch.device("cpu") if you are encountering OOM errors
     :param duo_scaling: whether to use duo scaling, which uses both input activations
-        and weights to determine the scaling factor
+        and weights to determine the scaling factor. Defaults to True
+        If True, both activations and weights are used.
+        If False, only activations are used.
+        If None, half the grid search is performed with duo_scaling=False and the
+        other half is performed with duo_scaling=True.
```
kylesayrs, line 119: I've never heard of this option? Is it a standard option that users will be able to easily understand and provide value to the algorithm? If it's not a standard and isn't shown to provide increased accuracy, I think this might just create more complexity, both for users and for code maintainers.

fynnsu: When I looked into the duo scaling previously I couldn't find much support for it. Here is the pr where it was introduced to autoawq https://github.com/casper-hansen/AutoAWQ/pull/216.

brian-dellabetta: `duo_scaling=None` is a new option. I changed the default back to original (duo_scaling=True) for now, this is what we'll want to test out. but @fynnsu  was seeing duo_scaling=False was helping, searching both with a coarser n_grid seems like a reasonable default once we do some benchmarking if we find it to be beneficial

kylesayrs: Can we give it a string/enum value instead of None? Also, if we find that this mixed duo scaling option doesn't help, then landing this code will have introduced more complexity and options that we have now to support which don't help accuracy or performance.

brian-dellabetta: Do you prefer this?
```python
duo_scaling: bool | Literal["both"] = True
```
I wanted to avoid having to deal with string types ("both" vs. "Both" etc.), and enum types would break backward compatibility

brian-dellabetta: It's like 8 lines of complexity when preparing the grid search sweep. we can always revert, but I think this is a reasonable config that would span a better search space. we won't be able to definitively say even after benchmarking, optimal choice could very well be different based on model

kylesayrs: @brian-dellabetta pydantic will produce a readable error if the user uses the wrong capitalization, if you think that's acceptable.

I think "both" makes a lot more sense to me than `None`, since it makes it clear that this is going to double the search runtime. I agree that we shouldn't be doubling search runtime unless it shows some benefits.

brian-dellabetta: I changed it to `"both"`, but left n_grid to be the total number of grid points searched. so "both" with n_grid=20 will search 10 points with duo_scaling=False, and 10 points with duo_scaling=True

fynnsu (approved):
Looks good!

src/llmcompressor/modifiers/awq/base.py
kylesayrs: That seems fair!

PR #2008, opened 2025-11-10
Implement `propagate_error` argument

kylesayrs:
## Changes ##
* Add `propagate_error` argument to sequential pipeline.
  * When True, the sequential pipeline propagates quantized outputs (outputs after module optimization/GPTQ apply)
  * When False, the sequential pipeline propagates unquantized outputs (full precision)
  * [GPTAQ](https://arxiv.org/pdf/2504.02692) authors report slightly better accuracy recovery when `propagate_error=False`. This aligns with the preliminary testing I've done myself, but more work will be required to make this a default argument.


brian-dellabetta (approved):
nice! nit -- `batch_idx` is a bit more readable than `b_idx`

kylesayrs:
@brian-dellabetta I agree, but I think `b_idx` is more readable than putting the loop on two separate lines 🙁

dsikka:
@mergifyio refresh

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -104,22 +104,27 @@ def __call__(
...
-                        inputs = activations.fetch(batch_idx, subgraph.input_names)
-                        subgraph.forward(model, **inputs)
+                    for b_idx in tqdm(range(len(dataloader)), desc=calib_desc):
+                        inputs = activations.fetch(b_idx, subgraph.input_names)
+                        outputs = subgraph.forward(model, **inputs)
 
-                    LifecycleCallbacks.sequential_epoch_end(subgraph)
+                        if not dataset_args.propagate_error:
+                            activations.update(b_idx, outputs)
+                            activations.delete(b_idx, subgraph.consumed_names)
```
HDCharles: this seems like it could be moved to the bottom since we have to do it in both cases, does doing it early save memory or something?

HDCharles (approved):
see comment but assuming there's a good reason, going to accept

HDCharles:
@kylesayrs is this ready to land? looks like you haven't touched this since 3 weeks?

kylesayrs:

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs: We cannot move it to the bottom, since by that time the intermediates from the forward pass have been garbage collected.

PR #2012, opened 2025-11-10
[TypeHint] Fix `format_calibration_data` type hint

kylesayrs:


src/llmcompressor/datasets/utils.py
brian-dellabetta, line 141:
```suggestion
    :return: DataLoader containing trimmed calibration samples
```

PR #2014, opened 2025-11-10
[Tests] Update ZP Check

dsikka:
SUMMARY:
- ZP are the only qparam that are initialized using the compressed dtype. Update the test to reflect this.

### DIFF
--- a/tests/llmcompressor/modifiers/calibration/test_observers.py
+++ b/tests/llmcompressor/modifiers/calibration/test_observers.py
@@ -54,7 +54,7 @@ def test_observers_update(shape, group_size, actorder):
 
         assert_alike(updated_scale, getattr(module, f"{location}_scale"))
         # zp params are initialized using the compressed dtype
-        updated_zero_point.shape == getattr(module, f"{location}_zero_point").shape
+        assert updated_zero_point.shape == getattr(module, f"{location}_zero_point").shape
 
 
 def assert_alike(a, b):


tests/llmcompressor/modifiers/calibration/test_observers.py
```diff
@@ -53,7 +53,8 @@ def test_observers_update(shape, group_size, actorder):
         updated_scale, updated_zero_point = observer(value)
 
         assert_alike(updated_scale, getattr(module, f"{location}_scale"))
-        assert_alike(updated_zero_point, getattr(module, f"{location}_zero_point"))
+        # zp params are initialized using the compressed dtype
+        updated_zero_point.shape == getattr(module, f"{location}_zero_point").shape
```
brian-dellabetta: missing an assert?
```suggestion
        assert updated_zero_point.shape == getattr(module, f"{location}_zero_point").shape
```

brian-dellabetta (dismissed):

kylesayrs:
It seems like [this function call](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/utils/helpers.py#L129-L131) should cast to the quantized type, or even better, not do the [zp dequantize step](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_args.py#L415) at all to avoid imprecision issues.

dsikka:
> It seems like [this function call](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/utils/helpers.py#L129-L131) should cast to the quantized type, or even better, not do the [zp dequantize step](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_args.py#L415) at all to avoid imprecision issues.

Whenever we clamp and round to a quantized type, we maintain the dense dtype after casting / rounding to the quantized dtype range. This is how we handle weights and scales. 

Zero-points are the only parameter that are initialized using the quantized dtype. We can add a step to cast to the zp dtype after that call but I dont think that's what the function should be doing.

kylesayrs:
> We can add a step to cast to the zp dtype after that call but I dont think that's what the function should be doing.

As it stands, the zero point will be implicitly cast back to the quantized dtype via the `copy_` operation, so this is effectively what we're doing.

I only mean to suggest ways to make this more explicit (or not do it at all, which would be best) and to maintain the rule that qparams generated by observers always match the qparams that are initialized.

kylesayrs:
> Zero-points are the only parameter that are initialized using the quantized dtype

I agree. I suggest that this fact should be reflected in the observers/calculate_qparams functions as well.

dsikka:
> > Zero-points are the only parameter that are initialized using the quantized dtype
> 
> I agree. I suggest that this fact should be reflected in the observers/calculate_qparams functions as well.

I can change it there.

PR #2015, opened 2025-11-10
[Bugfix] IntermediatesCache nested model inputs

kylesayrs:
## Purpose ##
* Fix bug where nested model inputs were not being properly offloaded by intermediates cache
  * This was caused by this line where the `value` may be nested, but the `IntermediateValue` constructor does not handle nested values

```python3
- values[key] = IntermediateValue(value=value, device=model_device)
+ values[key] = cls._offload_value(value, offload_device, model_device)
```

## Changes ##
* Use `_offload_value` when offloading model inputs from dataloader
  * This necessitates converting `_offload_value` and `_onload_value` to be class methods
  * `_offload_value` accepts an `onload_device` argument to override which device is used when the value in onloaded (in the case of values from the dataloader, this should be the device of the model)

## Testing ##
* Tested with model with nested inputs e2e
* Added additional unit tests to cover this case


brian-dellabetta (approved):
nice!

PR #2018, opened 2025-11-10
[tests] switch lm_eval invocation to use pre-loaded transformers model

brian-dellabetta:
SUMMARY:
`lm_eval==0.4.9.1` has a broken entrypoint when using a model with a compressed-tensors quantization config with `--model hf`:

```
FAILED tests/lmeval/test_lmeval.py::TestLMEval::test_lm_eval[tests/lmeval/configs/vl_w4a16_actorder_weight.yaml] - ValueError: The model is quantized with CompressedTensorsConfig but you are passing a dict config. Please make sure to pass the same quantization config class to `from_pretrained` with different loading attributes.
```

It has been resolved on main, though a separate issue persists that is resolved with this PR -- https://github.com/EleutherAI/lm-evaluation-harness/pull/3393.

While that is in transit, and to avoid having to use lm_eval main in our ci/cd, this PR resolves the issue by pre-loading the model with `AutoModelForCausalLM` rather than relying on lm_eval's strange model loading logic.

TEST PLAN:
tests run now, for some reason the vl test is super slow on ibm-h100-1. The same thing happens on main. I've seen this before, but I'm not sure what's causing it. It seemed to correct itself the following day

### DIFF
--- a/.github/workflows/test-check-transformers.yaml
+++ b/.github/workflows/test-check-transformers.yaml
@@ -97,6 +97,10 @@ jobs:
         if: (success() || failure()) && steps.install.outcome == 'success'
         run: |
           pytest -v tests/llmcompressor/transformers/gptq
+      - name: Running AutoRound Tests
+        if: (success() || failure()) && steps.install.outcome == 'success'
+        run: |
+          pytest -v tests/llmcompressor/transformers/autoround
       - name: Running ONESHOT Tests
         if: (success() || failure()) && steps.install.outcome == 'success'
         run: |
--- a/examples/autoround/llama3_example.py
+++ b/examples/autoround/llama3_example.py
@@ -1,0 +1,56 @@
+from auto_round.calib_dataset import get_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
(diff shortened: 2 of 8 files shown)


tests/lmeval/test_lmeval.py
```diff
@@ -152,11 +153,12 @@ def test_lm_eval(self, test_data_file: str):
     @log_time
     def _eval_base_model(self):
         """Evaluate the base (uncompressed) model."""
-        model_args = {**self.lmeval.model_args, "pretrained": self.model}
-
         results = lm_eval.simple_evaluate(
-            model=self.lmeval.model,
-            model_args=model_args,
+            model=lm_eval.models.huggingface.HFLM(
+                pretrained=AutoModelForCausalLM.from_pretrained(self.model),
```
dsikka: how does this handle VL models?

brian-dellabetta: ah good point, i'll have to update for that. Just need to use the class name in the configs

brian-dellabetta: ok, updated to account for transformers classes other than `AutoModelForCausalLM` and lm_eval model class other than `HFLM`

PR #2030, opened 2025-11-13
Fix Qwen3Next import errors on old transformers versions

fynnsu:
SUMMARY:
Currently, we support `transformers>=4.54.0`, however `Qwen3Next` was only added in `transformers==4.57.0`. This updates the code so that:
1. In `src/llmcompressor/modeling/qwen3_next_moe.py` we only used the imports for typing, so they've been moved into an `if TYPE_CHECKING` block. 
2. In `tests/llmcompressor/modeling/test_calib_qwen3_next.py` we wrap the import statement in a try-except and skip the Qwen3Next test if they fail to import. 

TEST PLAN:
Run CI.

### DIFF
--- a/tests/e2e/e2e_utils.py
+++ b/tests/e2e/e2e_utils.py
@@ -10,15 +10,12 @@ from tests.test_timer.timer_utils import log_time
 from tests.testing_utils import process_dataset
 
 
-@log_time
-def _load_model_and_processor(
-    model: str,
-    model_class: str,
-):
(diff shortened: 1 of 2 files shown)


brian-dellabetta (approved):
LGTM, pending test errors

shanjiaz (approved):

brian-dellabetta (approved):
🚀 thanks for fixing!

shanjiaz (approved):

PR #2032, opened 2025-11-13
Support RHAIIS images for the e2e tests

dhuangnm:
SUMMARY:
Add support to allow the e2e tests run with the RHAIIS images in our CI/CD infrastructure:

- This PR relies on the kuben8s infrastructure setup for our CI/CD. All the deployment related changes are in the llm-compressor-testing PR: https://github.com/neuralmagic/llm-compressor-testing/pull/174
- The e2e tests with the RHAIIS images is only supported when running in our CI/CD so we can only run this mode using [the workflows in the llm-compressor-testing repo](https://github.com/neuralmagic/llm-compressor-testing/actions/workflows/test-image.yml). We don't change the way to run the e2e tests with wheels.
- This PR also added an `rhaiis-e2e-smoke.list` file that lists the models as a smoke test for the e2e tests run with the RHAIIS images. 
- Added a new `run_tests_in_rhaiis.sh` to run the e2e tests with the RHAIIS images.

`
bash tests/e2e/vLLM/run_tests_in_rhaiis.sh -c tests/e2e/vLLM/e2e-smoke.list -t tests/e2e/vLLM/test_vllm.py -s <path to save results etc>
`

A successful run for the e2e tests using the RHAIIS images is here: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/19676158533

TEST PLAN:
All tests


dsikka (changes requested):
I think we want this separate from our normal e2e testing. 

Key things to change:
- We should keep run_tests.sh as is and create a separate rhaiis image testing bash script
- The vLLM test is already written in functions which we should pull in to a separate rhaiis-e2e tets

dhuangnm:
> I think we want this separate from our normal e2e testing.
> 
> Key things to change:
> 
> * We should keep run_tests.sh as is and create a separate rhaiis image testing bash script
> * The vLLM test is already written in functions which we should pull in to a separate rhaiis-e2e tets

This doesn't affect our normal e2e testing. We don't change any existing ways to run the e2e tests, just adding a new way to allow the tests run with RHAIIS images. The changes to the run_tests.sh is to allow a smoke list file be used, it still supports its current way of using the configs folder and nothing is changed there either.

tests/e2e/vLLM/rhaiis-e2e-smoke.list
```diff
@@ -0,0 +1,7 @@
+fp4_nvfp4.yaml
```
dsikka, line 1: rename file to be clear that this is rhaiis specific

tests/e2e/vLLM/run_tests.sh
```diff
@@ -16,8 +16,18 @@ while getopts "c:t:" OPT; do
   esac
```
dsikka, line 16: Create a new rhaiis specific script

tests/e2e/vLLM/test_vllm.py
```diff
@@ -76,18 +83,32 @@ def set_up(self, test_data_file: str):
...
         # GPU memory utilization - only set if explicitly provided in config
         self.gpu_memory_utilization = eval_config.get("gpu_memory_utilization")
-        # vllm python env - if same, use the current python env, otherwise use
-        # the python passed in VLLM_PYTHON_ENV
-        if VLLM_PYTHON_ENV.lower() != "same":
-            self.vllm_env = VLLM_PYTHON_ENV
-        else:
+        if VLLM_PYTHON_ENV.lower() == "same":
             self.vllm_env = sys.executable
+        else:
```
dsikka: I am generally not a fan of the what we're doing here with env. We should separate this into its own rhaiis-specific test file

tests/e2e/vLLM/rhaiis-e2e-smoke.list
dhuangnm: We can actually also use it to run the regular e2e tests if people want to run smaller number of models for the e2e tests. This is not changing any thing now, just providing a flexibility here if you want to run smaller number of models, you can put the list in a file and just provide this file to the run_tests.sh

tests/e2e/vLLM/run_tests.sh
dhuangnm: Same as above.

tests/e2e/vLLM/test_vllm.py
dhuangnm: A lot of the code paths can be shared across the scenarios we want to test the e2e, no matter it's using the vllm installed into the same python env, or a virtualenv, or the rhaiis image. We can definitely separate them out into different scripts, but it's duplicating code and adding more efforts to maintain. I'm fine either way if you have strong preference for one against the other.

tests/e2e/vLLM/test_vllm.py
```diff
@@ -21,8 +21,15 @@
     "TEST_DATA_FILE", "tests/e2e/vLLM/configs/int8_dynamic_per_token.yaml"
 )
 SKIP_HF_UPLOAD = os.environ.get("SKIP_HF_UPLOAD", "")
-# vllm python environment
+# vllm environment: same (default), the path of vllm virtualenv, deployed runner name
 VLLM_PYTHON_ENV = os.environ.get("VLLM_PYTHON_ENV", "same")
+IS_VLLM_IMAGE = False
+RUN_SAVE_DIR = os.environ.get("RUN_SAVE_DIR", "none")
+# when using vllm image, needs to save the generated model
+if VLLM_PYTHON_ENV.lower() != "same" and (not Path(VLLM_PYTHON_ENV).exists()):
```
kylesayrs, line 27: I'm a little confused what the condition is here?

kylesayrs (approved):

tests/e2e/vLLM/test_vllm.py
dhuangnm: This is to check if the provided vllm env is on the same local python env, or in a virtual env (where the path of the virtual env should exist on the same machine), or is a deployed image (which will be a runner name). For the first two cases, the tests will use the python env (either same or separate) on the test machine for vllm. For the third case, it'll use the vllm from the runner which was deployed with the RHAIIS image.

dsikka (commented):
I think we should do the following:

1. Separate the test into two steps: Model Generation (hanlded by `set-up`, `run_oneshot_for_e2e_testing` and uploading to the hub) following by vLLM Inference
2. Run the second step in either a vLLM image or  using vLLM nightly. This will use the models generated from (1) and will allow separtion between upstream and midstream flows

I think most of the changes are in the vLLM side and this is also the most reflective of the actual flow we'd expect when using upstream and RHAIIS.

tests/e2e/vLLM/rhaiis-e2e-smoke.list
dhuangnm: Updated.

tests/e2e/vLLM/run_tests.sh
dhuangnm: Created a new rhaiis specific script: `run_tests_in_rhaiis.sh`, to run the e2e tests in rhaiis images specifically.

tests/e2e/vLLM/test_vllm.py
dhuangnm: Removed the added RUN_SAVE_DIR env var in `test_vllm.py` so no changes here anymore.

dhuangnm:
> I think we should do the following:
> 
> 1. Separate the test into two steps: Model Generation (hanlded by `set-up`, `run_oneshot_for_e2e_testing` and uploading to the hub) following by vLLM Inference
> 2. Run the second step in either a vLLM image or  using vLLM nightly. This will use the models generated from (1) and will allow separtion between upstream and midstream flows
> 
> I think most of the changes are in the vLLM side and this is also the most reflective of the actual flow we'd expect when using upstream and RHAIIS.

Thanks Dipika for the suggestions. Updated the code accordingly:

1. Separated functions in test_vllm.py to do each step for set-up, compress-model, save-compressed-model, and run-vllm;
2. The run-vllm will use either a vLLM image or nightly;
3. Added a new script run_tests_in_rhaiis.sh to run tests in rhaiis images specifically.
4. Removed the originally added RUN_SAVE_DIR env var and took advantage of the save_dir we already have in the model config files.

Please review again.

tests/e2e/vLLM/rhaiis-e2e-smoke.list
dsikka, line 1: My only fear for this is that the location may change and this may break but we can deal with that later.

tests/e2e/vLLM/run_tests_in_rhaiis.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
dsikka, line 1: can we also update the `run_tests.sh` to be `run_test_in_workflow.sh` or something as well to help further differentiate between the two?

dhuangnm:
> This looks much better! Can we kick off a custom run to make sure this works e2e as expected? Otherwise, LGTM.
> 
> Great work

Thanks Dipika! Yes a successful run from the PR is here: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/19676158533

tests/e2e/vLLM/rhaiis-e2e-smoke.list
dhuangnm: Agreed, I can work on it in a follow-up PR to maybe add a new field in the config file to indicate if the config should be run in the smoke tests or not. The work can be in its own PR.

tests/e2e/vLLM/run_tests_in_rhaiis.sh
dhuangnm: Yep, renamed `run_tests.sh` to `run_tests_in_python.sh` to differentiate if the tests run in python or in rhaiis images.

dhuangnm:
@dsikka thanks for the comments! I updated the PR again. Please review.

tests/lmeval/test_lmeval.py
```diff
@@ -49,7 +44,7 @@ class LmEvalConfig(BaseModel):
 TIMINGS_DIR = os.environ.get("TIMINGS_DIR", "timings/lm-eval")
 
 
-# Will run each test case in its own process through run_tests.sh
+# Will run each test case in its own process through run_tests_in_python.sh
```
dsikka, line 52: Will we need to change llmcompressor-testing to make sure this new script is used during e2e tests?

dhuangnm: Yep the PR to update the workflows is here: https://github.com/neuralmagic/llm-compressor-testing/pull/174

kylesayrs (approved):
Please make sure that the runner that runs in the case of `IS_VLLM_IMAGE` has the permissions to modify the vllm bash script and run `kubectl`

dhuangnm:
> Please make sure that the runner that runs in the case of `IS_VLLM_IMAGE` has the permissions to modify the vllm bash script and run `kubectl`

Yep we have run the PR on the runners and the runners do have permissions to modify the vllm bash script and run `kubectl`.

PR #2040, opened 2025-11-17
Add handling for Qwen3VLMoe on older transformers versions

fynnsu:
SUMMARY:
#1981 added Qwen3VLMoe with associated tests, however this model isn't available on all transformers versions that we support. Therefore, (similar to #2030) this pr ensures we don't import or test the model when using a transformers version that doesn't support it. 

TEST PLAN:
Confirmed that this change fixes `import llmcompressor` when using oldest support transformers version `4.54.0`. Ran test with old transformers version (test gets skipped) and new transformers version (test passes).

### DIFF
--- a/experimental/README.md
+++ b/experimental/README.md
@@ -1,0 +1,3 @@
+# Experimental Features
+
+This folder aims to highlight features that are a work-in-progress or are supported in LLM Compressor and/or Compressed-Tensors but lack full support in downstream libraries like vLLM.
--- a/experimental/mxfp4/llama3_mxfp4.py
+++ b/experimental/mxfp4/llama3_mxfp4.py
@@ -1,0 +1,35 @@
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+from llmcompressor.utils import dispatch_for_generation
+
+MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
+
+# Load model.
+model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
+
+# Configure the quantization algorithm and scheme.
+# In this case, we:
+#   * quantize the weights to fp4 with per group 32 via ptq
+recipe = QuantizationModifier(targets="Linear", scheme="MXFP4", ignore=["lm_head"])
+
+# Apply quantization.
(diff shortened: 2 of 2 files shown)


src/llmcompressor/modeling/qwen3_vl_moe.py
```diff
@@ -19,7 +25,7 @@ class CalibrateQwen3VLMoeTextSparseMoeBlock(MoECalibrationModule):
 
     def __init__(
         self,
-        original: OriginalQwen3VLMoeTextSparseMoeBlock,
+        original: Qwen3VLMoeTextSparseMoeBlock,
```
kylesayrs:
```suggestion
        original: "Qwen3VLMoeTextSparseMoeBlock",
```

src/llmcompressor/modeling/qwen3_vl_moe.py
```diff
@@ -116,7 +122,7 @@ def __init__(self, config, original):
 
 def replace(
     config: Qwen3VLMoeConfig,
-    original: OriginalQwen3VLMoeTextSparseMoeBlock,
+    original: Qwen3VLMoeTextSparseMoeBlock,
```
kylesayrs:
```suggestion
    original: "Qwen3VLMoeTextSparseMoeBlock",
```

tests/llmcompressor/modeling/test_calib_qwen3_vl_moe.py
```diff
@@ -1,14 +1,24 @@
...
 from tests.testing_utils import requires_gpu
 
+try:
+    from transformers import Qwen3VLMoeConfig
+    from transformers.models.qwen3_vl_moe.modeling_qwen3_vl_moe import (
+        Qwen3VLMoeTextSparseMoeBlock,
+    )
+except ImportError:
+    Qwen3VLMoeConfig = None
+    Qwen3VLMoeTextSparseMoeBlock = None
```
kylesayrs: use pytest.importorskip

src/llmcompressor/modeling/qwen3_vl_moe.py
fynnsu: This isn't required with the `from __future__ import annotations` line at the top. Do we prefer this style?

kylesayrs: I think this style is due to be deprecated/removed in the future
https://peps.python.org/pep-0749/#the-future-of-from-future-import-annotations
https://docs.python.org/3/library/__future__.html#id1

kylesayrs (approved):

PR #2044, opened 2025-11-18
[Deprecation] Remove Training Support

dsikka:
SUMMARY:
- Remove all training support
- Remove the distillation modifier and any related parsing
- Move the `data` folder which was under `finetune` to be under the `transformers` folder
- Extend base and pytorch testing to run on 3.10 and 3.13
- Update transformers tests to run on Python 3.12
- Updated `examples/quantization_2of4_sparse_w4a16.py`
- Update the `entrypoints` README.md

Testing:
- Nightly passes


src/llmcompressor/args/dataset_arguments.py
```diff
@@ -78,8 +78,8 @@ class CustomDatasetArguments(DVCDatasetArguments):
 @dataclass
 class DatasetArguments(CustomDatasetArguments):
```
kylesayrs, line 79: Other arguments we should consider removing/deprecating are
* concatenate_data
* splits
* num_calibration_samples
* pad_to_max_length
* min_tokens_per_module (separate from training, but no longer used)

kylesayrs:
Also, we should get rid of training callbacks and events (later)

dsikka:
> Also, we should get rid of training callbacks and events (later)

I thought I got them? Can you link what I missed

HDCharles (approved):
The greatest achievements are born from the greatest sacrifice

src/llmcompressor/args/dataset_arguments.py
dsikka: I'll take a note and do in a follow-up

rahul-tuli:
🔥

PR #2051, opened 2025-11-19
Add: File Based Caching for `lm_eval` tests

rahul-tuli:
# LM Eval Caching System

## Overview

Caches base model evaluation results to speed up tests by ~30-50%. When multiple quantization tests share the same base model and evaluation parameters, cached results are reused instead of re-evaluating.

### Enable (Default)
```bash
pytest tests/lmeval/test_lmeval.py
```

First run caches results, subsequent runs use cached results for matching parameters.

### Disable
```bash
DISABLE_LMEVAL_CACHE=1 pytest tests/lmeval/test_lmeval.py
```

### Custom Cache Location
```bash
LMEVAL_CACHE_DIR=/tmp/cache pytest tests/lmeval/test_lmeval.py
```

## How It Works

Results are cached based on:
- Model ID (e.g., `TinyLlama/TinyLlama-1.1B-Chat-v1.0`)
- Task (e.g., `gsm8k`)
- Few-shot count
- Sample limit
- Batch size
- Model arguments (hashed)

### Cache Storage

Single CSV file: `.lmeval_cache/cache.csv`

```csv
model,task,num_fewshot,limit,batch_size,model_args_hash,result
TinyLlama/TinyLlama-1.1B-Chat-v1.0,gsm8k,5,1000,100,abc123def456,"{'results': {...}}"
```

## Usage in Tests

```python
from tests.testing_utils import cached_lm_eval_run

class TestLMEval:
    @cached_lm_eval_run
    def _eval_base_model(self) -> dict:
        return self._eval_model(self.model)
```

### Inspect Cache
```bash
cat .lmeval_cache/cache.csv

# Or with pandas
python -c "import pandas as pd; print(pd.read_csv('.lmeval_cache/cache.csv'))"
```

### Clear Cache
```bash
rm -rf .lmeval_cache/
```

### Ignore in Git
Already added to `.gitignore`:
```gitignore
.lmeval_cache/
```

## Environment Variables

| Variable | Values | Default | Description |
|----------|--------|---------|-------------|
| `DISABLE_LMEVAL_CACHE` | `1`, `true`, `yes` | disabled | Disable caching |
| `LMEVAL_CACHE_DIR` | path | `.lmeval_cache` | Cache directory |

## Performance Example

**Without cache:** 175s total
```
Test 1 (W4A16): Base (60s) + Compressed (30s) = 90s
Test 2 (W8A8):  Base (60s) + Compressed (25s) = 85s
```

**With cache:** 115s total (34% faster)
```
Test 1 (W4A16): Base (60s) + Compressed (30s) = 90s  [MISS]
Test 2 (W8A8):  Base (0.1s) + Compressed (25s) = 25s [HIT]
```

## Example Cache Logs

When running tests, you'll see these log messages:

**First test (cache miss):**
```
LM-Eval cache MISS: meta-llama/Meta-Llama-3-8B-Instruct/gsm8k
# ... evaluation runs ...
LM-Eval cache WRITE: meta-llama/Meta-Llama-3-8B-Instruct/gsm8k
```

**Second test with same base model (cache hit):**
```
LM-Eval cache HIT: meta-llama/Meta-Llama-3-8B-Instruct/gsm8k
# ... evaluation skipped, cached result returned ...
```

### Testing the Cache

Run two tests that share the same base model:

```bash
# Clean cache
rm -rf .lmeval_cache/

# Test 1: FP8_DYNAMIC scheme - Cache MISS
CUDA_VISIBLE_DEVICES=5 CADENCE=weekly \
TEST_DATA_FILE=tests/lmeval/configs/fp8_dynamic_per_token.yaml \
.venv/bin/python -m pytest tests/lmeval/test_lmeval.py::TestLMEval::test_lm_eval -v

# Test 2: FP8 static scheme (same base model) - Cache HIT
CUDA_VISIBLE_DEVICES=5 CADENCE=weekly \
TEST_DATA_FILE=tests/lmeval/configs/fp8_static_per_tensor.yaml \
.venv/bin/python -m pytest tests/lmeval/test_lmeval.py::TestLMEval::test_lm_eval -v

# Inspect cache
cat .lmeval_cache/cache.csv
```

Expected output shows the second test completes much faster due to cached base model evaluation.

### Cache Not Working

Check environment variable:
```bash
echo $DISABLE_LMEVAL_CACHE  # Should be empty
```

Verify cache file exists:
```bash
ls -lh .lmeval_cache/cache.csv
```

### Cache Always Misses

Ensure exact parameter match:
- Model ID (case-sensitive)
- Task name
- All numeric parameters (fewshot, limit, batch_size)
- Model arguments

### Corrupted Cache

Simply delete and recreate:
```bash
rm .lmeval_cache/cache.csv
pytest tests/lmeval/test_lmeval.py
```

### GitHub Actions
```yaml
- name: Run LM Eval tests with cache
  env:
    LMEVAL_CACHE_DIR: ${{ runner.temp }}/lm_cache
  run: pytest tests/lmeval/test_lmeval.py
```

### Why CSV?
- Matches existing timing data format (`timings/*.csv`)
- Simple pandas integration

### Error Handling
Failures are logged but don't break tests - cache simply falls back to uncached execution on any error.

Rebased version of #1900


tests/testing_utils.py
```diff
@@ -292,3 +309,123 @@ def requires_cadence(cadence: Union[str, List[str]]) -> Callable:
...
+class LMEvalCacheKey:
+    """Cache key for LM Eval results based on evaluation parameters."""
+
+    model: str
+    task: str
+    num_fewshot: int
+    limit: int
+    batch_size: int
+    model_args_hash: str
+
```
HDCharles, line 326: should we also record lm_eval version?

tests/testing_utils.py
```diff
@@ -292,3 +309,123 @@ def requires_cadence(cadence: Union[str, List[str]]) -> Callable:
...
+@dataclass(frozen=True)
+class LMEvalCacheKey:
+    """Cache key for LM Eval results based on evaluation parameters."""
+
+    model: str
+    task: str
+    num_fewshot: int
+    limit: int
+    batch_size: int
+    model_args_hash: str
```
HDCharles, line 323: looks like we also use a random seed specified by the config to try to make the test deterministic/reproducible, should that also be cached?

tests/testing_utils.py
rahul-tuli: Good callout, added!

tests/testing_utils.py
rahul-tuli: Added!

shanjiaz (approved):
Niiiiiiice

tests/lmeval/test_lmeval.py
```diff
@@ -161,8 +161,9 @@ def test_lm_eval(self, test_data_file: str):
         self.tear_down()
 
     @log_time
+    @cached_lm_eval_run
```
dsikka, line 164: very nice

PR #2054, opened 2025-11-20
[Performance] Batched calibration

kylesayrs:
## Purpose ##
* Reduce calibration runtime by providing users with options to increase performance
  * `batch_size` controls the batch size of calibration data
  * `offload_sequential_activations` controls whether calibration data is offloaded to the CPU between layers

## Prerequisites ##
* #2080
* #2081

### Batched Calibration ###
* Add `batch_size` argument
* Change `data_collator` default from the default data collator to a `"truncation"` collator
* The `data_collator_with_truncation` function truncates all samples to the shortest length sample in the batch.
  * Statistics about how many tokens are dropped using this method are in the tables below
  * The data collator can also be changed to "padding" instead to pad to the longest length sample in the batch
* In order to reduce the amount of excess truncation/padding, default to `LengthAwareSampler` which samples from the dataset such that samples with similar batch lengths are batched together

## Runtimes ##
These runtimes are for LLama8B with GPTQ

Batch Size | Time | % Speedup | % Deleted
-- | -- | -- | --
Original (1) | 11m17 | N/A | 0.0
1 | 11m17 | 0.0 | 0.0
2 | 10m48 | 4.2 | 0.2
4 | 10m39 | 5.6 | 0.5
8 | 10m39 | 5.6 | 1.1
16 | 10m58 | 2.8 | 2.6
64 | 11m4 | 11.2 | 12.0
128 | 9m29 | 16.0 | 23.9
512 | 7m39 | 37.3 | 75.3

* The speedup is relatively meager up until you start deleting significant portions of the dataset via truncation

### Disable Offloading ###
* Add `offload_sequential_activations` argument, defaults to True (no behavior change)
  * Enabling this option increases throughput but also increases memory usage

Batch Size | Time | % Speedup | % Deleted
-- | -- | -- | --
Original (1) | 11m17 | N/A | 0.0
1 | 10m14 | 9.3 | 0.0
2 | 9m46 | 13.4 | 0.2
4 | 9m36 | 14.9 | 0.5
8 | 9m48 | 13.1 | 1.1
16 | 9m26 | 16.3 | 2.6
32 | 9m27 | 16.2 | 5.8
128 | 8m34 | 24.0 | 23.9
512 | 6m40 | 40.9 | 75.3

* Memory requirement for 512 samples on Llama 8B is ~70Gb, which is equivalent to batch size 128
* With this option enabled and batch size 32, calibration runtime is less than 1s per layer (down from ~11s)
  * This implies that the theoretical maximum speedup from reducing calibration time alone is ~15% for this model + dataset

### Misc ###
* Fix examples
  * Fixed examples where there's issues between model dtypes and processor dtypes (Mixtral, Pixtral, Whisper)
  * For multimodal models which use multimodal datasets, remove their data collators, as the batch unwrapping is now done by  the`TextGenerationDataset`
* Remove `_mask_padding` from `IntermediatesCache`, as I do not believe that this method is effective in masking padding tokens from hessian calculations
* Fix AWQ
  * AWQ was hard coded to handle only batches of size 1

### Evaluation Regression ###
Batch Size | Eval Score | Difference | % Deleted
-- | -- | -- | --
Original (1) | 0.6573 | 0.000 | 0.0
1 | 0.6513 | -0.6 | 0.0
2 | 0.6513 | -0.6 | 0.2
4 | 0.6657 | +0.8 | 0.5
8 | 0.6513 | -0.6 | 1.1
16 | 0.6672 | +1.0 | 2.6
64 | 0.6338 | -2.4 | 12.0
128 | 0.6603 | +0.3 | 23.9
512 | 0.6391 | -1.8 | 75.3

Deleting significant portions of the dataset (delete longer sequences first) has a detrimental effect on recovery

### Modifiers ###
* GPTQ
  * Ran full regression tests, as shown above
* AWQ
  * Ran AWQ with batch size 32 and checked output sanity
* Quantization Modifier
  * Ran NVFP4 with batch size 10 and checked output sanity

### Calibration Regression Testing ###
I ran calibration for the following models (but did not evaluate recovery)

The following model examples can calibrate without issue:
* Llama3
* Gemma3
* Internvl3
* Mllama
* Llama4

The following models had a bug where processor and model dtypes were mismatched, but are now fixed by this PR:
* Mistral3
* Pixtral
* Whisper

The following models have an accelerate device offloading bug:
* Idefics3
* Phi3 Vision

The following model examples have an MoE replacement bug:
* qwen3-vl-30b-a3b-Instruct

## Future Work ##
While these options are a great place to start, the next step to improve runtime is to allow multi-GPU compression, likely via torch.distributed tensor parallelism

### DIFF
--- a/src/llmcompressor/utils/helpers.py
+++ b/src/llmcompressor/utils/helpers.py
@@ -1119,7 +1119,13 @@ def disable_lm_head(model: torch.nn.Module):
         def dummy_forward(self, input: torch.Tensor) -> torch.Tensor:
             return input.to("meta") @ dummy_weight.T
 
-        with patch_attr(lm_head, "forward", dummy_forward.__get__(lm_head)):
+        with contextlib.ExitStack() as stack:
+            lm_head_forward = dummy_forward.__get__(lm_head)
+            stack.enter_context(patch_attr(lm_head, "forward", lm_head_forward))
+
+            if hasattr(model, "_hf_hook"):
+                stack.enter_context(patch_attr(model._hf_hook, "io_same_device", False))
+
             yield
 
 


src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -317,8 +319,7 @@ def oneshot(
         max_seq_length.
     :param streaming: True to stream data from a cloud dataset.
     :param overwrite_cache: Whether to overwrite the cached preprocessed datasets.
```
HDCharles, line 330: batchsize docsting?

HDCharles (dismissed):
looks good aside from the missing docstring

src/llmcompressor/entrypoints/oneshot.py
kylesayrs: Thanks for the catch.

HDCharles (approved):

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -69,9 +67,27 @@ class CustomDatasetArguments(DVCDatasetArguments):
...
-    data_collator: Callable[[Any], Any] = field(
-        default_factory=lambda: DefaultDataCollator(),
-        metadata={"help": "The function to used to form a batch from the dataset"},
+    batch_size: int = field(
+        default=1,
+        metadata={
+            "help": (
+                "Calibration batch size. During calibration, LLM Compressor disables "
+                "lm_head output computations to reduce memory usage from large "
+                "calibration batch sizes. Large batch sizes may result excess padding "
```
dsikka: nit: `result in`

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -126,8 +142,8 @@ class DatasetArguments(CustomDatasetArguments):
         default=512,
         metadata={"help": "Number of samples to use for one-shot calibration"},
     )
-    shuffle_calibration_samples: bool | None = field(
-        default=True,
+    shuffle_calibration_samples: bool = field(
+        default=False,
```
dsikka, line 146: Why False by default?

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -214,6 +230,14 @@ class DatasetArguments(CustomDatasetArguments):
             "definition"
         },
     )
+    offload_sequential_activations: bool = field(
```
dsikka, line 233: Iirc, we have always been offloading activation for gptq but not awq?
We're enabling this for all modifiers now, including awq?

src/llmcompressor/datasets/utils.py
```diff
@@ -213,3 +171,116 @@ def make_dataset_splits(
...
+            continue
+
+        min_len = min(len(feature[key]) for feature in features)
+        for feature in features:
+            feature[key] = feature[key][:min_len]
+
+    return default_data_collator(features, return_tensors)
+
+
+class LengthAwareSampler(Sampler[int]):
```
dsikka, line 254: docstring

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -413,7 +413,7 @@ def cache_smooth_activations_hook(
             ):
                 self._smooth_activation_means[smooth_name] = _accumulate_mean(
                     # Assume that first argument is the input
-                    args[0].cpu().abs().detach().squeeze(),
+                    args[0].cpu().abs().detach().flatten(0, -2),
```
dsikka, line 405: why do we need to change?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -90,7 +89,11 @@ def __call__(
                 stack.enter_context(DisableQuantization(model))
 
             # prepare intermediates cache
-            activations = IntermediatesCache.from_dataloader(dataloader, model_device)
+            cache_offload = dataset_args.offload_sequential_activations
```
dsikka, line 92: Doesn't this mean awq will now offload activations, something Brian identified as a potential improvement for the modifier?

src/llmcompressor/args/dataset_arguments.py
kylesayrs:
```
logger.warning(
    "Shuffling the dataset can lead to unoptimal batching for sequence "
    "lengths of non-uniform sizes. When collating with truncation, this will "
    "delete a large number of tokens. When collating with padding, this "
    "will add a large number of padding tokens"
)
```

src/llmcompressor/args/dataset_arguments.py
kylesayrs: AWQ's activation offloading is entirely separate from the activation offloading that is done by the sequential pipeline.

AWQ caches and offloads Attention/MLP activations for performing MSE calculations.
The Sequential Pipeline caches and offloads activations in between layers

src/llmcompressor/modifiers/awq/base.py
kylesayrs: `squeeze()` is almost always bad practice, as it assumes that the outer dimensions are `1`.

In this case, this code was written with the assumption that the batch dim was `1`. However, now that we support batched calibration, the proper behavior is to flatten batch dim and sequence dim together. More generally, this means flattening all dimensions except the last (hidden size) dimension.

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs: AWQ does not share an activation cache with the sequential pipeline. Technically they could be shared in some cases, but not generically, since AWQ typically caches Attention/MLP activations and the sequential pipeline caches layer activations.

dsikka (approved):

dsikka (dismissed):

PR #2061, opened 2025-11-20
[bugfix] saving model without model.name_or_path

HDCharles:
Summary:

the way models were saved assumed that model.name_or_path was populated but this is not always the case. I think that if you generate a model and then immediately quantize it, you'll run into this problem.

This PR relaxes this assumption so that the model can be saved.

TEST PLAN:
ran https://gist.github.com/HDCharles/78c55da388076aa046ee893edfbd9df5

previously you'd get:

huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.

now:

no error

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Big updates have landed in LLM Compressor! To get a more in-depth look, check ou
 
 Some of the exciting new features include:
 
+* **AutoRound Quantization Support**: Added [`AutoRoundModifier`](examples/autoround/llama3_example.py) for quantization using [AutoRound](https://aclanthology.org/2024.findings-emnlp.662.pdf), an advanced post-training algorithm that optimizes rounding and clipping ranges through sign-gradient descent. This approach combines the efficiency of post-training quantization with the adaptability of parameter tuning, delivering robust compression for large language models while maintaining strong performance.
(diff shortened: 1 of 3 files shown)


src/llmcompressor/transformers/utils/helpers.py
```diff
@@ -57,7 +57,12 @@ def infer_recipe_from_model_path(model_path: Union[str, Path]) -> Optional[str]:
         - Hugging face model ID
     :return: The path to the recipe file if found, None otherwise.
     """
-    model_path = model_path.as_posix() if isinstance(model_path, Path) else model_path
+    model_path = (
+        model_path.as_posix() if isinstance(model_path, Path) else model_path.strip()
+    )
+    if model_path == "":
```
kylesayrs, line 63: Is this not already covered by the line below it?

```python3
if os.path.isdir(model_path) or os.path.isfile(model_path):
```

If not, can you put it on the same line? That way we reduce the number of if statements and get to share logic for logger messages

HDCharles: no it doesn't cover it. It doesn't go into that section since its not a path or file.

I also don't think unifying if statements is good here since each is handling a specific case.

you seem to be suggesting something like

```python
    if os.path.isdir(model_path) or os.path.isfile(model_path) or model_path.strip()=="":
        # Model path is a local path to the model directory or file
        model_path = (
            os.path.dirname(model_path) if os.path.isfile(model_path) else model_path
        )
        recipe = os.path.join(model_path, RECIPE_FILE_NAME)

        if os.path.isfile(recipe):
            logger.info(f"Found recipe in the model_path: {recipe}")
            return recipe
        logger.debug(f"No recipe found in the model_path: {model_path}")
        return None
```


but i'm not sure if that's better because this code is explicitly for when we detect a valid path/file. There are other cases like model_path='nm-testing/valid_model' which don't go in that path and do other things. So it seems like explicit handling for the case of 'not valid path and not valid recipe' should be separate too.

high level, the problem is line 93, 
what it should do: check if somehting is a valid model and if not, log and continue
what it does: makes a request to hf hub which returns an error that stops everything.

I initially fixed this with a try except but @fynnsu didn't like that and wanted to just handle the simpler case by detecting "" and exiting the logic.

PR #2070, opened 2025-11-25
[AWQ] use match_modules_set and fix logic

HDCharles:
Depends on https://github.com/vllm-project/compressed-tensors/pull/524

Summary:
- modified AWQ _set_resolved_mappings
  -  get smoothing and balance layers at same time using match_modules_set
  -  (bugfix) correct logic so that if any balance layers are incompatible, that matching is skipped
  -  added warnings
  -  get rid of tqdm and skip counting @kylesayrs 
  -  added helper for module_to_name
  -  remove hardcoded handling for single balance layer by updating get_lowest_common_module to handle that
- modified SmoothQuant _resolve_mappings
  - brought into alignment with AWQ
    - this is largely a horizontal move though there is handling for situations that would have been missed before like 
      - multiple smooth layer matches in a single set 
      - parent contexts further than 1 layer away.
  - updated mapping definitions to always be tuple(list[str],str) which is always the case but wasn't required unlike in AWQ
- removed get_lowest_common_parent
  - now we can use CT's get_lowest_common_ancestor_name so only need to check for module_list (it has a lot of bugfixes compared to the get_lowest_common_parent implementation in LLMC)
- updated test_base for AWQ and smoothquant
  - added test case for _set_resolved_mappings to check that partially skipped matches are handled correctly
  - added tests for MoE matching being handled correctly
  - added test cases for get_lowest_non_module_list_ancestor
  - imported Linear and used that instead of torch.nn.Linear
- reverted test_pytorch.py for logarithmic_equalizations and smoothquant
  - The test was updated in https://github.com/vllm-project/llm-compressor/pull/2084 by @rahul-tuli to ignore some modules but in general because of the way the new logic works, you need to ignore the whole set.
  - if you only ignore one element the matching logic would need to determine whether there's a full set or not *somehow* which it doesn't do. In the previous logic, this was possible because it was assumed the whole set had to be siblings of the smooth_layer, but the new util is trying to be more flexible and so relaxes this assumption which prevents the same approach from working. If this is a common need, perhaps we can add a util that checks for a context parent context of size N or something. 

TEST PLAN:
pytest /home/HDCharles/repos/llm-compressor/tests/llmcompressor/modifiers/awq/test_base.py
pytest /home/HDCharles/repos/llm-compressor/tests/llmcompressor/modifiers/smoothquant/test_base.py

### DIFF
--- a/src/llmcompressor/modeling/llama4.py
+++ b/src/llmcompressor/modeling/llama4.py
@@ -48,26 +48,36 @@ class SequentialLlama4TextMoe(MoECalibrationModule):
 
     def forward(self, hidden_states: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
         hidden_states = hidden_states.reshape(-1, self.hidden_dim)
-        router_scores, router_logits = self.router(hidden_states)  # transformers>=4.54
-
+        router_scores, router_logits = self.router(hidden_states)
         out = self.shared_expert(hidden_states)
 
-        for expert_index in range(self.num_experts):
-            # find expert scores
-            expert_score = router_scores[:, expert_index].unsqueeze(-1)
-            top_token_mask = expert_score[:, 0] > 0
+        _, router_indices = torch.topk(router_logits, self.top_k, dim=1)
+        expert_mask = torch.nn.functional.one_hot(
+            router_indices, num_classes=self.num_experts
+        ).permute(2, 1, 0)  #  (num_experts, top_k, batch_size * sequence_length)
 
-            # llama4 applies scores before expert relu
-            expert_in = hidden_states * expert_score
+        for i in range(self.num_experts):
+            # fetch relevant token indices for this expert
+            token_idx = torch.where(expert_mask[i].squeeze(0))
 
-            # calibrate experts
+            # Original Llama4 definition - apply score to hidden states
+            # before applying to expert this results in NaNs during calibration
+            # routed_in = hidden_states * router_scores[:, i].reshape(-1, 1)
+
             if self.calibrate_all_experts:
-                expert_out = self.experts[expert_index](expert_in)[top_token_mask]
+                # all tokens for this expert
+                expert_out = self.experts[i](hidden_states)[token_idx]
             else:
-                expert_out = self.experts[expert_index](expert_in[top_token_mask])
+                # only relevant tokens for this expert
+                expert_out = self.experts[i](hidden_states[token_idx])
 
-            # accumulate output
-            out[top_token_mask] += expert_out
+            if len(token_idx) > 0:
+                # Deviation from original Llama4 definition to avoid NaNs
+                # NaNs during calibration
+                weighted_output = expert_out * router_scores[:, i][token_idx].reshape(
+                    -1, 1
+                )
+                out[token_idx] += weighted_output
 
         return out, router_logits
 
--- a/tests/llmcompressor/modeling/test_calib_llama4.py
+++ b/tests/llmcompressor/modeling/test_calib_llama4.py
@@ -85,11 +85,11 @@ def test_calib_llama4_module():
     module = SequentialLlama4TextMoe(original, config, calibrate_all_experts=True)
     with calibration_forward_context(module):
         out, router_logits = module(sample)
-        assert torch.nn.functional.mse_loss(true_out, out) < 1e-10
-        assert torch.nn.functional.mse_loss(true_router_logits, router_logits) < 1e-10
+        assert torch.nn.functional.mse_loss(true_out, out) < 0.1
+        assert torch.nn.functional.mse_loss(true_router_logits, router_logits) < 0.1
 
     module = SequentialLlama4TextMoe(original, config, calibrate_all_experts=False)
     with calibration_forward_context(module):
         out, router_logits = module(sample)
-        assert torch.nn.functional.mse_loss(true_out, out) < 1e-10
-        assert torch.nn.functional.mse_loss(true_router_logits, router_logits) < 1e-10
+        assert torch.nn.functional.mse_loss(true_out, out) < 0.1
+        assert torch.nn.functional.mse_loss(true_router_logits, router_logits) < 0.1


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -312,68 +313,76 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
 
-            for smooth_name, smooth_layer in (
-                pbar := tqdm(
-                    match_named_modules(model, [mapping.smooth_layer], self.ignore)
-                )
+            # Use match_modules_set to find coherent sets of modules
+            target_patterns = (mapping.smooth_layer, *mapping.balance_layers)
+
+            for modules_set in (
+                pbar := tqdm(match_modules_set(model, target_patterns, self.ignore))
```
kylesayrs: I don't think a tqdm is necessary here? It shouldn't be an intensive process

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -312,68 +313,76 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
             num_skipped_mappings = 0
 
-            for smooth_name, smooth_layer in (
-                pbar := tqdm(
-                    match_named_modules(model, [mapping.smooth_layer], self.ignore)
-                )
+            # Use match_modules_set to find coherent sets of modules
+            target_patterns = (mapping.smooth_layer, *mapping.balance_layers)
+
+            for modules_set in (
```
```suggestion
            for smooth_layer, *balance_layers in (
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -312,68 +313,76 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
-                                    != balance_layer.in_features
-                                )
-                                or (
-                                    smooth_name.endswith(".qkv_proj")
-                                    and smooth_layer.out_features
-                                    != 3 * balance_layer.in_features
-                                )
+                # Get names using the pre-built mapping
+                smooth_name = module_to_name.get(smooth_layer)
+                if smooth_name is None:
```
kylesayrs: Under what circumstances would this value be `None`? Same with `balance_name`. I'm not sure how it would fail, but if it does, it should be a loud failure, not a silent skip, right?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -312,68 +313,76 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                                and smooth_layer.out_features
+                                != 3 * balance_layer.in_features
                             )
-                        ):
-                            num_skipped_mappings += 1
-                            continue
+                        )
+                    ):
+                        num_skipped_mappings += 1
+                        continue
```
kylesayrs: Can we replace and generalize this to
```python3
if any(smooth_layer.weight.size(0) % layer.weight.size(-1) != 0 for layer in balance_layers):
    continue
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -312,68 +313,76 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
-        For each activation in the mapping list, we find the corresponding weight to
-        balance by searching for the longest substring. For instance, if our balance
-        weight is ".*re:.*q_proj" and the activation is "re:.*self_attn_layer_norm" we
-        would match model.layer.0.p_proj to model.layer.0.self_attn_layer_norm and
-        repeat for model.layer.1 and so on
+        Uses match_modules_set to find coherent sets of (smooth_layer, *balance_layers)
+        that belong together in the model architecture.
         """
+        # Build a module-to-name mapping for efficient lookups
+        module_to_name = {module: name for name, module in model.named_modules()}
```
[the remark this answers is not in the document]
kylesayrs: This is probably fair, although a real edge case. The real solution would be to return names with something like `match_named_modules_set`, although i think what we have here is good for now.

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -312,68 +313,76 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+            # Use match_modules_set to find coherent sets of modules
+            target_patterns = (mapping.smooth_layer, *mapping.balance_layers)
+
+            for modules_set in (
+                pbar := tqdm(match_modules_set(model, target_patterns, self.ignore))
             ):
                 pbar.set_description(
                     f"Resolving mapping {mapping_idx+1}/{len(self.mappings)}"
                     f" ({num_skipped_mappings} skipped)"
                 )
```
kylesayrs: I don't think this is really necessary?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -312,68 +313,76 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                    balance_names.append(balance_name)
 
                 if len(balance_layers) == 0:
                     continue
 
-                elif len(balance_layers) == 1:
+                if len(balance_layers) == 1:
                     # for single balance layer, parent is the balance layer
-                    parent_name, parent = balance_name, balance_layer
+                    parent_name, parent = balance_names[0], balance_layers[0]
```
kylesayrs: This should be handled implicitly by `get_lowest_common_parent`, right?

src/llmcompressor/modifiers/awq/base.py
HDCharles: doesn't that not work for the qkv_proj? I think mlp layers can do similar with [gate_up_proj](https://github.com/huggingface/transformers/blob/06d52fe106dc6922815dc25bfa156a9d78b16959/src/transformers/models/llama4/modeling_llama4.py#L75-L76) too.

src/llmcompressor/modifiers/awq/base.py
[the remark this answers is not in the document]
HDCharles: i'll errorcheck since that util doesn't exist

kylesayrs (dismissed):
🔥

src/llmcompressor/modifiers/awq/base.py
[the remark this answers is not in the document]
HDCharles: added warning to check this

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,64 +320,48 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
 
-            for smooth_name, smooth_layer in (
-                pbar := tqdm(
-                    match_named_modules(model, [mapping.smooth_layer], self.ignore)
+        module_to_name = {}
+        for name, module in model.named_modules():
+            if module in module_to_name:
+                logger.info(
+                    f"Warning, {name} and {module_to_name[module]} both "
+                    "share the same module the same module, "
```
fynnsu: typo

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,64 +320,48 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                    )
 
-                if len(balance_layers) == 0:
                     continue
-
-                elif len(balance_layers) == 1:
-                    # for single balance layer, parent is the balance layer
-                    parent_name, parent = balance_name, balance_layer
                 else:
                     # for multiple balance layers, find lowest common parent
```
fynnsu: Unnecessary `else:` since the `if` body always calls `continue`

fynnsu (dismissed):
Looks good!

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,73 +325,57 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
                     continue
 
-                elif len(balance_layers) == 1:
-                    # for single balance layer, parent is the balance layer
-                    parent_name, parent = balance_name, balance_layer
-                else:
-                    # for multiple balance layers, find lowest common parent
-                    parent_name, parent = get_lowest_common_parent(balance_names, model)
+                ancestor_name = get_lowest_common_ancestor_name(balance_names)
+                ancestor_name, ancestor = get_lowest_non_module_list_ancestor(
```
kylesayrs: Consider writing as
```python3
while not isinstance((ancestor := model.get_submodule(ancestor_name)), ModuleList):
    ancestor_name = ancestor_name.rsplit(".", 1)[0]
```

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -198,27 +199,22 @@ def _resolve_mappings(self, model: Module) -> List[SmoothQuantMapping]:
...
         resolved_mappings = []
-        for to_balance, to_smooth in self.mappings:
-            to_smooth_list = [to_smooth] if isinstance(to_smooth, str) else to_smooth
-
-            for smooth_name, smooth_layer in match_named_modules(
-                model, to_smooth_list, self.ignore
+        module_to_name = get_module_to_name_dict(model)
+        for mapping in self.mappings:
+            for *nested_balance_layers, smooth_layers in match_modules_set(
+                model, tree_flatten(mapping)[0], self.ignore
```
kylesayrs:
```suggestion
                model, *mapping, self.ignore
```

src/llmcompressor/utils/pytorch/module.py
```diff
@@ -369,3 +370,15 @@ def get_layer_by_name(layer_name: str, module: Module) -> Module:
     if not layer_name:
         return module
     return attrgetter(layer_name)(module)
+
+
+def get_module_to_name_dict(model: Module) -> dict[Module:str]:
```
kylesayrs:
```suggestion
def get_module_to_name_dict(model: Module) -> dict[Module, str]:
```

kylesayrs (approved):
Sweet

src/llmcompressor/modifiers/smoothquant/base.py
HDCharles: I don't think this works [A,[B,C]] -> A,[B,C] we need to get [A,B,C]

kylesayrs: Ah I see.

fynnsu: Could switch to `pytree.tree_leaves` tho

src/llmcompressor/utils/pytorch/module.py
fynnsu: Why was this resolved?

`dict[t1:t2]` isn't a valid type. Mypy gives the below output:
```
error: "dict" expects 2 type arguments, but 1 given  [type-arg]
error: Invalid type comment or annotation  [valid-type]
note: did you mean to use ',' instead of ':' ?
```

src/llmcompressor/utils/pytorch/module.py
```diff
@@ -369,3 +370,15 @@ def get_layer_by_name(layer_name: str, module: Module) -> Module:
...
+    module_to_name = {}
+    for name, module in model.named_modules():
+        if module in module_to_name:
+            logger.info(
```
fynnsu: This should be `logger.warning`

src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -205,7 +206,9 @@ def _fuse_norms(self, model: PreTrainedModel):
             for norm, *linears in match_modules_set(
                 model, (mapping.norm, *mapping.linears)
             ):
-                fuse_norm_linears(norm, linears)
+                # match_modules_set returns a list of lists
+                assert len(norm) == 1
+                fuse_norm_linears(norm[0], tree_flatten(linears)[0])
```
fynnsu: `tree_leaves`

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -198,27 +199,22 @@ def _resolve_mappings(self, model: Module) -> List[SmoothQuantMapping]:
...
-                    resolved_mappings.append(
-                        SmoothQuantMapping(smooth_name, smooth_layer, balance_layers)
-                    )
+                assert len(smooth_layers) == 1, (
+                    "SmoothQuant mappings must match a single smooth layer for each "
+                    f"mapping but got {[module_to_name.get(s) for s in smooth_layers]}"
+                    f" for mapping: {mapping}"
+                )
+                smooth_layer = smooth_layers[0]
+                smooth_name = module_to_name.get(smooth_layers[0])
```
fynnsu, line 214:
```suggestion
                smooth_name = module_to_name.get(smooth_layer)
```

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -198,27 +199,22 @@ def _resolve_mappings(self, model: Module) -> List[SmoothQuantMapping]:
...
+                balance_layers = tree_flatten(nested_balance_layers)[0]
```
fynnsu: `tree_leaves`

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -198,27 +199,22 @@ def _resolve_mappings(self, model: Module) -> List[SmoothQuantMapping]:
...
-
-                if balance_layers:
-                    resolved_mappings.append(
-                        SmoothQuantMapping(smooth_name, smooth_layer, balance_layers)
-                    )
+                assert len(smooth_layers) == 1, (
+                    "SmoothQuant mappings must match a single smooth layer for each "
+                    f"mapping but got {[module_to_name.get(s) for s in smooth_layers]}"
+                    f" for mapping: {mapping}"
+                )
```
fynnsu: I think it's generally considered bad practice to use `assert` in production code. Instead we should use `if ...: raise ...`

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,73 +324,60 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                model, (mapping.smooth_layer, *mapping.balance_layers), self.ignore
             ):
-                pbar.set_description(
-                    f"Resolving mapping {mapping_idx+1}/{len(self.mappings)}"
-                    f" ({num_skipped_mappings} skipped)"
+                assert len(smooth_layers) == 1, (
+                    "AWQ mappings need to match a single smoothlayer for each "
+                    f"mapping but got {[module_to_name.get(s) for s in smooth_layers]}"
+                    f" for mapping: {mapping}"
+                )
```
fynnsu: Same as other comment, avoid using `assert` in production code.

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,73 +324,60 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
-                else:
-                    # for multiple balance layers, find lowest common parent
-                    parent_name, parent = get_lowest_common_parent(balance_names, model)
+                ancestor_name = get_lowest_common_ancestor_name(balance_names)
+                # no ModuleList ancestors
+                while not isinstance(
+                    (ancestor := model.get_submodule(ancestor_name)),
+                    torch.nn.ModuleList,
+                ):
+                    ancestor_name = ancestor_name.rsplit(".", 1)[0]
```
fynnsu: I think this could hit an infinite loop if you have a top level module list. 

i.e. `ancestor_name = "layers"` after `rsplit` is still `"layers"`.

fynnsu (commented):
Thanks for the pr. Added some comments below

src/llmcompressor/modifiers/smoothquant/base.py
HDCharles: do we have a dependency on Jax? i don't have it in my env atm

edit: nvm that's in the same dir, i thought you meant official pytree

src/llmcompressor/modifiers/awq/base.py
HDCharles: gah

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -740,6 +739,25 @@ def _check_layers_are_compatible(
     return True
 
 
+def get_lowest_ancestor_with_avoid(name: str, model: Module, avoid=torch.nn.Module):
```
fynnsu: The default for `avoid` seems risky to me. Maybe you meant it to be `torch.nn.ModuleList`? Otherwise this will probably just return the root `model` on every pytorch model unless `avoid` is explicitly set.

fynnsu (commented):
Thanks for updating. One small comment to resolve and then should be good to go

src/llmcompressor/utils/pytorch/module.py
```diff
@@ -368,4 +369,17 @@ def get_layer_by_name(layer_name: str, module: Module) -> Module:
     """
     if not layer_name:
         return module
+    print("CCC", layer_name)
```
kylesayrs: Remove

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -721,6 +712,54 @@ def _assert_all_activations_consumed(self):
...
+    NOTE: primarily used to exclude parents of type ModuleList, which don't play
+    nicely with hooks because their forward method is never directly
+    called for MoE models. See Qwen3MoeSparseMoeBlock for example, experts
+    are selected based on router output and their forward method is called.
+    https://github.com/huggingface/transformers/blob/v4.52.4/src/transformers/models/qwen3_moe/modeling_qwen3_moe.py#L233
+    """
+    while True:
+        if name == "":
+            return "", model
+        ancestor = get_layer_by_name(name, model)
```
```suggestion
        ancestor = model.get_submodule(name)
```

src/llmcompressor/utils/pytorch/module.py
HDCharles: mistake, fixed now

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,73 +324,58 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
-                elif len(balance_layers) == 1:
-                    # for single balance layer, parent is the balance layer
-                    parent_name, parent = balance_name, balance_layer
-                else:
-                    # for multiple balance layers, find lowest common parent
-                    parent_name, parent = get_lowest_common_parent(balance_names, model)
+                ancestor_name = get_lowest_common_ancestor_name(balance_names)
+                ancestor_name, ancestor = get_lowest_ancestor_with_avoid(
+                    ancestor_name, model, torch.nn.ModuleList
+                )
```
kylesayrs: It's weird to me that we need to get the lowest common ancestor twice

src/llmcompressor/utils/pytorch/module.py
kylesayrs: Should this not be a warning?

src/llmcompressor/modifiers/awq/base.py
HDCharles: i separated the function into 2 pieces since CT only needs the longest common substring functionality.

get_lowest_ancestor_with_avoid could  first call get_lowest_common_ancestor_name and then do what it does normally if you like that better?

HDCharles: ok changed

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,73 +324,57 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                if len(smooth_layers) > 1:
+                    raise ValueError(
+                        "AWQ needs to match a single smoothlayer for each mapping but "
+                        f"got {[module_to_name.get(s) for s in smooth_layers]}"
+                        f" for mapping: {mapping}"
+                    )
+                smooth_layer = smooth_layers[0]
+                smooth_name = module_to_name.get(smooth_layer)
+
+                # [[b00, b01, b02...], [b10, b11, b12,...], ...] v
```
rahul-tuli:
```suggestion
                # [[b00, b01, b02...], [b10, b11, b12,...], ...] 
```

HDCharles: its meant to be a down arrow, should I remove?

src/llmcompressor/utils/pytorch/module.py
HDCharles: fixed

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -319,73 +324,57 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                    raise ValueError(
+                        "AWQ needs to match a single smoothlayer for each mapping but "
+                        f"got {[module_to_name.get(s) for s in smooth_layers]}"
+                        f" for mapping: {mapping}"
+                    )
+                smooth_layer = smooth_layers[0]
+                smooth_name = module_to_name.get(smooth_layer)
+
+                # [[b00, b01, b02...], [b10, b11, b12,...], ...] ↓
+                #                             [b00, b01, b02, ..., b10, b11, b12, ...]
```
fynnsu, line 342: Something to consider in the future :)

Unicode `U+21B3`: ↳

HDCharles: strictly worse than U+2193: ↓

PR #2080, opened 2025-12-01
[Performance] [Utils] Refactor embeddings utils

kylesayrs:
## Purpose ##
* Prerequisite for detecting if the lm head can be skipped (which is necessary to support large batch calibration)
  * Breaking up the embeddings utils in this way makes it easier to implement `disable_lm_head` later

## Changes ##
* Generalize embedding utils
  * `_get_embeddings_or_warn` -> `get_embeddings`
    * Callers no longer have to use try-catch
    * Callers are responsible for warning
  * `untie_word_embeddings` is largely unchanged, slight clarity changes
* Update modifiers
  * `untie_if_target_shared_embedding(...)` -> `if targets_embeddings(...): untie_word_embeddings(...)`

## Testing ##
* Rename `test_model_shared_tensors` to `test_untie_word_embeddings`
* Add `test_targets_embeddings` to test targeting embeddings

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,8 +144,7 @@ setup(
             if BUILD_TYPE == "release"
             else "compressed-tensors>=0.12.3a2"
         ),
-        # TODO: replace it with the release version
-        ("auto_round @ git+https://github.com/intel/auto-round.git@llmc"),
+        ("auto-round==0.9.1"),
     ],
     extras_require={
         "dev": [
--- a/src/llmcompressor/modifiers/autoround/base.py
+++ b/src/llmcompressor/modifiers/autoround/base.py
@@ -105,6 +105,7 @@ class AutoRoundModifier(Modifier, QuantizationMixin):
     # AutoRound modifier arguments
     iters: int = 200
     enable_torch_compile: bool = True
+    batch_size: int = 8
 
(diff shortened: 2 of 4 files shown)


src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -151,14 +150,18 @@ def on_initialize(self, state: State, **kwargs) -> bool:
     @torch.no_grad()
     def on_start(self, state: State, event: Event, **kwargs):
         self.started_ = True
+        model = state.model
+
+        # untie embeddings and lm_head so they can be rotated separately
+        # in the future, these can remain tied to save memory
+        if targets_embeddings(model, self._get_targets(model)):
```
HDCharles: Need to always untie because center embeddings always affects the embeddings.

HDCharles:
Otherwise looks good

src/llmcompressor/modifiers/transform/spinquant/base.py
kylesayrs: Thanks for the catch. Theoretically "center embeddings" doesn't actually have an effect, but it's the default for now so we should always untie

PR #2081, opened 2025-12-01
[Performance] Disable LM Head

kylesayrs:
## Purpose ##
* In order to support calibrating with large batch sizes, the lm_head must be skipped, otherwise too much memory will be used 
* This implementation ensures that lm_head inputs can still be calibrated. While this technically invalidates calibrating output activations of the lm_head, this case is very outside of the scope of supported features that I think it can be safely ignored/ allowed to error.

| With LM Head | LM Head Disabled |
| - | - |
| without_disable | with_disable |

>  Llama 8B, 32 samples, sequence length 2048, batch size 16

## Prerequisites ##
* #2080

## Changes ##
* `disable_lm_head` replaces the module forward with a dummy forward function which returns meta outputs
  * Because calibration hooks trigger before forward is called, lm_head inputs can still be calibrated
  * This is safe with transformers implementations, as there are [no implementations](https://github.com/search?q=repo%3Ahuggingface%2Ftransformers+%22self.lm_head.weight%22&type=code) which access lm_head weights directly
  * This function is always enabled as a subcontext of `calibration_forward_context`

## Testing ##
* Added `test_disable_lm_head`
* Added lm head disabling tests for `test_calibration_forward_context`


src/llmcompressor/utils/helpers.py
```diff
@@ -1049,11 +1055,45 @@ def calibration_forward_context(model: torch.nn.Module):
...
+        return
+
+    else:
+        dummy_weight = lm_head.weight.to("meta")
+
+        def dummy_forward(self, input: torch.Tensor) -> torch.Tensor:
+            return input.to("meta") @ dummy_weight.T
+
+        with patch_attr(lm_head, "forward", dummy_forward.__get__(lm_head)):
+            yield
```
HDCharles, line 1093: still getting used to the code paradigms here, this nested context manager design is an interesting approach. I'm pretty sure i had similar situations in torchao and I never would have considered this.

HDCharles (approved):
looks good

src/llmcompressor/utils/helpers.py
kylesayrs: Yeah, I love the context manager pattern, as it makes it reminds implementers to clean up any side effects which might affect downstream code.

HDCharles (approved):

PR #2083, opened 2025-12-02
feat: add importance-aware mixed-precision quantization

wangwenmingaa:
This PR introduces and implements an importance-aware mixed-precision quantization method for large language models, with the following key features:

**Importance-Guided Quantization Strategy:**
We evaluate the contribution of each layer (or module) to the overall model performance and dynamically assign quantization precision accordingly. Empirical analysis shows that the first few and last few layers are typically more critical, so we keep them in full precision (e.g., FP16/FP32), while less sensitive middle layers are aggressively quantized.

**Mixed-Precision Configuration:**
High-importance layers: No quantization (kept in FP16/FP32)
Medium-importance layers: Quantized to FP8
Low-importance layers: Quantized to FP4
Strict Average Bitwidth Constraint (≤ 8-bit):
The overall average bitwidth across all layers is constrained to no more than 8 bits, ensuring efficient resource usage while maximizing performance.

**Outperforms FP8 per-channel Baseline:**
**Accuracy**:  Achieves an average +1.04% accuracy gain over FP8 per-channel quantization on challenging benchmarks including AIME 2024–2025, LiveCodeBench, and MMLU-Pro, demonstrating superior preservation of model capabilities under aggressive quantization.
**Inference Speed**: Runs faster than the FP8 per-channel quantized baseline under identical hardware conditions, thanks to reduced memory bandwidth pressure from mixed-precision weight layouts.
**Practical & Plug-and-Play**: Requires no architectural modifications—only a principled, importance-aware assignment of quantization policies—making it immediately deployable for real-world large language model inference.

### DIFF
--- a/examples/quantization_non_uniform/quantization_importance_aware.py
+++ b/examples/quantization_non_uniform/quantization_importance_aware.py
@@ -11,7 +11,7 @@ tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 
 # Configure the importance-aware mixed-precision quantization recipe.
 # In this case, we:
-#   * Keep the first 3 layers (0,1,2) and last 3 layers (62,63,51) in full precision
+#   * Keep the first 3 layers (0,1,2) and last 3 layers (51,62,63) in full precision
 #       (not quantized) due to their high sensitivity.
 #   * Exclude lm_head from quantization to preserve output quality.
 #   * Quantize weights of specific middle layers' self-attention and MLP blocks to fp4:
@@ -19,7 +19,7 @@ tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 #       - Modules: k_proj, o_proj, q_proj, v_proj, down_proj, gate_proj, up_proj
 #       - Scheme: fp4, symmetric, per-group (group_size=16), static (PTQ)
 #   * Quantize weights of other intermediate layers to fp8:
-#       - Layers: 3-14, 25-30, 32-36, 37-55, 61 (excluding already covered or ignored)
+#       - Layers: 3-14, 25-30, 32-45, 49, 52-55, 61
 #       - Same modules as above
 #       - Scheme: fp8, symmetric, per-channel, static (PTQ)
 #   * Additionally, dynamically quantize input activations for fp8-weighted layers:
@@ -29,8 +29,8 @@ tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 
 fp4_group = "15|16|17|18|19|20|21|22|23|24|31|46|47|48|50|56|57|58|59|60"
 fp8_group = (
-    "7|37|11|35|3|4|5|6|8|9|10|12|13|14|25|26|27|28|29|30|"
-    "32|33|34|36|38|39|40|41|42|43|44|45|49|52|53|54|55|61"
+    "3|4|5|6|7|8|9|10|11|12|13|14|25|26|27|28|29|30|"
+    "32|33|34|35|36|37|38|39|40|41|42|43|44|45|49|52|53|54|55|61"
 )
 
 recipe = f"""
--- a/examples/quantization_non_uniform/quantization_importance_aware.py
+++ b/examples/quantization_non_uniform/quantization_importance_aware.py
@@ -1,0 +1,94 @@
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.utils import dispatch_for_generation
+
+MODEL_ID = "Qwen/Qwen3-32B"
+
+# Load model.
+model = AutoModelForCausalLM.from_pretrained(MODEL_ID, torch_dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
+
+# Configure the importance-aware mixed-precision quantization recipe.
+# In this case, we:
+#   * Keep the first 3 layers (0,1,2) and last 3 layers (62,63,51) in full precision
+#       (not quantized) due to their high sensitivity.
+#   * Exclude lm_head from quantization to preserve output quality.
+#   * Quantize weights of specific middle layers' self-attention and MLP blocks to fp4:
+#       - Layers: 15-24, 31, 46-48, 50, 56-60
+#       - Modules: k_proj, o_proj, q_proj, v_proj, down_proj, gate_proj, up_proj
+#       - Scheme: fp4, symmetric, per-group (group_size=16), static (PTQ)
+#   * Quantize weights of other intermediate layers to fp8:
+#       - Layers: 3-14, 25-30, 32-36, 37-55, 61 (excluding already covered or ignored)
+#       - Same modules as above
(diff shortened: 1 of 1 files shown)


wangwenmingaa:
@dsikka @kylesayrs This PR is ready for review. Could someone please take a look?  Thanks!

mratsim:
I think if we give such an example it should be grounded on academic research so that people 5 years from now have an easier time to challenge the current status quo.

For example when doing importance aware quantization to quantize GLM-4.6 to fit in 192GB VRAM, this is what I mention: https://huggingface.co/mratsim/GLM-4.6-EXL3

> ## Quantization theory and heuristics for manual tuning
> 
> ### Layers to quantize
> 
> Quantization should be focused on Linear layers (also called Dense or Fully-Connected layers i.e. MatMul+Bias)
> In particular quantizing LayerNorm/RMSnorm layer is strongly discouraged, see [1]
> > LayerNorm in Quantization. Kovaleva et al. (2021); Wei et al. (2022) find that outliers in the
> > LayerNorm parameters of BERT (Devlin et al., 2019) cause difficulties in model compression.
> > Given the importance of LayerNorm, all the quantization methods we discuss above leave LayerNorm unquantized.
> 
> This is also reported in Intel and Nvidia repo:
> - https://github.com/intel/neural-compressor/issues/1963#issuecomment-2274873441
> - https://github.com/NVIDIA/TensorRT/issues/4084#issuecomment-2294513950
> 
> EXL3 can only quantize linear layers.
> 
> ### Tensors to up-quantize
> 
> If there is enough bits, down projections should be prioritized.
> 
> According to [4]
> > Fig. 3: Maximum absolute value over layers for a LLaMA3-8B.
> > Each color represent a different projection and we clearly see that down_proj has the biggest
> > spikes in input and output. We also observe that RMSNorm propagate spikes through the entire model
> 
> According to [5]
> > Figure 5(a) illustrates the extremal ratio across layers and modules in LLaMA2-7B, highlighting
> > that weight outliers are concentrated in the down-projection matrices Wdown
> > ℓ of the second layer and
> > the last two layers. Figures 5(b) and 5(c) provide detailed visualizations of these outliers in the last
> > two layers.
> 
> ### Mixture-of-Experts quantization (MoE)
> 
> Mixture-of-Experts require specific quantization techniques.
> 
> #### Mixed-precision quantization
> 
> Some layers have a higher impact on LLM performance.
> According to [2], spending more bits in attention layers results in large gain compared to spending them in FFN layers.
> According to [3] on 2-bit quantization:
> - quantizing expert FFN layers do not seriously impact model quality
> - quantizing cross-attention has some impact
> - quantizing self-attention has a large impact
> - quantizing dense FFN has a very significant impact
> 
> Hence to preserve model quality we should choose not to quantize dense FFN layers and self-attention layers.
> 
> We notice that:
> - official MXFP4 weights of gpt-oss-120b from OpenAI keep self-attention in BF16:
>   - https://huggingface.co/openai/gpt-oss-120b/blob/main/model.safetensors.index.json
> - NVFP4 weights of DeepSeek-R1 quantized by Nvidia also keep self-attention in BF16:
>   - https://huggingface.co/nvidia/DeepSeek-R1-0528-FP4/blob/main/model.safetensors.index.json
> 
> #### Layers with high-impact
> 
> According to [2], giving more bits to the first `k` blocks have a significantly higher impact on model quality than for the same last `k` blocks.
> 
> #### Expert quantization
> 
> When quantizing MoE, quantizing activations is tricky as only a subset of experts are activated per request.
> 
> EXL3 has the tooling in-place to ensure all experts are activated during quantization, though it is unsure if the dataset should be expanded to be diverse enough so that all experts have a high likelyhood of taking the full range of values they can exhibit to avoid clipping.
> 
> ## References
> 
> 1. Why Do Some Inputs Break Low-Bit LLM Quantization? (2025)\
>   Ting-Yun Chang, Muru Zhang, Jesse Thomason, Robin Jia\
>   https://arxiv.org/pdf/2506.12044
> 
> 2. Examining Post-Training Quantization for Mixture-of-Experts: A Benchmark (2024)\
>   Pingzhi Li, Xiaolong Jin, Yu Cheng, Tianlong Chen\
>   https://arxiv.org/pdf/2406.08155v1
> 
> 3. Mixture of Quantized Experts (MoQE): Complementary Effect of Low-bit Quantization and Robustness (2023)\
>   Young Jin Kim, Raffy Fahim, Hany Hassan Awadalla\
>   https://arxiv.org/pdf/2310.02410
> 
> 4. Precision Where It Matters: A Novel Spike\
>    Aware Mixed-Precision Quantization Strategy for\
>    LLaMA-based Language Models (2025)\
>    Lucas Maisonnave, Cyril Moineau, Olivier Bichler, and Fabrice Rastello\
>    https://arxiv.org/pdf/2504.21553
> 
> 5. Systematic Outliers in Large Language Models (2025)\
>    Yongqi An, Xu Zhao, Tao Yu, Ming Tang, Jinqiao Wang\
>    https://arxiv.org/pdf/2502.06415v2

brian-dellabetta:
Hi @wangwenmingaa , thanks for the contribution, but in terms of importance-aware mixed-precision quantization, rather than hard-coding it into the quantization config, we plan to add this as a more robust feature next year -- dynamically finding which layers can be quantized heavily and which need to be left more precise.

IEI-mjx:
hi @wangwenmingaa ,
does vllm support mixed-precision quantized models?

kylesayrs (commented):
This is a neat example, and I think that this example will improve the visibility around the power of mixed precision.

Could you please include evaluation results in the example? Evaluation results which demonstrate the performance-accuracy tradeoff of mixed precision?

kylesayrs:
@IEI-mjx Yes! For layers other than MoE layers.

brian-dellabetta:
Hi @wangwenmingaa , regarding your comment in your summary:
> We evaluate the contribution of each layer (or module) to the overall model performance and dynamically assign quantization precision accordingly.

Can you elaborate on this a bit? Do you quantize a single layer and see how much of an effect on recovery it has, and from that determine which are most important? I wonder if we can include some of your importance-aware tooling or processes rather than just an example with the result of the analysis

IEI-mjx:
> @IEI-mjx Yes! For layers other than MoE layers.

Unfortunately, now I am trying to keep several layers of our MoE model in bf16 while other layers are quantized to int4. 
By the way, here is the average loss per layer during quantization
image
and here is the mean abs_max per layer
image
So, i am trying to ignore layer0/4/22/23 (the first, last two layers and the abnormal high loss layer) and the mix-precision in MoE experts between layers may lead to model load error?

kylesayrs:
> The mix-precision in MoE experts between layers may lead to model load error

Yes, vLLM has logic which throws an error if you attempt to load a model where fused layers (such as moe, qkv, ect.) do not have the same quantization.

kylesayrs (approved):

kylesayrs:
Closing for staleness, please reopen if still interested.

I think the changes needed are
1. Use the python objects to create the recipe
2. Simplify the regexes
3. Link to justifications on why this particular recipe is good for this model

PR #2084, opened 2025-12-02
[Fix] SmoothQuant MoE Support: Smooth All Experts, Not Just expert.0

rahul-tuli:
## Problem

SmoothQuant only smoothed the first expert (`expert.0`) in Mixture of Experts (MoE) models, leaving all other experts unsmoothed. This caused severe accuracy degradation for MoE models

### Root Cause

The `_resolve_mappings()` method in `SmoothQuantModifier` used `get_matching_layer()`, which returns only the **first regex match** instead of iterating over all matches. For MoE models with regex patterns like `"re:.*experts.*w1"`, this meant only `expert.0.w1` was smoothed while experts 1-N were ignored.

```python
# Before (BUGGY)
for balance_suffix in to_balance:
    _, balance_layer = get_matching_layer(balance_suffix, layer_name, model)
    # ❌ Returns only expert.0, ignores experts 1-15
```

## Solution

Replace `get_matching_layer()` with `match_named_modules()` from `compressed_tensors.utils` to iterate over **ALL** matched layers. This follows the same proven pattern used in `AWQModifier`.

```python
# After (FIXED)
for balance_regex in to_balance:
    for _, balance_layer in match_named_modules(smooth_parent, [balance_regex], self.ignore):
        balance_layers.append(balance_layer)
    # ✅ Returns ALL experts (expert.0, expert.1, ..., expert.15)
```

### Key Changes

1. **Updated imports**: Use `match_named_modules` from `compressed_tensors.utils`
2. **Rewrote `_resolve_mappings()`**: Iterate over all matched layers instead of just the first

## Tests Added

Added unit tests to encompass the issue to verify MoE support, these tests fail on main but pass with current diff:

### 1. `test_moe_all_experts_smoothed`
Verifies all 8 experts in a single MoE layer are included in `balance_layers`:
```python
num_experts = 8
# ... create MoE model with 8 experts ...
resolved_mappings = sq._resolve_mappings(model)
assert len(mapping.balance_layers) == num_experts  # All 8 experts
```

### 2. `test_moe_multiple_layers_all_experts_smoothed`
Verifies correct behavior across multiple transformer layers:
```python
num_layers = 2
num_experts = 4
# ... create model with 2 layers, 4 experts each ...
assert len(resolved_mappings) == num_layers
for mapping in resolved_mappings:
    assert len(mapping.balance_layers) == num_experts  # All 4 experts per layer
```

## Test Results

All tests pass successfully:

```bash
$ python -m pytest tests/llmcompressor/modifiers/smoothquant/test_base.py -v

test_smooth_quant_is_registered                          ✅ PASSED
test_smooth_quant_defaults                               ✅ PASSED
test_override_defaults                                   ✅ PASSED
test_moe_all_experts_smoothed                            ✅ PASSED
test_moe_multiple_layers_all_experts_smoothed            ✅ PASSED

========================= 5 passed in 0.41s =========================
```

### Before Fix (Tests Failed)
```
AssertionError: Expected 8 balance layers, got 1
# Only expert.0 was smoothed ❌
```

### After Fix (Tests Pass)
```
All 8 experts smoothed ✅
All tests passing ✅
```

## Related Issues

Fixes the SmoothQuant MoE bug reported in the community discussion about MoE quantization support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,8 +144,7 @@ setup(
             if BUILD_TYPE == "release"
             else "compressed-tensors>=0.12.3a2"
         ),
-        # TODO: replace it with the release version
-        ("auto_round @ git+https://github.com/intel/auto-round.git@llmc"),
+        ("auto-round==0.9.1"),
     ],
     extras_require={
         "dev": [
--- a/src/llmcompressor/entrypoints/utils.py
+++ b/src/llmcompressor/entrypoints/utils.py
@@ -29,12 +29,12 @@ from llmcompressor.core import reset_session
 from llmcompressor.pytorch.model_load.helpers import parse_dtype
 from llmcompressor.transformers.compression.compressed_tensors_utils import (
     modify_save_pretrained,
-    untie_word_embeddings,
 )
 from llmcompressor.transformers.utils.helpers import (
     is_model_ct_quantized_from_path,
 )
 from llmcompressor.typing import Processor
+from llmcompressor.utils import untie_word_embeddings
 from llmcompressor.utils.fsdp.helpers import is_fsdp_model
 
 
--- a/src/llmcompressor/modifiers/autoround/base.py
+++ b/src/llmcompressor/modifiers/autoround/base.py
@@ -20,10 +20,8 @@ from llmcompressor.core import Event, EventType, State
 from llmcompressor.modifiers import Modifier
 from llmcompressor.modifiers.quantization.calibration import apply_calibration_status
 from llmcompressor.modifiers.quantization.quantization import QuantizationMixin
-from llmcompressor.transformers.compression.compressed_tensors_utils import (
-    untie_if_target_shared_embedding,
-)
-from llmcompressor.utils.pytorch.module import get_no_split_params
+from llmcompressor.utils import targets_embeddings, untie_word_embeddings
+from llmcompressor.utils.pytorch import get_no_split_params
 
 __all__ = ["AutoRoundModifier"]
 
@@ -107,9 +105,9 @@ class AutoRoundModifier(Modifier, QuantizationMixin):
     # AutoRound modifier arguments
     iters: int = 200
     enable_torch_compile: bool = True
+    batch_size: int = 8
 
     # private variables
-    _module_names: Dict[torch.nn.Module, str] = PrivateAttr(default_factory=dict)
     _all_module_input: Dict[str, List[Tuple]] = PrivateAttr(default_factory=dict)
     _q_input: Optional[torch.Tensor] = PrivateAttr(default=None)
 
@@ -124,10 +122,6 @@ class AutoRoundModifier(Modifier, QuantizationMixin):
             QuantizationMixin.initialize_quantization(self, state.model)
 
         # prepare module names
-        self._module_names = {
-            m: name
-            for name, m in match_named_modules(state.model, self.targets, self.ignore)
-        }
         self._add_temporary_names(state.model)
         # freeze all model parameters
         for _, param in state.model.named_parameters():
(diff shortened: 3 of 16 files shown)


src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -196,31 +192,38 @@ def _resolve_mappings(self, model: Module) -> List[SmoothQuantMapping]:
...
+                    if smooth_parent_name
+                    else model
+                )
+
+                balance_layers = []
+                for balance_regex in to_balance:
+                    for _, balance_layer in match_named_modules(
+                        smooth_parent, [balance_regex], self.ignore
+                    ):
+                        balance_layers.append(balance_layer)
```
fynnsu: I think we can just do 
```python
for _, balance_layer in match_named_modules(
    smooth_parent, to_balance, self.ignore
):
    balance_layers.append(balance_layer)
```
since `match_named_modules` already contains a loop over targets ([link](https://github.com/vllm-project/compressed-tensors/blob/73c2cf935b53e0078be7766c5ee064755d980d78/src/compressed_tensors/utils/match.py#L65)).

or maybe even
```python
balance_layers = [balance_layer for _, balance_layer in match_named_modules(smooth_parent, to_balance, self.ignore)]

```

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -196,31 +192,38 @@ def _resolve_mappings(self, model: Module) -> List[SmoothQuantMapping]:
...
+            for smooth_name, smooth_layer in match_named_modules(
+                model, to_smooth_list, self.ignore
+            ):
+                # Search for balance layers within the parent scope
+                smooth_parent_name = ".".join(smooth_name.split(".")[:-1])
+                smooth_parent = (
+                    get_layer_by_name(smooth_parent_name, model)
+                    if smooth_parent_name
+                    else model
+                )
```
fynnsu: This is totally fine as is but it might be cleaner if we updated `get_layer_by_name` so that `get_layer_by_name("", model)` returns `model`. 

This could be done by adding an if statement before [this line](https://github.com/vllm-project/llm-compressor/blob/034072486874bb7119930d1daa37ac0593f7b193/src/llmcompressor/utils/pytorch/module.py#L368) to handle the special case. 

That would allow us to remove the `if smooth_parent_name else model` logic.

fynnsu (approved):
Looks good!

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -196,31 +192,38 @@ def _resolve_mappings(self, model: Module) -> List[SmoothQuantMapping]:
...
-                        _, balance_layer = get_matching_layer(
-                            balance_suffix, layer_name, model
-                        )
-                        if balance_layer:
-                            balance_layers.append(balance_layer)
-                    # each mapping can contain multiple layers to balance, but only
-                    # one layer to smooth
-                    mapping = SmoothQuantMapping(
-                        layer_name, smooth_layer, balance_layers
+            to_smooth_list = [to_smooth] if isinstance(to_smooth, str) else to_smooth
```
HDCharles, line 202: i'm working on a matching util that will make this mapping-matching easier, fyi

rahul-tuli:
@fynnsu Great suggestions! I've applied both:

  1. Simplified the nested loop: You're absolutely right that we can pass to_balance directly to match_named_modules since it already iterates over the targets list.
  Changed to a list comprehension for cleaner code.
  2. Updated get_layer_by_name: Added special handling for empty string to return the module itself. This cleaned up the ternary logic nicely. Also added a test case to
  cover this behavior.

  Both changes are now committed. Thanks for the thoughtful review!

src/llmcompressor/modifiers/smoothquant/base.py
rahul-tuli: Accepted! see comment

src/llmcompressor/modifiers/smoothquant/base.py
rahul-tuli: Accepted! see comment

PR #2100, opened 2025-12-05
add kv quant for autoround example

mengniwang95:
SUMMARY:
Add kv cache quantization into autoround example

TEST PLAN:
Have test the example offline

meta-llama/Meta-Llama-3-8B-Instruct 
W4A16 tuning + cache quantization result:
| Tasks | Version | Filter           | n-shot | Metric      |     | Value |     | Stderr |
| ----- | ------: | ---------------- | -----: | ----------- | --- | ----: | --- | -----: |
| gsm8k |       3 | flexible-extract |      5 | exact_match | ↑   | 0.740 | ±   | 0.0139 |
|       |         | strict-match     |      5 | exact_match | ↑   | 0.742 | ±   | 0.0138 |

### DIFF
--- a/src/llmcompressor/core/session.py
+++ b/src/llmcompressor/core/session.py
@@ -7,7 +7,7 @@ registration, and state tracking.
 """
 
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable
 
 from loguru import logger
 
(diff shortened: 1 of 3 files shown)


hshen14:
@dsikka @kylesayrs @brian-dellabetta please help review the PR. Thanks.

yiliu30:
@mengniwang95  please add UTs, thanks! https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/transformers/autoround/test_autoround_oneshot.py

dsikka:
I think it would be a lot cleaner if we added a kv cache specific example

mengniwang95:
> I think it would be a lot cleaner if we added a kv cache specific example

Do you mean adding an example which only quantizes kv cache without autoround quantization?

dsikka:
Yes

mengniwang95:
@dsikka How about moving this autoround quantizer + kv cache quantization example to https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_kv_cache/ folder? I notice examples in this folder are all quantizer + kv cache quantization
or can we create sub-folders under auto-round to maintain different auto-round examples, e.g., kv_cache_quantization, quantization_w4a16, ...?

mengniwang95:
Hi @dsikka , I split the kv quant part to another folder since we will contribute more autoround examples. Could you please take a look?

examples/autoround/README.md
```diff
@@ -125,6 +124,7 @@ lm_eval --model vllm \
 
 We can see the resulting scores look good!
 
+```text
```
brian-dellabetta: If you are changing formatting style here, you should drop the line just below that has ```bash

examples/autoround/quantization_kv_cache/llama3_example.py
```diff
@@ -0,0 +1,68 @@
...
+
+ds = get_dataset(
+    tokenizer=tokenizer,
+    seqlen=MAX_SEQUENCE_LENGTH,
+    nsamples=NUM_CALIBRATION_SAMPLES,
+)
+
+# Configure the quantization algorithm to run.
+#   * quantize the weights to 4 bit with AutoRound with a group size 128
+#   * quantize the kv cache to fp8
```
brian-dellabetta, line 24: I am not sure if order matters but if you want it to match your comment, I believe you need to have AutoRound first in the list of quant_modifiers below.

dsikka (commented):
LGTM once Brian's comments are resolved

examples/autoround/README.md
mengniwang95: Thanks, deleted the ```text

examples/autoround/quantization_kv_cache/llama3_example.py
mengniwang95: Thanks, updated the comment to match the list of quant_modifiers

dsikka (approved):

PR #2104, opened 2025-12-09
[Bug fix] fix Qwen3VLMoe

Wangzheee:
SUMMARY:
Fix insufficient CPU memory: Do not create a new object to prevent insufficient CPU memory


dsikka:
Hi @Wangzheee we recently updated this definition, do you mind rebasing?

Wangzheee:
> Hi @Wangzheee we recently updated this definition, do you mind rebasing?

Thanks~ 
I rebased the latest code for main. It still seems like I need to use this PR to fix insufficient CPU memory.

brian-dellabetta (approved):
These changes make sense to me, though deferring to the VL expert -- @kylesayrs WDYT?

dsikka (changes requested):
We should test these with existing examples before merge

Wangzheee:
> We should test these with existing examples before merge

Thanks～ 
When quantifying Qwen3 vl, we found that a particularly large amount of CPU memory is allocated here

PR #2122, opened 2025-12-14
[LM Eval] Move vl configs to a different folder

dsikka:
SUMMARY:
This will make it easier to run the lm-eval tests in two different jobs, one focused on VL models and the other focused on a normal decoder model


kylesayrs (approved):

dsikka:
@dhuangnm could we separate the lm-eval test into two jobs as well, one which runs on the `configs` folder and one which works on the `vl_configs` folder? This should prevent future timeouts as well as during release

PR #2128, opened 2025-12-15
[awq] simplify compute_layer_means

HDCharles:
SUMMARY:
Previously I added _orient and _reorient functions to handle this, there was a TODO to use flatten_for_calibration, this was a little tricky since 
A) its difficult to use to reverse the transformation (see initial commit)
B) it changes the shape to [1, num_H, num_W, block_H*block_W] and while we can reshape that directly into the shape needed for the normalization easily enough, we'd like to record the 4D shape [num_H, num_W, block_H, block_W] which requires yet more steps to get and record.

Thus it seems simpler to do the unflatten and transpose manually, exactly how we need it, then record shapes, normalize and lastly use the recorded shapes to undo it all.

TEST PLAN:
pytest /home/HDCharles/repos/llm-compressor/src/llmcompressor/modifiers/awq/base.py


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -705,14 +704,36 @@ def _compute_layer_means(layers: list[Module]) -> torch.Tensor:
                 )
                 continue
 
-            # need to get to shape [num different chunks x size of each chunk]
-            weight = _orient_weight(weight, q_args)
-            # TODO ^ simplify logic and use flatten_for_calibration
+            match q_args.strategy:
+                # chunk size is the size of the size of the
```
fynnsu, line 708: typo

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -729,79 +750,6 @@ def validate_duo_scaling(cls, v):
...
-                num_heights, num_widths, block_height, block_width
-            )  # nH, nW, H, W
-            .transpose(1, 2)  # nH, H, nW, W
-            .reshape(orig_shape)  # nH*H=rows, nW*W=cols
-        )
-    else:
-        raise NotImplementedError(
-            "expected weight quantization strategy to be "
-            f"one of TENSOR, CHANNEL, GROUP, or BLOCK, got {q_args.strategy}"
-        )
```
fynnsu, line 801: Maybe add this back as the default case if non of the others are matched?

fynnsu (approved):
Approved but with some minor suggestions

src/llmcompressor/modifiers/awq/base.py
HDCharles: laater

src/llmcompressor/modifiers/awq/base.py
HDCharles: laater

PR #2138, opened 2025-12-16
make lm_eval test save to hf_hub by default

HDCharles:
makes things easier to run locally

TEST PLAN:
test run here: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/20280940142

### DIFF
--- a/tests/lmeval/test_lmeval.py
+++ b/tests/lmeval/test_lmeval.py
@@ -127,7 +127,12 @@ class TestLMEval:
         base_results = self._eval_base_model()
 
         if not self.save_dir:
-            self.save_dir = self.model.split("/")[1] + f"-{self.scheme}"
+            base_name = self.model.split("/")[1] + f"-{self.scheme}"
+            hf_cache = os.environ.get("HF_HUB_CACHE")
+            if hf_cache and os.path.isdir(hf_cache):
+                self.save_dir = os.path.join(hf_cache, base_name)
+            else:
+                self.save_dir = base_name
         oneshot_model, processor = run_oneshot_for_e2e_testing(
             model=self.model,
             model_class=self.model_class,


tests/lmeval/test_lmeval.py
```diff
@@ -127,7 +127,12 @@ def test_lm_eval(self, test_data_file: str):
         base_results = self._eval_base_model()
 
         if not self.save_dir:
-            self.save_dir = self.model.split("/")[1] + f"-{self.scheme}"
+            base_name = self.model.split("/")[1] + f"-{self.scheme}"
```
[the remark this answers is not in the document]
HDCharles: don't want to change any logic outside of adding the hf_hub_dir

tests/lmeval/test_lmeval.py
```diff
@@ -127,7 +127,12 @@ def test_lm_eval(self, test_data_file: str):
...
+            hf_cache = os.environ.get("HF_HUB_CACHE")
+            if hf_cache and os.path.isdir(hf_cache):
+                self.save_dir = os.path.join(hf_cache, base_name)
+            else:
+                self.save_dir = base_name
```
brian-dellabetta, line 135: Is this compatible with the format in which models are downloaded from HF hub? They have their own naming structure, e.g. a folder `models--deepseek-ai--DeepSeek-V3` for model id `deepseek-ai/DeepSeek-V3`. but maybe there's a helper in HF's cli codebase to help with this?

HDCharles: yeah i don't think this would have that same name structure. I'm mostly trying to move the save away from local dirs since in our devservers this saves it to /home which keeps filling up. Not sure if the hub cache would be problematic for other reasons, i've been mostly testing using this change for a while and haven't run into issues. Though maybe there's a better dir to check?

I usually set all of 

```
HF_HUB_CACHE HF_HOME TRANSFORMERS_CACHE HF_DATASETS_CACHE
```

to the same dir but maybe i shouldn't be?

also in these tests the model gets deleted at cleanup, its only when we get errors or something that it starts to become an issue.

brian-dellabetta: Discussed on slack, I don't see any helpers in the HF cli code to help with this. Since it's only our tests and it's an edge case I think it's fine to merge in. i'll approve

tests/lmeval/test_lmeval.py
```diff
@@ -127,7 +127,12 @@ def test_lm_eval(self, test_data_file: str):
         base_results = self._eval_base_model()
 
         if not self.save_dir:
-            self.save_dir = self.model.split("/")[1] + f"-{self.scheme}"
+            base_name = self.model.split("/")[1] + f"-{self.scheme}"
+            hf_cache = os.environ.get("HF_HUB_CACHE")
```
kylesayrs, line 131:
```python3
from huggingface_hub import constants
constants.HF_HUB_CACHE
```

FYI this is what transformers uses, and I think this should be more robust, as it [handles expanding the path and fallbacks](https://github.com/huggingface/huggingface_hub/blob/7e8595b6b1728e392bda85c9d151613b50112434/src/huggingface_hub/constants.py#L146)

HDCharles: Ok will update

PR #2141, opened 2025-12-16
Add HF token to prevent tests from skipping

dsikka:
Summary
- Many tests get skipped in our base tests because the workflow was missing the HF_TOKEN and did not have access to a gpu
- With this change, only the weekly and nightly tests outside of the transformers folder are skipped and everything else is running

### DIFF
--- a/tests/e2e/e2e_utils.py
+++ b/tests/e2e/e2e_utils.py
@@ -1,8 +1,10 @@
+from typing import Callable
+
 import torch
 import transformers
 from datasets import load_dataset
 from loguru import logger
-from transformers import AutoProcessor
+from transformers import AutoProcessor, DefaultDataCollator
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.quantization import GPTQModifier, QuantizationModifier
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,7 +39,7 @@ style:
 # run tests for the repo
 test:
 	@echo "Running python tests";
-	pytest tests $(PYTEST_ARGS)
+	pytest -r tests $(PYTEST_ARGS) --ignore tests/lmeval
```
dbarbuzzi: Shouldn't there be one or more letters after the `-r` flag? I’ve only ever used this to include more reasons in the summary, since the default is for failures, but in CI we often want to see reasons for skipped tests as well. We typically use `-ra` for CI purposes in this regard.

From `pytest --help`:

```
  -r chars              show extra test summary info as specified by chars:
                        (f)ailed, (E)rror, (s)kipped, (x)failed, (X)passed,
                        (p)assed, (P)assed with output, (a)ll except passed
                        (p/P), or (A)ll. (w)arnings are enabled by default
                        (see --disable-warnings), 'N' can be used to reset
                        the list. (default: 'fE').
```

dsikka: I ran it with just -r and it seemed to be sufficient. I can update to -ra

dbarbuzzi (approved):

PR #2142, opened 2025-12-17
Update lm-eval set-up to address regression

dsikka:
SUMMARY:
- Seems like setting the collator from tuncation to default + shuffling addresses the regression we're seeing in lm-eval
- Given the recovery values you see in these tests were determined using these settings, I think they should be how we evaluate our lm-eval tests for the time being

### DIFF
--- a/examples/awq/fp8_block_llama_example.py
+++ b/examples/awq/fp8_block_llama_example.py
@@ -1,0 +1,81 @@
+from datasets import load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.awq import AWQModifier
+from llmcompressor.utils import dispatch_for_generation
+
+# Select model and load it.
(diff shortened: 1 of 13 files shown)


HDCharles:
I'm unsure if we should make this change, if the new approach is lower variance and we're confident it's not an actual regression, should we just update the thresholds?

dsikka:
> I'm unsure if we should make this change, if the new approach is lower variance and we're confident it's not an actual regression, should we just update the thresholds?

Replied on slack

HDCharles:
ok, we can make this change for now and then decide what we want long term later

tests/e2e/e2e_utils.py
```diff
@@ -94,6 +100,7 @@ def data_collator(batch):
             )
 
     # Apply quantization.
+    breakpoint()
```
brian-dellabetta:
```suggestion
```

brian-dellabetta (commented):
Nice catch! One breakpoint to remove

HDCharles (approved):

kylesayrs (approved):

PR #2144, opened 2025-12-17
[Args] Shuffle data samples by default

kylesayrs:
## Purpose ##
* Better accuracy recovery is seen when shuffling data samples
* Assume that for batch_size=1, `data_collator_with_truncation` is equivalent to `default_data_collator`

### DIFF
--- a/src/llmcompressor/pipelines/sequential/ast_helpers.py
+++ b/src/llmcompressor/pipelines/sequential/ast_helpers.py
@@ -45,6 +45,14 @@ def autowrap_forward(module: torch.nn.Module, ignore: List[str]):
     :param module: module whose forward method should be replaced
     :param ignore: explicit list of function names to wrap
     """
+    # check forward method is implemented
+    if module.forward.__name__ == "_forward_unimplemented":
+        raise ValueError(
+            "Cannot calibrate model which does not implement `forward` method. Please "
+            "either implement a forward method on the model, or pass a submodule to "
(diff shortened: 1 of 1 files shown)


src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -335,6 +352,17 @@ def oneshot(
         model calibration. When True, all experts will see all tokens during
         calibration, ensuring proper quantization statistics. When False, only
         routed experts will be used. Only relevant for MoE models. Default is True.
+    :param pipeline: Calibration pipeline used to calibrate model Options:
+        ['basic', 'datafree', 'sequential', independent]
```
brian-dellabetta:
```suggestion
        ['basic', 'datafree', 'sequential', 'independent']
```

brian-dellabetta (dismissed):
one nit

examples/multimodal_vision/gemma3_example.py
```diff
@@ -39,6 +39,7 @@
     splits=DATASET_SPLIT,
     recipe=recipe,
     batch_size=BATCH_SIZE,
+    shuffle_calibration_samples=False,
```
HDCharles, line 42: Why is Gemma getting special handling here?

kylesayrs: Because it's the only example of batching we have. Batching is bad due to padding/removal if it's shuffled.

PR #2147, opened 2025-12-18
[Offload] Remove deprecated offload functions, fix smoothquant

kylesayrs:
## Purpose ##
* Remove deprecation warnings as a result of using deprecated offload functions
* Fix smoothquant? Was smoothquant ever doing anything to offloaded models?

## Changes ##
* Remove `align_module` and `align_modules`
  * Onloading is now done automatically via attribute access
  * Use `disable_offloading` where appropriate to reduce runtime cost
* Remove usages of `register_offload_parameter` and `delete_offload_parameter`
* Remove `dispatch_for_sequential` and use `offload_model` directly

### SmoothQuant ###
* It seems to me like smoothquant never actually modified offloaded models. The in-place operations `mul_` and `div_` only affect onloaded values when in the `align_module_device` context. This means that, without calling `update_offload_parameter`, offloaded values are never modified and therefore the final model is unmodified
  * These changes use `update_offload_parameter`, as is required to update the model


tests/llmcompressor/transformers/compression/test_quantization.py
```diff
@@ -8,7 +8,7 @@
 from llmcompressor.args import DatasetArguments
 from llmcompressor.pytorch.utils import tensors_to_device
 from llmcompressor.transformers.data import TextGenerationDataset
-from llmcompressor.utils.dev import dispatch_for_generation
+from llmcompressor.utils.dev import dispatch_model
```
brian-dellabetta: probably a mistake?

kylesayrs: Yes! Thanks

brian-dellabetta (dismissed):

HDCharles:
WTF is going on in smoothquant tests?

HDCharles:
Should we add a test that checks if smooth quant is smoothquanting?

Issue #2151, opened 2025-12-18
[Bug]: qwen3-vl-30b-a3b v0.9.0 awq no matches found for input_layernorm


tansor:
### ⚙️ Your current environment

```
### Environment Information ###
Operating System: `Linux-4.18.0-147.5.2.15.h1109.eulerosv2r10.x86_64-x86_64-with-glibc2.31`
Python Version: `3.10.10 (main, Mar 21 2023, 18:45:11) [GCC 11.2.0]`
llm-compressor Version: `0.9.0`
compressed-tensors Version: `0.13.0`
transformers Version: `4.57.1`
torch Version: `2.7.1+cu118`
CUDA Devices: `['NVIDIA A100-SXM4-80GB']`
AMD Devices: `None`
```

### 🐛 Describe the bug

When I run `llm-compressor/examples/awq/qwen3-vl-30b-a3b-Instruct-example.py`, I got error, can you help me?

```
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 13/13 [00:09<00:00,  1.39it/s]
The module name  (originally ) is not a valid Python identifier. Please rename the original module to avoid import issues.
2025-12-18T05:54:15.688944+0000 | reset | INFO - Compression lifecycle reset
2025-12-18T05:54:15.690668+0000 | moe_calibration_context | INFO - Found 48 MoE modules to replace
Replacing MoE modules for calibration: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 48/48 [07:21<00:00,  9.19s/it]
2025-12-18T06:01:36.719691+0000 | moe_calibration_context | INFO - Replaced 48 MoE modules for calibration
2025-12-18T06:01:36.719944+0000 | moe_calibration_context | INFO - 48/48 modules will remain in calibration form (permanent)
2025-12-18T06:01:36.720856+0000 | from_modifiers | INFO - Creating recipe from modifiers
2025-12-18T06:02:06.436522+0000 | on_initialize | INFO - No AWQModifier.mappings provided, inferring from model...
[...]

PR #2152, opened 2025-12-18
[AWQ][Smooth] mapping shouldn't use ignore

HDCharles:
Summary:

after the change to use [match_module_set](https://github.com/vllm-project/llm-compressor/pull/2070) i made it so AWQ would take into account the ignore list during module mapping, this was a mistake since the ignore list should be used for quantization.

A similar issue occured with smoothquant which is also fixed, new tests were added to verify this behavior

at the same time, added behavior that if all balance layers + smooth layer are ignored, then it skips that mapping.

these changes should fix https://github.com/vllm-project/llm-compressor/issues/2151 however i have not been able to get it to run, while the matching now works correctly, something is happening with the tracing, i'm investigating that separately as i'm able to run it when i switch to the basic pipeline so this isn't an issue with this 

Test Plan:
pytest /home/HDCharles/repos/llm-compressor/tests/llmcompressor/modifiers/smoothquant/test_base.py
pytest /home/HDCharles/repos/llm-compressor/tests/llmcompressor/modifiers/awq/test_base.py

### DIFF
--- a/examples/autoround/README.md
+++ b/examples/autoround/README.md
@@ -91,7 +91,6 @@ oneshot(
     shuffle_calibration_samples=False,
 )
 
-
 # Save to disk compressed.
 SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-W4A16-G128-AutoRound"
 model.save_pretrained(SAVE_DIR, save_compressed=True)
--- a/examples/autoround/quantization_kv_cache/llama3_example.py
+++ b/examples/autoround/quantization_kv_cache/llama3_example.py
@@ -2,7 +2,6 @@ from auto_round.calib_dataset import get_dataset
 from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
-from llmcompressor.modifiers.autoround import AutoRoundModifier
 from llmcompressor.utils import dispatch_for_generation
 
 # Select model and load it.
@@ -22,10 +21,24 @@ ds = get_dataset(
 )
 
 # Configure the quantization algorithm to run.
+#   * quantize the kv cache to fp8
 #   * quantize the weights to 4 bit with AutoRound with a group size 128
-recipe = AutoRoundModifier(
-    targets="Linear", scheme="W4A16", ignore=["lm_head"], iters=200
-)
+recipe = """
+quant_stage:
+  quant_modifiers:
+    QuantizationModifier:
+      kv_cache_scheme:
+        num_bits: 8
+        type: float
+        strategy: tensor
+        dynamic: false
+        symmetric: true
+    AutoRoundModifier:
+      targets: [Linear]
+      scheme: W4A16
+      ignore: [lm_head]
+      iters: 200
+"""
 
 # Apply algorithms.
 oneshot(


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -120,7 +120,7 @@ class AWQModifier(Modifier, QuantizationMixin):
         to smoothed) and the second entry is the layer whose output is scaled to
         achieve the smoothing.
         If regex is used, it matches layers with the largest overlap in module name.
-    :param ignore: list of layers to ignore, even if they match a regex in mappings.
+    :param ignore: list of layers to ignore during quantization (not smoothing),
```
```suggestion
    :param ignore: list of layers to ignore during quantization (not smoothing).
```

kylesayrs (dismissed):

fynnsu (approved):

ZewenShen-Cohere:
Hi Charles, I’m working on the same issue as well. I don’t think your change will work correctly for multimodal LLMs. In practice, we usually do not want to quantize the vision encoder. As vision encoder usually shares a similar transformer structure, the vision encoder would also be picked up by the AWQ mapping resolution logic when you remove the 'ignore', which is not desirable. In my [draft PR](https://github.com/vllm-project/llm-compressor/pull/2158), I added an additional field in the modifier to separate the logic.

HDCharles:
> Hi Charles, I’m working on the same issue as well. I don’t think your change will work correctly for multimodal LLMs. In practice, we usually do not want to quantize the vision encoder. As vision encoder usually shares a similar transformer structure, the vision encoder would also be picked up by the AWQ mapping resolution logic when you remove the 'ignore', which is not desirable. In my [draft PR](https://github.com/vllm-project/llm-compressor/pull/2158), I added an additional field in the modifier to separate the logic.

I don't think it is an issue, its just going to do smoothing which will be mathematically irrelevent, though likely slow. Also this was tested on python /home/HDCharles/repos/llm-compressor/examples/awq/qwen3-vl-30b-a3b-Instruct-example.py which is multimodal. I left comments on your PR, I think we can make further enhancements on top of this PR since they seem orthagonal and the changes in this PR are enabling behaviors we had previously and are expected.

ZewenShen-Cohere:
Hi Charles, thank you for the suggestion. I've made a PR to filter out these redundant mappings https://github.com/vllm-project/llm-compressor/pull/2179

HDCharles:
@ZewenShen-Cohere sorry i'm back from the break now.

i've added that functionality to this PR and added tests as well as consolidated the logic for smoothquant and AWQ

examples/awq/README.md
```diff
@@ -42,4 +42,6 @@ In order to target weight and activation scaling locations within the model, the
 ]
 ```
 
-To support other model families, you can add supply your own mappings via the `mappings` argument with instantiating the `AWQModifier`, or you can add them to the registry [here](/src/llmcompressor/modifiers/awq/mappings.py) (contributions are welcome!)
+Note: the mappings define which layers get smoothed whereas targets and ignore define which layers get quantized. So if you include a layer in the ignore list that is going to get matched due to the included mappings, it will get smoothed but not quantized.
+
+To support other model families, you can supply your own mappings via the `mappings` argument with instantiating the `AWQModifier`, or you can add them to the registry [here](/src/llmcompressor/modifiers/awq/mappings.py) (contributions are welcome!)
```
fynnsu:
```suggestion
To support other model families, you can supply your own mappings via the `mappings` argument when instantiating the `AWQModifier`, or add them to the registry [here](/src/llmcompressor/modifiers/awq/mappings.py) (contributions are welcome!)
```

src/llmcompressor/modifiers/smoothquant/base.py
```diff
@@ -200,9 +201,21 @@ def _resolve_mappings(self, model: Module) -> list[SmoothQuantMapping]:
         """
         resolved_mappings = []
         module_to_name = get_module_to_name_dict(model)
+        # Get names of modules that are not ignored
+        ignored_names = set()
+        if self.ignore:
+            from compressed_tensors.utils import match_named_modules
```
brian-dellabetta: nit -- for required deps like CT might be cleaner to import this at top-level.

brian-dellabetta (dismissed):
Updates and tests look good! One nit.

PR #2148, opened 2025-12-18
[Offload] Switch to torch offloader from accelerate

kylesayrs:
## Co-requisites ##
* Compressed Tensors: [[TorchOffloader] Remove Accelerate](https://github.com/vllm-project/compressed-tensors/pull/530)

## Changes ##
* Perform tracing in the `disable_onloading` context (any tensors that are mentioned will be referenced as meta tensors to avoid excess onloading)
* Dispatch model after tracing (not really necessary)
* `dispatch_for_sequential` is now an alias for `offload_model`
* `dispatch_for_generation` is now an alias for `dispatch_model`
* Add a `get_main_device` utility to help with centralized device getting
* `untie_word_embeddings` is now simpler
* Fix fusing test by initializing tensors without gradients

## Testing ##
* https://github.com/neuralmagic/llm-compressor-testing/actions/runs/21348973738
* Optional TODO: add more tracing tests

PR #2157, opened 2025-12-19
Enhance Autoround to support multiple cards tuning

yiliu30:
Given AutoRound uses block‑level reconstruction loss to fine‑tune quantization parameters, which requires running backward passes on each block. For large model, like Qwen3-235B, a single GPU often doesn’t have enough memory to hold an entire block during backward computation. To address this, we use the HF accelerator to dispatch the module across multiple devices.
In this PR, we enable this feature on LLMC side:
- [x] Add `device_ids` for tuning with multiple cards
- [x] Map `ignore` to Autoround skipping layers
- [x] Add `Qwen/Qwen3-235B-A22B` as example for multiple cards

#### Test plan
```bash
pytest -svv ./llmcompressor/transformers/autoround/test_autoround_oneshot.py -k test_oneshot_with_device_map
```
#### Example results

```bash
# vllm (pretrained=INC4AI/Qwen3-235B-A22B-W4A16-G128-AutoRound-ITERS1-LLMC-TEST-ONLY,tensor_parallel_size=2,max_model_len=8192,max_num_batched_tokens=32768,max_num_seqs=128,add_bos_token=True,gpu_memory_utilization=0.8,dtype=bfloat16,max_gen_toks=2048,enable_prefix_caching=False), gen_kwargs: (None), limit: 1000.0, num_fewshot: None, batch_size: 128
# |Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
# |-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
# |gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.928|±  |0.0082|
# |     |       |strict-match    |     5|exact_match|↑  |0.930|±  |0.0081|
  

# vllm (pretrained=INC4AI/Qwen3-235B-A22B-W4A16-G128-AutoRound-ITERS200-LLMC-TEST-ONLY,tensor_parallel_size=2,max_model_len=8192,max_num_batched_tokens=32768,max_num_seqs=128,add_bos_token=True,gpu_memory_utilization=0.8,dtype=bfloat16,max_gen_toks=2048,enable_prefix_caching=False), gen_kwargs: (None), limit: 1000.0, num_fewshot: None, batch_size: 128
# |Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
# |-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
# |gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.934|±  |0.0079|
# |     |       |strict-match    |     5|exact_match|↑  |0.915|±  |0.0088|

```

cc @hshen14 @thuang6 @wenhuach21

### DIFF
--- a/examples/autoround/qwen3_example.py
+++ b/examples/autoround/qwen3_example.py
@@ -25,7 +25,8 @@ ds = get_dataset(
 
 # Configure the quantization algorithm to run.
 #   * quantize the weights to 4 bit with AutoRound with a group size 128
-#   * For `Qwen/Qwen3-235B-A22B`, it requires more about 300 GB memory to run tuning with default settings.
+#   * For `Qwen/Qwen3-235B-A22B`, it requires about 300 GB memory
+#     to run tuning with default settings.
 recipe = AutoRoundModifier(
     targets="Linear",
     scheme="W4A16",
--- a/src/llmcompressor/modifiers/autoround/base.py
+++ b/src/llmcompressor/modifiers/autoround/base.py
@@ -63,8 +63,9 @@ def suspend_accelerate_hooks(model: nn.Module):
     Temporarily suspend Accelerate hooks from a model.
 
     This context manager detaches all Accelerate hooks (used for device offloading,
-    dtype casting, etc.) from the model, allowing Autoround to operate without interference.
-    On exit, the model is restored to its original device and all hooks are re-attached.
+    dtype casting, etc.) from the model, allowing Autoround to operate without
+    interference. On exit, the model is restored to its original device
+    and all hooks are re-attached.
     """
     saved_hooks = {}
     original_device = next(model.parameters()).device


src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -278,6 +321,17 @@ def on_finalize(self, state: State, **kwargs) -> bool:
 
         return True
 
+    def get_unquantized_layer_names(self, wrapped_model: torch.nn.Module) -> List[str]:
```
yiliu30, line 337: @mengniwang95 @xin3he You may need this function as well.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -54,6 +57,33 @@ def _wrap_decoding_layer(layer: torch.nn.Module) -> _PretrainModelWrapper:
...
+    for name, module in model.named_modules():
+        if hasattr(module, "_hf_hook"):
+            saved_hooks[name] = module._hf_hook
+
+    remove_hook_from_submodules(model)
+    try:
+        yield
+    finally:
+        remove_hook_from_submodules(model)
+        model.to(original_device)
```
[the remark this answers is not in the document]
yiliu30: Since the input is a decoding layer, we can assume it always includes parameters.

examples/autoround/qwen3_example.py
```diff
@@ -0,0 +1,68 @@
...
+recipe = AutoRoundModifier(
+    targets="Linear",
+    scheme="W4A16",
+    ignore=[
+        "lm_head",
+        "re:.*mlp.gate$",
+    ],
+    iters=ITERS,
+    enable_torch_compile=False,
+    device_map="0,1,2,3",  # Use 4 A100 GPUs
```
brian-dellabetta: I also think the name here is confusing, isn't device_map usually a different format like a dict that maps layer name to device id? If `device_map="0,1,2,3"` is valid in transformers, we can leave as is, otherwise `device_ids` may be a better name

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -54,6 +57,33 @@ def _wrap_decoding_layer(layer: torch.nn.Module) -> _PretrainModelWrapper:
     return wrapped_model
 
 
+@contextmanager
+def suspend_accelerate_hooks(model: nn.Module):
```
brian-dellabetta, line 61: just fyi, we are refactoring our usage of accelerate hooks for offloading. You can follow some of that in
* https://github.com/vllm-project/llm-compressor/pull/2148

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -215,15 +249,20 @@ def apply_autoround(self, state, subgraph):
...
             ar_quant_scheme = self._mapping_config_to_autoround()
+            fp_layers = self.get_unquantized_layer_names(decoding_layer)
             ar = AutoRound(
                 model=wrapped_model,
                 tokenizer="",
                 scheme=ar_quant_scheme,
                 iters=self.iters,
                 enable_torch_compile=self.enable_torch_compile,
                 batch_size=self.batch_size,
+                device_map=self.device_map,
```
brian-dellabetta: just fyi, we are looking into better parallelization support and will create an RFC in the new year to gather feedback on best approaches. See PR and comment [here](https://github.com/vllm-project/llm-compressor/pull/1837#issuecomment-3671659187)

yiliu30:
> Hi AutoRound team, I think these changes make sense, though we are refactoring some things that overlap with these changes. Please see comments.
> 
> Can you point me to the logic in the auto round repo that handles the multi-gpu parallelization work? I'd like to see how you're handling it

Hi @brian-dellabetta , here is the logic for multi-gpu devices, https://github.com/intel/auto-round/blob/b53ead7d77746385d700152c7f00960f18fb9d85/auto_round/compressors/base.py#L1560-L1562.

We take a block, its input, and the list of available devices, then assign each submodule to one of those devices. The accelerator’s `AlignDevicesHook` later used for dispatching the submodules accordingly.  

Inside [set_auto_device_map_for_block_with_tuning](https://github.com/intel/auto-round/blob/b53ead7d77746385d700152c7f00960f18fb9d85/auto_round/utils/device.py#L984), we estimate the block’s memory requirements based on its parameters, input, batch size, and a few heuristic factors. Using this estimate, we [assign devices to the submodules](https://github.com/intel/auto-round/blob/b53ead7d77746385d700152c7f00960f18fb9d85/auto_round/utils/device.py#L701) to make memory usage stays as balanced as possible across all GPUs. The final mapping is then attached to each module as its `tuning_device`.

examples/autoround/qwen3_example.py
yiliu30: Yes, the `device_map` is used in Transformers, and we follow a similar approach. Please refer to:
https://huggingface.co/docs/accelerate/en/usage_guides/big_modeling#accelerate

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Thanks! I noticed that change as well. We can adapt to it once it’s ready.
And could I ask what motivated you to implement that functionality yourself instead of using the accelerator hooks? I imagine it requires quite a bit of engineering effort.

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Thanks! I’ve looked through some of the multi‑card discussions in LLMC, and they’re quite insightful. 
In AutoRound, we currently chose the accelerator hooks because they’re general enough to work across most models without requiring explicit cross‑card communication ops or modeling changes. The downside, of course, is some communication overhead and limited overlap, which can affect performance. 

We’re also exploring more efficient ways to fully squeeze out GPU performance. Looking forward to the RFC from you all, hope it covers the tuning case if possible!

examples/autoround/qwen3_example.py
brian-dellabetta: I've only ever seen device_map be a string like "auto" or "sequential", or a dictionary mapping each module name to each device_id, like 
```python
device_map = {"block1": 0, "block2.linear1": 0, "block2.linear2": 1, "block2.linear3": 1}
```
What does it mean if `device_map="0,1,2,3"`? Is that like auto but only with the first 4 devices?

Reference: https://huggingface.co/docs/accelerate/en/concept_guides/big_model_inference#designing-a-device-map

src/llmcompressor/modifiers/autoround/base.py
brian-dellabetta: I don't think we have anything posted on the decision to move away from `accelerate` hooks, outside of it being a pain point and our usage of it being limited in scope. cc @kylesayrs in case there is any other information we can provide.

src/llmcompressor/modifiers/autoround/base.py
brian-dellabetta: we will be sure to share it out in the new year.

Can you elaborate on what you mean by the tuning case? Is this specific to the tuning stage mentioned in the SignRoundv2 paper?

yiliu30: Thanks! The tuning here refers to fine‑tuning the quantization parameters by evaluating the block‑wise reconstruction error. In this process, we compute the loss between the original floating‑point model and the Q‑DQ model, and then run a backward pass to update the gradients of the quantization parameters accordingly. This approach was introduced in SignRound v1. cc @wenhuach

image

For implementation details, please refer to the code here. https://github.com/intel/auto-round/blob/440288fd6b92509e84da337437a30997ac544735/auto_round/compressors/base.py#L2984

examples/autoround/qwen3_example.py
yiliu30: Oh, you're right. Updated to `device_ids`!

src/llmcompressor/modifiers/autoround/base.py
kylesayrs: There are many reasons why we're switching away from accelerate's offloading implementation (which are listed [here](https://github.com/vllm-project/compressed-tensors/pull/529)), but chief among them is that we're going to need to support distributed cpu and disk offloading, which is needed for [distributed computing](https://github.com/vllm-project/llm-compressor/issues/2180).

kylesayrs: FYI we have a utility which is similar to this called [`disable_hf_hook`](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/utils/offload.py#L272), but it looks like the functionality is the same. We can keep both implementations, since they'll need to be replaced later anyways.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -110,6 +140,10 @@ class AutoRoundModifier(Modifier, QuantizationMixin):
     iters: int = 200
     enable_torch_compile: bool = True
     batch_size: int = 8
+    # optional device map for layer dispatch during tuning
+    # examples: "0,1" for cuda:0,cuda:1; "auto" to use all available GPUs
+    # when None, no dispatching and the model stays on its current device
+    device_ids: Optional[str] = None
```
kylesayrs, line 159: Add docstring

kylesayrs (dismissed):
Related to #2180. I've done some basic modeling and it seems like AutoRound could have improved performance from deeper integration and utilizing the DP strategy detailed in the RFC.

In the meantime, this PR looks great, thanks!

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Thanks, updated.

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Thanks! I’d prefer to keep it as is and adopt the new change once it’s ready.

yiliu30:
Hi @dsikka could you help retrigger the CI. Thanks!

kylesayrs (approved):

brian-dellabetta (approved):
If you guys are in agreement to keep suspend_accelerate_hooks separate from our other implementation and remove in future, I'm good with these changes. 

Thanks for adding this!

src/llmcompressor/modifiers/autoround/base.py
brian-dellabetta: Ok sounds good, just wanted to make sure since fine-tuning has become an overloaded term. Thanks!

PR #2161, opened 2025-12-19
[AWQ] Fix _accumulate_mean bug, move AWQ activation averaging off CPU, and improve logging

ZewenShen-Cohere:
This PR addresses the following issues:

1. _accumulate_mean produces incorrect output on its first run.

2. cache_smooth_activations_hook previously performed the averaging computation on the CPU. When both the hidden dimension and sequence length are large, this makes AWQ calibration CPU-bound. The slowdown is especially severe when multiple AWQ quantization jobs run concurrently.

3. Added more informative logging to the AWQ calibration grid search, including per-mapping JSON logs.

This PR is a subset of https://github.com/vllm-project/llm-compressor/pull/2158


ZewenShen-Cohere:
@brian-dellabetta

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -147,6 +152,7 @@ class AWQModifier(Modifier, QuantizationMixin):
     offload_device: torch.device | None = None
     duo_scaling: bool | Literal["both"] = True
     n_grid: int = 20
+    log_output_path: str | None = None
```
brian-dellabetta: users are able to configure either a log file or log dir at a higher scope, see code [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/entrypoints/oneshot.py#L127-L141). So anything logged with `loguru.logger` should also be automatically added if those fields are set as inputs to `oneshot`. Have you tried that?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -260,9 +268,14 @@ def on_finalize(self, state: State, **kwargs) -> bool:
         if not self.ended_:
             self.on_end(state, None)
 
+        # Save error metrics to file if log_output_path is provided
+        if self.log_output_path is not None and len(self._error_metrics) > 0:
+            self._save_error_metrics()
```
brian-dellabetta: rather than relying on log dir to determine if something should be logged, i think it is preferred to use the log level. You can log these metrics as debug so the user will only see it if log level is debug. WDYT?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -638,6 +662,23 @@ def _compute_best_scale(
                 "module. If you encounter this error, raise an issue at "
                 "https://github.com/vllm-project/llm-compressor/issues"
             )
+        
+        err_reduction = best_error / initial_error
```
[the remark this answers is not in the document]
brian-dellabetta: this should never be hit, but given the type of initial_error is inferred to be `Optional[float]`, might make sense to add this. up to you

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -811,9 +957,10 @@ def _accumulate_mean(
     sum_added = inp.sum(dim=0)
     num_added = inp.size(0)
     if prev_mean_and_count is None:
-        return sum_added, num_added
+        return sum_added / num_added, num_added
```
brian-dellabetta, line 925: ah! this was the error you were talking about. yeah this makes sense, nice catch.

src/llmcompressor/modifiers/awq/base.py
ZewenShen-Cohere: That's a good idea. Will make the change!

ZewenShen-Cohere:
Thank you for the revision @brian-dellabetta. I've switched to use the logger.

dsikka:
Can you address the quality issues?

You can do this by running these steps in the root dir:

pip install -e .[dev]
make style
make quality

ZewenShen-Cohere:
> Can you address the quality issues?
> 
> You can do this by running these steps in the root dir:
> 
> pip install -e .[dev] make style make quality

Done

HDCharles (commented):
hey, I mentioned before that I fixed the accumulate means bug in a different PR, was that removed from the code but not the description?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -568,7 +574,14 @@ def _compute_best_scale(
                 for balance_layer in balance_layers_to_patch
             ],
         ):
-            for grid_idx, use_duo_scaling in product(range(n_grid), duo_scalings):
+            total_iterations = n_grid * len(duo_scalings)
+            pbar = tqdm(
```
HDCharles, line 609: unsure if we want additional output for AWQ, this step is relatively quick in practice and AWQ already does a lot of outputting

brian-dellabetta: Yeah, I agree. Any time we have to use `leave=False` for nested tqdm logs, output can get messy if any other things are logged. Did you find you needed this for some other reason, or can we remove this?

ZewenShen-Cohere: I use this to better understand the loss landscape of the AWQ optimization. It's a good indicator for understanding if the grid is too fine, etc. Does this sound reasonable?

HDCharles: ok i've accepted

dsikka:
@ZewenShen-Cohere do you mind fixing the quality issue, by running the command in your rootdir?

``` 
make style
make quality
```

ZewenShen-Cohere:
@dsikka Done!

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -705,7 +736,7 @@ def _compute_loss(
         # Compute the MSE loss for each batch
         for fp16_batch, int_w_batch in zip(fp16_outputs, int_w_outputs):
             loss += torch.nn.functional.mse_loss(
-                fp16_batch, int_w_batch.to(fp16_batch.device)
+                fp16_batch, int_w_batch.to(fp16_batch.device), reduction="sum"
```
brian-dellabetta, line 739: thanks for catching this!

PR #2175, opened 2025-12-29
Use natural sort for ordered grouping by layer

antbob:
SUMMARY:
The current code is implicitly relying on ordered grouping by layer however that is not always the case in practice and also since the API contract allows sets explicitly. When the order is broken it leads to:

ValueError: Matched a re:.*mlp\.gate_proj\.weight$ twice before completing set (language_model.model.layers.12.mlp.gate_proj.weight, language_model.model.layers.4.mlp.gate_proj.weight)

TEST PLAN:
The https://huggingface.co/mistralai/Ministral-3-14B-Reasoning-2512 can be used to test this and is a real life example on which the ValueError above has tripped.

### DIFF
--- a/tests/llmcompressor/modeling/test_calib_llama4.py
+++ b/tests/llmcompressor/modeling/test_calib_llama4.py
@@ -1,5 +1,4 @@
 import contextlib
-import os
 from functools import partial
 
 import pytest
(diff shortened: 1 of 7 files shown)


kylesayrs (approved):
Looks good, thanks for the catch!

PR #2177, opened 2025-12-31
[Audio] Qwen2 Audio Example

kylesayrs:
## Purpose ##
* Support qwen2_audio, which previously could not be supported due to accelerate offloading

### DIFF
--- a/examples/awq/llama_example_ddp.py
+++ b/examples/awq/llama_example_ddp.py
@@ -1,0 +1,113 @@
+#############################################################################
+# This script is adapted from ./llama_example.py and adds DDP functionality.
+# run this with `torchrun --nproc_per_node=2 llama_example_ddp.py`
+# or change nproc_per_node to your desired configuration
+# to adapt other examples to use DDP, see the 2 altered sections below
+#############################################################################
+
+import time
+
+import torch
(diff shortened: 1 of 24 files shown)


examples/multimodal_audio/qwen2_audio.py
```diff
@@ -0,0 +1,134 @@
...
+    return {key: value[0] for key, value in inputs.items()}
+
+
+ds = ds.map(preprocess, remove_columns=ds.column_names)
+
+# Recipe
+recipe = GPTQModifier(
+    targets="Linear",
+    scheme="W4A16",
+    ignore=["lm_head", "re:audio_tower.*"],
```
HDCharles, line 84: If we're ignoring the whole audio tower, should we have an example/documentation that shows how to do e.g. GPTQ and then does model free quant on the model tower?

kylesayrs: Quantizing the audio/vision tower is almost never done at this point, see our other examples

kylesayrs: We should probably revisit this policy in the future though, since it used to be that audio/vision towers could only be used once (one image per conversation), but are now used at all stages

PR #2182, opened 2026-01-02
Add missing assert in regression tests to prevent silent failures

JasonZhao47:
### Motivation
I noticed that the test case ```test_compressed_matches_decompressed__hf_quantizer``` currently calls ```torch.equal()``` without an ```assert``` statement. 

I am adding the missing assertion in ```tests/llmcompressor/transformers/compression/test_run_compressed.py``` to prevent silent failures in future regressions.

## TEST PLAN
The tests still pass, verified by running ```pytest -v tests/llmcompressor/transformers/compression/test_run_compressed.py``` locally. 

This confirms that the current models and kernels are indeed aligned.

## Notes for Reviewers 
I am currently familiarizing myself with the ```llm-compressor``` codebase. 
While exploring Issue 1853 for a potential contribution, I happened to notice that this test performs a check without an actual assert statement.
I’d appreciate any feedback if I missed anything regarding the contribution guidelines. Thanks!

### DIFF
--- a/tests/llmcompressor/transformers/compression/test_run_compressed.py
+++ b/tests/llmcompressor/transformers/compression/test_run_compressed.py
@@ -174,6 +174,6 @@ class Test_Compressed_CompressedLinear_Decompressed_Linear:
 
         # Compare outputs for each input
         for idx in range(len(SAMPLE_INPUT)):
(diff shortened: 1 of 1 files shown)


dsikka (approved):

dsikka:
Thank you!

PR #2188, opened 2026-01-05
[AWQ] speed improvements

HDCharles:
SUMMARY:
We identified several fixes, TODOs and improvements after the AWQ generalization [PR](https://github.com/vllm-project/llm-compressor/pull/1961) to increase the AWQ speed. This largely implements them, details below.

speedup on:
python /home/HDCharles/repos/llm-compressor/examples/awq/llama_example.py
OLD:
(8.00 minutes)
GPU Memory - Peak: 10.00 GB
NOW:
(7.09 minutes)
GPU Memory - Peak: 13.18 GB

RESULT:
11.37% speedup, memory increase is expected and primarily due to change #4 and #1 below

changes:
1) instead of recording the fp16_baseline_output during apply_smoothing, we add a hook so that the output is captured during sequential pipeline execution. also keep it on device to avoid unnecessary device on/offloading
2) we concatenate outputs into a single tensor for faster error calculation
3) instead of recording the entire state dict during compute_best_scale, we only record the state of the balance layers (also keep them onloaded on gpu instead of offloading since we're storing significantly less now)
4) previously we would write the stored value to the balance layer, then update that value based on the scale factor (2 writes), now we calculate hte scaled balance layer and update directly (1 write)
5) don't need to update the offload parameter when calculating best scale, only local
6) improvement #4 also allows us to be more targeted, previously during the first write we would update the whole state dict which is no longer necessary

other changes which were tested:
torch compiling the best_scales_loop (device offloading prevented compilation)
calculating mse_loss progressively as each sample is run (slower)

TEST PLAN:
ran AWQ tests and examples to verify correctness

### DIFF
--- a/examples/autoround/qwen3_example.py
+++ b/examples/autoround/qwen3_example.py
@@ -1,0 +1,69 @@
+from auto_round.calib_dataset import get_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.autoround import AutoRoundModifier
+from llmcompressor.utils import dispatch_for_generation
+
+# Select model and load it.
+model_id = "Qwen/Qwen3-235B-A22B"
+model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+# Select calibration dataset.
+NUM_CALIBRATION_SAMPLES = 128
+MAX_SEQUENCE_LENGTH = 2048
+ITERS = 200
+# Get aligned calibration dataset.
+
+ds = get_dataset(
+    tokenizer=tokenizer,
+    seqlen=MAX_SEQUENCE_LENGTH,
+    nsamples=NUM_CALIBRATION_SAMPLES,
+)
+
+
+# Configure the quantization algorithm to run.
+#   * quantize the weights to 4 bit with AutoRound with a group size 128
+#   * For `Qwen/Qwen3-235B-A22B`, it requires about 300 GB memory
+#     to run tuning with default settings.
+recipe = AutoRoundModifier(
+    targets="Linear",
+    scheme="W4A16",
+    ignore=[
+        "lm_head",
+        "re:.*mlp.gate$",
+    ],
+    iters=ITERS,
+    enable_torch_compile=False,
+    device_ids="0,1,2,3",  # Use 4 A100 GPUs
+)
+
+
+# Apply algorithms.
+oneshot(
+    model=model,
+    dataset=ds,
+    recipe=recipe,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+    shuffle_calibration_samples=False,
+)
+
+
+# Save to disk compressed.
+SAVE_DIR = model_id.rstrip("/").split("/")[-1] + "-W4A16-G128-AutoRound"
+print(f"save to {SAVE_DIR}")
+model.save_pretrained(SAVE_DIR, save_compressed=True)
+tokenizer.save_pretrained(SAVE_DIR)
+
+
+# Confirm generations of the quantized model look sane.
+print("\n\n")
+print("========== SAMPLE GENERATION ==============")
+dispatch_for_generation(model)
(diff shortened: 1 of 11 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -525,10 +560,23 @@ def _compute_best_scale(
...
+            for balance_layer in mapping.balance_layers
+            if hasattr(balance_layer, "quantization_scheme")
+            and hasattr(balance_layer.quantization_scheme, "weights")
+        ]
+
+        # Save state of the balance layers so we can restore later
+        # Using id(layer) as key works correctly with MoE because each expert
+        # is a separate module instance with unique id
+        orig_layer_states = {
+            id(layer): {
```
kylesayrs: Technically you don't need to use `id`, since modules are already hashed to ids

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -594,24 +636,26 @@ def _compute_best_scale(
...
                     call_observer(
                         balance_layer,
                         "weight",
                         balance_layer.weight,
                         # TODO test should_calculate_gparam for nvfp4 support
                     )
-                    update_offload_parameter(
-                        balance_layer,
-                        "weight",
+                    balance_layer.weight.data = (
```
kylesayrs, line 676: Nice job avoiding writing to the offload

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -594,24 +636,26 @@ def _compute_best_scale(
...
                             balance_layer,
-                            balance_layer.weight.data,
+                            balance_layer.weight,
                             "weight",
                             w_qscheme,
                         )
-                        / _scalesview,
-                    )
+                        / _scalesview
+                    ).to(balance_layer.weight.dtype)
```
kylesayrs, line 684: You can avoid this by using `copy_` instead of assignment.

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -647,23 +689,17 @@ def _compute_best_scale(
...
-                fp16_batch, int_w_batch.to(fp16_batch.device)
-            ).item()
-            num_elements += fp16_batch.numel()
-
-        # Normalize the loss by the total number of elements
-        loss /= num_elements
-
-        return loss
+        # Concatenate all tensors and compute MSE in a single operation
+        # (faster than tensor by tensor)
```
kylesayrs: Nice.

kylesayrs (dismissed):
Looks great, awesome improvements

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: So we don't need `update_offload_parameter` here because it all happens on the exec device, and the smooth function is done elsewhere after best_scales are calculated?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -349,6 +358,17 @@ def cache_parent_kwargs_hook(
             values = inspect.signature(module.forward).bind(*args, **kwargs)
             self._parent_args_cache[module].append(values.arguments)
 
+        def cache_fp16_baseline_hook(
```
brian-dellabetta, line 361: if we are now caching output activations for every mapping in a given subgraph, wouldn't this increase memory requirements quite a bit, especially for MoE models? For which model are you seeing the 30% memory increase  that you mention in the summary?

brian-dellabetta (approved):
Looks good! One clarifying question and a question on the increased memory requirements

src/llmcompressor/modifiers/awq/base.py
HDCharles: hold on i'm rewriting this, i didn't realize by default we don't enable offloading so all my measurements were off. we do need to cache this but not on gpu and we can offload it

src/llmcompressor/modifiers/awq/base.py
HDCharles: yeah we can just keep in memory and not mess with that.

dsikka:
@mergifyio refresh

HDCharles:
see https://github.com/vllm-project/llm-compressor/pull/2265 for further discussion and contains a better version of these changes

PR #2195, opened 2026-01-07
Switch the pytorch tests to run on the L4 runners

dhuangnm:
SUMMARY:
We have the GCP L4 runners back and expanded to multiple zones. Switching the pytorch tests to run on the L4 runners so the test jobs can be distributed on more runners and not only rely on the H100 runners. The base tests in the workflow ran OOM on L4 so we'll still run them on the H100.

TEST PLAN:
Had a successful run for the test-check.yaml workflow: https://github.com/vllm-project/llm-compressor/actions/runs/20787678167/job/59701632932


dbarbuzzi (approved):

dhuangnm:
> Should we do this for the base tests as well?

I tried to run the base tests with L4 runners but they ran OOM so I still keep the base tests to run on H100 and only run the pytorch tests on L4.

kylesayrs (approved):

PR #2196, opened 2026-01-07
[Performance] Memoryless observers for weights

kylesayrs:
## Purpose ##
* Reduce memory usage when calibrating weights

## Changes ##
* Always use `memoryless_minmax` and `memoryless_mse` when performing weight quantization

## Testing ##
* lm-eval 🟢: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/20797105387/job/59733406907

### DIFF
--- a/src/llmcompressor/modifiers/quantization/calibration.py
+++ b/src/llmcompressor/modifiers/quantization/calibration.py
@@ -64,8 +64,16 @@ def initialize_observer(
     # training is no longer supported: always use memoryless for weights
     if base_name == "weight" and args.observer in ("static_minmax", "minmax"):
         observer = "memoryless_minmax"
+        logger.warning(
+            "Overriding weight observer for lower memory usage "
+            f"({args.observer} -> {observer})"
+        )
     if base_name == "weight" and args.observer in ("mse",):
         observer = "memoryless_mse"
+        logger.warning(
+            "Overriding weight observer for lower memory usage "
+            f"({args.observer} -> {observer})"
+        )
 
     if args is not None and args.dynamic is not True:
         observer = Observer.load_from_registry(


src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -43,6 +43,8 @@ def initialize_observer(
     The name is then used to load the observer from the registry and attached
     to the module. The name of the observer uses the base_name provided.
 
+    This function always initializes memoryless observers for weights
```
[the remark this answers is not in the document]
kylesayrs: This covers all of the possible observers

brian-dellabetta:
Do we want to change the presets in CT as well? Seems strange to me that we have logic in [presets](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_scheme.py#L178), in [QuantizationArgs validation](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_args.py#L339-L372), and here in LC as well

kylesayrs:
@brian-dellabetta Yeah I agree, I'll change the presets too. I think this PR should still land though, since it also safeguards against users accidentally using a moving-average observer for weights, when in reality it won't do anything except use extra memory.

kylesayrs:
https://github.com/vllm-project/compressed-tensors/pull/540

@brian-dellabetta There's no need to wait for this, we can land these changes now

dsikka:
@kylesayrs are we concerned around algorithms beyond the QuantModifier requiring a non-memoryless observer for weights?

E.g Does the AutoRound modifier need to track values over time for weight updates?

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -57,9 +59,17 @@ def initialize_observer(
     args: QuantizationArgs = getattr_chain(
         module, f"quantization_scheme.{arg_name}", None
     )
+    observer = args.observer
+
+    # training is no longer supported: always use memoryless for weights
+    if base_name == "weight" and args.observer in ("static_minmax", "minmax"):
+        observer = "memoryless_minmax"
+    if base_name == "weight" and args.observer in ("mse",):
```
dsikka, line 71: actually it probably makes sense to put up a warning instead of silently updating

kylesayrs: Yep

kylesayrs:
@dsikka Autoround does not use LC observers, as evidenced by how `QuantizationMixin.start_calibration` is never called.

dsikka (approved):

brian-dellabetta (approved):

PR #2203, opened 2026-01-08
feat(datasets): add batch token statistics logging to LengthAwareSampler

jwpark33:
SUMMARY:
This PR introduces batch-level token statistics logging to the LengthAwareSampler. When the batch_size is greater than 1, the sampler now calculates and logs the token overhead incurred by padding and truncation within each batch, as well as the total overhead for the entire dataset.

Key Changes:
- LengthAwareSampler Enhancement: Added a `batch_size` parameter and a private method `_calculate_and_log_batch_stats` to track token dynamics.
- Log Detailed Overhead: Provides DEBUG level logs for both per-batch and cumulative token additions (padding) or removals (truncation).
- Integration: Updated `_make_sampler` in `utils.py` to correctly propagate the `batch_size` to the sampler instance.
- Improved Observability: Helps users understand the efficiency of their batching strategy in terms of token utilization.

TEST PLAN:
```
pytest tests/llmcompressor/datasets/test_length_aware_sampler.py
```

### DIFF
--- a/src/llmcompressor/datasets/utils.py
+++ b/src/llmcompressor/datasets/utils.py
@@ -299,7 +299,7 @@ class LengthAwareSampler(Sampler[int]):
             f"{self.num_samples} samples with batch size {self.batch_size}"
         )
 
-        sorted_lengths = sorted(lengths, reverse=True)[: self.num_samples]
+        sorted_lengths = [lengths[i] for i in self.order][: self.num_samples]
         total_tokens_removed = 0
         total_tokens_added = 0
 


src/llmcompressor/datasets/utils.py
```diff
@@ -284,6 +288,47 @@ def __init__(
 
         lengths = [len(sample) for sample in data_source[feature_name]]
         self.order = torch.argsort(torch.tensor(lengths), descending=True).tolist()
+        self._calculate_and_log_batch_stats(lengths)
+
+    def _calculate_and_log_batch_stats(self, lengths: list[int]):
```
kylesayrs, line 293: I think you should be able to use the `self.order` value directly, essentially by grouping `self.order` into chunks of `batch_size`, and then summing the difference of (max_seq_len_in_batch - seq_len)

PR #2209, opened 2026-01-09
[Examples] Correct out-of-date warning for kv cache examples

kylesayrs:
## Purpose ##
* As of the attention refactor, CT inference with kv cache quantization is supported. Fix incorrect information

## Changes ##
* Remove note about CT inference not being supported
* Standardize sample generation code
* Remove note about gemma in transformers==4.49.0 (out of supported versions)


dsikka:
@mergifyio refresh

PR #2228, opened 2026-01-13
[model_free_ptq] Enhance to work with previously quantized checkpoints like `nvidia/DeepSeek-R1-NVFP4`

brian-dellabetta:
Prerequisites (tests will fail until merged):
- [x] https://github.com/vllm-project/compressed-tensors/pull/607

SUMMARY:

This PR enhances the `model_free_ptq` entrypoint to work with previously quantized checkpoints. The added example extends the `nvidia/DeepSeek-R1-NVFP4` nvfp4-quantized checkpoint to:
- [x] convert modelopt's NVFP4 format to create CT's, for corresponding mlp/expert layers.
- [x] quantize all compatible linear self_attn layers to FP8_BLOCK, including ones with shape not exactly divisible by block_size[1].
- [x] merge the two quantization_configs into a single compressed-tensors config in `config.json` "quantization_config"

Changes to src:
- [x] removes the targets must be "Linear" constraint from model_free_ptq, as it is no longer an issue in vllm.
- [x] import Converter abstraction from compressed-tensors convert_checkpoint entrypoint so that conversion from modelopt NVFP4 to CT format can happen at the same time as converting layers to some compressed form.
- [x] Some helper code moved to CT and imported here instead.

TEST PLAN:
- [x] Checkpoint (and script to run) available at https://huggingface.co/bdellabe/DeepSeek-R1-NVFP4-FP8-BLOCK. Works in vllm 0.15.1
- [x] Confirmed checkpoint is equivalent when running `convert_checkpoint(..., converter=...) + model_free_ptq(..., converter=None)` vs. `model_free_ptq(..., converter=...)`
- [x] Updated model_free_ptq tests to ensure they work when stacked (equivalently, when a CT checkpoint is used as the input model to model_free_ptq), that the quantization config is correct

### DIFF
--- a/tests/llmcompressor/modeling/test_calib_deepseek_v3.py
+++ b/tests/llmcompressor/modeling/test_calib_deepseek_v3.py
@@ -66,6 +66,8 @@ def test_calib_deepseekv3_module():
     config = DeepseekV3Config()
     with torch.device("cuda"):
         original = OriginalDeepseekV3MoE(config).eval()
+        for param in original.parameters():
+            param.data.normal_(mean=0.0, std=0.02)
 
     # Create dummy input tensor that simulates hidden_states
     hidden_dim = config.hidden_size
@@ -76,14 +78,12 @@ def test_calib_deepseekv3_module():
     with calibration_forward_context(original):
         true_output = original(sample)
 
-    assert not torch.isnan(true_output).any(), "true_output contains NaN values"
-
     module = CalibrationDeepseekV3MoE(original, config, calibrate_all_experts=True)
     with calibration_forward_context(module):
         output = module(sample)
-        assert torch.allclose(true_output, output, atol=1e-6)
+        assert torch.nn.functional.mse_loss(true_output, output) < 1e-10
 
     module = CalibrationDeepseekV3MoE(original, config, calibrate_all_experts=False)
     with calibration_forward_context(module):
         output = module(sample)
-        assert torch.allclose(true_output, output, atol=1e-6)
+        assert torch.nn.functional.mse_loss(true_output, output) < 1e-10


examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
```diff
@@ -0,0 +1,61 @@
...
+    ),
+    ignore=[
+        # NOTE: self_attn.kv_a_proj_with_mqa has incompatible shape 576x7168 with block size 128x128
+        # This regex excludes all layers that aren't
+        # - self_attn.kv_b_proj
+        # - self_attn.o_proj
+        # - self_attn.q_a_proj
+        # - self_attn.q_b_proj
+        "re:^(?!.*self_attn.(kv_b_proj|o_proj|q_a_proj|q_b_proj)).*$"
+    ],
```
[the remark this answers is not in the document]
brian-dellabetta: updated

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
```diff
@@ -0,0 +1,61 @@
...
+        ),
+        targets=[
+            # NOTE: self_attn.kv_a_proj_with_mqa has incompatible shape 576x7168 with block size 128x128
+            # This regex includes all layers that aren't
+            # - self_attn.kv_b_proj
+            # - self_attn.o_proj
+            # - self_attn.q_a_proj
+            # - self_attn.q_b_proj
+            # "re:.*self_attn.(kv_b_proj|o_proj|q_a_proj|q_b_proj).*"
+        ],
```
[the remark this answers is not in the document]
brian-dellabetta: updated

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
```diff
@@ -0,0 +1,56 @@
...
+MODEL_ID = "nvidia/DeepSeek-R1-NVFP4"
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-FP8-BLOCK"
+
+# Apply FP8-Block to the model's compatible self_attn Linear layers
+# Once quantized, the model is saved
+# using compressed-tensors to the SAVE_DIR.
+model_free_ptq(
+    model_stub=MODEL_ID,
+    save_directory=SAVE_DIR,
+    scheme=QuantizationScheme(
```
dsikka, line 21: Just use the pre-set scheme

brian-dellabetta: we need to set targets

dsikka: you can set targets outside of the scheme?

brian-dellabetta: [This](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/entrypoints/model_free/__init__.py#L36-L54) is the model_free_ptq API. I don't think so

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
```diff
@@ -0,0 +1,88 @@
...
+                block_structure=[128, 128],
+            ),
+            input_activations=QuantizationArgs(
+                num_bits=8,
+                type=QuantizationType.FLOAT,
+                strategy=QuantizationStrategy.GROUP,
+                symmetric=True,
+                dynamic=True,
+                observer=None,
+                group_size=128,
```
kylesayrs: Is this not just "FP8_BLOCK"?

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
```diff
@@ -0,0 +1,88 @@
...
+                #   - self_attn.q_b_proj
+                "re:.*self_attn.(o_proj|q_a_proj|q_b_proj).*"
+            ],
+        ),
+        max_workers=8,
+        device="cuda:0",
+    )
+
+
+def merge_configs():
```
kylesayrs: It might be simpler to just build the config from scratch. Since this flow is very specialized to deepseek anyways.

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -55,6 +55,7 @@ def model_free_ptq(
     # validate arguments
     model_files = get_checkpoint_files(model_stub)
     scheme_name, scheme = validate_scheme(scheme)
+    ignore = ignore or []
```
kylesayrs: We can just make this (or an empty tuple) the default value, since the ignore list is never mutated.

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -43,9 +43,46 @@ def process_file(
...
+        # - model.layers.0.mlp.down_proj.weight
+        # - model.layers.0.mlp.gate_proj.weight
+        # - model.layers.0.mlp.up_proj.weight
+        # - model.layers.3.mlp.shared_experts.down_proj.weight
+        # - model.layers.3.mlp.shared_experts.gate_proj.weight
+        # - model.layers.3.mlp.shared_experts.up_proj.weight
+        # - model.layers.3.mlp.experts.0.down_proj.weight
+        # - model.layers.3.mlp.experts.0.gate_proj.weight
+        # - model.layers.3.mlp.experts.0.up_proj.weight
+        if _match_name(module_name, "re:.*mlp.*\.(gate|up|down)_proj$"):
```
kylesayrs: Hopefully we don't plan to leave this in the source code.

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -104,8 +141,11 @@ def process_file_microscale_scheme(
     for name in list(tensors.keys()):
         module_name, param_name = name.rsplit(".", 1)
         is_linear_weight = param_name == "weight" and not module_name.endswith("norm")
+        is_targeted = (is_linear_weight and "Linear" in scheme.targets) or any(
```
kylesayrs: Isn't it more robust and explicit to just make ["Linear"] the default target list?

src/llmcompressor/entrypoints/model_free/__init__.py
kylesayrs: Maybe should also consider lm_head

src/llmcompressor/entrypoints/model_free/validate.py
```diff
@@ -47,7 +47,8 @@ def validate_scheme(scheme: QuantizationScheme) -> tuple[str, QuantizationScheme
     # target all modules; filter by ignore list
     # technically this should be "re:.*", but vllm's
     # ct moe layer has a hard coded check for "Linear"
-    scheme.targets = ["Linear"]
+    if len(scheme.targets) == 0:
+        scheme.targets.append("Linear")
```
kylesayrs, line 51: I think we should force users to be explicit about targeting linear layers.

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
brian-dellabetta: yes but i need to set targets, I can do this more cleanly with `**FP8_BLOCK`

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
brian-dellabetta: yeah, this is more to just show what i'm doing, i don't expect this pr to ever land like this. we can cherry-pick the changes in another PR

src/llmcompressor/entrypoints/model_free/process.py
brian-dellabetta: when i'm done i plan to ask team how frequently we expect requests like this to see if we should look into abstractions to help with this. i don't plan to land this PR as is

src/llmcompressor/entrypoints/model_free/process.py
brian-dellabetta: it's just hard to do with the API, user might use preset schemes which don't set targets

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
brian-dellabetta: Cleaned this up a bit to use `**FP8_BLOCK`

src/llmcompressor/entrypoints/model_free/__init__.py
brian-dellabetta: updated

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
brian-dellabetta: updated

src/llmcompressor/entrypoints/model_free/process.py
dsikka: I agree with Kyle. We can create  a dedicated model_opt conversion tool in ct

src/llmcompressor/entrypoints/model_free/validate.py
kylesayrs: Nm all good

src/llmcompressor/entrypoints/model_free/process.py
kylesayrs: Sg

src/llmcompressor/entrypoints/model_free/process.py
brian-dellabetta: Cleaned this up, it's ready for review again. all the custom DSR1 logic has been moved to the converter, the diff and example should look pretty clean now

examples/model_free_ptq/dsr1_nvfp4_fp8_block.py
brian-dellabetta: cleaned up, please take a look

src/llmcompressor/entrypoints/model_free/save_utils.py
```diff
@@ -86,3 +86,76 @@ def update_safetensors_index(
             indent=2,
             sort_keys=True,
         )
+
+
+def create_quant_config(
```
dsikka: Do we not already have a pathway to create a config in model_free_ptq? Why are we adding a new function?

brian-dellabetta: It was previously embedded in the function just above, but moving it out into a helper function, since it has to be more complex now to support creating a config fresh vs. appending to a previously created quantization config

HDCharles: i think if there's no converter we do, if there's a converter i think we need to merge things together

(oh i was too slow)

HDCharles (approved):
looks good to me

src/llmcompressor/entrypoints/model_free/save_utils.py
brian-dellabetta: right. the 3 pathways should be listed in the docstring. previously we were just creating a quantization config from scratch, not trying to add it to a pre-existing one

PR #2239, opened 2026-01-14
[Mergify] Update / add mergify conditions

dsikka:
Summary
- Mark PRs as stale if opened for > 60 days
- Remove documentation comment
- Add a `quality-failed` label if the job fails and remove once addressed

Will follow-up with closing stale PRs if these all look ok in the next day or so


kylesayrs (changes requested):
Hold

PR #2247, opened 2026-01-15
[Workflows][CI] Mark PRs as stale and close if no activity

dsikka:
Summary
- Marks PRs as stale if no activity for the past 60 days
- Closes them if still no activity after 30 additional days
- Does not impact drafts

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,12 +1,11 @@
-name: 'Close inactive issues and PRs'
+name: 'Close inactive PRs'
 
 on:
   schedule:
-    - cron: '0 12 * * *'
+    - cron: '0 17 * * *'
 
 jobs:
-  close-issues-and-pull-requests:
-    # Prevents triggering on forks or other repos
+  close-pull-requests:
     if: github.repository == 'vllm-project/llm-compressor'
     permissions:
       issues: write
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  close-pull-requests:
+    if: github.repository == 'vllm-project/llm-compressor'
+    permissions:
+      issues: write
+      pull-requests: write
+      actions: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d
```
dbarbuzzi, line 16: nit: it would be nice to have the pinned version as a comment for clarity when reading the workflow/potentially updating it, e.g.:


```suggestion
      - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d  # v10.1.1
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          operations-per-run: 1000
```
dbarbuzzi, line 18: This seems a bit high, are we concerned with being rate-limited? The docs list `1000` as an example debugging value only because when used with the prescribed debugging configuration, it would not actually perform the actions/hit the rate limit due to the value.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,30 @@
...
+          exempt-draft-pr: true
+
+          days-before-pr-stale: 60
```
dhuangnm, line 21: Will we unstale the PR if it gets activity after 60 days?

Issue #2160, opened 2025-12-19
[RFC][AWQ] Add masking support for MSE loss & mean activation calculation


ZewenShen-Cohere:
Chat templates have become increasingly long, and when they are used during AWQ calibration, these repetitive segments can occupy a large fraction of the calibration dataset. As a result, the mean activation statistics and MSE loss used in AWQ become heavily biased toward the chat templates rather than the actual task content. To address this, we would like to introduce masking to exclude these template tokens from the calibration statistics.

Currently, llm-compressor has no support for masking. I made an initial attempt in this direction in [this draft PR](https://github.com/ZewenShen-Cohere/llm-compressor-fork/pull/1/changes), summarized below:

1. I introduced a use_loss_masking flag in dataset_arguments.py and oneshot.py (see [here](https://github.com/ZewenShen-Cohere/llm-compressor-fork/pull/1/changes#diff-00ec9366d79c2b2daef7bbcaa939b343f129c934f26e1b209f3fe348b1293c37R250) and [here](https://github.com/ZewenShen-Cohere/llm-compressor-fork/pull/1/changes#diff-8b25e0e7dfb3c229926f751ac7d9dff4784e3e6692cc1038b050e0c1cb21792eR343)). When this flag is enabled, the input dataset is expected to provide a loss_mask column. During calibration, the pipeline retrieves this mask alongside input_ids and makes it available to downstream components (see [here](https://github.com/ZewenShen-Cohere/llm-compressor-fork/pull/1/changes#diff-7fa7c4bb4a7a6087e1af538b307f86d166cff0365bfe9977f9512fd8777df0a4R115)).

2. Making the AWQ hooks aware of loss_mask is nontrivial. My initial idea was to pass it directly via subgraph.forward(model, **inputs) and have the hooks inspect kwargs. However, this is not supported by the captured fx.graph. As a workaround, I used a context variable that is active only during the forward pass, allowing the mask to be accessed by the AWQ modifier (see [here](https://github.com/ZewenShen-Cohere/llm-compressor-fork/pull/1/changes#diff-7fa7c4bb4a7a6087e1af538b307f86d166cff0365bfe9977f9512fd8777df0a4R115) and [here](https://github.com/ZewenShen-Cohere/llm-compressor-fork/pull/1/changes#diff-a79ed86c2a25b361e101f22c2cdcc549cf2dbc09b48de91c752f5edc4cf925a7R408)).

[...]

PR #2250, opened 2026-01-15
[AWQ] Add Token Masking Support for Calibration

ZewenShen-Cohere:
This PR addresses https://github.com/vllm-project/llm-compressor/issues/2160 by introducing token-level masking infrastructure for calibration, enabling modifiers to focus optimization on specific tokens (e.g., assistant responses only) while ignoring others during calibration.

  ### Motivation

  When quantizing instruction-tuned models, calibration data typically contains both user prompts and assistant responses. However, quantization should primarily preserve the quality of model outputs (assistant responses), not the processing of inputs. Token masking allows modifiers to compute loss only on tokens that matter, improving quantization quality for instruction-tuned models.

  ### Usage

  Users provide a `loss_mask` field in their dataset (1 for tokens to include, 0 to exclude) and enable masking with `use_loss_mask=True`:

  ```python
  oneshot(
      model=model,
      dataset=ds,  # dataset with "loss_mask" field
      recipe=recipe,
      use_loss_mask=True,
  )
```

### Results 

Model: meta-llama/Meta-Llama-3-8B-Instruct

Benchmark: gsm8k

AWQ + int4a16 + gs128 + asymmetric quantization 
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7202|±  |0.0124|
|     |       |strict-match    |     5|exact_match|↑  |0.7202|±  |0.0124|
```

Same setup with token masking (see `examples/awq/llama_example_with_masking.py` for details)
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.7255|±  |0.0123|
|     |       |strict-match    |     5|exact_match|↑  |0.7263|±  |0.0123|
```

### Implementation

  The implementation adds core infrastructure that can be leveraged by any modifier:

  - State extension: loss_masks and current_batch_idx fields allow pipelines to pass mask data to modifiers during calibration
  - Pipeline integration: Both basic and sequential pipelines collect loss masks from batches and track the current batch index
  - AWQ integration: AWQ modifier updated to use masks in activation accumulation and loss calculation as a reference implementation

  ### File Changes

  - `src/llmcompressor/core/state.py` extends the State class with two new fields for token masking:
    - loss_masks: stores the list of loss mask tensors for each batch
    - current_batch_idx: tracks which batch is being processed, allowing modifiers to retrieve the correct mask
  - `src/llmcompressor/args/dataset_arguments.py` adds the use_loss_mask argument to enable token masking during calibration.
  - Pipeline updates to collect and provide loss masks to modifiers:
    - `src/llmcompressor/pipelines/basic/pipeline.py` collects loss masks from each batch and updates current_batch_idx before each forward pass.
    - `src/llmcompressor/pipelines/sequential/pipeline.py` populates loss masks from cached activations and tracks batch index during subgraph calibration.
  - `src/llmcompressor/modifiers/awq/base.py` integrates token masking into AWQ as a reference implementation:
    - Activation accumulation hook filters activations using the loss mask
    - Loss calculation applies masking to compute MSE only on relevant tokens
  - `src/llmcompressor/modifiers/utils/pytorch_helpers.py` adds get_loss_mask_from_batch helper function to extract loss masks from batch dictionaries.
  - `src/llmcompressor/datasets/utils.py` updates the data collator to handle the loss_mask field during truncation.
  - `examples/awq/llama_example_with_masking.py` demonstrates token masking with Llama-3-8B-Instruct, showing how to create masks that target assistant responses in chat data.

  ### Pending

- ~~Currently token masking doesn't support the `up_proj <-> down_proj` mapping in MoE models, as masking dispatch is required. I will add an error message when this case happens.~~ Done
- ~~There are several AWQ-related PRs pending review, need to resolve the merge conflicts later. ~~ Done
- ~~@HDCharles recommended moving the masking logic into the `_run_sample` function in `awq/base.py` instead of the `_compute_loss` function. I will address this later.~~ I think it doesn't matter for now.

### CODE REVIEW

File: src/llmcompressor/pipelines/basic/pipeline.py [hunk 1]
```diff
@@ -38,15 +41,35 @@ def __call__(
...
 
         LifecycleCallbacks.calibration_epoch_start()
 
         with contextlib.ExitStack() as stack:
             stack.enter_context(calibration_forward_context(model))
-            for batch in tqdm.tqdm(dataloader, desc="Calibrating"):
+            for batch_idx, batch in enumerate(
+                tqdm.tqdm(dataloader, desc="Calibrating")
+            ):
                 batch = apply_pad_mask_to_batch(batch)
```
kylesayrs: Let's remove `apply_pad_mask_to_batch`. In my understanding, it is a misguided attempt to do what this PR accomplishes.

examples/awq/llama_example_with_masking.py
```diff
@@ -0,0 +1,140 @@
...
+            and input_ids[i + 1] == assistant_id
+            and input_ids[i + 2] == end_header_id
+        ):
+            i += 3  # Skip header tokens
+            # Mark content tokens until eot_id
+            while i < len(input_ids) and input_ids[i] != eot_id:
+                mask[i] = 1
+                i += 1
+        else:
+            i += 1
```
kylesayrs, line 99: Do you think that we can generalize this logic and move it from the example script to within LLM Compressor? For example, something like

```python3
oneshot(..., loss_mask=["padding", "sys_prompt", "thinking"]
```

Or do you think this abstraction would be too hard/ not useful?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -390,8 +390,28 @@ def cache_smooth_activations_hook(
...
+                loss_mask = (
+                    loss_masks[batch_idx] if loss_masks and batch_idx >= 0 else None
+                )
+
+                if loss_mask is not None:
+                    # Mask: [batch, seq] -> [batch, seq, 1]
+                    mask = loss_mask.to(activations.device).unsqueeze(-1)
+                    flat_activations = activations.flatten(0, -2)  # [batch*seq, hidden]
+                    flat_mask = mask.flatten(0, -2).squeeze(-1)
+                    masked_activations = flat_activations[flat_mask.bool()]
```
kylesayrs, line 441: masked_scatter isn't relevant here?

examples/awq/llama_example_with_masking.py
ZewenShen-Cohere: I don’t know an easy way to implement this, since the tokenizer and chat template vary significantly across different models. If needed, we probably need to provide a plugin system like what we did for calibrate_all_experts

src/llmcompressor/modifiers/awq/base.py
ZewenShen-Cohere: The source tensor and the mask tensor have the same shape, so I think it's unnecessary

src/llmcompressor/pipelines/basic/pipeline.py
ZewenShen-Cohere: Done!

examples/awq/llama_example_with_masking.py
ZewenShen-Cohere: I think we can have a dictionary that specifies the start token and the end response token of each model. This allows us to mask the response generated by the LLM. We should also allow for custom mask.

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -416,6 +436,14 @@ def cache_smooth_activations_hook(
...
             # storing inputs to first balance layer is sufficient
             # other balance layers get the same input
+
+            # The line below is useful for models that use parallel transformer block,
+            # such as gemma 3, command A. Need a better way to integrate it to the code.
+            # layer_to_hook = (
+            #     mapping.parent.mlp
+            #     if hasattr(mapping.parent, 'mlp')
+            #     else mapping.balance_layers[0]
+            # )
```
brian-dellabetta, line 478: This could be added as an optional field to AWQMapping. If provided, hook to this layer, otherwise the first balance layer.

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -38,15 +41,34 @@ def __call__(
...
-                batch = apply_pad_mask_to_batch(batch)
+            for batch_idx, batch in enumerate(
+                tqdm.tqdm(dataloader, desc="Calibrating")
+            ):
+
+                # Collect loss mask from this batch before moving to device
+                if use_loss_mask:
+                    session.state.loss_masks.append(
+                        get_loss_mask_from_batch(batch, use_loss_mask)
+                    )
```
brian-dellabetta: Since we are already only doing this inside an `if use_loss_mask` conditional, i would argue against including `use_loss_mask` as a field to `get_loss_mask_from_batch`. This would clean up the utility function a bit and allow us to clean up the type
```python
loss_masks: list[torch.Tensor | None] | None = None
```
to
```python
loss_masks: list[torch.Tensor] | None = None
```

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -38,15 +41,34 @@ def __call__(
         :param dataloader: loads data for calibration
         :param dataset_args: dataset arguments relevant to pipelines
         """
+        session = active_session()
         dispatch_for_generation(model)  # basic dispatch is identical to generation
         model_device = get_execution_device(model)
+        use_loss_mask = (
+            getattr(dataset_args, "use_loss_mask", False) if dataset_args else False
```
brian-dellabetta, line 44: I believe we've changed this so that `dataset_args` will always be passed into the pipeline's call method, as shown [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/entrypoints/oneshot.py#L218-L226). You can update the typehint on L30 above to be
```python
dataset_args:"DatasetArguments",
```
and remove the if check on this line

brian-dellabetta (commented):
Thanks @ZewenShen-Cohere , things are looking good! I have a few comments

src/llmcompressor/pipelines/basic/pipeline.py
ZewenShen-Cohere: https://github.com/ZewenShen-Cohere/llm-compressor-fork/blob/6cdc06067e47a8d182eb3fa0df56a656a1de6c94/src/llmcompressor/pipelines/basic/pipeline.py#L80

This line passes dataset_args=None to the pipeline

src/llmcompressor/modifiers/awq/base.py
ZewenShen-Cohere: I wanted to do something like this, but this logic will apply to all mappings. Feels like a bit hacky

src/llmcompressor/pipelines/basic/pipeline.py
brian-dellabetta: ah ok, thanks for pointing that out. this is some strange legacy code from sparse gpt. we can skip this for now

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta: got it, yeah just a thought. autoawq had extra stuff in their resolved mappings data model ([example](https://github.com/casper-hansen/AutoAWQ/blob/main/awq/models/qwen3_moe.py#L28-L40)) that we pruned and just infer for the sake of simplicity. if it comes to it

ZewenShen-Cohere: I’ll leave it as is for now.

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -97,6 +97,16 @@ def __call__(
...
 
+            # Populate loss_masks once from cached activations for AWQ masking support
+            use_loss_mask = getattr(dataset_args, "use_loss_mask", False)
+            if use_loss_mask:
+                session.state.loss_masks = [
+                    activations.fetch(batch_idx, ["loss_mask"]).get("loss_mask")
+                    for batch_idx in range(len(dataloader))
+                ]
+            else:
+                session.state.loss_masks = None
```
kylesayrs, line 108: In the future, we should move this logic to a helper, or call this within `IntermediatesCache.from_dataloader`

PR #2257, opened 2026-01-16
[CI] Update stale PR / Issue conditions

dsikka:
SUMMARY:
- Add a message for stale issues
- Bump min days from 60 to 90 to be consistent with vLLM
- Add keep-open condition to ignore PRs / Issues with this label


shanjiaz (dismissed):

kylesayrs (approved):
@dsikka 
FYI, I think users report not having permissions to reopen issues. We should change that

PR #2265, opened 2026-01-20
[AWQ] Modifier Speedups

HDCharles:
We identified a number of inefficiencies and fixes after the AWQ generalization [PR](https://github.com/vllm-project/llm-compressor/pull/1961), this PR largely implements them, see details below. Note I previously made this [speed improvements PR](https://github.com/vllm-project/llm-compressor/pull/2188) which had some issues that have been fixed in this one, that PR is going to be closed.

# BENCHMARKS

to iterate more quickly i ran these tests on models with most of their [layers removed](https://github.com/vllm-project/llm-compressor/blob/1f036248f0310b8e95d488fc5d20831bcc9b62b7/examples/awq/llama_example.py#L69), the actual improvement should be a bit better since the layers which were removed are where the improvement happens. To replicate these numbers see the [first commit](https://github.com/vllm-project/llm-compressor/pull/2265/changes/1f036248f0310b8e95d488fc5d20831bcc9b62b7#diff-208ced55cba2d38b1bc9b03b5f79ae3483c0849cc708a6c1131c231aae5d4b3dR69)

| Runtime (min) | Improvement | PR    | Base  |
|---------------|-------------|-------|-------|
| llama_no_off  |        8.7% |  6.17 |  6.76 |
| llama_off     |        5.6% | 10.46 | 11.08 |
| moe_no_off*    |        2.8% |  6.67 |  6.86 |
| moe_off*       |        1.9% |  7.57 |  7.72 |

| Memory (GB)     |        |       |       |
|-----------------|--------|-------|-------|
| llama_PR_no_off |   7.8% |  9.22 |    10 |
| llama_PR_off    |  17.6% |  3.66 |  4.44 |
| moe_PR_no_off   |  -5.2% | 11.61 | 11.04 |
| moe_PR_off**      | -24.3% |  2.92 |  2.35 |

\*The actual speedup for MoE is going to be higher than this. These numbers are for a single layer being quantized so the calibration overhead is going to attenuate the gains.

\*\* This worsening of memory is due to the weights being cached on device, the layernorm -> up + gate proj mapping has to cache the up + gate linears for the entire MLP layer which is ~.6GB (same as when there's no offloading) which is large relative to the other device usage when offloading is enabled.

# SUMMARY:
changes:

- targetted weight cache, no offloading
  - previously in compute_best_scale we would record the entire state dict of the parent and store it on cpu
  - now only record the balance layer weights and store those on device since they are generally small
- reduce load/write/rewrite weights
  - during grid search we have to repeatedly updaste the weight to use scaled and fake quantized versions of the weight. Previously this was done by writing the original value, calculating the scaled value and then writing that (2 writes)
  - we instead calculate the scaled value directly from the on-device cached value and write it once
- fake quantize only on device weight
  - previously the on/offloaded balance layer weight was updated
  - we now just update the on-device value
- compute loss
  - we slightly optimize  compute loss to reduce device movement
  - note a number of approaches to improve the loss computation were attempted including
    - progressively calculating the loss while running the samples to get int_w_outputs to avoid storing the whole int_w_output on device (was slower and seems to saves the same memory as deleting int_w_outputs after its used)
    - using torch.cat to combine the fp16 outputs and int_w_outputs into a single tensors so we can do only a single mse calculation (torch.cat briefly doubles memory usage so it created a significant memory increase)
    - avoiding torch.cat by preallocating a flat tensor of the needed size and progressively storing chunks into it (slow)
    - torch compile on compute loss and/or run samples (would likely speed up the run samples code but the offloading framework doesn't work well with it)
- del int_w_outputs, simply deleting this intermediate value after its used saves a significant amount of memory
- change default offload device behavior, normally None, now by default we check whether we use the default MoE Mapping and if so, offload to cpu by default

TEST PLAN:
(see first commit tests)

### DIFF
--- a/examples/quantization_2of4_sparse_w4a16/README.md
+++ b/examples/quantization_2of4_sparse_w4a16/README.md
@@ -1,5 +1,7 @@
 # `int4` Weight Quantization of a 2:4 Sparse Model
 
+> **DEPRECATION WARNING:** The `marlin_24` compression format is deprecated and will be removed in a future release, as vLLM no longer supports marlin_24 models. See [issue #2267](https://github.com/vllm-project/llm-compressor/issues/2267) for more details.
+
 `llm-compressor` supports quantizing weights while maintaining sparsity patterns for memory savings and inference acceleration with `vLLM`
 
 > `2:4 sparisty + int4/int8` mixed precision computation is supported in vLLM on Nvidia capability > 8.0 (Ampere, Ada Lovelace, Hopper).
--- a/examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
+++ b/examples/quantization_2of4_sparse_w4a16/llama7b_sparse_w4a16.py
@@ -5,10 +5,14 @@
 # fine tuning practices
 # https://developers.redhat.com/articles/2025/06/17/axolotl-meets-llm-compressor-fast-sparse-open
 
+# DEPRECATION WARNING: The marlin24 compression format is deprecated and will
+# be removed in a future release, as vLLM no longer supports marlin24 models.
+# See https://github.com/vllm-project/llm-compressor/issues/2267 for details.
+
+import warnings
 from pathlib import Path
 
 import torch
-from loguru import logger
 from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
@@ -64,9 +68,10 @@ quantized_model.save_pretrained(
 )
 tokenizer.save_pretrained(f"{output_dir}/quantization_stage")
 
-logger.info(
-    "llmcompressor does not currently support running ",
-    "compressed models in the marlin24 format. "
-    "The model produced from this example can be ",
-    "run on vLLM with dtype=torch.float16.",
+warnings.warn(
+    "The marlin24 compression format is deprecated and will be removed in a future "
+    "release, as vLLM no longer supports marlin24 models. "
+    "See https://github.com/vllm-project/llm-compressor/issues/2267 for details.",
+    DeprecationWarning,
+    stacklevel=2,
 )
--- a/src/llmcompressor/modifiers/quantization/calibration.py
+++ b/src/llmcompressor/modifiers/quantization/calibration.py
@@ -66,13 +66,15 @@ def initialize_observer(
         observer = "memoryless_minmax"
         logger.warning(
             "Overriding weight observer for lower memory usage "
-            f"({args.observer} -> {observer})"
+            f"({args.observer} -> {observer})",
+            log_once=True,
         )
     if base_name == "weight" and args.observer in ("mse",):
         observer = "memoryless_mse"
         logger.warning(
             "Overriding weight observer for lower memory usage "
-            f"({args.observer} -> {observer})"
+            f"({args.observer} -> {observer})",
+            log_once=True,
         )
 
     if args is not None and args.dynamic is not True:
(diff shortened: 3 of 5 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -643,17 +691,15 @@ def _compute_best_scale(
...
                             balance_layer,
-                            balance_layer.weight.data,
+                            balance_layer.weight,
                             "weight",
                             w_qscheme,
                         )
-                        / _scalesview,
-                    )
+                        / _scalesview
+                    ).to(balance_layer.weight.dtype)
```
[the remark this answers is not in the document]
HDCharles: this is deliberate

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -727,17 +783,15 @@ def _compute_loss(
         loss = 0.0
         num_elements = 0
 
-        # Compute the MSE loss for each batch
+        # # Compute the MSE loss for each batch
```
brian-dellabetta:
```suggestion
        # Compute the MSE loss for each batch
```

HDCharles:
> One super nit, and a question on the memory increase for MoE models. That seems like a big memory hit, and users frequently report OOM issues with AWQ. I wonder if it is worth the tradeoff. Did you try with batched calibration? It seems like the fp16 cache introduced in here saves us some forward passes, but with batched calibration a lot of that runtime is reduced.

i'm not sure that it is a big memory hit, users who are hitting an OOM error should be enabling offloading so i'm not sure if its relevent in that case.

As for the other situation, the memory increase is significant with respect to teh incredibly tiny memory usage in this example. If you have a larger calibration set, this memory increase won't scale with that. I'm not sure that those numbers are at a scale we're worried about do you?

The best way to solve this would be to add an additional option to enable weight offloading, do you want me to add this?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -203,6 +205,20 @@ def on_initialize(self, state: State, **kwargs) -> bool:
                 architecture=state.model.__class__.__name__
             )
 
+        # Set default offload_device
+        if self.offload_device == Sentinel("not_provided"):
+            # Check if we're using MoE mappings
+            if self.mappings is _moe_default_mappings:
```
brian-dellabetta: to capture all MoEs, i think we could use `llmcompressor.modifiers.utils.pytorch_helpers:is_moe_model(state.model)`

brian-dellabetta (approved):
Looks good! Thanks for updating

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -413,6 +445,17 @@ def cache_smooth_activations_hook(
...
                 )
 
+                # Also cache FP16 baseline outputs
+                self._fp16_baseline_cache[mapping.parent] = IntermediatesCache(
+                    None,
+                    self.offload_device,
+                )
+                self.register_hook(
+                    mapping.parent,
+                    cache_fp16_baseline_hook,
```
kylesayrs: Isn't this redundant with `cache_parent_kwargs_hook`? ie,

```python3
def cache_activations_hook(module, args, kwargs, output):
  input_args = inspect.signature(module.forward).bind(*args, **kwargs)
  self._parent_args_cache[module].append(input_args.arguments)

  output = (output[0] if isinstance(output, tuple) else output).detach()
  self._fp16_baseline_cache[module].append({"output": output})

self.regiseter_hook(parent, cache_activations_hook, "forward", with_kwargs=True)
```

HDCharles: outdated

src/llmcompressor/modifiers/awq/base.py
brian-dellabetta, line 128: we should update this docstring, but i can do that in my follow-up too

PR #2282, opened 2026-01-23
[torch] Pin to 2.9.1

dsikka:
Summary
- Temporarily pin to torch 2.9.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,11 +121,11 @@ setup(
             else "requests>=2.32.2"
         ),
         ("tqdm>=4.66.3,<=4.67.1" if BUILD_TYPE == "release" else "tqdm>=4.66.3"),
-        ("torch>=2.7.0,<=2.9.1" if BUILD_TYPE == "release" else "torch>=2.7.0"),
+        ("torch>=2.7.0,<=2.9.1" if BUILD_TYPE == "release" else "torch>=2.7.0,<=2.9.1"),
         (
             "transformers>=4.56.1,<=4.57.5"
             if BUILD_TYPE == "release"
-            else "transformers>=4.56.1,<=4.57.5"
+            else "transformers>=4.56.1"
         ),
         ("datasets>=4.0.0,<=4.4.1" if BUILD_TYPE == "release" else "datasets>=4.0.0"),
         (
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -121,11 +121,11 @@ def localversion_func(version: ScmVersion) -> str:
             else "requests>=2.32.2"
         ),
         ("tqdm>=4.66.3,<=4.67.1" if BUILD_TYPE == "release" else "tqdm>=4.66.3"),
-        ("torch>=2.7.0,<=2.9.1" if BUILD_TYPE == "release" else "torch>=2.7.0"),
+        ("torch>=2.7.0,<=2.9.1" if BUILD_TYPE == "release" else "torch>=2.7.0,<=2.9.1"),
         (
             "transformers>=4.56.1,<=4.57.5"
             if BUILD_TYPE == "release"
-            else "transformers>=4.56.1,<=4.57.5"
+            else "transformers>=4.56.1"
```
kylesayrs, line 128: Seems a little premature :)

dsikka:
> I think we're fine to leave this off once [vllm-project/compressed-tensors#546](https://github.com/vllm-project/compressed-tensors/pull/546) is merged. it will know the torch pin from there, and we don't have to mess with torchvision/torchcodec

I think with the order that the CI installs packages, we may end up with incorrect torchvision / torchcodec versions

setup.py
dsikka: lol for now they are just releasing rcs

brian-dellabetta:
> > I think we're fine to leave this off once [vllm-project/compressed-tensors#546](https://github.com/vllm-project/compressed-tensors/pull/546) is merged. it will know the torch pin from there, and we don't have to mess with torchvision/torchcodec
> 
> I think with the order that the CI installs packages, we may end up with incorrect torchvision / torchcodec versions

tests are passing?

dsikka:
> > > I think we're fine to leave this off once [vllm-project/compressed-tensors#546](https://github.com/vllm-project/compressed-tensors/pull/546) is merged. it will know the torch pin from there, and we don't have to mess with torchvision/torchcodec
> > 
> > 
> > I think with the order that the CI installs packages, we may end up with incorrect torchvision / torchcodec versions
> 
> tests are passing?

I mean for our e2e and weekend testing

Issue #2286, opened 2026-01-25
Support Fp8 Block Quant for shapes not divisible by 128


robertgshaw2-redhat:
**Is your feature request related to a problem? Please describe.**
* Certain models (e.g. DeepSeekR1 + GLM-4.7 Flash) have shapes not divisible by 128
* these models can still run with FP8 block quant on VLLM (e.g. the AutoFP8 ones in DSR1 can)

**Describe the solution you'd like**
* I would like llm-compressor to support this

**Describe alternatives you've considered**
* using a different quant scheme, but then its not compatible with DeepGEMM and certain TRTLLM kernels

**Additional context**
* this would be a great feature

Etelis:
I can try.
If you don't think it's too hard

Etelis:
Instead of padding the weights directly and saving (which will increase the memory requirement for the model) I will add padding here (llm-compressor) during calibration and will create a PR on compressed-tensors to add padding during the forward pass in inference. 

Sounds reasonable?

brian-dellabetta:
Hey @Etelis , thanks for taking a look at this. i had a draft to do this last friday (#2283), but yours is more elegant with F.pad. Let's get yours in and i'll close mine. i'll review this morning

Etelis:
> Hey [@Etelis](https://github.com/Etelis) , thanks for taking a look at this. i had a draft to do this last friday ([#2283](https://github.com/vllm-project/llm-compressor/pull/2283)), but yours is more elegant with F.pad. Let's get yours in and i'll close mine. i'll review this morning

Sorry I haven't seen yours. 
I don't mind merging parts of mine / yours ..

brian-dellabetta:
> Sorry I haven't seen yours. I don't mind merging parts of mine / yours ..

no worries, i should've let Rob know. But i like yours better than mine with F.pad, and you already have tests. i can try this out today

robertgshaw2-redhat:
I think two motivating examples would be:
- create an FP8 Block DeepSeek-R1 for the Attn layers with NVFP4 for the MoE layers
- create a DeepSeekV2-Lite block Fp8 model (would be great for CI)
- create a GLM 4.7-Lite block Fp8 model (would be great)

robertgshaw2-redhat:
thanks guys!

Etelis:
If we can first merge the PR on [compressed-tensors](https://github.com/vllm-project/compressed-tensors/pull/547)
[...]

PR #2283, opened 2026-01-23
[Observers] Allow for case when weight shape and block size are not evenly divisble

brian-dellabetta:
SUMMARY:
Update observer logic for block strategy when weight shape is not divisible by block size

Prerequisite:
- [x] https://github.com/vllm-project/compressed-tensors/pull/547

TEST PLAN:
- [x] Quantized checkpoint made with this branch (and above CT branch) runs on vllm main for flashinfer, deepgemm and default kernels -- https://huggingface.co/bdellabe/DeepSeek-V2-Lite-FP8-BLOCK

Run script below with 
- `VLLM_BLOCKSCALE_FP8_GEMM_FLASHINFER=1 VLLM_USE_DEEP_GEMM=1` for flashinfer
- `VLLM_BLOCKSCALE_FP8_GEMM_FLASHINFER=0 VLLM_USE_DEEP_GEMM=1` for deepgemm
- `VLLM_BLOCKSCALE_FP8_GEMM_FLASHINFER=0 VLLM_USE_DEEP_GEMM=0` for default

```python
if __name__ == "__main__":
    from vllm import LLM, SamplingParams

    prompts = ["The Swiss Alps are", "Brad Marchand is", "The Toronto Maple Leafs are"]

    # Create a sampling params object for greedy sampling
    sampling_params = SamplingParams(
        temperature=0.80, top_p=0.95, max_tokens=40, min_tokens=10
    )
    llm = LLM(
        "bdellabe/DeepSeek-V2-Lite-FP8-BLOCK",
        max_model_len=4096,
        enforce_eager=True,
    )
    output = llm.generate(prompts, sampling_params)
    for out in output:
        print(out.outputs[0].text)

    print("COMPLETE")
```

PR #2290, opened 2026-01-26
Support FP8 Block Quantization for Non-Divisible Shapes

Etelis:
- Add padding support for FP8 block quantization when tensor dimensions are not divisible by the block size (128)
- Enables quantization of models like DeepSeek-R1, GLM-4 Flash, and Falcon-7B that have non-standard dimension sizes

## Related Issues/PRs

- **Fixes:** https://github.com/vllm-project/llm-compressor/issues/2286
- **Depends on:** https://github.com/vllm-project/compressed-tensors/pull/547

## Test Plan

- [x] Verified padding logic works for divisible dimensions (256x512)
- [x] Verified padding logic works for non-divisible dimensions (4544x768 - Falcon-7B style)
- [x] Verified padding logic works for both dimensions non-divisible (192x320)
- [ ] End-to-end test with actual model (requires compressed-tensors PR to be merged)

## Notes

This PR requires the corresponding changes in `compressed-tensors` (PR #547) to handle padding during the forward quantization pass. The llm-compressor changes handle calibration, while compressed-tensors handles inference.


src/llmcompressor/observers/helpers.py
```diff
@@ -76,8 +77,15 @@ def _flatten_weight(
...
         block_height, block_width = args.block_structure
         rows, cols = value.shape
-        block_rows = strategy_cdiv(rows, block_height, args.strategy, strict=True)
-        block_cols = strategy_cdiv(cols, block_width, args.strategy, strict=True)
+
+        # Pad tensor if dimensions are not divisible by block size
+        pad_rows = (block_height - rows % block_height) % block_height
+        pad_cols = (block_width - cols % block_width) % block_width
+        if pad_rows > 0 or pad_cols > 0:
+            value = F.pad(value, (0, pad_cols, 0, pad_rows), mode="constant", value=0)
```
brian-dellabetta, line 84: we won't be able to set to zero, as it will alter the downstream qparams calculations in undesired ways if the block has a bunch of zeros appended to it. we need to set the padding values to some reasonable reduction, like the mean or median of the non-zero values of the block, so the min/max values are not altered. I have this in my draft pr #2283 , perhaps we should use that instead but with the F.pad helper like you have

HDCharles: Is this True? 

it looks like [we manually clamp the min/max to include 0](https://github.com/vllm-project/compressed-tensors/blob/cc0acb38c0ae1219d18f38c5303944f492f3dfdc/src/compressed_tensors/quantization/utils/helpers.py#L83-L86). I believe the reason for this is that most kernels, even not for symmetric quantization require 0 to be exactly representable for masking tricks and stuff.

so this seems safe to me.

brian-dellabetta: @Etelis I talked to @HDCharles about this, and it sounds like zeros are preferred to mean values for MSE observer as well. We can probably confirm this with a test after the other work. it makes for a lot cleaner code than my #2283

src/llmcompressor/utils/block_quant_padding.py
```diff
@@ -0,0 +1,91 @@
...
+
+    :param shape: Original tensor shape (out_features, in_features)
+    :param block_structure: Block size as [block_n, block_k]
+    :return: Tuple of (pad_n, pad_k) representing padding for each dimension
+    """
+    block_n, block_k = block_structure
+    out_features, in_features = shape
+
+    pad_n = (block_n - out_features % block_n) % block_n
+    pad_k = (block_k - in_features % block_k) % block_k
```
HDCharles, line 37: Is the second modulo necessary? Seems like just the parenthesis is sufficient.

src/llmcompressor/observers/helpers.py
```diff
@@ -76,8 +77,15 @@ def _flatten_weight(
         # (1, num_block_rows, num_block_cols, block_width * block_height)
         block_height, block_width = args.block_structure
         rows, cols = value.shape
-        block_rows = strategy_cdiv(rows, block_height, args.strategy, strict=True)
-        block_cols = strategy_cdiv(cols, block_width, args.strategy, strict=True)
+
+        # Pad tensor if dimensions are not divisible by block size
```
HDCharles, line 81: Shouldnt this section use the helpers for padding that you wrote in block_quant_padding py?

If those helpers don't fit here, I think they should be generalized or broken up until they do.

Etelis:
Image now loads in VLLM
https://huggingface.co/Etelis/DeepSeek-V2-Lite-FP8-BLOCK-padded

src/llmcompressor/utils/block_quant_padding.py
Etelis: The second modulo handles the case when the dimension is already divisible by the block size. For example:

- If `out_features=256` and `block_n=128`:
  - Without second modulo: `128 - (256 % 128) = 128 - 0 = 128` (wrong, we want 0)
  - With second modulo: `(128 - 0) % 128 = 0` (correct)

- If `out_features=200` and `block_n=128`:
  - `(128 - (200 % 128)) % 128 = (128 - 72) % 128 = 56 % 128 = 56` (correct)

src/llmcompressor/observers/helpers.py
Etelis: Good point! Just pushed a commit to refactor this section to use `calculate_block_padding()` and `needs_block_padding()` from the `block_quant_padding` module. Also updated to derive block dimensions from the padded tensor shape directly using integer division, as suggested by Gemini.

src/llmcompressor/entrypoints/model_free/save_utils.py
```diff
@@ -20,13 +21,122 @@
...
 __all__ = ["update_config", "update_safetensors_index"]
 
+# Layer name patterns for detecting which config fields to update
+# These are "safe-to-pad" layers where padding zeros doesn't affect model correctness
+# because the extra dimensions will produce zero outputs that can be truncated
+
+# MLP intermediate layers - padding these is safe because:
+# - gate_proj/up_proj: extra output channels produce zeros (no input contribution)
+# - down_proj: extra input channels receive zeros (no output contribution)
+_MLP_GATE_UP_PATTERNS = ("gate_proj", "up_proj", "w1", "w3", "fc1")
```
HDCharles, line 31: Should we also specify gate_up_proj like iirc transformers has for qwen3 moe?

src/llmcompressor/entrypoints/model_free/save_utils.py
```diff
@@ -20,13 +21,122 @@
...
+_MLP_DOWN_PATTERNS = ("down_proj", "w2", "fc2")
+_MLP_INTERMEDIATE_PATTERNS = _MLP_GATE_UP_PATTERNS + _MLP_DOWN_PATTERNS
+
+# MoE layer patterns
+_MOE_PATTERNS = ("experts.", "mlp.experts", "block_sparse_moe")
```
HDCharles, line 36: Is that period intentional?

src/llmcompressor/utils/block_quant_padding.py
```diff
@@ -0,0 +1,91 @@
...
+            f"Expected 2D weight tensor, got shape {weight.shape}. "
+            "Block quantization padding only supports 2D weight matrices."
+        )
+
+    out_features, in_features = weight.shape
+    original_shape = (out_features, in_features)
+
+    pad_n, pad_k = calculate_block_padding(original_shape, block_structure)
+
+    if pad_n > 0 or pad_k > 0:
```
HDCharles, line 83: Feels more natural to just always do the padding branch since that will become a noop. Also simplifies to a single output type rather than None.

brian-dellabetta:
Closing in favor of #2283

PR #2125, opened 2025-12-14
[Examples] QwenOmni Example

kylesayrs:
## Purpose ##
* Resolves #1872
* Support `Qwen3OmniMoeForConditionalGeneration`

## Changes ##
* Pass `model.thinker` to `oneshot`, since `model` does not implement a forward method (the thinker module is a `PreTrainedModel` that contains all of the parameters worth quantizing)
* Patch the `fast_pos_embed_interpolate` to support accelerate offloading
* Patch dataloader to squeeze `image_grid_thw`, but leave `pixel_values` and other inputs unsqueezed
* Save example audio generation

## Testing ##
* Ran example e2e and produced sane outputs
[output.wav](https://github.com/user-attachments/files/24154836/output.wav)

```
['Based on the image provided, the animal is a small, white kitten.\n\nHere are some specific details about the animal:\n\
n*   **Species and Age:** It is a very young domestic cat, commonly known as a kitten.\n*   **Color and Coat:** The kitten
 has a pure white coat. Its fur appears soft and fluffy.\n*   **Posture and Location:** It is curled up and sleeping sound
ly on a white computer keyboard, which is placed on a wooden desk.\n*   **Features:** The kitten has its eyes closed, indi
cating it is asleep. It has small, pointed ears and a small nose.']
```

PR #2302, opened 2026-01-28
[AWQ ] Qwen3-omni-30b-a3b-example

JartX:
Example to quantize Qwen3 Omni with the AWQ method

### DIFF
--- a/examples/awq/qwen3-omni-30b-a3b-instruct-example.py
+++ b/examples/awq/qwen3-omni-30b-a3b-instruct-example.py
@@ -80,6 +80,7 @@ recipe = AWQModifier(
     ],
     duo_scaling=False,
     config_groups={
+        # Using W4A16 group 32 scheme to work with Exllama kernel
         "group_0": {
             "targets": ["Linear"],
             "input_activations": None,


examples/awq/qwen3-omni-30b-a3b-instruct-example.py
```diff
@@ -0,0 +1,112 @@
...
+            "input_activations": None,
+            "output_activations": None,
+            "weights": {
+                "num_bits": 4,
+                "type": "int",
+                "symmetric": True,
+                "strategy": "group",
+                "group_size": 32,
+                "observer": "mse",
+            }
```
brian-dellabetta, line 94: any reason why you're not using the `W4A16` preset here? was the `"mse"` observer important for you?

JartX: Yep for use ExllamaKernel and the group size 32 also, because cant exec It with ConchLinearKernel, only ExllamaKernel that works on 32

JartX:
> Thanks for the contribution! One question, otherwise i think it would be good to include this in our examples. we had a number of users ask about qwen omni.
> 
> This loads up in vllm?

Yes using the last vllm main, i go to upload the quant model

examples/awq/qwen3-omni-30b-a3b-instruct-example.py
brian-dellabetta: Oh ok, group size 32 is important. what about `"observer": "mse"`? We've previously not recommended as a default setting in examples because it can lead to longer runtimes with questionable benefit

JartX: > Oh ok, group size 32 is important. what about `"observer": "mse"`? We've previously not recommended as a default setting in examples because it can lead to longer runtimes with questionable benefit

I've always generated them with MSE, just tell me which one you want me to put and I'll change it :)

brian-dellabetta: Got it, ok. Just remove the line for now, it's fine to use but we usually stick to our recommended defaults in examples.

brian-dellabetta:
> > Thanks for the contribution! One question, otherwise i think it would be good to include this in our examples. we had a number of users ask about qwen omni.
> > This loads up in vllm?
> 
> Yes using the last vllm main, i go to upload the quant model

@JartX awesome! please let link the checkpoint when you upload, would love to check it out

JartX:
@brian-dellabetta https://huggingface.co/jart25/Qwen3-Omni-30B-A3B-Instruct-AWQ-W4A16

examples/awq/qwen3-omni-30b-a3b-instruct-example.py
```diff
@@ -0,0 +1,111 @@
...
+        "re:.*mlp\.gate$",
+        "re:.*shared_expert_gate$",
+        "re:.*input_layernorm$",
+        "re:.*post_attention_layernorm$",
+        "re:.*norm$",
+        "re:.*lm_head$"
+    ],
+    duo_scaling=False,
+    config_groups={
+        "group_0": {
```
brian-dellabetta, line 84: A comment would be good to add here to explain why we're not using the preset, based on our other thread would this be appropriate? 
```suggestion
        # Using W4A16 group 32 scheme to work with Exllama kernel
```

JartX: @brian-dellabetta do it

kylesayrs:
Hi @JartX

There is already a qwen3 omni example https://github.com/vllm-project/llm-compressor/pull/2125
Given that using the model with AWQ is not substantially different than using it with GPTQ (just change the modifier), I don't we need a separate example for this case.

JartX:
Hi @kylesayrs , I can't run the model generated with GPTQ in vllm, so I quantized it in AWQ.

https://github.com/vllm-project/llm-compressor/pull/2125#issuecomment-3807787353

Please, since this is just an example and the quantized model has also been published, and it took me a lot of work, study, and testing, I ask that you reconsider reintroducing it.

For example, GPTQ runs ConchLinearKernel for me, which also gives me an error, while with group_size 32 I can run the quantized model using llm-compressor.

Thank you very much.

PR #2314, opened 2026-01-30
move smoothquant to transforms

### LINKED ISSUE #2306: Move Smoothquant to Transforms folder (and AWQ eventually)
In llm-compressor we have a separation between modifiers and transforms

i.e. for a linear op $Y=X.W$:

a transform is a modification such as scaling or rotating that can be applied independenly of quantization, i.e.

- $Y=(Xs).(W/s)$  rescaling or equalization, used in smoothquant and AWQ
- $Y=(X.rotation).(unrotation.W)$ rotation is used in spinquant and quip

If you applied a transform to a model it theoretically wouldn't change any of the numerics aside from small floating point precision differences, each change to W comes with a comparable change upsteam in the model that will modify X in the opposite way. In practice these transforms take into account which quantization scheme will eventually be applied to W (and sometimes X) in order to make it easier to quantize.

Whereas a modifier is something that applies compression, i.e a function that applies quantization or sparsification to the weights, either with or without calibration data. These include round-to-nearest, GPTQ and autoround

The intended flow is to apply your transforms (if desired) to make the model weights easier to quantize, and then apply quantization.

## Problem

As illustrated above, Smoothquant and AWQ are in-theory transforms, but currently reside in the modifiers folder, are described as modifiers in several docs and especially for AWQ, operate like a modifier rather than a transform. 

### Smoothquant
Smoothquant is already implemented like a transform, after applying smoothquant your model is numerically the same, and then you apply quantization. Smoothquant has to be bundled with a quantizationmodifier as in [this example](https://github.com/vllm-project/llm-compressor/blob/dd221a45cfa3f235219b03384e16fff34c2b7eda/examples/quantization_w8a8_int8/llama3_example.py#L57-L59) since it doesn't do quantization on its own. This allows you to combine your transform with an arbitrary modifier i.e. smoothquant + GPTQ as in the example.

As of now Smoothquant is functionally a transform but a majority of our docs and the folder structure imply otherwise. This could be rectified relatively straightforwardly (by moving it to modifiers/transforms) though would require care to not break external references.

### AWQ
AWQ is not implemented like a transform, it is currently designed to apply quantization itself which limits its functionality you can only ever to AWQ + RTN. While [This PR](https://github.com/vllm-project/llm-compressor/pull/2206#pullrequestreview-3718801019) is in progress to enable AWQ to behave more like a transform, once that functionality is complete it should probably be reconfigured to operate like that by default and moved to the transforms dir.

Etelis:
## Summary

Moves `SmoothQuantModifier` from `modifiers/smoothquant/` to `modifiers/transform/smoothquant/` to correctly categorize it as a transform rather than a modifier.

Closes #2306

## Changes

- Moved SmoothQuant source files to `modifiers/transform/smoothquant/`
- Moved corresponding test files
- Updated all imports across examples, docs, and dependent code
- Exported `SmoothQuantModifier` from `modifiers.transform`

```python
from llmcompressor.modifiers.transform.smoothquant import SmoothQuantModifier
```

### DIFF
--- a/src/llmcompressor/modifiers/smoothquant/__init__.py
+++ b/src/llmcompressor/modifiers/smoothquant/__init__.py
@@ -1,0 +1,23 @@
+# ruff: noqa
+"""
+Backwards compatibility shim for SmoothQuantModifier.
+
+This module has been moved to llmcompressor.modifiers.transform.smoothquant.
+This shim will be removed in a future version.
+"""
+
+import warnings
+
+warnings.warn(
+    "Importing from 'llmcompressor.modifiers.smoothquant' is deprecated. "
+    "Please update your imports to use 'llmcompressor.modifiers.transform.smoothquant' "
+    "or 'llmcompressor.modifiers.transform' instead. "
+    "This compatibility shim will be removed in a future version.",
+    DeprecationWarning,
+    stacklevel=2,
+)
+
+from llmcompressor.modifiers.transform.smoothquant import *
+from llmcompressor.modifiers.transform.smoothquant.base import SmoothQuantModifier
+
+__all__ = ["SmoothQuantModifier"]
--- a/docs/developer/index.md
+++ b/docs/developer/index.md
@@ -14,7 +14,7 @@ LLM Compressor is an open-source project that values community contributions. We
 
     Our community guidelines ensure that participation in the LLM Compressor project is a positive, inclusive, and respectful experience for everyone.
 
-    [:octicons-arrow-right-24: Code of Conduct](code-of-conduct.md)
+    [:octicons-arrow-right-24: Code of Conduct](../../CODE_OF_CONDUCT.md)
 
 - :material-source-pull:{ .lg .middle } Contributing Guide
 
@@ -22,6 +22,6 @@ LLM Compressor is an open-source project that values community contributions. We
 
     Learn how to effectively contribute to LLM Compressor, including reporting bugs, suggesting features, improving documentation, and submitting code.
 
-    [:octicons-arrow-right-24: Contributing Guide](contributing.md)
+    [:octicons-arrow-right-24: Contributing Guide](../../CONTRIBUTING.md)
 
 </div>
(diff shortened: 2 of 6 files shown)


Etelis:
> looks good, lets maintain BC for at least a version, can you either edit the modifiers init to overwrite getattr and capture smoothquant imports and warn and then redirect or leave a shim in the old smoothquant init location and warn and reimport?
> 
> a bunch of links are broken and should be fixed irrespective of the BC fix

All addressed, anything else you want me to do?

HDCharles (approved):

Etelis:
@brian-dellabetta the deprecated import shim is already in place — `from llmcompressor.modifiers.smoothquant import SmoothQuantModifier` still works, just emits a DeprecationWarning.

re: using compressed-tensors transform primitives — yeah that makes sense but feels like a separate PR, this one is just the organizational move.

src/llmcompressor/modifiers/smoothquant/__init__.py
```diff
@@ -1,3 +1,23 @@
...
+    "Importing from 'llmcompressor.modifiers.smoothquant' is deprecated. "
+    "Please update your imports to use 'llmcompressor.modifiers.transform.smoothquant' "
+    "or 'llmcompressor.modifiers.transform' instead. "
+    "This compatibility shim will be removed in a future version.",
+    DeprecationWarning,
+    stacklevel=2,
+)
+
+from llmcompressor.modifiers.transform.smoothquant import *
+from llmcompressor.modifiers.transform.smoothquant.base import SmoothQuantModifier
```
kylesayrs, line 21: Do we need both this and the line above?

HDCharles:
> Hey guys, this would be a breaking change if users can no longer do
> 
> ```python
> from llmcompressor.modifiers.smoothquant import SmoothQuantModifier
> ```
> 
> I think if we want to do this we should still allow a deprecated import from that module.
> 
> I was under the impression a migration to transform submodule would make use of [these primitives](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/transform/spinquant/base.py#L6-L11) in the compressed-tensors transforms submodule, like the other transform-based modifiers do?

i didn't realize they used different primitives so didn't include it in the issue, we can make that a separate issue though

kylesayrs:
@Etelis Could you please rebase on main? I'll make sure this gets in

Etelis:
> @Etelis Could you please rebase on main? I'll make sure this gets in

Sure give me an hour

Etelis:
> @Etelis Could you please rebase on main? I'll make sure this gets in

Done :)

PR #2319, opened 2026-01-30
Update vLLM GPU Utilization 

dsikka:
SUMMARY:
- Appers as if oneshot memory is not freed resulting in gpu_memory_utilization errors with vLLM which requires 90% of the gpu to launch the engine
- We can set this value to something lower but likely requires some investigation as was not previously needed
cc @kylesayrs


kylesayrs:
I believe this should be fixed by https://github.com/vllm-project/llm-compressor/pull/2375

PR #1561, opened 2025-06-17
Use torch.compile to speed up GPTQ algo

aladerran:
SUMMARY:
In response to #1496, this PR uses torch.compile to speed up the GPTQ quantization process in gptq_quantize.py, along with simple benchmarking tools.

I tested on a single NVIDIA A100-SXM4-80GB, with:
PyTorch version: 2.7.0+cu126
CUDA version: 12.6
cuDNN version: 90501

[gptq_baseline_profile.txt](https://github.com/user-attachments/files/20777203/gptq_baseline_profile.txt)
[gptq_tc_profile.txt](https://github.com/user-attachments/files/20777204/gptq_tc_profile.txt)

TEST PLAN:
```bash
# Run with torch.compile
CUDA_VISIBLE_DEVICES=0 python benchmarks/gptq_runtime_profiler.py > gptq_tc_profile.txt 2>&1
# Baseline run without torch.compile
CUDA_VISIBLE_DEVICES=0 python benchmarks/gptq_runtime_profiler.py > gptq_baseline_profile.txt 2>&1
```

First-time contributor here, please let me know if you have any tips!

Issue #1496, opened 2025-06-01
Use torch.compile to speed up GPTQ algorithm


dsikka:
The bulk of the runtime associated with the GPTQ algorithm is implemented in this helpers file [src/llmcompressor/modifiers/quantization/gptq/gptq_quantize.py](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/gptq/gptq_quantize.py)

Maybe there are some opportunities to compile portions of this code?

Please include some runtime benchmarks to validate the changes

dsikka:
@ved1beta

colldata79:
can i look at this ? 

@ved1beta  is this back to active or can i pick this up?

kylesayrs:
@colldata79 Please see #1561 thread

Getting torch compile to work consistently across all users has been difficult. If you have an interest/experience with torch.compile, I can assign this ticket to you!

colldata79:
yes please assign it to me

colldata79:
@kylesayrs  i have a couple of quick questions before i jump onto this week

“What performance bar makes this worth merging? (e.g., ≥1.3× on A100? measured on which model sizes?)”
“What correctness bar? (bitwise match, or ‘no meaningful perplexity regression’?)”
“Which CI/tests should cover it—unit tests only, or an E2E perf test?”

and please can you or @dsikka  please assign this to me

HDCharles:
good questions:

we landed an [AWQ PR](https://github.com/vllm-project/llm-compressor/pull/2265) for performance improvements and that may be a good idea to compare with

The speedups were relatively small, though consistently better than before across architectures and settings. If some models do better and others worse, it'll likely require a larger average speedup to land.

I would look at examples like https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/llama3_example.py and https://github.com/vllm-project/llm-compressor/blob/main/examples/quantizing_moe/qwen_example.py and do similar tests, you can see how i modified them in https://github.com/vllm-project/llm-compressor/pull/2265/changes/d1347840586ccbe6cb9b017cfbb719937136e542

As far as correctness i think I would probably compare compiled and uncompiled outputs, though if e2e to easier then that seems fine too.

other notes: I tried using torch.compile for the AWQ speedup and ran into issues with device on/offloading so watch out for that.
[...]

PR #2320, opened 2026-01-31
Benchmark torch.compile optimization for GPTQ

colldata79:
## Summary
Benchmark and validate `torch.compile` optimization for the GPTQ quantization algorithm.

Addresses #1496
Supersedes #1561

## What was done
- Added `enable_torch_compile` flag to `GPTQModifier`
- Added `@torch.compile(dynamic=True)` decorated `_process_block` function
- Added `quantize_weight_optimized` compiled code path
- Created benchmark harness for reproducibility
- Ran benchmarks on Qwen2.5-3B (V100) and TinyLlama-1.1B
- Validated correctness (compiled vs uncompiled outputs)
- Validated memory (no inflation, no leaks)

## Benchmark Results (Qwen2.5-3B on V100)

| Metric | Value |
|--------|-------|
| Baseline | 879s (14m 39s) |
| Compiled (warm) | 478s (7m 58s) |
| **Speedup** | **1.84x** |
| Compile overhead | ~35s (one-time) |
| Memory overhead | 0% |
| Graph breaks | 0 |

## Validation (TinyLlama-1.1B)

| Test | Result |
|------|--------|
| Memory Sentinel | ✅ 0.03% error |
| Storage-Size Proof | ✅ No inflation |
| Numerical Check | ✅ Functionally equivalent |


colldata79:
This PR is ready for review. Could a maintainer please add the `ready` label?

HDCharles:
Hey, this looks good!

I'd say for this to land there are a few final steps

1) lm eval performance of e.g. the qwen model you generated vs the normal gptq one, or something along those lines. I know you have numerical results but ultimately this is where we will/won't see regression in our [lm eval tests](https://github.com/vllm-project/llm-compressor/tree/main/tests/lmeval)
2) get rid of the majority of the benchmark code, hopefully you didn't spend too long on this. if you want to add a benchmarking folder with some dedicated tests i don't think its a bad idea but what you have is complicated and much of it is not necessary, i.e. checking storage size has nothing to do with the GPTQ algorithm, that would be the model compressor which wasn't modified, many of the other features are similar. A dedicated benchmark script to make it easier to compare compiled and non compiled gptq could be useful (but not necessary) but something as complicated as this makes it harder, not easier to do those tests for a random person looking to do some work. At worst case if we really want the code again, its logged in this commit. I often do this, i.e. I tend to leave a bunch of tests in my [first commit](https://github.com/vllm-project/llm-compressor/pull/2265/changes/d1347840586ccbe6cb9b017cfbb719937136e542) so if someone wants to repro its there, though its not clogging up the actual codebase.

let me know if you need help doing the evaluation

HDCharles (changes requested):
looks good, see comment

colldata79:
Great feedback . i will look at this later in the week to close it out

HDCharles:
another point, we should probably extract some/most of the functionality into helper functions so we don't have the exact same code duplicated across quantize_weight and quantize_weight_optimized. This will be a real pain to maintain otherwise.

HDCharles:
for the API lets expand https://github.com/vllm-project/llm-compressor/pull/2384/changes#diff-8b25e0e7dfb3c229926f751ac7d9dff4784e3e6692cc1038b050e0c1cb21792eR25 and just make this 'enable_torch_compile`

Issue #1853, opened 2025-09-22
[Help Wanted] Update models used in tests


fynnsu:
Some of the models we use in tests have grown outdated. Others are larger than necessary for the nature of the test. Many of our per-commit / nightly tests are also focused on ensuring the model can be compressed/run but don't test the quality of the compressed model. This means we can use a very small "smoke llm" that has the same architecture as a regular llm but is too small to produce real outputs.

#1833 does exactly this to replace one of the test models with a 6-layer llm with less than 0.5M weights. However, there are also other models (such as `TinyLlama/TinyLlama-1.1B-intermediate-step-1431k-3T`) that could potentially be replaced.

# Goals
1. Update tests to use more modern models
2. Reduce the runtime of tests (particularly the transformers check test)

# Suggested Approach
1. Determine the most common test models in use in the repo's tests.
2. Determine if these models could be replaced with a dummy "smoke llm" based on how they are used in tests
3. (If possible) evaluate which models/tests account for the greatest amount of test runtime
4. Replace models in tests with a smaller/newer model from the same model family (or a smoke model designed for that purpose)

sudiptap:
can you please assign this to me?

fynnsu:
Thanks for your help! Let me know if you have any questions

ojeda-e:
Although I understant this issue was assigned last month, I'd like to work on this @fynnsu. Let me know if it's possible to post some questions. Otherwise, it's all good :)

fynnsu:
Hi @sudiptap, are you actively working on this issue or is it okay to hand it off?

sudiptap:
I am working on this, if its ok I would take 3 more days, then I can hand over.

fynnsu:
Hi @sudiptap, any update on this?

sudiptap:
@fynnsu sorry, someone else take this over.

fynnsu:
No worries! @ojeda-e if you still want to work on this, it's yours. Happy to answer any questions you may have!

ojeda-e:
Thanks @fynnsu, yes, I am interested. 

Given the scope of this issue, my questions are not only to comfirm I understant the problem, but also about the strategy to address all the multiple parts it comprises, and to avoid large PRs.

From the suggested approach, 

Q1. The first bullet point:
> - Determine the most common test models in use in the repo's tests.

Is this intended to be included in the documentation (development guide), or just for the purpose of working on this issue and post it here as part of the discussion?

[...]

PR #2322, opened 2026-01-31
[Tests] Update test models to smoke variants for faster runs

LudovicoYIN:
## Summary
- Replace several test model stubs with smoke models to speed up nightly/e2e runs.
- Update SparseGPT configs to use smaller models.
- Rename weights_only nightly config to reflect smoke usage.

## Testing
- `make style`
- `pytest -q tests/llmcompressor/transformers/kv_cache/test_kv_cache.py -vv -s`
- `CADENCE=nightly TEST_DATA_FILE=<e2e_yaml_filename> pytest -q tests/e2e/vLLM/test_vllm.py::TestvLLM::test_vllm -vv -s`
- `CADENCE=nightly TEST_DATA_FILE=llama_consec_runs.yaml pytest -q tests/llmcompressor/transformers/sparsegpt/test_consecutive_runs.py -vv -s`
- `CADENCE=nightly TEST_DATA_FILE=weights_only_smoke_nightly.yaml pytest -q tests/llmcompressor/transformers/compression/test_quantization.py::test_quantization_reload -vv -s`

part of #1853

### DIFF
--- a/src/llmcompressor/datasets/__init__.py
+++ b/src/llmcompressor/datasets/__init__.py
@@ -12,4 +12,5 @@ from .utils import (
     get_calibration_dataloader,
     get_processed_dataset,
     make_dataset_splits,
+    get_rank_partition,
 )
--- a/src/llmcompressor/datasets/utils.py
+++ b/src/llmcompressor/datasets/utils.py
@@ -7,6 +7,7 @@ calibration data preparation, and dataloader creation for both training and
 one-shot calibration workflows.
 """
 
+import math
 import re
 from collections.abc import Iterator, Sized
 from typing import Any, Callable, Optional
@@ -14,6 +15,7 @@ from typing import Any, Callable, Optional
 import torch
 from datasets import Dataset
 from loguru import logger
+from torch import distributed as dist
 from torch.utils.data import DataLoader, RandomSampler, Sampler
 from transformers.data import DataCollatorWithPadding, default_data_collator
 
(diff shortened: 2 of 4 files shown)


tests/e2e/vLLM/configs/qwen3_fp4_nvfp4.yaml
```diff
@@ -1,6 +1,6 @@
 cadence: "nightly"
 test_type: "regression"
-model: Qwen/Qwen3-30B-A3B
+model: nm-testing/tinysmokeqwen3
```
dsikka: We want to use qwen / an MoE here. The architecture of this model does not contain experts / a gate

tests/e2e/vLLM/configs/qwen3_fp8_dynamic_per_token.yaml
```diff
@@ -1,4 +1,4 @@
 cadence: "nightly"
 test_type: "regression"
-model: Qwen/Qwen3-30B-A3B
+model: nm-testing/tinysmokeqwen3
```
dsikka: Same as above

dsikka (changes requested):
The e2e models have been selected to validate specific vLLM integrations. Going too small can also be confusing as we end up generating bad outputs through the tests and it’s hard to know if it’s due to a weight loading issue in vLLM, a quantization issue or just because the model is too small.

I would suggest just starting with nightly cases first

LudovicoYIN:
@dsikka 
Thanks review. I’ve reverted all e2e model changes and the kv_cache change. The only nightly changes left are weights_only_smoke_nightly and the SparseGPT consecutive‑runs smoke model.

I checked the remaining nightly configs — most have perplexity thresholds or fixed stubs. The only open candidate is llama_7b_sparse.yaml (no perplexity). Should I switch it to a smoke model or keep the 7B model?

HDCharles (approved):
these changes seem safe and seems like we can do others in a different PR if desired.

also in general we'be been discussing getting ride of sparse gpt, so moving those tests to smoke tests should be more than sufficient.

Issue #2296, opened 2026-01-27
[AWQ] add option to take smooth layer quantization into accout


HDCharles:
normally the way AWQ works is to pick a layer that is going to be quantized, try a bunch of scale factors to find the one that minimizes quantization error when that layer is quantized and then do an [inverse rescale](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L520) on the preceeding layer which is normally not quantized. However a problem arises for the [up_proj -> down_proj](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/mappings.py#L51-L54) mapping
because both the smooth and balance layers are targeted for quantization. Since we only take into account the quantization of the balance layers in our current AWQ implementation, we could be making the smooth layer harder to quantize with our choice of scale factor for the balance layer since the smooth layer is basically ignored during the quantization error calculation for quantizing the balance layer.

We should

1) test if this has a significant impact
2) add an option to enable this feature if its beneficial

STEPS

A) add a check [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L378) for whether smooth_name is in targeted_names and if so, change the get_lowest_common...etc search to include the smooth layer (this is how we determine what module is run to determine the quantization error, so we need smooth layer to be run if we're taking its quantization into accout)
B) add a flag to [compute_best_scale](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L567) for if the smooth layer is targeted
C) if necessary add the smooth layer to [this dict](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L497)
D) move [the rescale weight code](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L664-L696) into a function which is called for each balance layer
E) if necessary, call the rescale weight code for 1/_scalesview on the smooth_layer
F) check whether this has an impact on lm_eval performance on some small set of models.
G) check how this affects the runtime of AWQ for those models.

if its beneficial then put up a PR with those changes demonstrating what was tested and how it affects things.

Etelis:
Assign me here as well

Ramshankar07:
Can I work on it @Etelis , I'm eager to learn and contribute

Etelis:
[...]

PR #2323, opened 2026-01-31
[AWQ] Add option to consider smooth layer quantization in scale search

Ramshankar07:
SUMMARY

This PR adds an option to take smooth layer quantization into account when computing AWQ scales, e.g. for the up_proj → down_proj mapping in transformer FFN blocks.

This is PR for : #2296

Problem

AWQ picks scale factors by minimizing quantization error only for the balance layer (e.g. down_proj) and applies an inverse rescale to the smooth layer (e.g. up_proj), which is usually assumed unquantized. When both are quantization targets, the scale chosen for the balance layer can worsen quantization of the smooth layer because the smooth layer’s quantization error is not included in the objective.

Solution

- `smooth_layer_quantization` flag (default: `False`) on `AWQModifier`. When enabled and the smooth layer is in the quantization target list:
 - Ancestor used for forward / error:** The module used for running samples (and thus for the quantization error) is chosen via `get_lowest_common_ancestor_with_avoid` over both balance layer names and  the smooth layer name.
 - Scale search: `_compute_best_scale` is called with `smooth_layer_targeted=True`. The smooth layer is added to `orig_layer_weights`, and during the grid search we rescale it by 1/s, quantize to Q(W/s).
 - Applying scales: After the best scale is chosen, the inverse rescale is applied to the smooth layer (and balance layers) in the existing `_smooth` logic. When the smooth layer is in `orig_layer_weights`, we apply rescaling from `orig_layer_weights[smooth_layer]` so the stored weights are W/s for later calibration, avoiding double rescaling of the quantized grid-search weights.


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -654,37 +743,58 @@ def _compute_best_scale(
...
-                        / _scalesview
-                    ).to(balance_layer.weight.dtype)
+                    # Quantize the rescaled weights: Q(W / s)
+                    # The quantized weights will be used as-is in the forward pass
+                    smooth_layer.weight.data = forward_quantize(
+                        smooth_layer,
+                        smooth_layer.weight,
+                        "weight",
+                        w_qscheme,
+                    ).to(smooth_layer.weight.dtype)
```
HDCharles: this should all be coming from the helper function

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -518,11 +554,18 @@ def _smooth(
...
                     elif module == smooth_layer:
-                        if module.weight.ndim == 1:
+                        # When smooth_layer_quantization is enabled, smooth_layer.weight
+                        # was overwritten during grid search (Q(W/s)); use original so we
+                        # apply W/s for later calibration. Otherwise use current weight.
+                        if smooth_layer in orig_layer_weights:
+                            w = orig_layer_weights[smooth_layer].to(scales.device)
+                        else:
+                            w = smooth_layer.weight
+                        if smooth_layer.weight.ndim == 1:
```
HDCharles: do the same thing as we do for the balance layers part of the if branch, no reason to store to memory, then multiply and store to memory again when you can just multiply and store to memory once.

HDCharles: also need to modify the else branch below in the same way

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -654,37 +743,58 @@ def _compute_best_scale(
                 scales[torch.isinf(scales)] = 1
                 scales[torch.isnan(scales)] = 1
 
-                # Q(W * s)
+                # Q(W * s) for balance layers
                 for balance_layer in balance_layers_to_patch:
```
HDCharles: better to iterate over layers_to_patch and then just check if its the smooth layer and change the _scalesview if so

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -564,11 +607,47 @@ def _run_samples(self, module: Module) -> list[torch.Tensor]:
...
+            )
+            / _scalesview
+        ).to(balance_layer.weight.dtype)
+
     def _compute_best_scale(
         self,
         mapping: ResolvedMapping,
         fp16_outputs: list[torch.Tensor],
         orig_layer_weights: dict[torch.nn.Module, torch.Tensor],
+        smooth_layer_targeted: bool = False,
```
HDCharles: need a better name for this,

we aren't doing 'smooth_layer_targeted'

we are doing smooth_layer quantization or taking smooth layer quantization error into accout because the smooth layer is targeted for quantization.

if we have a function that does X because Y is true, we want the flag to be named X rather than Y if possible.

better name is 

quantize_both or also_quantize_smooth_layers or something like that.

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -616,17 +697,25 @@ def _compute_best_scale(
             if hasattr(balance_layer, "quantization_scheme")
             and hasattr(balance_layer.quantization_scheme, "weights")
         ]
+        smooth_layer_to_patch = None
+        if mapping.smooth_layer in orig_layer_weights:
+            smooth_layer_to_patch = mapping.smooth_layer
+
+        layers_to_patch = balance_layers_to_patch
+        if smooth_layer_to_patch is not None:
+            layers_to_patch = balance_layers_to_patch + [smooth_layer_to_patch]
```
HDCharles: the logic here is a bit mixed up,

you're passing in a flag that is supposed to determine whether you are doing smooth layer quantization but then you don't use it while you do this logic.

since smooth_layer_targeted is useless if smooth layer isn't in orig_layer weights, we should have an assert.

1) add an assert to make sure the user didn't pick that boolean without including the smooth layer in the orig_layer_weights
2) otherwise this section should just be changed to be layers to patch with specific balance layer or smooth layer logic depending on a check of whether its the smooth layer or not.

src/llmcompressor/modifiers/awq/base.py
HDCharles: like this should just be layers_to_patch and include the smooth_layer if necessary

HDCharles (changes requested):
thanks for your contribution! overall looks good but see comments.

otherwise we should run some of the eval tests and see the results, let me know if you need help with that

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: Moved it to helper function

Ramshankar07:
I'll do the lm_eval and update by tomorrow

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -531,7 +575,7 @@ def _smooth(
                             # in this case, default to scaling the last output features
                             # because the desired smooth layer is v_proj
                             # https://github.com/casper-hansen/AutoAWQ/blob/main/awq/quantize/scale.py#L123
-                            weight = module.weight
+                            weight = w.to(module.weight.device).clone()
```
HDCharles: you already get w above, right? can reuse that, no need to clone it

can clean this up and make naming more consistent:

```suggestion
                        weight = (
                            orig_layer_weights[smooth_layer]
                            if smooth_layer in orig_layer_weights
                            else smooth_layer.weight
                        )
                        if smooth_layer.weight.ndim == 1:
                            update_offload_parameter(
                                module,
                                "weight",
                                weight.to(module.weight.device) / scales,
                            )
                        else:
                            # NOTE: edge case when smooth layer number of out_features
                            # is not equal to balance layer number of in_features
                            # e.g. when fused qkv_proj is used to smooth o_proj
                            # in this case, default to scaling the last output features
                            # because the desired smooth layer is v_proj
                            # https://github.com/casper-hansen/AutoAWQ/blob/main/awq/quantize/scale.py#L123
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -564,11 +608,94 @@ def _run_samples(self, module: Module) -> list[torch.Tensor]:
...
+            forward_quantize(
+                balance_layer,
+                balance_layer.weight,
+                "weight",
+                w_qscheme,
+            )
+            / _scalesview
+        ).to(balance_layer.weight.dtype)
+
+    def _apply_smooth_layer_quantization_in_grid_search(
```
HDCharles: this is suplerfluous

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -564,11 +608,94 @@ def _run_samples(self, module: Module) -> list[torch.Tensor]:
             for output in outputs
         ]
 
+    def _apply_balance_layer_quantization_in_grid_search(
```
HDCharles: rename this to rescale_layer

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -655,43 +800,22 @@ def _compute_best_scale(
...
-                            "weight",
-                            w_qscheme,
+                for layer in layers_to_patch:
+                    if layer == smooth_layer_to_patch:
+                        self._apply_smooth_layer_quantization_in_grid_search(
+                            layer, orig_layer_weights, scales, device
+                        )
+                    else:
+                        self._apply_balance_layer_quantization_in_grid_search(
+                            layer, orig_layer_weights, _scalesview
```
HDCharles: just use a single function and invert scalesview for if its the smoothlayer

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -655,43 +800,22 @@ def _compute_best_scale(
...
-                    balance_layer.weight.data = (
-                        forward_quantize(
-                            balance_layer,
-                            balance_layer.weight,
-                            "weight",
-                            w_qscheme,
+                for layer in layers_to_patch:
+                    if layer == smooth_layer_to_patch:
+                        self._apply_smooth_layer_quantization_in_grid_search(
+                            layer, orig_layer_weights, scales, device
```
HDCharles: don't use scales

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -610,23 +741,37 @@ def _compute_best_scale(
...
+                "orig_layer_weights; include the smooth layer in orig_layer_weights "
+                "when enabling smooth layer quantization in the grid search."
+            )
+            layers_to_patch = layers_to_patch + [mapping.smooth_layer]
+        smooth_layer_to_patch = (
+            mapping.smooth_layer if also_quantize_smooth_layers else None
+        )
+        balance_layers_in_patch = [
+            layer for layer in layers_to_patch if layer in mapping.balance_layers
+        ]
```
HDCharles: neither of these are necessary

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -531,7 +575,7 @@ def _smooth(
                             # in this case, default to scaling the last output features
                             # because the desired smooth layer is v_proj
                             # https://github.com/casper-hansen/AutoAWQ/blob/main/awq/quantize/scale.py#L123
-                            weight = module.weight
+                            weight = w.to(module.weight.device).clone()
                             weight[-scales.size(0) :].div_(scales.view(-1, 1))
```
HDCharles, line 621: don't do this in place

HDCharles (changes requested):
see comments

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: I removed the conversion

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: changed to _rescale_balance_layer

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: unified into a single function

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: Changed it to scales_view

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -603,11 +647,59 @@ def _run_samples(self, module: Module) -> list[torch.Tensor]:
...
+        so that rescale is W * (1/s) = W/s.
+        """
+        w_qscheme = layer.quantization_scheme.weights
+        orig = orig_layer_weights[layer].to(scales_view.device)
+
+        if is_smooth_layer and layer.weight.ndim > 1:
+            # Edge case: only scale the last scales.size(0) rows (divide by s)
+            scales_1d = (1.0 / scales_view).view(-1)
+            w = orig.clone()
+            w[-scales_1d.size(0) :].div_(scales_1d.view(-1, 1))
```
HDCharles: why are we inverting this 2 times, shouldn't it be 0 times since you calculate _scalesview_inv outside the function?

HDCharles:
ok looks good aside from that final bit, now lets see if we can show that this actually improves accuracy, do you need help running evaluations?

HDCharles:
@ZewenShen-Cohere do you have any results with this technique that we should compare with

ZewenShen-Cohere:
> @ZewenShen-Cohere do you have any results with this technique that we should compare with

Not at the moment. The AWQ code has gone through substantial changes since October, so my earlier results wouldn’t align well for a fair comparison.

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: This branch is for multi dimension, we need scales again to implement the “only last k rows divided by s” behavior, so we invert back to just take last K.

Ramshankar07:
> ok looks good aside from that final bit, now lets see if we can show that this actually improves accuracy, do you need help running evaluations?

I just got some modal credits. I got struck up with some work. I'll make two scripts to find the difference and time. 

Can suggest me what model and dataset to pick and try.
I thought to go with llama3.2 1b and some small dataset

HDCharles:
our existing tests include:

https://github.com/vllm-project/llm-compressor/blob/main/tests/lmeval/configs/w4a16_awq_sym.yaml

so if you can do that, i would pick that one, if thats too large then i'd probably do tinyllama or something similar.

Ramshankar07:
I just finished the testing using 
``` 
DATASET_SPLIT = "train_sft"
NUM_CALIBRATION_SAMPLES = 256
MAX_SEQUENCE_LENGTH = 2048 
```
Here are the results:


```
========== AWQ RUNTIME COMPARISON ==========
  Baseline (no smooth_layer_quantization): 470.78s
  With smooth_layer_quantization:           550.98s
  Difference: +80.20s (+17.0%)
=============================================


========== LMEVAL METRICS COMPARISON ==========
  alias: baseline=gsm8k  with_smooth=gsm8k
  exact_match,flexible-extract: baseline=0.72  with_smooth=0.69
  exact_match,strict-match: baseline=0.72  with_smooth=0.7
  exact_match_stderr,flexible-extract: baseline=0.04512608598542127  with_smooth=0.04648231987117316
  exact_match_stderr,strict-match: baseline=0.04512608598542127  with_smooth=0.046056618647183814
===============================================

========== LMEVAL THRESHOLD CHECKS ==========
  exact_match,strict-match: PASS (baseline=0.720, with_smooth=0.700, target>=0.70, recovery>=0.91x)
  exact_match,flexible-extract: FAIL (baseline=0.720, with_smooth=0.690, target>=0.70, recovery>=0.91x)
===============================================
```

HDCharles:
hmm mixed results is surprising, @ZewenShen-Cohere, thoughts?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -533,22 +555,38 @@ def _apply_smoothing(self, model: Module) -> None:
...
+                    also_quantize_smooth_layers=smooth_layer_targeted,
                 )
 
                 @torch.no_grad()
                 def _smooth(
                     module: Module, orig_layer_weights: dict[Module, torch.Tensor]
                 ):
-                    scales = best_scales.to(module.weight.device)
+                    scales = best_scales.to(
+                        device=module.weight.device, dtype=module.weight.dtype
```
HDCharles: is the dtype shift necessary?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -533,22 +555,38 @@ def _apply_smoothing(self, model: Module) -> None:
...
                 orig_layer_weights = {
                     balance_layer: balance_layer.weight.clone()
                     for balance_layer in mapping.balance_layers
                     if hasattr(balance_layer, "quantization_scheme")
                     and hasattr(balance_layer.quantization_scheme, "weights")
                 }
+                if (
+                    smooth_layer_targeted
+                    and hasattr(smooth_layer, "quantization_scheme")
+                    and hasattr(smooth_layer.quantization_scheme, "weights")
```
HDCharles: shouldn't these just be in smooth_layer_targeted?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -570,9 +616,9 @@ def _smooth(
                             # in this case, default to scaling the last output features
                             # because the desired smooth layer is v_proj
                             # https://github.com/casper-hansen/AutoAWQ/blob/main/awq/quantize/scale.py#L123
-                            weight = module.weight
-                            weight[-scales.size(0) :].div_(scales.view(-1, 1))
-                            update_offload_parameter(module, "weight", weight)
+                            w = w.to(module.weight.device).clone()
```
HDCharles: why clone?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -687,50 +798,41 @@ def _compute_best_scale(
...
+                    balance_layers := [
+                        layer
+                        for layer in layers_to_patch
+                        if layer in mapping.balance_layers
+                    ]
+                ) and all(
                     getattr(layer.quantization_scheme.weights, "strategy", None)
                     == QuantizationStrategy.TENSOR_GROUP
-                    for layer in balance_layers_to_patch
+                    for layer in balance_layers
```
HDCharles:
```suggestion
                    for layer in layers_to_patch
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -603,11 +649,59 @@ def _run_samples(self, module: Module) -> list[torch.Tensor]:
             for output in outputs
         ]
 
+    def _rescale_layer(
```
HDCharles: should probably be _rescale_and_fake_quantize_layer

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -603,11 +649,59 @@ def _run_samples(self, module: Module) -> list[torch.Tensor]:
...
+        orig = orig_layer_weights[layer].to(scales_view.device)
+
+        if is_smooth_layer and layer.weight.ndim > 1:
+            # Edge case: only scale the last scales.size(0) rows (divide by s)
+            scales_1d = scales_view.reciprocal().view(-1)
+            w = orig.clone()
+            w[-scales_1d.size(0) :].div_(scales_1d.view(-1, 1))
+            layer.weight.data.copy_(w)
+        else:
+            layer.weight.data.copy_(orig * scales_view)
```
HDCharles:
```suggestion
        target = layer.weight.data
        if is_smooth_layer and layer.weight.ndim > 1:
            partial_slice = slice(scales_view.numel(), None)
            orig = orig[*partial_slice]
            target = layer.weight.data[*partial_slice]
        target.copy_(orig * scales_view)
```

this double inversion and cloning and the in place and everything can be simplified

HDCharles:
can you measure the result on main vs the result without taking smooth layer into account? are they the same?

src/llmcompressor/modifiers/awq/base.py
HDCharles: the inversion doesn't change the dimensionality though

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: yes, I was getting dtype error previously due to promotion of dtypes. This helps in matching it.

src/llmcompressor/modifiers/awq/base.py
Ramshankar07: yup! my bad. I just moved it :)

src/llmcompressor/modifiers/awq/base.py
Ramshankar07:
```
w[-scales.size(0) :].div_(scales.view(-1, 1))
```
The next line is a in-place operation, so I thought it might cause problems if I directly change it.

Ramshankar07:
I'm getting some problems while comparing main llm-composer repo vs without smooth quant, I'll debug closely and try to replicate main repo's result.

to add: I'm getting 0 as error.

HDCharles:
ok lets get the results working and see if the algorithm is useful and we can discuss the code a bit more later. the more i look at this, the more i realize I'm basically walking you through a full code refactor which should be beyond the scope of this issue.

At this point the code is good enough to focus on the results and making sure we didn't break things vs main and seeing if there's a clear use case somewhere. If we get some good results I can take a more hands on approach with the refactor to make this comment revision wackamole easier.

let me know if you need help, especially if you have scripts that you'd want to run but lack gpu resources. you can also reach out to me on the vllm slack (same name/picture) if that's easier

Ramshankar07:
I just finished the testing using 
``` 
DATASET_SPLIT = "train_sft"
NUM_CALIBRATION_SAMPLES = 256
MAX_SEQUENCE_LENGTH = 2048 
```
Here are the results:

| Config       | AWQ time (s) | Exact match (strict) | StdErr (strict) | Exact match (flex) | StdErr (flex) |
|-------------|--------------|----------------------|------------------|--------------------|---------------|
| without smooth   | 479.54       | 0.20                 | 0.0402           | 0.21               | 0.0409        |
| With smooth | 595.44       | 0.24                 | 0.0429           | 0.25               | 0.0435        |

With-smooth looks modestly better in accuracy on gsm8k (about +4 percentage points but it takes some more time

HDCharles:
something is up here, on main for gsm8k 

we have regression tests:

https://github.com/vllm-project/llm-compressor/blob/main/tests/lmeval/configs/w4a16_awq_sym.yaml

we get .70 at least on gsm8k and i've been doing gptq which is giving .702.

Can you provide your repro?

Ramshankar07:
https://github.com/Ramshankar07/llm-compressor/tree/awq-smooth-layer-quantization this is the repo link

HDCharles:
i'll take a look and see, we're trying to get out a release right now though so may be a day or two. Also can you reach out to me on vllm slack under the same username?

Ramshankar07:
Sure.

PR #2331, opened 2026-02-05
[AutoRound] Add DP Support

yiliu30:
SUMMARY:
Add DDP support for Autoround and use Qwen as example.

TEST PLAN:


yiliu30:
Clean the implementation in https://github.com/vllm-project/llm-compressor/pull/2411

PR #2332, opened 2026-02-05
[bug][awq] fix inf handling

HDCharles:
Must have been a bad merge or rebase at some point, scalesview was being set before the inf/nan check

TEST PLAN:
CI


HDCharles:
yeah if .to moves the device then its going to be an issue

PR #2333, opened 2026-02-06
[GPTQ][ddp] enabling DDP for GPTQ

HDCharles:
After the changes in https://github.com/vllm-project/compressed-tensors/pull/572 https://github.com/vllm-project/compressed-tensors/pull/534 https://github.com/vllm-project/llm-compressor/pull/2340 we're ready to start rolling out DDP implementations of various modifiers

# API:
The Api we've landed on attempts to maintain the normal flow with minimal changes necessary to enable DDP:

1) the user will call torchrun --nproc_per_node==<num_threads> script.py to start the script
2) the user will initialize the distributed context, (they can use the helper init_dist to do this)
3) the user will load the model using the new context manager, setting the device map as outlined [here](https://github.com/vllm-project/compressed-tensors/pull/572). (For most users this will be "auto_offload")
4) (optional) the user can partition the dataset at load time using get_rank_partition or just load as normal and oneshot will partition the data later (will load 1 copy of dataset into cpu memory for each rank which may be onerous)
```python
from compressed_tensors.offload import load_offloaded_model, init_dist
init_dist()
with load_offloaded_model(): 
    model = AutoModelForCausalLM.from_pretrained(MODEL_ID, device_map="auto_offload")
...
ds = load_dataset(
    DATASET_ID, split=get_rank_partition(DATASET_SPLIT, NUM_CALIBRATION_SAMPLES)
```

# Implementation
Adding the DDP process to GPTQ has relatively straightforward though optimizing it for speed was a bit trickier. There are 4 steps

1) assigning each module to a rank which it will be compressed by
2) for each module assigned to a rank, having all hessian information sent by other ranks to the assigned rank
3) each rank compresses the modules that it was assigned
4) broadcast the final quantized values to all ranks

Step 1 required the largest optimization, without any load balancing, we ran into situations where 1 rank could be doing twice as much work as another. Thus we implemented basic load balancing and time estimation that seems to be working well in practice. The other major optimization was using asynchronous ops for thread to thread communication. Before these optimizations, 2 thread GPTQ was as fast as 1 thread GPTQ for llama3-8B, afterward it results in a 27% speedup despite being a relatively small model.

| model_id | world_size | max_time | max_memory | save_time | flex_extract | eval_time |
|----------|-------------|----------|------------|-----------|--------------|-----------|
| Meta-Llama-3-8B-Instruct |  1 | 745.03 | 5.82 | 19.57 | 0.7066 | 95.28 |
| Meta-Llama-3-8B-Instruct | 2 | 372.20 | 5.57 | 49.10 | 0.7089 | 95.24 |
| Meta-Llama-3-8B-Instruct | 4 | 264.07 | 5.82 | 52.50 | 0.7180 | 96.74 |
| Qwen3-30B-A3B | 1 | 14207.53 | 6.56 | 748.23 | 0.8704 | 209.93 |
| Qwen3-30B-A3B | 2 | 7018.25 | 6.36 | 696.65 | 0.8810 | 205.89 |
| Qwen3-30B-A3B | 4 | 3694.46 | 6.36 | 723.05 | 0.8832 | 217.62 |

# GPTQ Changes

while validating numerical accuracy of the DDP technique, we noticed that accuracy improved significantly for each thread added. After some debugging we realized this was because the existing [hessian calculation](https://github.com/vllm-project/llm-compressor/pull/2333/changes#diff-18d1319f01629ca65cc54f955dc6177f6dd025f057013932b2ed29842854f3ecL61-L65) was causing an accumulation of floating point errors. By rewriting the hessian calculation to sum the intermediate hessians and only divide by num_samples at the end, we improved the GSM8K evaluation from (.67, .66) to (.71, .71). You can repro these results [here](https://github.com/vllm-project/llm-compressor/pull/2333/changes#diff-d31ce0453051853c17ba2a5225b3d1bfab548e095bab0967d6acfd1b3ce1b35d)

### DIFF
--- a/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
+++ b/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
@@ -9,6 +9,7 @@ from compressed_tensors import (
     SparsityCompressionConfig,
 )
 from compressed_tensors.config import CompressionFormat
+from compressed_tensors.offload import is_rank0
 from loguru import logger
 from transformers import PreTrainedModel
 
@@ -88,22 +89,23 @@ def modify_save_pretrained(model: PreTrainedModel):
             if compressor is not None:
                 compressor.compress_model(model)
 
-            # save (compressed) model structure
-            original_save_pretrained.__get__(model, model_class)(
-                save_directory,
-                safe_serialization=safe_serialization,
-                **kwargs,
-            )
+            if is_rank0():
+                # save (compressed) model structure
+                original_save_pretrained.__get__(model, model_class)(
+                    save_directory,
+                    safe_serialization=safe_serialization,
+                    **kwargs,
+                )
 
-            # update config to reflect compression
-            if compressor is not None:
-                compressor.update_config(save_directory)
+                # update config to reflect compression
+                if compressor is not None:
+                    compressor.update_config(save_directory)
 
-            # update existing recipe
-            update_and_save_recipe(model.name_or_path, save_directory)
+                # update existing recipe
+                update_and_save_recipe(model.name_or_path, save_directory)
 
-            # copy python files from cache dir to save_path if any
-            copy_python_files_from_model_cache(model, save_directory)
+                # copy python files from cache dir to save_path if any
+                copy_python_files_from_model_cache(model, save_directory)
 
         save_pretrained_wrapper._overridden = True
         return save_pretrained_wrapper


src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -263,35 +268,100 @@ def compress_modules(self):
         """
         Quantize modules which have been calibrated
         """
-        for module in list(self._num_samples.keys()):
+        ### Not Distributed
+        if not is_distributed():
+            self.compress_module_list(list(self._num_samples.keys()))
+
```
kylesayrs, line 279: Do we need an `else` here?

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -263,35 +268,100 @@ def compress_modules(self):
...
                 align_module_device(module),
                 self._maybe_onload_hessian(module),
                 CompressionLogger(module) as comp_logger,
             ):
-                loss, quantized_weight, scale, zero_point, g_idx = quantize_weight(
+                loss, q_param_dict = quantize_weight(
                     module=module,
                     quant_args=quant_args,
-                    hessians_dict=self._hessians,
+                    hessian=self._hessians[module] / self._num_samples[module],
```
kylesayrs: Consider passing `num_samples` as an arg to `quantize_weight`

HDCharles: i'm unsure why this was implemented by passing entire dicts originally. Seems like i'd rather make the function more explicit on what its acting on i.e. what i have here.

src/llmcompressor/modifiers/quantization/gptq/base.py
HDCharles: added return

src/llmcompressor/modifiers/quantization/gptq/base.py
kylesayrs: I agree that not passing a dictionary is cleaner, but it comes at a memory cost since we cannot "move" the hessian memory. This is an instance where I feel like (better behavior) is preferable to (cleaner code)

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -263,35 +268,101 @@ def compress_modules(self):
...
+                )
+                n_comm = dist.reduce(
+                    n, op=dist.ReduceOp.SUM, dst=target_rank, async_op=True
+                )
+                pending_comms.append(h_comm)
+                pending_comms.append(n_comm)
+                if rank == target_rank:
+                    wait_for_comms(pending_comms)
+                    self._hessians[module] = H
+                    self._num_samples[module] = n
```
kylesayrs: I don't fully understand this logic. Why is this step needed? Can't you just write to the memory address directly using
```python3
for module in module_list:
    h_comm = dist.reduce(
        self._hessians[module],
        op=dist.ReduceOp.SUM,
        dst=target_rank,
        async_op=True
    )

    pending_comms.append(h_comm)

wait_for_comms(pending_comms)
```

This way seems to maximize throughput more than how it's written now, right?

kylesayrs: It seems like you do a similar approach when broadcasting.

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -263,35 +268,101 @@ def compress_modules(self):
...
+
+            # Get parameters from module
+            to_broadcast = []
+            for attr in _GPTQ_Q_PARAMS:
+                if getattr(module, attr, None) is not None:
+                    to_broadcast.append(getattr(module, attr, None))
+
+            # Broadcast each tensor asynchronously
+            # note: update in place, since compress_module_list updated the offload
+            for tensor in to_broadcast:
```
kylesayrs: What's the benefit of splitting into two for loops? Why not just write

```python3
for module in module_list:
    for attr in _GPTQ_Q_PARAMS:
        if (tensor := getattr(module, attr, None) is not None:
            pending_comms.append(dist.broadcast(tensor, ...))
````

File: src/llmcompressor/modifiers/quantization/gptq/base.py (hunk 2 above)
Comment (kylesayrs) [reply src/llmcompressor/modifiers/quantization/gptq/base.py]: Spoke offline, we agreed to pop the value from the dict.

File: src/llmcompressor/utils/dist.py [hunk 5]
```diff
@@ -0,0 +1,54 @@
+from typing import Callable, Hashable, TypeVar
+
+import torch.distributed as dist
+
+T = TypeVar("T", bound=Hashable)
+
+
+def greedy_bin_packing(
```
kylesayrs, line 8: Beautiful

examples/quantization_w4a16/llama3_ddp_example.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+
+ds = ds.map(tokenize, remove_columns=ds.column_names)
+
+recipe = GPTQModifier(targets="Linear", scheme="W4A16", ignore=["lm_head"])
+
+
+torch.cuda.reset_peak_memory_stats()
+start_time = time.time()
```
kylesayrs, line 73: Did you mean to keep these?

src/llmcompressor/modifiers/quantization/gptq/base.py
```diff
@@ -263,35 +268,101 @@ def compress_modules(self):
...
+            self._num_samples.pop(module, None)
+
+    def _reduce_hessian_to_target_rank(self, module_list, module_to_rank):
+        rank = dist.get_rank()
+        pending_comms = []
+        for module in module_list:
+            target_rank = module_to_rank[module]
+            with self._maybe_onload_hessian(module):
+                H = self._hessians[module]
+                n = torch.Tensor([self._num_samples.get(module)]).to(H.device)
```
kylesayrs: Do you want to refactor `_num_samples` into scalar torch tensors? That way you don't have to do any casting.

examples/quantization_w4a16/llama3_ddp_example.py
HDCharles: Yeah for a purely quant speed improvement feature, measuring the quant time seemed important

src/llmcompressor/modifiers/quantization/gptq/base.py
HDCharles: Yeah local minima code after redactors that should be fixed

src/llmcompressor/modifiers/quantization/gptq/base.py
HDCharles: Yeah holdover from dist_impl

PR #2335, opened 2026-02-06
[MoE] MiniMax-M2/M2.1 calibration follow-up

LudovicoYIN:
## Summary
Add initial MiniMax-M2 / M2.1 support in llm-compressor.

### Changes
- Add MiniMax-M2 MoE calibration integration.
- Add MiniMax-M2 calibration tests.
- Add MiniMax-M2 quantization example.

### Files
- `src/llmcompressor/modeling/minimax_m2.py`
- `tests/llmcompressor/modeling/test_calib_minimax_m2.py`
- `examples/quantizing_moe/minimax_m2_example.py`

## Test Plan
- `CADENCE=weekly /root/miniforge3/envs/llm-compressor/bin/python -m pytest tests/llmcompressor/modeling/test_calib_minimax_m2.py -k minimax -rs`
- `/root/miniforge3/envs/llm-compressor/bin/python examples/quantizing_moe/minimax_m2_example.py`

### DIFF
--- a/examples/quantization_w4a4_fp4/qwen3_5_example.py
+++ b/examples/quantization_w4a4_fp4/qwen3_5_example.py
@@ -1,0 +1,85 @@
+from datasets import load_dataset
+from transformers import AutoTokenizer, Qwen3_5MoeForConditionalGeneration
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+
+# NOTE: This example requires transformers >= v5
+
+MODEL_ID = "Qwen/Qwen3.5-122B-A10B"
+
+# Load model.
+model = Qwen3_5MoeForConditionalGeneration.from_pretrained(MODEL_ID, dtype="auto")
+processor = AutoTokenizer.from_pretrained(MODEL_ID)
+
+
+recipe = QuantizationModifier(
+    targets="Linear",
+    scheme="NVFP4",
+    ignore=[
+        "re:.*lm_head",
+        "re:visual.*",
+        "re:model.visual.*",
+        "re:.*mlp.gate$",
+        "re:.*embed_tokens$",
+        "re:.*shared_expert_gate$",
+        "re:.*linear_attn.*",
+    ],
+)
+
+NUM_CALIBRATION_SAMPLES = 256
+MAX_SEQUENCE_LENGTH = 4096
+
+# Load datasets and preprocess.
+samples_per_dataset = NUM_CALIBRATION_SAMPLES
+
+ds_ultrachat = load_dataset(
+    "HuggingFaceH4/ultrachat_200k",
+    split=f"train_sft[:{samples_per_dataset}]",
+)
+
+# Both datasets share a "messages" column with the same chat format.
+# Keep only that column so we can concatenate them.
+ds = ds_ultrachat.select_columns(["messages"])
+ds = ds.shuffle(seed=42)
+
+
+def preprocess(example):
+    return {
+        "text": processor.apply_chat_template(
+            example["messages"],
+            tokenize=False,
+        )
+    }
+
+
+ds = ds.map(preprocess)
+
+
+# Tokenize inputs.
+def tokenize(sample):
+    return processor(
+        sample["text"],
+        padding=False,
+        max_length=MAX_SEQUENCE_LENGTH,
+        truncation=True,
+        add_special_tokens=False,
+    )
+
+
+ds = ds.map(tokenize, remove_columns=ds.column_names)
+
+# Apply quantization.
+oneshot(
+    model=model,
+    recipe=recipe,
+    dataset=ds,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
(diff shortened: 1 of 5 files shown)


LudovicoYIN:
> Thank for your contribution. I need to take a closer look into the modeling code but everything else looks good. Were you able to generate a checkpoint with the example?

Thanks for your review @dsikka . I'm generating a checkpoint and will upload it to Hugging Face soon.

One question: MiniMax-M2 uses FP8 weights by default. I've converted it to BF16 using model = AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16). Would this approach be acceptable?

dsikka:
> > Thank for your contribution. I need to take a closer look into the modeling code but everything else looks good. Were you able to generate a checkpoint with the example?
> 
> Thanks for your review @dsikka . I'm generating a checkpoint and will upload it to Hugging Face soon.
> 
> One question: MiniMax-M2 uses FP8 weights by default. I've converted it to BF16 using model = AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16). Would this approach be acceptable?

I think it may automatically decompress the model to bfloat16 without explicitly requiring the dtype (I believe this is currently the case for gpt-oss, for example) but you may have to test. But either way, this is fine to do as long is in bfloat16 for quantization

src/llmcompressor/modeling/minimax_m2.py
```diff
@@ -0,0 +1,99 @@
...
+        # in find_tied_parameters (caused by holding self.original)
+        self.gate = original.gate
+        self.experts = original.experts
+
+        # MiniMax specific parameters
+        self.jitter_noise = original.jitter_noise
+        self.num_experts = config.num_local_experts
+        self.top_k = original.top_k
+        # Use unbound function so this module's buffers are used.
+        self._route_tokens_to_experts = type(original).route_tokens_to_experts
```
dsikka, line 44: Is this in the model definition? https://github.com/huggingface/transformers/blob/cc88192bc5ca0b66e7fc1152bb372fb90c6bc2ce/src/transformers/models/minimax/modeling_minimax.py#L518

LudovicoYIN:
Hi @dsikka — quick clarification on why I changed direction in this follow-up.

In the initial implementation I followed the Transformers 5.x MiniMax-M2 structure, where expert weights are stored as packed `nn.Parameter` tensors (e.g. `gate_up_proj` / `down_proj`) instead of explicit per-expert `nn.Linear` modules (`w1/w2/w3`).

llm-compressor AWQ matching is currently module-based (e.g. `Linear` modules / module regex), and the calibration flow expects module-level `weight` semantics. Because of this, the packed-parameter structure leads to incomplete expert weight matching and expert weights are skipped during quantization.

The original MiniMax repo exposes experts as explicit linear layers, which aligns with existing AWQ mappings and produces stable end-to-end quantization. For now, I converged to that structure to ensure the workflow is functional and verifiable (in transformers==4.57.6). https://huggingface.co/MiniMaxAI/MiniMax-M2.1/blob/main/modeling_minimax_m2.py

I also experimented with forcing a structural rewrite on the Transformers 5.x path (`is_permanent=True`) to convert experts into explicit Linear-style modules. This fixes quantization matching, but it changes the saved model structure and vLLM fails to load the checkpoint due to MiniMax architecture mismatch.

src/llmcompressor/modeling/minimax_m2.py
LudovicoYIN: In my initial commits, I followed the Transformers 5.x MiniMax-M2 structure, where expert weights are packed as `nn.Parameter` tensors (e.g., `gate_up_proj` / `down_proj`) instead of explicit `nn.Linear` layers (`w1/w2/w3`).

However, after validating quantization end-to-end, I found that the packed representation doesn't align well with our current AWQ module matching (which targets `Linear` modules), leading to incomplete expert weight coverage.

So, in subsequent commits, I switched to the original MiniMax structure where experts are explicitly defined as `w1/w2/w3` `nn.Linear` layers (https://huggingface.co/MiniMaxAI/MiniMax-M2.1/blob/main/modeling_minimax_m2.py).

LudovicoYIN:
hi @dsikka 
MiniMax-M2 adaptation is now complete.
- Integrated MiniMax-M2 using the official HF repo/modeling path (`MiniMaxAI/MiniMax-M2`, `trust_remote_code=True`).
- MiniMax calibration tests are passing.
- MiniMax quantization example has been run successfully end-to-end.
- Quantized artifacts have been uploaded to Hugging Face: https://huggingface.co/ludovicoYIN/MiniMax-M2-BF16-W4A16
- Current environment uses `transformers==4.57.6`.

HDCharles:
looks good to me, also @LudovicoYIN if you are looking for a review reach out to me/ @kylesayrs / @dsikka in vllm slack by DM because i think this fell through the cracks and shouldn't have taken us a month to review.

LudovicoYIN:
Thanks for the review and the tip — I’ll reach out on Slack next time to help keep things moving.  @HDCharles @kylesayrs

PR #2336, opened 2026-02-06
Improve how we identify and run e2e smoke tests

dhuangnm:
SUMMARY:
Currently we use the file `tests/e2e/vLLM/rhaiis-e2e-smoke.list` to mark the configs for smoke tests that we use to run for the RHAIIS image. This is vulnerable as we need to keep the list in this file up-to-date to any changes in the config yaml files and this is error-prone.

This PR removes this `tests/e2e/vLLM/rhaiis-e2e-smoke.list` file and use the config yaml file directly to mark the smoke tests. We added a new field `test_group` to the yaml file and updated the `run_tests_in_*.sh` scripts to parse this field and filter out tests if a test group (-g) is specified. This allows both python and rhaiis image testing be able to run smoke and full tests for the configs.

To be more specific:

$# to run e2e tests for all configs (default)
`
bash tests/e2e/vLLM/run_tests_in_python.sh -c tests/e2e/vLLM/configs -t tests/e2e/vLLM/test_vllm.py
`

$# to run e2e tests for configs with smoke only
`
bash tests/e2e/vLLM/run_tests_in_python.sh -c tests/e2e/vLLM/configs -t tests/e2e/vLLM/test_vllm.py -g rhaiis
`

Similar commands for the `run_tests_in_rhaiis.sh` script.

Going forward, for any newly added configs for the e2e tests, if we want to include them into the smoke tests for the RHAIIS image, we need to remember to add the `test_group: "smoke"` into their yaml file under the configs/ so we can automatically pick it up for the RHAIIS image testing.

TEST PLAN:
A successful run of the smoke tests is here: 
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/21727920814


tests/e2e/vLLM/configs/fp4_nvfp4.yaml
```diff
@@ -1,7 +1,8 @@
 cadence: "nightly"
 test_type: "regression"
+test_group: "smoke"
```
dsikka: I think we want to call it "rhaiis" or something to indicate that this is also a test that runs for rhaiis e2e tests. Smoke is a bit misleading

dhuangnm: Good point, updated the PRs accordingly to use "rhaiis".

tests/e2e/vLLM/run_tests_in_python.sh
```diff
@@ -2,39 +2,60 @@
 
 SUCCESS=0
 
-while getopts "c:t:" OPT; do
+while getopts "c:t:g:" OPT; do
```
dsikka, line 5: why do we need to change this file at all?

dsikka (changes requested):

dsikka (approved):

PR #2342, opened 2026-02-10
[cicd] move check ready action to on pull_request_target

brian-dellabetta:
SUMMARY:
Change the check ready label ci/cd action to run on `pull_request_target` so that it runs more robustly for community user PRs. From [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target):

> This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

TEST PLAN:
Unfortunately this needs to be merged into main in order to run and fully validate. The action doesn't seem to get run on this branch

### DIFF
--- a/.github/workflows/ready-label-check.yaml
+++ b/.github/workflows/ready-label-check.yaml
@@ -1,7 +1,7 @@
 name: Check Ready Label
 
 on:
-  pull_request:
+  pull_request_target:
     branches: [ main , 'release/*' ]
     types:
       - labeled


HDCharles:
it can't seem to figure out if the ready label was applied?

PR #2344, opened 2026-02-10
AWQ: orig_layer_weights should save all balance layer weights

ZewenShen-Cohere:
Currently, orig_layer_weights only clones weights for layers that have a quantization scheme and are listed in `mapping.balance_layers`. This becomes a problem when we disable quantization for a layer that is still in `mapping.balance_layers`: all balance layers still need to be smoothed at the end, but orig_layer_weights does not store the original weights for all of them. As a result, the smoothing step fails (see where the error is triggered: https://github.com/ZewenShen-Cohere/llm-compressor-fork/blob/e9e3d3191f7598198f070c5f8269f08ec89e0b2f/src/llmcompressor/modifiers/awq/base.py#L554
).


HDCharles:
weird, don't we have a skip layers test, howd that pass?

shouldn't e.g. https://github.com/vllm-project/llm-compressor/blob/b49fbfda933e168f7b58f10ff45e019b3f24baee/tests/llmcompressor/transformers/compression/recipes/new_quant_full.yaml#L5 fail?

ZewenShen-Cohere:
I think a common use case is to quantize only the Q and K projections, while leaving the V projection unquantized. This is also important for the parallel transformer block used in Command A, where we quantize only the MoE and not the attention.

PR #2346, opened 2026-02-10
[AWQ] Add activation_hook_target field for custom activation cache hooking

ZewenShen-Cohere:
## Summary

- Adds an optional `activation_hook_target` field to `AWQMapping` that lets users specify which submodule (relative to the parent/LCA) to hook for activation caching, replacing the hardcoded `hasattr(parent, 'mlp')` workaround for MoE models with parallel transformer blocks.
- When `activation_hook_target` is `None` (default), behavior is unchanged: the hook is placed on `balance_layers[0]`. When set (e.g. `"mlp"`), it resolves to the corresponding submodule on the parent via `getattr_chain`.

## Motivation

In parallel transformer architectures, attention and MLP run in parallel from the same input. The existing code always hooks `balance_layers[0]` for activation caching, which captures the wrong activations when balance layers span both attention and MLP branches. There was a commented-out `hasattr(parent, 'mlp')` workaround, but it was brittle and not generalizable. This change makes the hook target explicitly configurable per mapping.

## Test 

I've tested this change with our internal models, and it aligns with previous results.

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -704,11 +704,9 @@ class AWQModifier(Modifier, QuantizationMixin):
                 else:
                     scales = x_mean.pow(ratio).clamp(min=1e-4).view(-1)
                 scales = scales / (scales.max() * scales.min()).sqrt()
-                _scalesview = scales.view(1, -1).to(device)
-
-                # avoid scaling values that overflow
                 scales[torch.isinf(scales)] = 1
                 scales[torch.isnan(scales)] = 1
+                _scalesview = scales.view(1, -1).to(device)
 
                 # Q(W * s)
                 for balance_layer in balance_layers_to_patch:


src/llmcompressor/modifiers/awq/mappings.py
```diff
@@ -16,10 +16,21 @@ class AWQMapping:
...
+        parent module (lowest common ancestor of balance_layers) specifying which
+        submodule to hook for activation caching. Useful for parallel transformer
+        blocks (e.g. Cohere, Gemma 3) where the first balance layer is not the
+        correct place to capture activations. When ``None`` (default), the hook
+        is placed on ``balance_layers[0]``.
     """
 
     smooth_layer: str
     balance_layers: list[str]
+    activation_hook_target: str | None = None
```
brian-dellabetta, line 33: would be good to update one of the default mappings, like the one for cohere or gemma 3 that you call out above, so users can see an example, if you have seen benefit from this

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -76,6 +76,9 @@ class AWQModifier(Modifier, QuantizationMixin):
           balance_layers: ["re:.*q_proj", "re:.*k_proj", "re:.*v_proj"]
         - smooth_layer: "re:.*final_layer_norm"
           balance_layers: ["re:.*fc1"]
+        # For parallel transformer blocks (e.g. Cohere), specify which
```
brian-dellabetta: Can you include an example usage of activation_hook_target here? this comment might be better inside the docstring of :param mappings:

src/llmcompressor/modifiers/awq/mappings.py
ZewenShen-Cohere: Hi @brian-dellabetta, thank you for the review. This change is only useful for MoE models with parallel transformer blocks. As far as I know, currently there's no open source model that uses this architecture, so the situation is a bit tricky here.

brian-dellabetta: Got it. Might also be good to include your response as a comment in the code -- "This change is only useful for MoE models with parallel transformer blocks."

ZewenShen-Cohere: Done!

HDCharles: i feel like we should maybe least include an example mapping with this, something like

parallel_transformer_block_mappings or something like that so there's a clear example even if there's not an OSS model that it would apply to.

ZewenShen-Cohere:
> i'll approve but i think this could use an example mapping even if there's not a real model that it applies to

Thank you for the feedback. I've added such an example.

PR #2349, opened 2026-02-11
DataLoader options, single-pass weight calibration, optional sequential prefetch

GOavi101:
## Summary
Performance-oriented changes for calibration and weight quantization: DataLoader tuning when workers are used, a single-pass weight calibration in `QuantizationModifier`, and an optional sequential prefetch to overlap onload with forward. Defaults stay safe for low RAM/GPU.

### 1. DataLoader (`datasets/utils.py`)
- **`pin_memory`**: Set to `True` only when CUDA is available **and** `dataloader_num_workers > 0` (avoids extra pinned memory when `num_workers=0`).
- When `num_workers > 0`: set `persistent_workers=True` and `prefetch_factor=2` for faster calibration.

### 2. Optional sequential prefetch (`pipelines/sequential/pipeline.py`, `args/dataset_arguments.py`, `entrypoints/oneshot.py`)
- New argument **`sequential_prefetch`** (default **`False`**).
- When `False`: same as before — one batch on GPU at a time (low peak memory).
- When `True`: prefetch next batch in a background thread to overlap onload with forward (faster when GPU memory allows two batches).

### 3. Dataset/oneshot args
- `dataloader_num_workers` default remains **0** (low-memory safe); help text updated.
- `sequential_prefetch` added to `DatasetArguments` and `oneshot()` with default `False`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,11 +107,11 @@ Note that the model can be swapped for a local or remote HF-compatible checkpoin
 Quantization is applied by selecting an algorithm and calling the `oneshot` API.
 
 ```python
+from compressed_tensors.offload import dispatch_model
 from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.quantization import QuantizationModifier
-from llmcompressor.utils import dispatch_for_generation
 
 MODEL_ID = "Qwen/Qwen3-30B-A3B"
 
@@ -134,7 +134,7 @@ oneshot(model=model, recipe=recipe)
 
 # Confirm generations of the quantized model look sane.
 print("========== SAMPLE GENERATION ==============")
-dispatch_for_generation(model)
+dispatch_model(model)
 input_ids = tokenizer("Hello my name is", return_tensors="pt").input_ids.to(
     model.device
 )
--- a/docs/.nav.yml
+++ b/docs/.nav.yml
@@ -9,6 +9,20 @@ nav:
     - Compressing your Model: getting-started/compress.md
     - Deploying with vLLM: getting-started/deploy.md
     - FAQ: getting-started/faq.md
+  - Key Models:
+    - key-models/index.md
+    - Llama 4:
+      - key-models/llama4/index.md
+      - FP8 Example: key-models/llama4/fp8-example.md
+    - Qwen3:
+      - key-models/qwen3/index.md
+      - FP8 Example: key-models/qwen3/fp8-example.md
+    - Kimi-K2:
+      - key-models/kimi-k2/index.md
+      - FP8 Example: key-models/kimi-k2/fp8-example.md
+    - Mistral Large 3:
+      - key-models/mistral-large-3/index.md
+      - FP8 Example: key-models/mistral-large-3/fp8-example.md
   - Guides:
     - Compression Schemes: guides/compression_schemes.md
     - Saving a Model: guides/saving_a_model.md
--- a/docs/getting-started/compress.md
+++ b/docs/getting-started/compress.md
@@ -24,7 +24,7 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.quantization import QuantizationModifier
-from llmcompressor.utils import dispatch_for_generation
+from compressed_tensors.offload import dispatch_model
 
 MODEL_ID = "Qwen/Qwen3-30B-A3B"
 
@@ -47,7 +47,7 @@ oneshot(model=model, recipe=recipe)
 
 # Confirm generations of the quantized model look sane.
 print("========== SAMPLE GENERATION ==============")
-dispatch_for_generation(model)
+dispatch_model(model)
 input_ids = tokenizer("Hello my name is", return_tensors="pt").input_ids.to(
     model.device
 )
--- a/docs/key-models/index.md
+++ b/docs/key-models/index.md
@@ -1,0 +1,39 @@
+# Key Models
+
+The following models are among the most commonly used with LLM Compressor: Llama 4, Qwen3, Kimi-K2, and Mistral Large 3. Each model page contains quantization examples with tested configurations and recommended parameters.
+
+<div class="grid cards" markdown>
+
+  - **Llama 4**
+
+      ---
+
+      Meta's Llama 4 Scout multimodal model.
+
+      [:octicons-arrow-right-24: Llama 4](llama4/index.md)
+
+  - **Qwen3**
+
+      ---
+
+      Qwen3-VL MoE vision-language model.
+
+      [:octicons-arrow-right-24: Qwen3](qwen3/index.md)
+
+  - **Kimi-K2**
+
+      ---
+
+      Kimi-K2 Thinking model.
+
+      [:octicons-arrow-right-24: Kimi-K2](kimi-k2/index.md)
+
+  - **Mistral Large 3**
+
+      ---
+
+      Mistral's 675B parameter model.
+
+      [:octicons-arrow-right-24: Mistral Large 3](mistral-large-3/index.md)
+
+</div>
--- a/docs/key-models/kimi-k2/fp8-example.md
+++ b/docs/key-models/kimi-k2/fp8-example.md
@@ -1,0 +1,37 @@
+## Kimi-K2 FP8 Example
+
+### Code Walkthrough
+
+Let's walk through the main steps of the quantization process:
+1. Load model
+2. Apply quantization
+
+### 1. Load Model
+
+```python
+from llmcompressor import model_free_ptq
+
+MODEL_ID = "unsloth/Kimi-K2-Thinking-BF16"
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-FP8-BLOCK"
+```
+
+### 2. Apply Quantization
+
+Once quantized, the model is saved. This uses compressed-tensors to the SAVE_DIR.
+
+```python
+model_free_ptq(
+    model_stub=MODEL_ID,
+    save_directory=SAVE_DIR,
+    scheme="FP8_BLOCK",
+    ignore=[
+        "re:.*gate$",
+        "lm_head",
+        "re:.*kv_a_proj_with_mqa$",
+        "re:.*q_a_proj$",
+        "model.embed_tokens",
+    ],
+    max_workers=15,
+    device="cuda:0",
+)
+```
--- a/docs/key-models/kimi-k2/index.md
+++ b/docs/key-models/kimi-k2/index.md
@@ -1,0 +1,5 @@
+# Kimi-K2
+
+Quantization examples for the Kimi-K2 Thinking model.
+
+- [FP8 Example](fp8-example.md)
--- a/docs/key-models/llama4/fp8-example.md
+++ b/docs/key-models/llama4/fp8-example.md
@@ -1,0 +1,71 @@
+## Llama4 FP8 Example
+
+### Code Walkthrough
+
+Let's walk through the main steps of the quantization process:
+1. Load model
+2. Configure quantization algorithm and scheme
+3. Apply quantization
+4. Confirm generations of the quantized model look sane
+5. Save to disk in compressed-tensors format
+
+### 1. Load Model
+
+Load the model using `AutoModelForCausalLM`:
+
+```python
+from compressed_tensors.offload import dispatch_model
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+
+MODEL_ID = "meta-llama/Llama-4-Scout-17B-16E-Instruct"
+
+model = AutoModelForCausalLM.from_pretrained(MODEL_ID, dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
+```
+
+### 2. Configure the Quantization Algorithm and Scheme
+
+```python
+recipe = QuantizationModifier(
+    targets="Linear",
+    scheme="FP8_BLOCK",
+    ignore=[
+        "re:.*lm_head",
+        "re:.*self_attn",
+        "re:.*router",
+        "re:.*vision_model.*",
+        "re:.*multi_modal_projector.*",
+        "Llama4TextAttention",
+    ],
+)
+```
+
+### 3. Apply Quantization
+
+```python
+oneshot(model=model, recipe=recipe)
+```
+
+### 4. Confirm Generations of the Quantized Model Look Sane
+
+```python
+print("========== SAMPLE GENERATION ==============")
+dispatch_model(model)
+input_ids = tokenizer("Hello my name is", return_tensors="pt").input_ids.to(
+    model.device
+)
+output = model.generate(input_ids, max_new_tokens=20)
+print(tokenizer.decode(output[0]))
+print("==========================================")
+```
+
+### 5. Save to Disk in Compressed-Tensors Format
+
+```python
+SAVE_DIR = MODEL_ID.split("/")[1] + "-FP8-BLOCK"
+model.save_pretrained(SAVE_DIR)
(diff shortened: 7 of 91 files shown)


src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -87,7 +81,10 @@ def on_start(self, state: State, event: Event, **kwargs):
         for module in state.model.modules():
             update_fused_layer_weight_global_scales(module)
 
+        # Single pass over targeted modules: global scale then zp/scale (avoids
+        # two full iterations over named_modules)
         for _, module in tqdm.tqdm(named_modules, desc="Calibrating weights"):
+            update_weight_global_scale(module)
```
kylesayrs: I don't think this will work, because updating the global scale needs to happen before `update_fused_layer_weight_global_scales`. `update_fused_layer_weight_global_scales` is dependent on `update_weight_global_scale` being called first, since it fuses QKV, Up/Gate weights.

One other optimization here might be to fuse `update_weight_global_scale` and `update_fused_layer_weight_global_scales` together for better cache locality. This would mean doing some kind of dfs on the model, detecting fused layers, and jointly calibrating + fusing them

GOavi101: Yes, that makes sense. I’ll try to implement the DFS / fused-layer joint calibration in a separate PR and have left it documented in the performance guide for follow-up.

GOavi101:
with out changes
```
2026-02-11T18:17:02.818623+0000 | get_model_compressor | INFO - skip_sparsity_compression_stats set to True. Skipping sparsity compression statistic calculations. No sparsity compressor will be applied.
Compressing model: 224it [00:59,  3.78it/s]

real	5m58.574s
user	12m10.075s
sys	2m59.391s
```


with  changes
```
2026-02-11T18:26:02.660886+0000 | get_model_compressor | INFO - skip_sparsity_compression_stats set to True. Skipping sparsity compression statistic calculations. No sparsity compressor will be applied.
Compressing model: 224it [00:51,  4.36it/s]

real	5m42.606s
user	11m41.606s
sys	3m9.245s
```

HDCharles:
Still failing quality checks

If you do pip install -e.[dev] it will install the correct version of ruff

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -113,25 +138,47 @@ def __call__(
...
                 # reduce memory movement by keeping modules onloaded
+                num_batches = len(dataloader)
+                use_prefetch = getattr(dataset_args, "sequential_prefetch", False)
                 with disable_offloading():
                     # do a preliminary pass to trigger modifier hooks
-                    for batch_idx in tqdm(range(len(dataloader)), desc=calib_desc):
-                        inputs = activations.fetch(batch_idx, subgraph.input_names)
-
-                        # Set current batch index before forward pass for AWQ masking
+                    calib_batches = (
```
HDCharles: nit: i prefer a normal if else for something this complicated, only use x if y else z for 1 or 2 line things

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -29,6 +30,30 @@
 __all__ = ["SequentialPipeline"]
 
 
+def _batches_with_prefetch(
```
HDCharles: this is really cool, when we did performance improvements for AWQ we discussed this as an option but didn't get around to implementing it, would theoretically be an even larger improvement because we have to run samples many times for each module and those are often offloaded to cpu

https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L597

wonder if there's an easy way to modify our intermediates [offload cache](https://github.com/vllm-project/llm-compressor/blob/c37fcfa081daa024f865a3f4798db029a0a67d43/src/llmcompressor/pipelines/cache.py#L30) to do this automatically

HDCharles:
> with out changes
> 
> ```
> 2026-02-11T18:17:02.818623+0000 | get_model_compressor | INFO - skip_sparsity_compression_stats set to True. Skipping sparsity compression statistic calculations. No sparsity compressor will be applied.
> Compressing model: 224it [00:59,  3.78it/s]
> 
> real	5m58.574s
> user	12m10.075s
> sys	2m59.391s
> ```
> 
> with changes
> 
> ```
> 2026-02-11T18:26:02.660886+0000 | get_model_compressor | INFO - skip_sparsity_compression_stats set to True. Skipping sparsity compression statistic calculations. No sparsity compressor will be applied.
> Compressing model: 224it [00:51,  4.36it/s]
> 
> real	5m42.606s
> user	11m41.606s
> sys	3m9.245s
> ```

can you provide your repro? 

If you do something like https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/llama3_example.py but set recipe = None and pipeline="sequential" on the oneshot call, you can isolate the changes from this PR without the overhead from quantization.

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -258,9 +258,19 @@ class DatasetArguments(CustomDatasetArguments):
...
-            "help": "Number of worker processes for data loading. Set to 0 to disable "
-            "multiprocessing. Note: Custom data collators may not work with "
-            "multiprocessing. Default is 0."
+            "help": "Number of worker processes for data loading. Default is 0 (safe "
+            "for low CPU/GPU memory). Set to 2 or more for faster calibration if you "
+            "have sufficient RAM. Custom data collators may not work with "
+            "multiprocessing."
+        },
+    )
+    sequential_prefetch: bool = field(
```
HDCharles, line 267: as a followup, this defaulting to true may be a good idea depending on how the memory benchmarks actually look. It feels like this should be a relatively small increase in memory and we have so many configurable options that only a few are going to be actually used.

src/llmcompressor/datasets/utils.py
```diff
@@ -127,13 +127,24 @@ def format_calibration_data(
...
+    # Pin memory only when using workers (saves RAM for low-memory users when
+    # num_workers=0; when num_workers>0, pin_memory speeds CPU->GPU transfer)
+    num_workers = args.dataloader_num_workers
+    pin_memory = torch.cuda.is_available() and num_workers > 0
+    # persistent_workers avoids worker respawn between epochs; prefetch_factor
+    # allows loading next batches in parallel (only used when num_workers > 0)
+    kwargs: dict[str, Any] = {}
+    if num_workers > 0:
+        kwargs["persistent_workers"] = True
+        kwargs["prefetch_factor"] = 2
```
brian-dellabetta: nit-looks like this is already the default value when num_workers>1, docs [here](https://docs.pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader). Would help readability to remove or add a comment that we're setting this to default behavior

brian-dellabetta (approved):
Thanks for the contribution! One nit, otherwise LGTM pending other comments

src/llmcompressor/pipelines/sequential/pipeline.py
GOavi101: @HDCharles , I can add optional prefetch to the intermediates cache and have AWQ turn it on when offloading, so iteration overlaps onload with compute without extra API surface.

If that sounds good to you, I can implement it and add tests as well.

examples/quantization_w4a4_fp4/llama3_example_prefetch.py
```diff
@@ -0,0 +1,53 @@
...
+        padding=False,
+        max_length=MAX_SEQUENCE_LENGTH,
+        truncation=True,
+        add_special_tokens=False,
+    ),
+    remove_columns=ds.column_names,
+)
+
+# No recipe: run sequential pipeline only (cache + subgraph passes, no quantization).
+oneshot(
```
HDCharles, line 48: would be better for the exampe to have sequential prefetch enabled

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -777,9 +778,9 @@ def _compute_best_scale(
             }
         )
 
-        assert (
-            torch.isnan(best_scales).sum() == 0
-        ), f"Nan found in scales: {best_scales}"
+        assert torch.isnan(best_scales).sum() == 0, (
```
HDCharles: looks like your formatting isn't working correctly

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -299,6 +299,7 @@ def oneshot(
     sequential_targets: list[str] | None = None,
     sequential_offload_device: str = "cpu",
     quantization_aware_calibration: bool = True,
+    sequential_prefetch: bool = True,
```
HDCharles: we need clear benchmark results across a variety of models and use cases that this doesn't explode memory and is compatible with DDP changes...etc, 

I meant that in the future it may make sense to make this true, not necessarily right now.

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -113,28 +138,56 @@ def __call__(
...
+                            calib_desc,
+                        )
+                        for batch_idx, inputs in calib_batches:
+                            session.state.current_batch_idx = batch_idx
+                            subgraph.forward(model, **inputs)
+                    else:
+                        for batch_idx in tqdm(range(num_batches), desc=calib_desc):
+                            inputs = activations.fetch(batch_idx, subgraph.input_names)
+                            session.state.current_batch_idx = batch_idx
+                            subgraph.forward(model, **inputs)
```
HDCharles: lets simplify all this, changes function to _get_batches( ... , use_prefetch=False) and edit the logic there to work with/without prefetch

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -29,6 +30,30 @@
...
+        for batch_idx in tqdm(range(num_batches), desc=desc):
+            if future is not None:
+                inputs = future.result()
+            else:
+                inputs = activations.fetch(batch_idx, input_names)
+            if batch_idx + 1 < num_batches:
+                future = executor.submit(activations.fetch, batch_idx + 1, input_names)
+            else:
+                future = None
+            yield batch_idx, inputs
```
HDCharles:
```suggestion
def _get_batches(
    activations: IntermediatesCache,
    num_batches: int,
    input_names: list[str],
    desc: str,
    use_prefetch: bool = True,
) -> Iterator[tuple[int, dict]]:
    """
    Yield (batch_idx, inputs) with the next batch optionally prefetched in a
    background thread to overlap fetch (onload from offload device) with the
    main-thread forward pass.
    """
    if not use_prefetch:
        for batch_idx in tqdm(range(num_batches), desc=desc):
            inputs = activations.fetch(batch_idx, input_names)
            yield batch_idx, inputs
        return

    with ThreadPoolExecutor(max_workers=1) as executor:
        future = None
        for batch_idx in tqdm(range(num_batches), desc=desc):
            if future is not None:
                inputs = future.result()
            else:
                inputs = activations.fetch(batch_idx, input_names)
            if batch_idx + 1 < num_batches:
                future = executor.submit(activations.fetch, batch_idx + 1, input_names)
            else:
                future = None
            yield batch_idx, inputs
```

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -113,28 +138,56 @@ def __call__(
...
+                            for batch_idx in tqdm(range(num_batches), desc=prop_desc):
+                                inputs = activations.fetch(
+                                    batch_idx, subgraph.input_names
+                                )
+                                output = subgraph.forward(model, **inputs)
+                                if subgraph_index < num_subgraphs - 1:
+                                    activations.update(batch_idx, output)
+                                    activations.delete(
+                                        batch_idx, subgraph.consumed_names
+                                    )
```
HDCharles: simplify this like above

src/llmcompressor/pipelines/cache.py
```diff
@@ -89,7 +106,7 @@ def from_dataloader(
             for batch in tqdm(dataloader, desc="Preparing cache")
         ]
 
-        return cls(batch_intermediates, offload_device)
+        return cls(batch_intermediates, offload_device, prefetch=False)
```
HDCharles: shouldn't this also accept prefetch as an argument?

HDCharles:
this is getting complicated, lets take the intermediates cache changes and move that to a separate PR. I left some comments to clean this up, once that's done we can probably land the calibration changes.

I don't think we can make pretech true as a default without 

A) benchmarks for memory and runtime across a few model classes (tinyllama, llama8B and Qwen MoE 30B at minimum)
B) verify whether this conflicts with DDP

but i think to land the calibration changes, as long as we have memory/runtime measurements on something we can probably land it

examples/quantization_w4a4_fp4/llama3_example_prefetch.py
```diff
@@ -0,0 +1,53 @@
+"""
```
HDCharles, line 1: lets just make this into an example with prefetch, so name it along the lines of the other examples in that folder and add _prefetch at the end of the title

examples/quantization_w4a4_fp4/repro_benchmark_sequential.py
```diff
@@ -0,0 +1,53 @@
...
+# No recipe: run sequential pipeline only (cache + subgraph passes, no quantization).
+oneshot(
+    model=model,
+    dataset=ds,
+    recipe=None,
+    pipeline="sequential",
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+)
+print("Done.")
```
HDCharles: I would add something along the lines of

https://github.com/vllm-project/llm-compressor/pull/2265/changes/d1347840586ccbe6cb9b017cfbb719937136e542#diff-208ced55cba2d38b1bc9b03b5f79ae3483c0849cc708a6c1131c231aae5d4b3dR80-R88 to the example so its clear where you are getting the measurements

HDCharles:
still failing quality checks, 

in llm-compressor root do:

pip install -e.[dev]
make quality
make style

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -113,25 +145,33 @@ def __call__(
                 prop_desc = f"({subgraph_index + 1}/{num_subgraphs}): Propagating"
 
                 # reduce memory movement by keeping modules onloaded
+                num_batches = len(dataloader)
+                use_prefetch = getattr(dataset_args, "sequential_prefetch", False)
```
HDCharles, line 149: nit: would be better to keep the name consistent

(sequential_prefetch -> use_prefetch)

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -113,25 +145,33 @@ def __call__(
...
                         subgraph.forward(model, **inputs)
 
                     LifecycleCallbacks.sequential_epoch_end(subgraph)
 
                     # this pass does not trigger modifier hooks
                     # and is only used for capturing outputs of newly compressed modules
                     with HooksMixin.disable_hooks():
-                        for batch_idx in tqdm(range(len(dataloader)), desc=prop_desc):
-                            inputs = activations.fetch(batch_idx, subgraph.input_names)
+                        for batch_idx, inputs in _get_batches(
```
brian-dellabetta, line 167: Just want to confirm that the tqdm call in _get_batches will work the same here, it will keep track of the time taken to run calibration/propagation. looks like this is true

GOavi101: Yes. _get_batches uses tqdm(range(num_batches), desc=desc) in both the prefetch and non-prefetch paths, so when we call it with calib_desc or prop_desc we get one progress bar per phase with the same timing/throughput behavior as before. The progress bar still reflects the time to run calibration or propagation.

src/llmcompressor/modifiers/quantization/quantization/base.py
kylesayrs: https://github.com/vllm-project/llm-compressor/issues/2388

PR #2350, opened 2026-02-11
Swap to use CPU runners

dsikka:
SUMMARY:
- Swap ubuntu runners to use our cpu runner
- Remove 2 year old docker build workflow that we never use

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -9,6 +9,10 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+concurrency:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-publish-release-images.yaml
fynnsu, line 1: This workflow is still actively being run for releases:
https://github.com/vllm-project/llm-compressor/actions/workflows/build-and-publish-release-images.yaml

fynnsu: Nevermind, I see the steps are always skipped.

dsikka: we do not support docker builds

PR #2353, opened 2026-02-11
feat: early group-size divisibility check with layer FQNs

### LINKED ISSUE #1983: W4A16 quantization option to handle/ignore layers with columns%groupsize != 0
When you apply W4A16 quantization to a model, all layers get quantized, if you then try to save the compressed model you'll run into errors saving layers where the groupsize doesn't evenly divide the columns.  

https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/lifecycle/forward.py#L278

this has tripped up users before:

e.g. https://github.com/vllm-project/compressed-tensors/issues/447

It feels like there should be an option to ignore or pad layers that don't have the necessary dimension so you don't have to manually ignore all the layers with that shape.

GOavi101:
## Summary
Add an early check so users hit a clear error at `initialize()` (before long calibration e.g. GPTQ) when using group/tensor-group quantization on layers whose weight columns are not divisible by `group_size`, instead of failing at save with an opaque message.

## What changed
- **Policy:** Only GROUP and TENSOR_GROUP require strict divisibility (those kernels don’t support non-divisible shapes). BLOCK is intentionally not checked (block kernels support non-divisible). This is centralized in `group_size_validation.py`.
- **Early error:** We fail during `initialize_quantization()` and raise with:
  - The exact layer FQNs and `(columns, group_size)` for each problematic layer
  - Instructions to add those names to the modifier’s `ignore` list
- **Tests:** Added tests for the validation helper and for the modifier (raises with expected message, succeeds when layers are ignored or all divisible).

 Fixes #1983

### DIFF
--- a/docs/.nav.yml
+++ b/docs/.nav.yml
@@ -9,6 +9,20 @@ nav:
     - Compressing your Model: getting-started/compress.md
     - Deploying with vLLM: getting-started/deploy.md
     - FAQ: getting-started/faq.md
+  - Key Models:
+    - key-models/index.md
+    - Llama 4:
+      - key-models/llama4/index.md
+      - FP8 Example: key-models/llama4/fp8-example.md
+    - Qwen3:
+      - key-models/qwen3/index.md
+      - FP8 Example: key-models/qwen3/fp8-example.md
+    - Kimi-K2:
+      - key-models/kimi-k2/index.md
+      - FP8 Example: key-models/kimi-k2/fp8-example.md
+    - Mistral Large 3:
+      - key-models/mistral-large-3/index.md
+      - FP8 Example: key-models/mistral-large-3/fp8-example.md
   - Guides:
     - Compression Schemes: guides/compression_schemes.md
     - Saving a Model: guides/saving_a_model.md
--- a/docs/key-models/index.md
+++ b/docs/key-models/index.md
@@ -1,0 +1,39 @@
+# Key Models
+
+The following models are among the most commonly used with LLM Compressor: Llama 4, Qwen3, Kimi-K2, and Mistral Large 3. Each model page contains quantization examples with tested configurations and recommended parameters.
+
+<div class="grid cards" markdown>
+
+  - **Llama 4**
+
+      ---
+
+      Meta's Llama 4 Scout multimodal model.
+
+      [:octicons-arrow-right-24: Llama 4](llama4/index.md)
+
+  - **Qwen3**
+
+      ---
+
+      Qwen3-VL MoE vision-language model.
+
+      [:octicons-arrow-right-24: Qwen3](qwen3/index.md)
+
+  - **Kimi-K2**
+
+      ---
+
+      Kimi-K2 Thinking model.
+
+      [:octicons-arrow-right-24: Kimi-K2](kimi-k2/index.md)
+
+  - **Mistral Large 3**
+
+      ---
+
+      Mistral's 675B parameter model.
+
+      [:octicons-arrow-right-24: Mistral Large 3](mistral-large-3/index.md)
+
+</div>
--- a/docs/key-models/kimi-k2/fp8-example.md
+++ b/docs/key-models/kimi-k2/fp8-example.md
@@ -1,0 +1,37 @@
+## Kimi-K2 FP8 Example
+
+### Code Walkthrough
+
+Let's walk through the main steps of the quantization process:
+1. Load model
+2. Apply quantization
+
+### 1. Load Model
+
+```python
+from llmcompressor import model_free_ptq
+
+MODEL_ID = "unsloth/Kimi-K2-Thinking-BF16"
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-FP8-BLOCK"
+```
+
+### 2. Apply Quantization
+
+Once quantized, the model is saved. This uses compressed-tensors to the SAVE_DIR.
+
+```python
+model_free_ptq(
+    model_stub=MODEL_ID,
+    save_directory=SAVE_DIR,
+    scheme="FP8_BLOCK",
+    ignore=[
+        "re:.*gate$",
+        "lm_head",
+        "re:.*kv_a_proj_with_mqa$",
+        "re:.*q_a_proj$",
+        "model.embed_tokens",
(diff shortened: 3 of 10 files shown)


GOavi101:
```
(.venv) root@openshiftai-vllm:~/llm-comp-1/llm-compressor# pytest tests/llmcompressor/modifiers/quantization/test_group_size_validation.py -v
================================================== test session starts ===================================================
platform linux -- Python 3.12.3, pytest-9.0.2, pluggy-1.6.0 -- /root/llm-comp-1/llm-compressor/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /root/llm-comp-1/llm-compressor
configfile: pyproject.toml
plugins: mock-3.15.1, anyio-4.12.1, rerunfailures-16.1
collected 5 items                                                                                                        

tests/llmcompressor/modifiers/quantization/test_group_size_validation.py::test_get_layers_indivisible_by_group_size_empty PASSED [ 20%]
tests/llmcompressor/modifiers/quantization/test_group_size_validation.py::test_get_layers_indivisible_by_group_size_finds_layer PASSED [ 40%]
tests/llmcompressor/modifiers/quantization/test_group_size_validation.py::test_initialize_quantization_raises_early_for_indivisible PASSED [ 60%]
tests/llmcompressor/modifiers/quantization/test_group_size_validation.py::test_initialize_quantization_succeeds_when_indivisible_ignored PASSED [ 80%]
tests/llmcompressor/modifiers/quantization/test_group_size_validation.py::test_initialize_quantization_succeeds_when_all_divisible PASSED [100%]

```

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -213,6 +216,25 @@ def initialize_quantization(self, model: torch.nn.Module):
...
+                for fqn, cols, gs in indivisible
+            ]
+            raise ValueError(
+                "The following layers have weight column counts not divisible by "
+                "group_size. Group and tensor-group quantization require "
+                "columns % group_size == 0; compressed-tensors will error when saving "
+                "or running forward. Add these layer names to the modifier's `ignore` "
+                "list and re-run.\n\n" + "\n".join(lines)
+            )
+
```
HDCharles: lets encapsulate this whole thing in a single function call to keep this part of the lifecycle cleaner

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -213,6 +220,13 @@ def initialize_quantization(self, model: torch.nn.Module):
 
         apply_quantization_config(model, self.resolved_config)
 
+        validate_group_size_divisibility(
```
HDCharles: nit: probably cleaner to have the bypass if statment be here rather than in the function

src/llmcompressor/modifiers/quantization/group_size_validation.py
```diff
@@ -0,0 +1,144 @@
...
+        s = getattr(strategy, attr, None)
+        if s is not None and str(s).lower() in _GROUP_STRATEGY_STRINGS:
+            return True
+    s = str(strategy).lower()
+    if s in _GROUP_STRATEGY_STRINGS:
+        return True
+    # Enum repr e.g. "quantizationstrategy.group"
+    if s.split(".")[-1] in _GROUP_STRATEGY_STRINGS:
+        return True
+    return False
```
HDCharles: does the first check not suffice? how often do we end up here?

src/llmcompressor/modifiers/quantization/group_size_validation.py
```diff
@@ -0,0 +1,144 @@
...
+    "get_layers_indivisible_by_group_size",
+    "validate_group_size_divisibility",
+]
+
+# Strategies for which we error on indivisible columns (no kernel support).
+# BLOCK is intentionally excluded: block kernels support non-divisible.
+_GROUP_STRATEGY_STRINGS = ("group", "tensor_group")
+
+
+def _is_group_or_tensor_group_strategy(strategy) -> bool:
```
HDCharles: seems like you might be able to get rid of this whole function and just use line 48

HDCharles (changes requested):
some final changes and this will be good to go, great work!

src/llmcompressor/modifiers/quantization/group_size_validation.py
```diff
@@ -0,0 +1,122 @@
...
+- CHANNEL, TENSOR, TOKEN, ATTN_HEAD: No group_size divisibility requirement;
+  we do not run this validation.
+
+See: compressed-tensors forward.py (GROUP/TENSOR_GROUP ValueError),
+strategy_cdiv in compressed_tensors.quantization.utils.helpers.
+"""
+
+from __future__ import annotations
+
+from typing import Set, Tuple
```
brian-dellabetta: nit -- you can just use the `set`/`tuple` built-ins for python 3.10+

tests/llmcompressor/modifiers/quantization/test_group_size_validation.py
```diff
@@ -0,0 +1,153 @@
...
+    state.update(model=model, device="cpu")
+    modifier = QuantizationModifier(scheme="W4A16", targets=["Linear"])
+
+    with torch.no_grad():
+        try:
+            modifier.on_initialize(state)
+            pytest.skip(
+                "no indivisible layers targeted (CT may not attach to simple models)"
+            )
+        except ValueError as exc:
```
brian-dellabetta: nit -- pytest has a built-in `with pytest.raises` helper for this. example:

```python
import pytest

def another_func():
    raise ValueError("Invalid value: oh no!")

def test_exc_info():
    with pytest.raises(ValueError) as exc_info:
        another_func()
    
    # Assert on the string representation of the exception value
    assert "oh no!" in str(exc_info.value)

    # You can also check the exact message
    assert str(exc_info.value) == "Invalid value: oh no!"

    # Or check other attributes, like the type
    assert exc_info.type is ValueError
```

brian-dellabetta (approved):
Couple nits, otherwise looks good!

brian-dellabetta (approved):

PR #2366, opened 2026-02-13
[Sequential Pipeline] only cache unique offloaded values

### LINKED ISSUE #2363: [Bug]: Possible memory leak, server crashes completely with two llm-compressor jobs on it

`requirements.txt` available at [https://github.com/neuralmagic/GuardBench/blob/main/llama4guard_vllm_requirements.txt](https://github.com/neuralmagic/GuardBench/blob/main/llama4guard_vllm_requirements.txt)

### 🐛 Describe the bug

I've tried to quantize `meta-llama/Llama-Gurad-4-12B` with LLM-Compressor and spawned two jobs in parallel on a single 8xH100 server. The entire server collapsed with "out of memory" msgs in logs.
Could there be a some kind of memory leak in LLM-Compressor?

Script for quantization:
```python
from transformers import AutoProcessor, Llama4ForConditionalGeneration
import argparse
from compressed_tensors.quantization import QuantizationScheme, QuantizationArgs, QuantizationType, QuantizationStrategy

def parse_actorder(value):
    # Interpret the input value for --actorder
    if value.lower() == "false":
        return False
    elif value.lower() == "group":
        return "group"
    elif value.lower() == "weight":
        return "weight"
    else:
        raise argparse.ArgumentTypeError("Invalid value for --actorder. Use 'group', 'weight', or 'False'.")

def parse_sym(value):
    if value.lower() == "false":
        return False
    elif value.lower() == "true":
        return True
    else:
        raise argparse.ArgumentTypeError(f"Invalid value for --sym. Use false or true, but got {value}")

parser = argparse.ArgumentParser()
parser.add_argument('--model_path', type=str, required=True)
parser.add_argument('--quant_path', type=str, required=True)

parser.add_argument('--group_size', type=int, required=True)
parser.add_argument('--calib_size', type=int, required=True)
parser.add_argument('--dampening_frac', type=float, required=True)
parser.add_argument('--observer', type=str, required=True) # mse or minmax
parser.add_argument('--sym', type=parse_sym, required=True) # true or false
parser.add_argument('--actorder', type=parse_actorder, required=True) # group or weight or false
parser.add_argument('--pipeline', type=str, default="basic") #['basic', 'datafree', 'sequential', independent]

args = parser.parse_args()
print(f"[DEBUGGING ARGS] {args}")

model = Llama4ForConditionalGeneration.from_pretrained(
    args.model_path,
    torch_dtype="auto",
    trust_remote_code=True,
)
processor = AutoProcessor.from_pretrained(args.model_path, trust_remote_code=True)

def preprocess_fn(example):
    # prepare for multimodal processor
    for msg in example["messages"]:
        msg["content"] = [{'type': 'text', 'text': msg['content']}]

    return {"text": processor.apply_chat_template(example["messages"], add_generation_prompt=False, tokenize=False)}

ds = load_dataset("neuralmagic/LLM_compression_calibration", split="train")
ds = ds.map(preprocess_fn)

print(f"================================================================================")
print(f"[For debugging] Calibration data sample is:\n{repr(ds[0]['text'])}")
print(f"================================================================================")

quant_scheme = QuantizationScheme(
    targets=["Linear"],
    weights=QuantizationArgs(
        num_bits=4,
        type=QuantizationType.INT,
        symmetric=args.sym,
        group_size=args.group_size,
        observer=args.observer,
        actorder=args.actorder
    ),
    input_activations=None,
    output_activations=None,
)

recipe = [
    GPTQModifier(
        targets=["Linear"],
        ignore=[
            "re:.*lm_head",
            "re:.*multi_modal_projector",
            "re:.*vision_model",
        ],
        dampening_frac=args.dampening_frac,
        config_groups={"group_0": quant_scheme},
    )
]
oneshot(
    model=model,
    dataset=ds,
    recipe=recipe,
    num_calibration_samples=args.calib_size,
    max_seq_length=8192,
    pipeline=args.pipeline,
)

SAVE_DIR = args.quant_path
print(f"Model saved to {SAVE_DIR}")
```

with:
```bash
python llmcompressor_int4.py --model_path meta-llama/Llama-Guard-4-12B --quant_path meta-llama/Llama-Guard-4-12B-quantized.w4a16_calib512_damp0.03_obsMinMax_sym_actorderWeight_pipelineSequential --group_size 128 --calib_size 512 --dampening_frac 0.03 --observer minmax --sym True --actorder weight --pipeline independent
```

### 🛠️ Steps to reproduce

_No response_

kylesayrs:
Updated by @brian-dellabetta 

SUMMARY:
The SequentialPipeline offloads subgraph outputs as part of normal usage. Occasionally these outputs share duplicates in kwargs that point to the same memory location on the onloaded device. When offloading is enabled, there was previously no check to see if any tensors to be offloaded had already previously been offloaded, which can cause a huge increase in memory requirements in some models, as reported in #2363. This PR 
- [x] adds an offload map to IntermediatesCache to ensure tensors are not redundantly offloaded
- [x] wraps the map in an override to ensure `torch.equal` is used rather than `torch.eq` (which is the one used with `==` checks). `torch.eq` can return multiple boolean values depending on the tensors being compared, resulting in an error. This override, which should only be used when the tensors are immutable (the case here), allows us to retain the original hashing function and have an `O(1)` lookup. Our other attempts to circumvent the issue added to runtime or required `O(N)` lookup.

Resolves #2363 

TEST PLAN:
- [x] Unit test added for `OverrideEqMode`
- [x] Script from #2363 runs with ~81GB CPU RAM after first layer propagation, increased to ~88GB CPU RAM used by layer 11/49, and then stays consistently <89GB CPU RAM used by layer 25/49. On current main, this script would hit ~750GB CPU RAM usage during first layer propagastion

### DIFF
--- a/src/llmcompressor/pipelines/cache.py
+++ b/src/llmcompressor/pipelines/cache.py
@@ -34,7 +34,7 @@ class OverrideEqMode(TorchDispatchMode):
         # Check if the operation is equality
         if func is torch.ops.aten.eq.Tensor:
             # Override to use torch.equal
-            assert len(args)==2, "Exactly 2 args must be provided"
+            assert len(args) == 2, "Exactly 2 args must be provided"
 
             # NOTE: Errors out without cast to torch.tensor
             return torch.tensor(id(args[0]) == id(args[1]))


src/llmcompressor/pipelines/cache.py
```diff
@@ -42,6 +44,10 @@ class IntermediatesCache:
     batch_intermediates: list[IntermediateValues]
     offload_device: torch.device | None
 
+    # onload value -> offload value
+    # used to avoid excess memory usage when shared tensors are offloaded
+    offload_values: WeakKeyDictionary[torch.Tensor, torch.Tensor] = WeakKeyDictionary()
```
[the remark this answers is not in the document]
brian-dellabetta: Everything in this class is set up as a class attribute so that onload/offload can be used as classmethods, refactoring should be considered separately

src/llmcompressor/pipelines/cache.py
```diff
@@ -5,11 +5,42 @@
...
+    """
+
+    def __torch_dispatch__(self, func, _types, args=(), kwargs=None):
+        kwargs = kwargs or {}
+
+        # Check if the operation is equality
+        if func is torch.ops.aten.eq.Tensor:
+            # Override to use torch.equal
+            # NOTE: Errors out without cast to torch.tensor
+            return torch.tensor(torch.equal(*args, **kwargs))
```
kylesayrs: Torch equal is very slow. Shouldn't we just be comparing ids?

```suggestion
            return id(args[0]) == id(args[1])
```

tests/llmcompressor/pipelines/test_cache.py
```diff
@@ -162,3 +162,18 @@ def deep_equal(a, b) -> bool:
...
+    b = torch.tensor([1, 2, 3])
+    c = torch.tensor([2, 2, 2])
+
+    with pytest.raises(RuntimeError):
+        assert a == b
+    with pytest.raises(RuntimeError):
+        assert not (a == c)
+
+    with OverrideEqMode():
+        assert a == b
```
kylesayrs, line 178: If we compare ids, this should be False

src/llmcompressor/pipelines/cache.py
brian-dellabetta: updated

tests/llmcompressor/pipelines/test_cache.py
brian-dellabetta: updated after changing to id

kylesayrs (commented):
lgtm! This was very annoying to implement lol

PR #2376, opened 2026-02-17
[Examples] Deprecate `dispatch_for_generation` in favor of `dispatch_model`

kylesayrs:
## Purpose ##
* Start using `dispatch_model` as a primitive instead of `dispatch_for_generation`, which doesn't add anything but indirection

## Changes ##
* Find and replace `dispatch_for_generation` -> `dispatch_model`
* Add deprecation warning to `dispatch_for_generation`

### DIFF
--- a/docs/key-models/llama4/fp8-example.md
+++ b/docs/key-models/llama4/fp8-example.md
@@ -14,11 +14,11 @@ Let's walk through the main steps of the quantization process:
 Load the model using `AutoModelForCausalLM`:
 
 ```python
+from compressed_tensors.offload import dispatch_model
 from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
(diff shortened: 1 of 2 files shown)


docs/key-models/llama4/fp8-example.md
```diff
@@ -18,7 +18,7 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.quantization import QuantizationModifier
-from llmcompressor.utils import dispatch_for_generation
+from compressed_tensors.offload import dispatch_model
```
brian-dellabetta: looks like you need to run `make style` on these?

kylesayrs: Ha yeah, they're markdown examples but I can isort them

PR #2384, opened 2026-02-18
perf: make MSE observer compatible with torch.compile

Bias92:
## summary
torch.compile support for MSE observer with chunked execution, opt-in via `set_torch_compile(True)`. `Default off` (no behavior change).

### e2e 

(64 samples, max_seq=4096 for activation quantization)

|                              | +activation (Qwen3-8B) | -activation (Qwen3-8B) |  -activation (Qwen3-30B-A3B) |
|------------------------------|-------------|-------------|--------------|
| compile (this PR)            | 319s*        | 24s         |  65.68s*              |
| eager (this PR)              | 603s        | 19s *         | 149.74s |
| eager on main (`7a52e9e2`)   | 602s        | 19s         | N/A |

For fast quantization runs, compile can be worse by at most 10s, for longer quantization runs, the significant speedup of compiled MSE makes it run significantly faster.

best case (+act, static activation MSE): ~1.9x speedup with compile.
worst case (-act, weight-only): ~5s slower with compile (DataFreePipeline path, calibration too short to amortize compile overhead). first compile run has cold cache penalty (~28s); settled value ~24s.

memory (8B): ~2180 MB (+act), ~16000 MB (-act, DataFreePipeline keeps full model weights resident on GPU). Compile/eager identical at each setting.

### Inner-loop microbenchmark (`benchmarks/bench_mse_compile.py` at 8ca9342)

Isolated `_compute_chunk` on (4096, 4096) tensor, chunk_size=5:

| n  | eager (ms) | compiled warm (ms) | speedup |
|----|------------|--------------------|---------| 
| 8  | 11.05      | 2.03               | 5.45x   |
| 15 | 19.09      | 3.18               | 6.01x   |

range: 4.88x to 6.01x across n=8..15. Removing the `clone` + `mark_unbacked` guards (1 -> 12 recompiles) brought steady-state speedup from ~3x to ~5-6x. overhead is 5-10s and speedup is 5-6x resulting in needing somewhere between 300-1000 reps to be faster.

## recompiles
12 across the calibration run (rank mismatch + duck sizing + stride). stable across sequence length variation (tested with seq 360~3019). `torch._dynamo.config.recompile_limit` default is 8; once exceeded, the function falls back to dynamic shape mode without further recompilation.

note: when techniques are employed to reduce recompiles, overall performance degrades significantly.

## reproduction
```python
from llmcompressor.observers.compile_config import set_torch_compile

set_torch_compile(True)  # or False for eager baseline
oneshot(
    model="Qwen/Qwen3-8B",
    dataset="ultrachat-200k",
    splits={"calibration": "train_sft"},
    num_calibration_samples=64,
    max_seq_length=4096,
    recipe="""
    quant_stage:
        quant_modifiers:
            QuantizationModifier:
                ignore: ["lm_head"]
                config_groups:
                    group_0:
                        targets: ["Linear"]
                        weights:
                            num_bits: 8
                            type: "int"
                            symmetric: true
                            strategy: "group"
                            group_size: 128
                            observer: "mse"
                        input_activations:
                            num_bits: 8
                            type: "int"
                            symmetric: true
                            strategy: "tensor"
                            dynamic: false
                            observer: "mse"
    """,
)
```


HDCharles:
Thanks for your contribution!

# comments

high level i'm not sure if we can land with the stategy of removing early stopping. In my tests i remember the early stopping preventing us from searching ~90% of the search space, so in practice i think this will result in a 10x slowdown for the non compiled performance. Given that, your numbers here are a bit off because you're comparing PR to PR whereas the actual number that would be useful is main to PR, if my experience is correct, you've effectively slowed your baseline by 10x so the 35x speedup is really more like 3.5x of what it was before this PR. Thats still great but it seems like it ignores the compile overhead and so we'll need to make sure this is actually a positive change with some e2e testing.

I'm also a bit unsure about the end game for this improvement since it just makes the observer compilable but doesn't actually make it compiled. We'd want this to be something a user could disable if they run into trouble as well.

a good comparison PR might be this [compile optimizations for GPTQ PR](https://github.com/vllm-project/llm-compressor/pull/2320) as far as how they implement the compilation and the flag.

also what benchmark are you running? is this an actual example with real data or is a toy example? I'm not sure what Benchmark (CPU, inductor, shape=(1,1,4096)) means in practice or how i could run it myself.


--------------

# next steps

i think we most likely want the compile and non compiled path to be reasonable and by removing early stopping we're basically destroying the non compiled path. Given that, I think there are 4 strategies we should look at for how compilation could work here, 

## A - bring back early stopping, compile the inner loop
walk back the removal early stopping. instead, extract a compilable [inner loop](https://github.com/vllm-project/llm-compressor/pull/2384/changes#diff-b9472770e3291ffd5f7e0adbc074a5b2f83ca00e8dc937e230dcc365f4e2f954R201-R225) of the code and exclude the data dependent part from compilation. 

## B - bring back early stopping, hide it behind a flag so the whole thing can be compiled
do early stopping if its not compiled, but don't do early stopping if it is compiled. I think you can leave data dependent control flow but hide it behind a flag and if i remember correctly, compile will not see the data dependent flow unless the flag changes. may require some finagling to get compile to not try to look and see the data dependent control flow.

## C - chunked early stopping
we can partially remove early stopping, we can chunk the grid search into a few chunks and run those and only do a data dependent exit after calculating the mse for the whole chunk. This has the potential to actually be the fastest in terms of compile performance since it gets you both early stopping while still compiling a larger chunk in one go.

## D
leave as is with no early stopping but do a ton of testing to make sure its bulletproof so we won't have to disable compilation by default because we get a bunch of issues because GPTQ isn't working and then get really terrible performance.

I would probably do A by default but could see doing B or C depending on benchmark comparisons between them.

## These steps will be required regardless
1) add functionality to compile the optimized function
2) Add a flag to enable_observer_compilation either as a oneshot arg or on the modifiers (ideally enabled by default)  
3) get some benchmarks for the change in e2e speed for one of our examples using the MSE observer so we can see what happens when we take compiler overhead into account.
4) edit the existing observer tests to test with/without compile

HDCharles (changes requested):
see comment

Bias92:
Thanks for the detailed feedback! I agree that preserving the non-compiled path performance is important. I'll look into the GPTQ compile PR you referenced and implement strategy A (bring back early stopping, compile the inner loop) as a starting point. I'll also add the enable_observer_compilation flag and run e2e benchmarks with real data. Will update the PR soon.

kylesayrs:
Hi @Bias92, could you please post your original baseline performance metrics, and post the script you're using to benchmark? I don't fully understand where your speedups are coming from, as you do not actually add a torch.compile decorator anywhere?

Bias92:
"hi @kylesayrs, ye that's right , I'll share the benchmark script and original baseline metrics. the speedups come from externally wrapping the function with `torch.compile(fullgraph=True)`, which wasn't included in the PR itself. as @HDCharles suggested, I'll add proper compilation functionality with an `enable_observer_compilation flag` in the next update."

Bias92:
Updated the implementation following the GPTQ PR #2320 pattern:

- Restored original `_grid_search_mse` with early stopping + `patch_attr` (non-compiled path preserved)
- Added `_grid_search_mse_compiled` as a separate torch.compile-compatible path
- Added `enable_torch_compile` flag via `observer_kwargs` (default `False`)
- Added `_call_grid_search` helper to reduce code duplication across observer classes

Non-compiled path performance is now identical to the original baseline. Will work on e2e benchmarks and lint fixes next.

Bias92:
Updated the implementation following the GPTQ PR #2320 pattern:

- Restored original `_grid_search_mse` with early stopping + `patch_attr` (non-compiled path preserved)
- Added `_grid_search_mse_compiled` as a separate torch.compile-compatible path
- Added `enable_torch_compile` flag via `observer_kwargs` (default `False`)
- Added `_call_grid_search` helper to reduce code duplication across observer classes

Non-compiled path performance is now identical to the original baseline. Will work on e2e benchmarks and lint fixes next.

Bias92:
### E2E Benchmark Results (RTX 4060 Ti, CUDA)

**TinyLlama-1.1B, FP8 W8A8, MSE observer, 64 calibration samples:**

| Metric | Baseline | Compiled | Delta |
|---|---|---|---|
| Time (s) | 26.5 | 24.4 | -7.8% |
| Peak memory (MB) | 2200 | 2200 | +0.0% |
| Quantized layers | 154 | 154 | - |
| Scales all_close | - | - | True |
| Max abs diff | - | - | 0.00e+00 |
| Matching layers | - | - | 154/154 |

**Calibration-only speedup**: ~1.15x (5.0s → 4.35s). E2E speedup is 1.09x because model loading + tokenization (~21s) dominates total time. Larger models where calibration is a bigger fraction of total time should see greater benefit.

Key findings:
- Zero memory overhead
- Numerically identical weight scales (154/154 layers match exactly)
- Non-compiled path preserves original baseline performance with early stopping

Benchmark script: `tests/llmcompressor/observers/benchmark_mse_compile.py`

src/llmcompressor/observers/mse.py
```diff
@@ -215,7 +252,7 @@ def _grid_search_mse(
                 global_scale=global_scale,
             ).to(observed.dtype)
             # Note that due to forward quantization implementation, token quant,
-            # unlike tensor_group, requires extra dtype cast
+            # unlike tensor_group, requires extra dtype cast
```
HDCharles: what is going on with the formatting here?

src/llmcompressor/observers/mse.py
```diff
@@ -234,3 +271,81 @@ def _grid_search_mse(
...
+        ).to(observed.dtype)
+
+        err = torch.sum((q - observed).abs().pow(norm), dim=(0, -1))
+
+        # Use torch.where instead of boolean indexing + torch.any for
+        # torch.compile compatibility (avoids data-dependent control flow)
+        improved = err < best_error
+        best_error = torch.where(improved, err, best_error)
+        best_min_val = torch.where(improved, shrinked_min_val, best_min_val)
+        best_max_val = torch.where(improved, shrinked_max_val, best_max_val)
```
HDCharles: this should be extracted into a separate function and used in both compiled and non-compiled paths so that we don't duplicate code and have to maintain the same functionality in 2 places.

I think both your change to getting rid of patch and the change to use torch.where are fine for the hotpath.

tests/llmcompressor/observers/test_observer_compile.py
```diff
@@ -0,0 +1,21 @@
...
+@pytest.mark.parametrize("shape", [(1, 1, 128), (1, 4, 128), (2, 1, 256)])
+def test_observer_torch_compile(observer_cls, shape):
+    args = QuantizationArgs(num_bits=8, type="int", symmetric=True, strategy="tensor")
+    observer = observer_cls(base_name="weight", args=args)
+    x = torch.randn(*shape)
+    eager_scale, eager_zp = observer(x)
+    compiled = torch.compile(observer, fullgraph=True, backend="eager")
+    compiled_scale, compiled_zp = compiled(x)
+    torch.testing.assert_close(eager_scale, compiled_scale)
+    torch.testing.assert_close(eager_zp, compiled_zp)
```
HDCharles: this can just be added to the existing observer mse tests file insteaed of creating a new one.

Bias92:
Addressed all review feedback:

1. **Extracted shared helper**: `_compute_candidate_error()` used by both compiled and non-compiled paths — no more code duplication
2. **Removed `patch_attr`** from both paths (pre-created `token_args` instead)
3. **Preserved early stopping** in non-compiled path (original baseline performance unchanged)
4. **Moved compile test** into `test_mse.py`, removed `test_observer_compile.py`

All 53 tests passing locally (0 failures, 4 skipped).

src/llmcompressor/observers/mse.py
Bias92: Fixed — this comment is now inside the shared `_compute_candidate_error()` helper. The previous diff was just a line-number shift.

HDCharles:
can you fix quality issues?

Bias92:
Done — ran `make style` and `make quality`, both passing now.

Bias92:
Hi @HDCharles, friendly ping — I've addressed all the feedback from your review:

Restored original _grid_search_mse with early stopping (non-compiled baseline unchanged)
Added _grid_search_mse_compiled as a separate path with enable_torch_compile flag (default False)
Extracted _compute_candidate_error() shared helper to eliminate duplication
Moved compile test into test_mse.py, removed separate test file
Quality checks passing

Would you be able to re-review when you have a moment? Also, CI is pending workflow approval if you could approve that as well. Thanks!

Bias92:
Hi @HDCharles, @dsikka, @kylesayrs — friendly ping!
I've addressed all the feedback from the previous review:

Restored original _grid_search_mse with early stopping (non-compiled path unchanged)
Added _grid_search_mse_compiled as a separate torch.compile-compatible path
Added enable_torch_compile flag via observer_kwargs (default False)
Extracted _compute_candidate_error() shared helper to reduce duplication
Fixed all lint/formatting issues (make style + make quality both passing)

The only remaining blocker is workflow approval for CI — once that's approved the checks can actually run. Would appreciate a re-review when you get a chance!

HDCharles:
Hey sorry about the wait, reach out to me on vllm slack if i don't respond in ~24 hours

HDCharles (commented):
why is findstr in the PR?

tests/llmcompressor/observers/benchmark_mse_compile.py
```diff
@@ -0,0 +1,312 @@
...
+        torch.cuda.empty_cache()
+
+    return {
+        "elapsed_s": elapsed,
+        "peak_memory_mb": peak_memory_mb,
+        "weight_scales": weight_scales,
+    }
+
+
+def compare_scales(
```
HDCharles: don't think this is needed, its unit tested in test_mse

HDCharles:
this isn't torch.compiling the function?

HDCharles:
can you reach out to me on vllm slack there's some additional things to discuss

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -406,9 +408,10 @@ def oneshot(
 
     # pass all args directly into Oneshot
     local_args = {
-        k: v for k, v in locals().items() if k not in ("local_args", "kwargs")
+        k: v for k, v in locals().items() if k not in ("local_args", "kwargs", "enable_observer_compile")
```
HDCharles: lets just make this global, enable_torch_compile and then you can choose the branch in observer based on this global state

Bias92: sounds good, i'll remove the oneshot params and just use a global enable_torch_compile flag that the observer reads directly

Bias92: removed the oneshot params. its just a global enable_torch_compile now that the observer reads directly.

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -406,7 +406,13 @@ def oneshot(
 
     # pass all args directly into Oneshot
     local_args = {
-        k: v for k, v in locals().items() if k not in ("local_args", "kwargs")
+        k: v
```
HDCharles: Get rid of random unrelated changes like this

src/llmcompressor/observers/mse.py
```diff
@@ -119,55 +136,217 @@ def __init__(self, *args, **kwargs):
...
-            self.maxshrink,
-            self.patience,
-            self.grid,
-            self.norm,
-            global_scale=None,
-            optimize_global_scale=True,
+        return self._call_grid_search(observed, None, True)
+
+
+def _compute_candidate_error(
```
HDCharles: Probably only need a full doc string on either this or chunk function rather than both

src/llmcompressor/observers/mse.py
```diff
@@ -178,59 +357,90 @@ def _grid_search_mse(
...
+            # the entire chunk, increment by chunk size towards patience.
+            if torch.equal(prev_best, best_error):
+                no_improve_count += current_chunk
+                if no_improve_count >= patience:
+                    break
+            else:
+                no_improve_count = 0
+            idx = chunk_end
+    else:
+        # Eager or single-step compile path.
```
HDCharles: Why is single step compile separated out, shouldn't it just be chunk=1?

HDCharles (changes requested):
Looks good just some minor things, also put your benchmarks for cold start if you can

Bias92:
## Benchmarks (A100 SXM 80GB)

### E2E: Qwen3-8B, weight group quantization (group_size=128) + tensor activation, static MSE, 64 cal samples
| | time | memory | speedup |
|---|---|---|---|
| eager | 1190.6s | 2069 MB | — |
| compiled cold | 881.7s | 2069 MB | 1.35x |
| compiled warm | 888.1s | 2069 MB | 1.34x |

### Isolated _compute_chunk microbenchmark (group strategy, group_size=128, synthetic inner-loop)
| hidden_dim | eager | compiled | speedup |
|---|---|---|---|
| 512 | 3.12ms | 0.16ms | 20.0x |
| 4096 | 3.14ms | 0.17ms | 19.1x |
| 16384 | 3.14ms | 0.16ms | 19.3x |

Recompiles: 4→1 with clone + mark_unbacked. The remaining recompile is a rank mismatch on the weight→activation transition.

src/llmcompressor/observers/mse.py
```diff
@@ -119,55 +136,195 @@ def __init__(self, *args, **kwargs):
...
-            self.maxshrink,
-            self.patience,
-            self.grid,
-            self.norm,
-            global_scale=None,
-            optimize_global_scale=True,
+        return self._call_grid_search(observed, None, True)
+
+
+def _compute_candidate_error(
```
HDCharles: the MSE stuff is now dominating the file, lets move all this to like helpers or mse_quant or something like that

HDCharles:
> ## Benchmarks (A100 SXM 80GB)
> ### E2E: Qwen3-8B, weight group quantization (group_size=128) + tensor activation, static MSE, 64 cal samples
> 	time 	memory 	speedup
> eager 	1190.6s 	2069 MB 	—
> compiled cold 	881.7s 	2069 MB 	1.35x
> compiled warm 	888.1s 	2069 MB 	1.34x
> ### Isolated _compute_chunk microbenchmark (group strategy, group_size=128, synthetic inner-loop)
> hidden_dim 	eager 	compiled 	speedup
> 512 	3.12ms 	0.16ms 	20.0x
> 4096 	3.14ms 	0.17ms 	19.1x
> 16384 	3.14ms 	0.16ms 	19.3x
> 
> Recompiles: 4→1 with clone + mark_unbacked. The remaining recompile is a rank mismatch on the weight→activation transition.

can you include instructions to how you got these numbers so they're reproducible?

also no need to include warm/cold if benchmarks are close, cold is the more important one

HDCharles:
also update the PR description, it looks like it has old benchmarks?

HDCharles:
failing quality

Bias92:
i jsut have fixed locally and pushed a follow-up commit. make style and make quality both pass on my side.

src/llmcompressor/observers/compile_config.py
```diff
@@ -0,0 +1,16 @@
+_enable_torch_compile = False
+_compile_chunk_size = 5
+
+
+def set_torch_compile(enabled: bool, chunk_size: int = 5):
+    global _enable_torch_compile, _compile_chunk_size
+    _enable_torch_compile = enabled
+    _compile_chunk_size = chunk_size
```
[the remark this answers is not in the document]
HDCharles: Yeah, assert this >0

src/llmcompressor/observers/compile_config.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def get_torch_compile() -> bool:
+    return _enable_torch_compile
+
+
+def get_compile_chunk_size() -> int:
+    return _compile_chunk_size
```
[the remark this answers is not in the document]
HDCharles: Nah this seems fine

src/llmcompressor/observers/mse.py
```diff
@@ -178,59 +335,92 @@ def _grid_search_mse(
...
-        shrinked_min_val = p * min_val
-        shrinked_max_val = p * max_val
+    if enable_compile:
+        # [recompile fix] Eliminate stride/duck-sizing guards from view tensors
+        observed = observed.clone()
 
-        if optimize_global_scale:
-            global_scale = generate_gparam(shrinked_min_val, shrinked_max_val)
+        # [recompile fix] Prevent size specialization on group-size dim
+        torch._dynamo.decorators.mark_unbacked(observed, observed.ndim - 1)
```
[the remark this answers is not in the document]
Bias92: acknowledged. coderabbit's own analysis confirms there s no public stable 
alternative. `mark_unbacked` is the documented way to mark unbacked 
dimensions in `PyTorch 2.x.` The call is required to prevent size 
specialization on the `group-size dim` , which would otherwise cause 
recompiles. the existing inline comment notes the intent. i'll monitor 
torch release notes; if a public API replacement is exposed i can switch.

src/llmcompressor/observers/mse.py
```diff
@@ -178,59 +353,92 @@ def _grid_search_mse(
...
+                best_max_val,
+            )
+            # Chunk-level early stopping: if no channel improved across
+            # the entire chunk, increment by chunk size towards patience.
+            if torch.equal(prev_best, best_error):
+                no_improve_count += current_chunk
+                if no_improve_count >= patience:
+                    break
+            else:
+                no_improve_count = 0
```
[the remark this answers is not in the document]
Bias92: intentional trade-off. chunked compile path checks patience at chunk 
boundaries (`chunk_size=5`) to amortize compile overhead. `best_error` is 
preserved across remaining iterations within the chunk, so result 
correctness is unchanged. only timing differs by up to `chunk_size-1` 
steps. documented in the docstring (commit 6877d508).

src/llmcompressor/observers/mse.py
```diff
@@ -119,55 +136,195 @@ def __init__(self, *args, **kwargs):
...
+# reducing per-call overhead by the chunk factor.
+_compute_chunk_compiled = torch.compile(_compute_chunk, dynamic=True)
+
 
 def _grid_search_mse(
     observed: torch.Tensor,
     args: QuantizationArgs,
+    token_args: QuantizationArgs,
     maxshrink: float,
     patience: float,
```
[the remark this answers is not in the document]
Bias92: fixed in commit 64398009. patience type changed from float to int.

tests/llmcompressor/observers/test_mse.py
```diff
@@ -87,3 +87,36 @@ def test_mse_fp4():
...
+    set_torch_compile(True)
+    compiled_scale, compiled_zp = observer(x)
+
+    torch.testing.assert_close(eager_scale, compiled_scale)
+    torch.testing.assert_close(eager_zp, compiled_zp)
+
+    # cleanup
+    set_torch_compile(False)
+    torch.testing.assert_close(eager_scale, compiled_scale)
+    torch.testing.assert_close(eager_zp, compiled_zp)
```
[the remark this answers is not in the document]
Bias92: fixed in commit a341c9ed. wrapped in `try/finally` and removed the stale post cleanup assertions.

src/llmcompressor/observers/mse.py
Bias92: moved to mse_quant.py in commit 555b921d.

tests/llmcompressor/observers/benchmark_mse_compile.py
Bias92: removed in `force-push` (befb2e2d).

src/llmcompressor/observers/mse.py
Bias92: addressed in your refactor commit 3106833c.

src/llmcompressor/observers/mse.py
Bias92: addressed during the refactor. chunked path docstring is in mse_quant.py.

src/llmcompressor/entrypoints/oneshot.py
Bias92: cleaned up during rebase. current diff against main is empty.

src/llmcompressor/observers/compile_config.py
```diff
@@ -0,0 +1,18 @@
+_enable_torch_compile = False
+_compile_chunk_size = 5
+
+
+def set_torch_compile(enabled: bool, chunk_size: int = 5):
```
HDCharles: this should be enabled/disabled in oneshot, can you add a enable_compile flag in oneshot that can modify this?

HDCharles: also this shouldn't include chunk size, chunk size should be an observer specific arg

src/llmcompressor/observers/mse.py
```diff
@@ -51,32 +48,38 @@ def __init__(self, *args, **kwargs):
         self.grid = observer_kwargs.get("grid", 100.0)
         self.norm = observer_kwargs.get("norm", 2.4)
```
HDCharles, line 27: chunk size should be handled in the same way as patience.

HDCharles (changes requested):
1) add so that this can be enabled/disabled in oneshot
2) move chunk size into observer
3) move compile into llm compressor proper

after that this should be good

src/llmcompressor/observers/compile_config.py
HDCharles: also lets move compile config into llmcompressor rather than observers

HDCharles:
fix quality failures

src/llmcompressor/observers/compile_config.py
brian-dellabetta: agree, ideally this is in the oneshot args instead of a global var. We should also scope it as appropriate. if it is compiling only the MSE grid search, the var name should be clear

src/llmcompressor/observers/mse_quant.py
```diff
@@ -0,0 +1,286 @@
...
+    """
+    min_val = torch.amin(observed, dim=(0, -1))
+    max_val = torch.amax(observed, dim=(0, -1))
+    best_error = torch.full_like(min_val, torch.finfo(min_val.dtype).max)
+    best_min_val = min_val.clone()
+    best_max_val = max_val.clone()
+
+    total_steps = int(maxshrink * grid)
+
+    dispatch = _grid_search_compiled if get_torch_compile() else _grid_search_eager
```
brian-dellabetta: rather than carrying around two separate implementations, can't we just do something like this?
```suggestion
    dispatch = _grid_search_eager if get_torch_compile() else torch.compile(_grid_search_eager)
```
I thought that was the original intention with torch.compile, it can act as a decorator or be applied directly to a function/module?

brian-dellabetta (commented):
Thanks for the contribution! a couple comments/questions on user experience and the way we're using torch.compile here

src/llmcompressor/observers/mse_quant.py
HDCharles: i want to have a single compile flag for everything, so i'd rather not scope it so limitedly

HDCharles: nah the code change you're suggesting doesn't work. We have to have a compile specific approach because the major part of compile's speedup is unrolling the for loop but when you do that in eager, it makes things slower

brian-dellabetta: oh i thought torch compile handled for loop unrolling, and barring that torch.vmap since these are all the same vectorizable operation with equivalent tensor shapes.

HDCharles: I think my comment was a bit inaccurate:

The problem is that we have data dependent flow, we exit based on whether error did/didn't improve and that's cancer to compile so we can

A) compile inside the data dependent flow (compile each loop sepeeately)
B) chunk multiple loops together and check the data dependent flow after the chunk finishes.

@Bias92 tested both and the latter was much faster.

So that means we have to do the chunked approach in eager (slower) or have 2 separate paths.

brian-dellabetta: Gotcha, yeah if there are early stopping or other conditional constraints, then what you have (a torch-compiled compute chunk) should be the same as torch.vmap. I will mark as resolved, bummer that we have to carry around 2 impls though

HDCharles: Agree

src/llmcompressor/observers/mse.py
Bias92: `chunk_size` now reads from `observer_kwarsg` in __init__ alongside patience. at  commit 8ff11ac4

src/llmcompressor/observers/compile_config.py
Bias92: addressed in commits 0ce3a2bc, 8ff11ac4 , 8f76d878. `compile_config` moved to top-level, `chunk_size` now an observer kwarg, enable_compile flag added to oneshot.

src/llmcompressor/entrypoints/oneshot.py
HDCharles: its still here

HDCharles (approved):

HDCharles:
Ahh I need to move the enable compile and maybe have parse args work correctly with it.

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -69,6 +70,7 @@ def on_start(self, state: State, event: Event, **kwargs):
         Begin calibrating activations.
         """
         self.started_ = True
+        set_torch_compile(state.enable_compile)
```
brian-dellabetta: this is only called during QuantizationModifier, but needs to be propagated to any QuantizationMixin modifier like GPTQModifier. Better suited inside `QuantizationMixin.start_calibration`?

src/llmcompressor/observers/mse_quant.py
```diff
@@ -0,0 +1,297 @@
...
+from compressed_tensors.quantization.lifecycle import fake_quantize
+from compressed_tensors.quantization.utils import calculate_qparams, generate_gparam
+
+from llmcompressor.observers.base import MinMaxTuple
+
+_enable_torch_compile = False
+
+
+def set_torch_compile(enabled: bool):
+    global _enable_torch_compile
```
brian-dellabetta: doesn't having field `state.enable_compile` remove the need for a global var? This seems better set as a class var on `MemorylessMSEObserver` when state is accessed.

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -290,6 +290,13 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
+    enable_compile: bool = field(
+        default=False,
+        metadata={
+            "help": "If True, use torch.compiled MSE observer inner loop for faster "
+            "calibration. Default False."
```
brian-dellabetta: If this is scoped to just MSE observer, seems like it should be a kwarg on observer. if it's intended to be used as a global setting, the help should reflect that. Are there cases where users would want torch.compile enabled for one function and disabled for another?
```suggestion
            "help": "If True, use torch.compiled functions where available for faster "
            "calibration. Default False."
```

brian-dellabetta (commented):
Thanks for the contribution @Bias92 . A few comments/questions

src/llmcompressor/modifiers/quantization/quantization/base.py
HDCharles: oof, yeah

src/llmcompressor/args/dataset_arguments.py
HDCharles: @Bias92 is also working on a GPTQ compile PR, i'd like to unify the compile settings. We already have too many flags and knobs that are completely ignored by even ourselves, I wanted to consolidate all the compile stuff together.

src/llmcompressor/observers/mse_quant.py
HDCharles: yeah

src/llmcompressor/args/dataset_arguments.py
brian-dellabetta: Can we update the help then as i suggested or similar?

HDCharles: yeah, done

brian-dellabetta (approved):

PR #2387, opened 2026-02-18
[Bugfix] Fix circular references when activation offload device is cuda

kylesayrs:
## Background ##
#2366 introduced a `WeakKeyDictionary` layer which caches shared tensors. This is a good approach, but has an edge case where, if the value of entry is identical to the key of the entry, then the key will never be garbage collected.

This can occur if the user specifies `sequential_offload_device="cuda"`, or if the AWQ offload device is "cuda" (default true in most cases).

## Purpose ##
* Fix memory leak in AWQ which led to very high CUDA memory usage

## Changes ##
* Guard against entries into the `WeakKeyDictionary` where the key and value are identical
* Misc
  * Move `OverrideEqMode` to the bottom of the `pipelines/cache.py`
  * Remove `_fp16_baseline_cache`, which was not being used

## Testing ##
| Before Changes | After Changes |
| - | - |
| awq_before | awq_after |

### DIFF
--- a/examples/model_free_ptq/qwen3_fp8_block.py
+++ b/examples/model_free_ptq/qwen3_fp8_block.py
@@ -1,0 +1,19 @@
+from llmcompressor import model_free_ptq
+
+MODEL_ID = "Qwen/Qwen3-0.6B"
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-FP8-BLOCK"
+
+# Apply FP8-Block to the model
+# Once quantized, the model is saved
+# using compressed-tensors to the SAVE_DIR.
+model_free_ptq(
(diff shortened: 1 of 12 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -172,10 +172,6 @@ class AWQModifier(Modifier, QuantizationMixin):
     )
     # List to store error metrics for each layer
     _error_metrics: list[dict] = PrivateAttr(default_factory=list)
-    # Cache FP16 baseline outputs for each parent module, one list of tensors per batch
-    _fp16_baseline_cache: dict[Module, IntermediatesCache] = PrivateAttr(
```
HDCharles, line 176: oops

PR #2390, opened 2026-02-20
Add support for passing a custom DataLoader to oneshot()

SorenDreano:
SUMMARY:
Adds a `dataloader` argument to the `oneshot` entrypoint.

Allow users to pass a pre-built PyTorch DataLoader directly via the `dataloader` parameter, bypassing the internal dataset-to-dataloader conversion. This is useful for custom data pipelines where users already have a prepared DataLoader and don't need get_calibration_dataloader().

Rather than using `self.dataloader = kwargs.pop("dataloader", None)`, we could also add a `dataloader` argument/attribute to `DatasetArguments` if you prefer.

TEST PLAN:
This change is fairly trivial, I made sure [https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/README.md](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/README.md) could still run and that passing the DataLoader too:

```python
from transformers import AutoTokenizer, AutoModelForCausalLM

NUM_CALIBRATION_SAMPLES=512
MAX_SEQUENCE_LENGTH=2048

# Load dataset.
ds = load_dataset("HuggingFaceH4/ultrachat_200k", split=f"train_sft[:{NUM_CALIBRATION_SAMPLES}]")
ds = ds.shuffle(seed=42)

# Preprocess the data into the format the model is trained with.
def preprocess(example):
    return {"text": tokenizer.apply_chat_template(example["messages"], tokenize=False,)}
ds = ds.map(preprocess)

# Tokenize the data (be careful with bos tokens - we need add_special_tokens=False since the chat_template already added it).
def tokenize(sample):
    return tokenizer(sample["text"], padding=False, max_length=MAX_SEQUENCE_LENGTH, truncation=True, add_special_tokens=False)
ds = ds.map(tokenize, remove_columns=ds.column_names)

from llmcompressor.datasets import get_calibration_dataloader

dataset_args = DatasetArguments(
    dataset=ds,
)

dataloader = get_calibration_dataloader(dataset_args, tokenizer)

# Apply quantization.
oneshot(
    model=model, 
    recipe=recipe,
    dataloader=dataloader,
)

# Save to disk compressed.
SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-W4A16-G128"
```

This is the exact same code from the documentation, with the DataLoader built outside of the `oneshot` call (`dataloader = get_calibration_dataloader(dataset_args, tokenizer)`) and passed directly to `oneshot`.


SorenDreano:
I'll happily review Gemini's propositions (especially the docstring one) if you approve this new `dataloader` argument, depending on whether you prefer to add it to `DatasetArguments` or the way I did it with `pop()`.

brian-dellabetta:
Hi @SorenDreano , rather than adding a new field for `dataloader` can you instead expand the options so that a `torch.utils.data.DataLoader` type to be passed in as `dataset`, and skip the call to `get_calibration_dataset` if `isinstance(dataset, torch.utils.data.DataLoader)`? I think that keeps the API cleaner.

cc @HDCharles @kylesayrs if you disagree

SorenDreano:
It makes sense @brian-dellabetta , applied the changes in my latest commit. Let me know the way you prefer:
- separate dataloader argument
- separate dataloader argument in DatasetArgs
- dataset that can be a DataLoader

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -101,12 +103,13 @@ class DatasetArguments(CustomDatasetArguments):
     arguments to be able to specify them on the command line
     """
 
-    dataset: str | None = field(
+    dataset: str | DataLoader | None = field(
```
brian-dellabetta: This was actually wrong before, we can update here too
```suggestion
    dataset: str | Dataset | DatasetDict | DataLoader | None = field(
```

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -10,6 +10,10 @@
 from dataclasses import dataclass, field
 from typing import Callable
 
+from torch.utils.data import DataLoader
+
+from datasets import Dataset, DatasetDict
+
```
brian-dellabetta, line 15: To avoid these imports at runtime, we can wrap in an `if TYPE_CHECKING` loop:
```suggestion
from __future__ import annotations
from typing import Callable, TYPE_CHECKING

if TYPE_CHECKING:
    from torch.utils.data import DataLoader
    from datasets import Dataset, DatasetDict

```

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -182,9 +182,12 @@ def __call__(self):
...
 
-        calibration_dataloader = get_calibration_dataloader(
-            self.dataset_args, self.processor
-        )
+        if isinstance(self.dataset_args.dataset, DataLoader):
+            calibration_dataloader = self.dataset_args.dataset
+        else:
+            calibration_dataloader = get_calibration_dataloader(
+                self.dataset_args, self.processor
+            )
```
HDCharles: instead of handling this ad-hoc in Oneshot, better to extend the get_calibration_dataloader to handle both branches.

also that would keep the lower level logic away from this area which is a hotpath we'd like to keep highlevel

SorenDreano:
I moved the `isinstance` check as requested by @HDCharles 

Also, I corrected the typing of the function `get_calibration_dataloader` since it can just `return` in the case of calibration-free quantization.
Also, according to [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations):

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable)

So I replaced the `return` by `return None` in `get_calibration_dataloader`

HDCharles (approved):

PR #2391, opened 2026-02-22
[Distributed] Extend QuantizationModifier to support distributed activation calibration

### LINKED ISSUE #2220: Extend modifiers to support weight-parallel optimization  - QuantizationModifier

Etelis:
Closes #2220

Adds DDP support to `QuantizationModifier` for activation observer synchronization across multiple GPUs during calibration.

At `SEQUENTIAL_EPOCH_END` and `CALIBRATION_EPOCH_END`, activation observer min/max values are all-reduced across ranks. Scale/zp are then recomputed from the global statistics so all ranks have identical quantization parameters.

### Changes

- Add `synchronize()`, `recompute_qparams()`, `recompute_global_scale()` to `Observer` base class
- Add `sync_activation_observers()` to `QuantizationMixin` (shared by `QuantizationModifier` and `GPTQModifier`)
- Batch all async `dist.all_reduce` operations and wait once, matching GPTQ DDP pattern
- Remove `recompute_qparams_from_observer` from `calibration.py` (now encapsulated in Observer methods)
- Align distributed example with existing DDP patterns (`init_dist`, `get_rank_partition`)

### Runtime & Evaluation Results

Model: `Meta-Llama-3-8B-Instruct`, W8A8 (static input activations), 256 calibration samples

| Setup | Quantization Time | Speedup | Perplexity (wikitext) |
|-------|------------------:|--------:|----------------------:|
| 1x A100-80GB (baseline) | 270.1s | 1.0x | 37.51 |
| 2x A100-80GB (DDP) | 138.0s | 1.96x | 37.64 |
| 4x A100-80GB (DDP) | 72.1s | 3.75x | 36.40 |

- Near-linear scaling with GPU count
- Perplexity matches across all configurations, confirming activation observer sync produces correct quantization parameters

### Test plan

- [x] Unit tests: `pytest tests/llmcompressor/utils/test_distributed.py` (8 tests)
- [x] Multi-GPU tests: `torchrun --nproc_per_node=2 -m pytest tests/llmcompressor/modifiers/quantization/test_quantization_ddp.py` (2 tests)
- [x] End-to-end quantization + eval on 1/2/4 GPUs (see results table above)

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -47,6 +47,11 @@ pull_request_rules:
           - files~=^[^/]+\.md$
           - files~=^docs/
           - files~=^examples/
+      - -files~=^src/
+      - -files~=^tests/
+      - -files~=^\.github/
+      - -files~=^Makefile$
+      - -files~=^pyproject\.toml$
     actions:
       label:
         add:


src/llmcompressor/utils/distributed.py
```diff
@@ -0,0 +1,194 @@
+"""
+Distributed utilities for multi-GPU (DDP) calibration and optimization.
```
kylesayrs: Consolidate with `src/llmcompressor/utils/dist.py`

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -84,21 +106,107 @@ def on_start(self, state: State, event: Event, **kwargs):
...
+            update_weight_global_scale(module)
+
+        # broadcast global_scales so all ranks can run the fuse step
+        for _, module in named_modules:
+            src_rank = module_to_rank[module]
+            broadcast_module_parameter(module, "weight_global_scale", src_rank)
+
+        # fuse global_scales (all ranks, idempotent)
+        for module in model.modules():
+            update_fused_layer_weight_global_scales(module)
```
kylesayrs: This part should be tested with NVFP4. The problem is that there may be cases where submodules which need to be fused are assigned to different ranks. This is why I suggested breaking this out for @GOavi101 to focus on.

kylesayrs: My suggestion is to do the weight calibration independently on each rank for now, and focus on the activation calibration.

src/llmcompressor/utils/distributed.py
```diff
@@ -0,0 +1,194 @@
...
+    All-reduce a tensor with MIN op across all ranks. No-op if not distributed.
+
+    :param tensor: tensor to reduce
+    :return: reduced tensor
+    """
+    if not is_distributed():
+        return tensor
+
+    device = tensor.device
+    if device.type == "cpu":
```
kylesayrs: Why does the tensor need to be moved to the gpu?
I also don't see observer values being on the cpu as a common case, I'm wondering when you ever saw this?

src/llmcompressor/utils/distributed.py
```diff
@@ -0,0 +1,194 @@
...
+        return
+
+    # NCCL requires each rank to use its own GPU
+    device = torch.device(f"cuda:{dist.get_rank()}")
+    tensor = param.data.to(device)
+    dist.broadcast(tensor, src=src_rank)
+    update_offload_parameter(module, param_name, tensor)
+
+
+def all_reduce_min(tensor: torch.Tensor) -> torch.Tensor:
```
kylesayrs: I think the code would be a bit clearer if we didn't break this function out, and instead just used `dist.all_reduce` directly.

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -84,21 +106,107 @@ def on_start(self, state: State, event: Event, **kwargs):
...
+
+        for _, module in match_named_modules(model, self.resolved_targets, self.ignore):
+            for base_name in ("input", "output", "q", "k", "v"):
+                observer = getattr(module, f"{base_name}_observer", None)
+                if observer is None:
+                    continue
+
+                # all-reduce accumulated min/max across ranks
+                if (
+                    hasattr(observer, "past_min_vals")
```
kylesayrs: Instead of having many `hasattr` calls, which is not very maintainable, consider implementing a `synchronize()` method on the observers directly. I also think that you need to synchronize global scales.

In addition, having many synchronization ops has lots of runtime cost. Consider implementing `synchronize()` with a return value of the comms, similar to the GPTQ implementation https://github.com/vllm-project/llm-compressor/blob/5f63d7a9a6ae0f9944e69ff87ea5cce31f923ae2/src/llmcompressor/modifiers/quantization/gptq/base.py#L349

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -84,21 +106,107 @@ def on_start(self, state: State, event: Event, **kwargs):
...
 
+        if event.type_ == EventType.SEQUENTIAL_EPOCH_END:
+            self._sync_activation_observers(state.model)
+
         if event.type_ == EventType.CALIBRATION_EPOCH_END:
+            self._sync_activation_observers(state.model)
             if not self.ended_:
                 self.on_end(state, None)
 
+    def _sync_activation_observers(self, model):
```
kylesayrs: I think that you'll want to implement this method on the `QuantizationMixin`, that way DDP activation calibration logic can be shared with other modifiers like `GPTQModifier`

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -235,6 +236,41 @@ def calibrate_value_hook(module: Module, value_states: torch.Tensor):
     calibrate_activations(module, value_states, base_name="v")
 
 
+def recompute_qparams_from_observer(module: Module, base_name: str):
```
kylesayrs: Is this function not redundant with [`call_observer`](https://github.com/vllm-project/llm-compressor/blob/5f63d7a9a6ae0f9944e69ff87ea5cce31f923ae2/src/llmcompressor/modifiers/quantization/calibration.py#L87)?

examples/big_models_with_sequential_onloading/llama3_8b_w8a8_distributed.py
```diff
@@ -0,0 +1,98 @@
...
+# Increasing the number of samples can improve accuracy.
+NUM_CALIBRATION_SAMPLES = 256
+MAX_SEQUENCE_LENGTH = 2048
+
+# Initialize distributed.
+# Usage: torchrun --nproc_per_node=2 llama3_8b_w8a8_distributed.py
+dist.init_process_group(backend="nccl")
+rank = dist.get_rank()
+world_size = dist.get_world_size()
+torch.cuda.set_device(rank)
```
kylesayrs: Please use `init_distributed` util

src/llmcompressor/utils/distributed.py
Etelis: Done. Deleted `distributed.py` entirely — all functions were either for weight partitioning (removed) or `all_reduce` wrappers (inlined into `Observer.synchronize()`). No functions needed to move to `dist.py`.

src/llmcompressor/modifiers/quantization/quantization/base.py
Etelis: Done. Removed distributed weight calibration entirely. Weight calibration now runs independently on each rank (identical results since each rank sees the full model). This avoids the NVFP4 fusion issue you flagged.

src/llmcompressor/utils/distributed.py
Etelis: Removed. The CPU-to-GPU movement and the `all_reduce` wrapper are both gone — `Observer.synchronize()` now calls `dist.all_reduce` directly on the observer tensors (which are always on GPU in practice).

src/llmcompressor/utils/distributed.py
Etelis: Done. Removed `all_reduce_min`/`all_reduce_max` wrappers. `Observer.synchronize()` calls `dist.all_reduce` directly with the appropriate `ReduceOp`.

src/llmcompressor/modifiers/quantization/quantization/base.py
Etelis: Done. Added `synchronize()` on the `Observer` base class that returns a list of `dist.Work` handles (matching the GPTQ async pattern). All `all_reduce` ops are batched and waited on once via `wait_for_comms`. Also added `recompute_global_scale()` and `recompute_qparams()` to encapsulate the recomputation from accumulated state. Memoryless observers (no `past_*` attributes) return empty list/None automatically via `getattr` with defaults.

src/llmcompressor/modifiers/quantization/quantization/base.py
Etelis: Done. `sync_activation_observers()` is now on `QuantizationMixin`, so both `QuantizationModifier` and `GPTQModifier` can use it.

src/llmcompressor/modifiers/quantization/calibration.py
Etelis: Not directly redundant — `call_observer` needs a tensor value to run the forward pass through the observer, while `recompute_qparams_from_observer` recomputes scale/zp from already-accumulated `past_min_vals`/`past_max_vals` state (needed after DDP sync). Removed `recompute_qparams_from_observer` and moved the logic into `Observer.recompute_qparams()` and `Observer.recompute_global_scale()` methods instead.

examples/big_models_with_sequential_onloading/llama3_8b_w8a8_distributed.py
Etelis: Done. Example now uses `init_dist()` from `compressed_tensors.offload`, along with `load_offloaded_model()`, `get_rank_partition()`, and `dispatch_model()` — matching the existing `llama3_ddp_example.py` pattern.

Etelis:
All review comments have been addressed:

- Removed distributed weight calibration — weight calibration runs identically on each rank
- Focus is exclusively on activation observer synchronization
- `synchronize()`, `recompute_qparams()`, `recompute_global_scale()` added to `Observer` base class
- `sync_activation_observers()` moved to `QuantizationMixin` (shared with `GPTQModifier`)
- All async `all_reduce` ops batched + waited once via `wait_for_comms()` (GPTQ pattern)
- Deleted `distributed.py`, removed `all_reduce_min`/`all_reduce_max` wrappers, removed CPU-to-GPU tensor movement
- Removed `recompute_qparams_from_observer` from `calibration.py`
- Example uses `init_dist()`, `get_rank_partition()`, `load_offloaded_model()`

**Runtime & eval results added to PR description** (Llama-3-8B, W8A8 static activations, 256 samples, A100-80GB):

| Setup | Time | Speedup | Perplexity (wikitext) |
|-------|-----:|--------:|----------------------:|
| 1x GPU | 270.1s | 1.0x | 37.51 |
| 2x GPU (DDP) | 138.0s | 1.96x | 37.64 |
| 4x GPU (DDP) | 72.1s | 3.75x | 36.40 |

src/llmcompressor/modifiers/quantization/quantization/base.py
Etelis: Done. Weight calibration now runs identically on every rank — no partitioning or broadcasting. PR focuses exclusively on activation observer synchronization.

src/llmcompressor/observers/base.py
```diff
@@ -133,6 +134,63 @@ def _get_module_param(self, name: str) -> Optional[torch.nn.Parameter]:
...
+
+        :return: list of async communication handles
+        """
+        comms = []
+        for attr, op in [
+            ("past_min_vals", dist.ReduceOp.MIN),
+            ("past_max_vals", dist.ReduceOp.MAX),
+            ("past_global_min_vals", dist.ReduceOp.MIN),
+            ("past_global_max_vals", dist.ReduceOp.MAX),
+        ]:
```
kylesayrs, line 153: I agree that I think this approach is more elegant than reimplementing for each subclass

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -95,7 +100,11 @@ def on_event(self, state: State, event: Event, **kwargs):
             if not self.started_:
                 self.on_start(state, None)
 
+        if event.type_ == EventType.SEQUENTIAL_EPOCH_END:
+            QuantizationMixin.sync_activation_observers(self, state.model)
```
kylesayrs, line 104: I think we'll need to add these to GPTQ and AWQ, right?

src/llmcompressor/observers/base.py
```diff
@@ -133,6 +134,63 @@ def _get_module_param(self, name: str) -> Optional[torch.nn.Parameter]:
...
+            val = getattr(self, attr, None)
+            if val is not None:
+                comms.append(dist.all_reduce(val, op=op, async_op=True))
```
HDCharles: Think we need the fp8 trick here from GPTQ base.py

src/llmcompressor/observers/base.py
HDCharles: dont we need to do this for every subclass? this strategy only makes sense for an single observer: static_minmax

memoryless_minmax and memoryless_mse it also works i guess, because they don't have any of those values

minmax (moving average) and mse (moving average) - it makes no sense, you probably need to average across ranks, though it wouldn't be hard in theory to do:

$$f(x_0,...,x_{n-1}) = w*\sum_{i=0}^{n-1} x_i(1-w)^{n-1-i}$$ (rank 0 avg)
$$f(x_n,...,x_{2n-1}=w*\sum_{i=n}^{2n-1} x_i(1-w)^{2n-1-i}$$ (rank 1 avg)
$$f(x_0,...,x_{2n-1}) = w*\sum_{i=0}^{2n-1} x_i(1-w)^{2n-1-i}$$ (alldata avg)
$$= (1-w)^{n} w \sum_{i=0}^{n-1} x_i(1-w)^{n-1-i} + w\sum_{i=n}^{2n-1} x_i(1-w)^{2n-1-i} $$
$$= f(x_0,...,x_{n-1})(1-w)^n + f(x_n,...,x_{2n-1})$$ accumulate in terms of rank averages

HDCharles: Yeah let's just average for now

Etelis:
@HDCharles Thanks for the feedback! Addressed both points in the latest push:

1. Added `_all_reduce_fp8_safe()` — upcasts FP8 tensors to float32 before NCCL all-reduce, then casts back. Used by all observer `synchronize()` calls.

2.`MovingAverageObserverBase` now overrides `synchronize()` with a SUM all-reduce + divide by `world_size`. The base `Observer.synchronize()` still uses MIN/MAX for static observers.

Both tests pass on 2x A100 via `torchrun --nproc_per_node=2`.

src/llmcompressor/observers/base.py
```diff
@@ -142,3 +200,37 @@ def _check_has_global_scale(self, global_scale: Optional[torch.nn.Parameter]):
...
+
+    Returns a list of async work handles. For FP8 tensors the reduction is
+    performed synchronously (upcast -> reduce -> downcast) so the returned
+    list is empty.
+
+    :param tensor: tensor to reduce **in-place**
+    :param op: reduction operation
+    :return: list of async communication handles
+    """
+    if tensor.dtype in _FP8_DTYPES:
```
HDCharles: you don't need to reimplement this, can just use as_broadcastable

e.g. https://github.com/vllm-project/llm-compressor/blob/cf3bd6463e8d471ad6c8cc20a6a9b053c178e555/src/llmcompressor/modifiers/gptq/base.py#L361

src/llmcompressor/observers/moving_base.py
```diff
@@ -97,6 +98,52 @@ def get_global_min_max(self, observed: torch.Tensor) -> MinMaxTuple:
...
+            val = getattr(self, attr, None)
+            if val is not None:
+                comms.extend(_all_reduce_fp8_safe(val, op=dist.ReduceOp.SUM))
+
+        # After the SUM completes we divide by world_size to get the average.
+        # Store the divisor so _finalize_synchronize can apply it.
+        self._sync_world_size = world_size
+        return comms
+
+    def finalize_synchronize(self):
```
HDCharles: Is this needed, can't you do the division before comms?

src/llmcompressor/observers/base.py
```diff
@@ -142,3 +200,37 @@ def _check_has_global_scale(self, global_scale: Optional[torch.nn.Parameter]):
...
+        orig_dtype = tensor.dtype
+        fp32 = tensor.float()
```
HDCharles: Incidentally I'm fairly sure this won't actually work because .float isn't a view

Etelis:
Thanks for the latest round of feedback! Addressed all three points:

src/llmcompressor/observers/moving_base.py
```diff
@@ -97,6 +99,33 @@ def get_global_min_max(self, observed: torch.Tensor) -> MinMaxTuple:
...
+            "past_max_vals",
+            "past_global_min_vals",
+            "past_global_max_vals",
+        ):
+            val = getattr(self, attr, None)
+            if val is not None:
+                val.div_(world_size)
+                comms.append(
+                    dist.all_reduce(
+                        as_broadcastable(val), op=dist.ReduceOp.SUM, async_op=True
```
kylesayrs: This is meant to be an average, right?

```suggestion
                        as_broadcastable(val), op=dist.ReduceOp.AVG, async_op=True
```

kylesayrs:
Looks good to me! Please use an average when synchronizing moving average observers across ranks

Issue #2374, opened 2026-02-17
add prefetch to intermediates cache


HDCharles:
https://github.com/vllm-project/llm-compressor/pull/2349 looks sweet, lets do that for intermediates cache

HDCharles:
@GOavi101

GOavi101:
sure

PR #2392, opened 2026-02-22
Feature/intermediates cache prefetch

GOavi101:
Optional prefetch was added to the intermediates cache and wired into AWQ when offloading.
IntermediatesCache
New method iter_prefetch() iterates over batches like iter() but prefetches the next batch in a background thread so onload from the offload device overlaps with use of the current batch, reducing wall‑clock time when offloading to CPU.
**AWQ**
When offload_device is set, _run_samples() uses cache.iter_prefetch() instead of the cache iterator so CPU→device onload overlaps with the forward pass over cached parent args during smoothing.
**Tests**
Two tests were added: one that prefetch yields the same batches as iter(), and one that prefetch on an empty cache yields nothing. No new public API; prefetch is used automatically when AWQ offloads.

Fix: #2374


HDCharles:
Looks good, probably want the same gating as the other pr and also probably want to unify the two to use the same util and make the tests be specific to the shared functionality.

HDCharles (changes requested):
see comment

tests/llmcompressor/pipelines/test_cache.py
```diff
@@ -53,6 +53,53 @@ def test_initialization(sample_dataloader):
...
+        return all(deep_equal(a[k], b[k]) for k in a)
+
+    via_iter = list(sample_cache.iter())
+    via_prefetch = list(sample_cache.iter_prefetch())
+    assert len(via_iter) == len(via_prefetch)
+    for i, (b_iter, b_prefetch) in enumerate(zip(via_iter, via_prefetch)):
+        assert batch_dicts_equal(b_iter, b_prefetch), f"batch {i} differs"
+
+
+def deep_equal(a, b) -> bool:
```
HDCharles: why is this moved? better to leave it where it was

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -138,6 +138,10 @@ class AWQModifier(Modifier, QuantizationMixin):
         requirements but requires more time to move data between cpu and execution
         device. Defaults to None, so cached args are not offloaded. Consider setting
         to torch.device("cpu") if you are encountering OOM errors
+    :param prefetch: when offloading, prefetch the next batch in a background thread
```
HDCharles: since both prefetches were merged together, this is probably best done on a global level, i.e. we can get rid of the gating from here and use teh one in data args and make that one global

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -40,25 +39,18 @@ def _get_batches(
...
-            yield batch_idx, inputs
+    batch_source = (
+        activations.iter_prefetch(input_names)
+        if use_prefetch
+        else activations.iter(input_names)
+    )
+    for batch_idx, inputs in tqdm(
+        enumerate(batch_source), total=num_batches, desc=desc
+    ):
+        yield batch_idx, inputs
```
HDCharles, line 53: nice consolidation

tests/llmcompressor/pipelines/test_cache.py
GOavi101: sure

src/llmcompressor/modifiers/awq/base.py
GOavi101: sure

tests/llmcompressor/pipelines/test_cache.py
HDCharles: This is still moved and adds a bunch of line changes that will alter the blame and history for no reason

tests/llmcompressor/pipelines/test_cache.py
```diff
@@ -41,6 +41,30 @@ def sample_cache(sample_dataloader):
 ]
 
 
+def deep_equal(a, b) -> bool:
```
HDCharles: once you get rid of these line changes, we'll be good to land

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -139,14 +131,16 @@ def __call__(
             else:
                 session.state.loss_masks = None
 
+            use_prefetch = getattr(dataset_args, "sequential_prefetch", False)
```
brian-dellabetta: nit -- I prefer we use a single config var name for this for consistency

GOavi101:
```
import time

import torch

from llmcompressor.pipelines.cache import IntermediatesCache

N_BATCHES = 32
BATCH_SHAPE = (1, 2048, 4096)  # LLaMA-3 8B hidden states
N_WARMUP = 3
N_RUNS = 10

# Tensors must be created on CUDA so _offload_value records onload_device=cuda
# and iter_prefetch() performs real H2D transfers in the background thread.
# Creating on CPU would make onload a no-op (CPU->CPU) and show no difference.
print(f"Building cache: {N_BATCHES} x {BATCH_SHAPE} (CUDA -> offload to pinned CPU)")
cache = IntermediatesCache.empty(N_BATCHES, torch.device("cpu"))
for i in range(N_BATCHES):
    cache.update(i, {"hidden_states": torch.randn(*BATCH_SHAPE, device="cuda")})

# Simulate a subgraph forward pass
model = torch.nn.Linear(4096, 4096, bias=False).cuda().to(torch.float16)


def consume(iterator):
    for batch in iterator:
        x = batch["hidden_states"].to(torch.float16)  # already on CUDA
        with torch.no_grad():
            _ = model(x)
    torch.cuda.synchronize()


def bench(label, fn, warmup=N_WARMUP, runs=N_RUNS):
    for _ in range(warmup):
        consume(fn())
    torch.cuda.synchronize()
    t0 = time.perf_counter()
    for _ in range(runs):
        consume(fn())
    torch.cuda.synchronize()
    elapsed = (time.perf_counter() - t0) / runs * 1000
    print(f"  {label:30s}: {elapsed:.1f} ms / iteration")


print("\n=== Results ===")
bench("iter (baseline)", cache.iter)
bench("iter_prefetch   ", cache.iter_prefetch)
```


Run | baseline | prefetch | speedup
-- | -- | -- | --
1 | 54.0ms | 43.6ms | 1.24x
2 | 54.0ms | 43.6ms | 1.24x
3 | 53.9ms | 43.5ms | 1.24x

HDCharles:
@brian-dellabetta no I haven't, at some point i'm going to go through and look at all our performance changes like sequential prefetch and see if its safe to enable by default and do e2e tests

PR #2394, opened 2026-02-23
Feature/calibrate weights dfs fused modules

GOavi101:
## Summary
- Replaces the three separate calibration loops (global scale, fused scales, zp/scale) with a single **DFS traversal** (`calibrate_weights(model)`) for better cache locality and fewer CPU→GPU onloads when using offloading.
- Introduces a **centralized source of truth** for fused vLLM modules (`fused_modules.py`): traditional attention (q/k/v), fused QKV, **MLA** (q_a + kv_a_proj_with_mqa), and MLP (gate/up, fused gate_up).
- **MLA (Multi-head Latent Attention)** fused global-scale support is added (previously not handled).

## Changes
- **`calibrate_weights(model, named_modules=..., update_zp_scale=..., desc=..., show_progress=...)`** in `calibration.py`: single stack-based DFS; pre-order `update_weight_global_scale` for targets, post-order `update_fused_layer_weight_global_scales` and `update_weight_zp_scale` for targets. API supports DDP via `named_modules` subset (#2220).
- **`fused_modules.py`**: `get_fused_attention_linears()`, `get_fused_mlp_linears()`, plus `is_fused_attention_module` / `is_fused_mlp_module`. Defines traditional, fused QKV, MLA, and gate/up layouts.
- **`helpers.py`**: `update_fused_layer_weight_global_scales()` refactored to use `get_fused_*_linears()`; MLA handled in the same path as traditional attention.
- **QuantizationModifier, AWQModifier, GPTQModifier**: all use `calibrate_weights()` instead of the three loops.


src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +159,91 @@ def update_weight_zp_scale(module: Module):
...
+            if id(module) in target_set:
+                update_weight_global_scale(module)
+            stack.append((module, True))
+            for child in reversed(list(module.children())):
+                stack.append((child, False))
+        else:
+            # Post-order: fused global scales (Attention/MLP), then zp_scale for targets
+            update_fused_layer_weight_global_scales(module)
+            if update_zp_scale and id(module) in target_set:
+                update_weight_zp_scale(module)
```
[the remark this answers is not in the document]
GOavi101: update_weight_global_scale is defined in this file, not imported

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +158,93 @@ def update_weight_zp_scale(module: Module):
...
+
+    try:
+        import tqdm
+    except ImportError:
+        tqdm = None
+
+    if show_progress and desc is not None and tqdm is not None and total_targets > 0:
+        pbar = tqdm.tqdm(total=total_targets, desc=desc)
+    else:
+        pbar = None
```
kylesayrs: Assume that tqdm is available

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +158,93 @@ def update_weight_zp_scale(module: Module):
...
+                "calibrate_weights requires either named_modules or both "
+                "targets and ignore"
+            )
+        from compressed_tensors.utils import match_named_modules
+
+        named_modules = list(match_named_modules(model, targets, ignore))
+    else:
+        named_modules = list(named_modules)
+
+    target_set = {id(m) for _, m in named_modules}
```
kylesayrs: A set of modules is equivalent to a set of ids of modules (more or less). You don't need the `id` call.

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +158,93 @@ def update_weight_zp_scale(module: Module):
...
+    stack: list[Tuple[Module, bool]] = [(model, False)]
+
+    while stack:
+        module, children_done = stack.pop()
+
+        if not children_done:
+            # Pre-order: global scale for target modules (FP4 / TENSOR_GROUP)
+            if id(module) in target_set:
+                update_weight_global_scale(module)
+            stack.append((module, True))
```
kylesayrs: Please test this traversal with models which have shared modules. You may have to memoize modules to avoid traversing twice.

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +158,93 @@ def update_weight_zp_scale(module: Module):
...
+    :param desc: Progress bar description; None to disable progress bar.
+    :param show_progress: If True and desc is not None, show a tqdm progress bar.
+    """
+    if named_modules is None:
+        if targets is None or ignore is None:
+            raise ValueError(
+                "calibrate_weights requires either named_modules or both "
+                "targets and ignore"
+            )
+        from compressed_tensors.utils import match_named_modules
```
kylesayrs: Please put imports at the top

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +158,93 @@ def update_weight_zp_scale(module: Module):
...
+        hooks (e.g. GPTQ).
+    :param desc: Progress bar description; None to disable progress bar.
+    :param show_progress: If True and desc is not None, show a tqdm progress bar.
+    """
+    if named_modules is None:
+        if targets is None or ignore is None:
+            raise ValueError(
+                "calibrate_weights requires either named_modules or both "
+                "targets and ignore"
+            )
```
kylesayrs: Can you avoid this check by using default tuple values?

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +158,93 @@ def update_weight_zp_scale(module: Module):
...
+        pbar = None
+
+    # Stack-based DFS: (module, children_visited)
+    stack: list[Tuple[Module, bool]] = [(model, False)]
+
+    while stack:
+        module, children_done = stack.pop()
+
+        if not children_done:
+            # Pre-order: global scale for target modules (FP4 / TENSOR_GROUP)
```
kylesayrs: Really great job with defining pre-order and post-order functions.

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +158,93 @@ def update_weight_zp_scale(module: Module):
     call_observer(module=module, base_name="weight")
 
 
+def calibrate_weights(
```
kylesayrs, line 211: I think the pre-order and post-order framing is very elegant, but it may get in the way of sharing weight offloading between both `calculate_gparam` and `calculate_qparam`.

src/llmcompressor/modeling/fused_modules.py
```diff
@@ -0,0 +1,121 @@
+"""
```
kylesayrs, line 1: Consider putting this in the `modeling` folder. Otherwise nice job with the fused definition.

kylesayrs (commented):
I think this looks nice, I especially like the fused module utilities. However, there still a few things to consider.

1. Weights are still onloaded twice, once to calculate `update_weight_global_scale` and a second time to calculate `update_weight_zp_scale`. Ideally, we'd be able to onload the weight once and see significant performance benefits for NVFP4.

```python3
weight = module.weight  # onload once
weight_global_scale = calculate_gparam(weight)
weight_scale, weight_zero_point = calculate_qparams(weight)
```

2. It's still unclear how this algorithm could be used to perform parallel weight calibration. Answering this problem is definitely tricky, and we can leave it aside for now, but something to consider.

src/llmcompressor/modeling/fused_modules.py
GOavi101: done

src/llmcompressor/modifiers/quantization/calibration.py
GOavi101: the pre/post structure was getting in the way of a single onload. I’ve added _update_weight_calibration_once(module, update_zp_scale) which onloads module.weight once and calls call_observer(..., value=value, should_calculate_gparam=..., should_calculate_qparams=...) so both use the same tensor. The sequential DFS now uses this in pre-order for target modules and no longer calls update_weight_zp_scale in post-order for them, so we get one onload per module for NVFP4.

HDCharles:
ok, taking a look

src/llmcompressor/modifiers/utils/helpers.py
```diff
@@ -8,103 +8,84 @@
...
-            "weight_global_scale",
-            global_scale,
-        )
-        update_offload_parameter(submodule.up_proj, "weight_global_scale", global_scale)
 
-        del global_scale
+def _apply_fused_global_scale(lin: Linear, g_prime: torch.Tensor) -> None:
+    """Set weight_global_scale to g'; rescale weight_scale so q = x/(s*g) unchanged."""
+    old_g = lin.weight_global_scale.data
+    update_parameter_data(lin, g_prime, "weight_global_scale")
```
HDCharles, line 85: this way is deprecated

```suggestion
update_offload_parameter(lin, "weight_global_scale", g_prime)
```

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +208,121 @@ def update_weight_zp_scale(module: Module):
...
+                    seen_pre.add(module)
+                    _update_weight_calibration_once(module, update_zp_scale)
+                stack.append((module, True))
+                for child in reversed(list(module.children())):
+                    stack.append((child, False))
+            else:
+                update_fused_layer_weight_global_scales(module)
+                if update_zp_scale and module in target_set and module not in seen_post:
+                    seen_post.add(module)
+                    if pbar is not None:
```
HDCharles, line 319: i think this progress bar update should be unindented or else it only updates when zp_scale is hit

src/llmcompressor/modifiers/utils/helpers.py
HDCharles: you have it correct below

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +208,121 @@ def update_weight_zp_scale(module: Module):
...
+    boundaries (PR #2391); weight calibration does not all-reduce weight
+    observer state—each rank calibrates its subset and can broadcast
+    quantized params afterward (e.g. GPTQ-style) if needed. Fused groups
+    (q/k/v, gate/up) must be assigned to the same rank so
+    update_fused_layer_weight_global_scales sees the full group. For
+    balanced wall time, assign by weight size (e.g. greedy_bin_packing with
+    item_weight_fn=lambda m: m.weight.numel(); see GPTQ DDP #2333 which uses
+    hessian shape for the same idea).
+
+    Benchmark: See tests/benchmark_calibrate_weights.py for onload count and
```
HDCharles, line 249: is this a real file?

src/llmcompressor/modifiers/utils/helpers.py
```diff
@@ -8,103 +8,84 @@
...
 
-        update_offload_parameter(
-            submodule.gate_proj,
-            "weight_global_scale",
-            global_scale,
-        )
-        update_offload_parameter(submodule.up_proj, "weight_global_scale", global_scale)
 
-        del global_scale
+def _apply_fused_global_scale(lin: Linear, g_prime: torch.Tensor) -> None:
```
HDCharles, line 82: probably need unit test for this and other new functions that are going to be in the hotpath

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -156,6 +208,121 @@ def update_weight_zp_scale(module: Module):
...
+
+    Two modes:
+    - Sequential (parallel=False): DFS over the model. Pre-order: one onload per
+      target via _update_weight_calibration_once (gparam + qparams). Post-order:
+      update_fused_layer_weight_global_scales (no extra onload for targets).
+    - Parallel (parallel=True): Phase 1 runs gparam + qparams per target
+      (order-independent, parallelizable). Phase 2 applies fused global scales
+      and rescales per-tensor scale s' = s * (g' / g).
+
+    DDP: Works with distributed setups. Pass named_modules as this rank's
```
HDCharles, line 236: I'm not seeing any of this functionality

HDCharles (changes requested):
see comments

PR #2400, opened 2026-02-24
[DDP][GPTQ] Fixes for big models

HDCharles:
big models were failing with DDP for a few reasons, primarily related to overloading shared memory or having too many mmaps.

this was primarily an issue with DDP + cpu offloading but even with disk offloading, the moe context stuff would not use the same offloading as the original module and would revert to cpu offloading and cause problems.

Additionally storing all the original modules could still cause mmap issues so those are only stored if needed now.

Finally when saving i saw situations where one thread would to the saving and another thread would go past it and then timeout so i added a barrier there.

this PR depends on https://github.com/vllm-project/compressed-tensors/pull/650

test plan: https://github.com/vllm-project/llm-compressor/pull/2400/changes/596c6b737fa0e540d4067e8c40e274ace88938d4

### DIFF
--- a/tests/llmcompressor/modeling/test_moe_context.py
+++ b/tests/llmcompressor/modeling/test_moe_context.py
@@ -9,6 +9,22 @@ from llmcompressor.modeling.moe_context import (
 )
 
 
+class MockOffloadCache(OffloadCache):
+    """Mock implementation of OffloadCache for testing."""
+
+    offload_device = "cpu"
+
+    def onload(self, offloaded):
+        return offloaded
+
+    def offload(self, tensor):
+        return tensor
+
+    def update_offload(self, offloaded, data):
+        if offloaded is not None and data is not None:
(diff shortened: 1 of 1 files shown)


HDCharles:
its a draft! 

image

src/llmcompressor/modeling/moe_context.py
```diff
@@ -111,8 +116,18 @@ def moe_calibration_context(
                 config=model.config,
                 calibrate_all_experts=calibrate_all_experts,
             )
+            # Apply the same offloading settings as the original module
+            _apply_offloading_to_replacement(module, replacement)
```
brian-dellabetta, line 120: do we need to do this to the (just merged) norm context as well?

HDCharles: The impact is going to be wayyy smaller

Issue #2327, opened 2026-02-02
Restructure and simplify the AWQModifier similar to SmoothQuant


dsikka:
# Summary

Similar to SmoothQuant, AWQ determines and applies smoothing scales. This process is facilitated by a set of model specific mappings which indicate the activation layers to smooth. Once smoothing is complete, the modifier applies redundant code that is already present within the QuantizationModifier to determine optimal quantization scales and zeros.

The scope of this task is to simplify this modifier by:

1. Removing all the code responsible for generating the quantization scales and zero-points. AWQ, like SmoothQuant, should have the ability to be used in a stackable manner with other quantization modifiers such as the GPTQModifier and QuantizationModifier. 

Examples 1:

```python

recipe = [
    AWQModifier(...),
    QuantizationModifier(...)
]
```
Examples 2:

```python

recipe = [
    AWQModifier(...),
    GPTQModifier(...)
]
```

Example 3:

```python

recipe = [
    AWQModifier(...),
    QuIPModifier(
        rotations=["v", "u"], transform_block_size=128, transform_type="hadamard"
    ),
    QuantizationModifier(targets="Linear", scheme="W4A16", ignore=["lm_head"]),
]
```

2. Simplify the modifier. This can be done by targeting key functionality that is quite complex, such as `_compute_best_scale`

# Implementation Steps

1. Remove quantization scale and zero-point generation code and ensure AWQ can be used in a stackable manner with other quantization modifiers. Validate this through additional test scripts.
2. Make other simplifications within the modiifer to improve the complexity of the modifier
3. Update all examples and tests using the AWQModifier to now use `AWQModifier` stacked with the `QuantizationModifier`. Validate performance and accuracy remains the same
4. As the AWQModifier will still require quantization arguments to generate the means, potentially add a recipe validation step to validate that these arguments match the quantization arguments passed to the subsequent quantization modifier, such as GPTQModifier and QuantizationModifier

colldata79:
Hello @dsikka  can i take a look at this ?

colldata79:
Please before I jump into this i have a few questions. 

1. After refactor, should AWQ still be usable standalone (i.e., produces a quantized model), or will it become smoothing-only and require stacking with QuantizationModifier/GPTQ?

2. When you say ‘similar to SmoothQuant’, do you mean only that AWQ applies smoothing in-place and does no quantization, or do you also want AWQ to adopt SmoothQuant’s lifecycle / mapping & cached-state structure?

[...]

PR #2206, opened 2026-01-08
[AWQ] Option to disable quantization

brian-dellabetta:
SUMMARY:
Prerequisite:
* https://github.com/vllm-project/compressed-tensors/pull/539

This PR 
- [x] adds a `disable_quantization` option to AWQModifier so that it can be used in conjunction with GPTQ or NVFP4 pathways. In this event, quantization config is still applied so AWQ can find best scales according to user's targeted scheme and layers, but is completely removed calibration end event (the weights remain modified). The subsequent modifier will re-add a quantization config.
- [x] To support this, QuantizationMixin's lifecycle is reorganized. `apply_quantization_config` (and anything else updating quantization status/enabled flags) is moved from on_initialize (which is run for every modifier before doing anything) to on_start (i.e. at the beginning of the modifier's run). This allows users to run AWQ w/ quant disabled, followed by another quant modifier on the same targets, in a single `oneshot` (but multiple pipelines). This seems more appropriate, @kylesayrs mentioned this was originally because of issues related to FSDP and QAT, which are no longer used, but wanted to elevate this so everyone is aware.
- [x] Some minor updates to reduce number of warnings users (see #2192)
- [x] Everything runs without issue after rebasing against the accelerate offloader refactor. 🔥 

Resolves #2192 

TEST PLAN:
The following flows were run manually
- [x] AWQ
- [x] AWQ w/ disable_quant + GPTQ
- [x] AWQ w/ disable_quant + Quant NVFP4 has better performance than AWQ, as expected because cached activations are invalidated during AWQ. See results below
- [x] I had to modify AutoRoundModifier on_initialize/on_start code to reflect changes, as it can't call directly into QuantizationMixin. Tests are passing, maintainers notified
- [x] lm_eval / e2e tests -- All tests pass except for some errors related to uploads to HF hub and another OOM issue that was previously known. Full logs at https://github.com/neuralmagic/llm-compressor-testing/actions/runs/21528701447/job/62038989634
- [x] merge after code-freeze / release

Results on `meta-llama/Meta-Llama-3-8B-Instruct` using https://github.com/vllm-project/vllm/blob/main/tests/evals/gsm8k/gsm8k_eval.py
QuantizationModifier NVFP4
```
Results:
Accuracy: 0.657
Invalid responses: 0.003
Total latency: 7.569 s
Questions per second: 174.263
Total output tokens: 123849
Output tokens per second: 16362.633
```

AWQ disable_quantization=False NVFP4 (slightly worse)
```
Results:
[...]

PR #2402, opened 2026-02-24
refactor(awq): restructure AWQModifier to be similar to SmoothQuantCl…

vishnuprasanth-j:
SUMMARY:

This PR restructures and simplifies AWQModifier to be similar to SmoothQuant. This is PR for: #2327

Solution:

This refactoring removes the quantization code from AWQModifier, making it perform smoothing only. AWQModifier now inherits from Modifier and HooksMixin (instead of QuantizationMixin), and must be stacked with a separate quantization modifier (QuantizationModifier, GPTQModifier, or QuIPModifier) to produce a quantized model. The changes include: (1) Removing QuantizationMixin inheritance and related code, (2) Adding recipe validation to ensure AWQ comes before quantization modifiers in the recipe, (3) Adding a warning when AWQ is used standalone without a subsequent quantization modifier, (4) All existing AWQ examples updated to use the new stacked pattern.

All existing AWQ examples updated and tested with new stacked pattern.


brian-dellabetta (commented):
Hi @vishnuprasanth-j , thanks for the contribution. I think this implementation is preferable to my draft PR #2206 to add a disable_quantization option to AWQModifier

We originally wanted AWQ to also quantize due to user feedback, but I think it complicates things more than it's worth, and we can always dynamically append a `QuantizationModifier` during recipe.validate if AWQ is found without an accompanying QuantizationModifier. 

I will discuss with the team

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -221,8 +240,6 @@ def on_initialize(self, state: State, **kwargs) -> bool:
                     "setting offload_device in your recipe."
                 )
             else:
-                # For non-MoE models, convert sentinel to None
-                # (no offloading by default)
```
brian-dellabetta, line 225: it's hard to read through the changes to your diff because you're making a lot of changes that are unrelated to the restructure. please revert so the diff is clean and scoped to your PR changes

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -105,91 +116,100 @@ class AWQModifier(Modifier, QuantizationMixin):
...
+    :param n_grid: number of grid points for best scales search. Defaults to 20.
+    :param config_groups: quantization config groups used only for simulating
+        quantization during the AWQ grid search. These args should match the
+        quantization args passed to the subsequent QuantizationModifier or
+        GPTQModifier. They do NOT produce final quantization scales/zero-points.
     """
 
     # Allow arbitrary types because AWQMapping has fields of type torch.nn.Module
-    model_config: ConfigDict = ConfigDict(arbitrary_types_allowed=True)
+    model_config: ConfigDict = ConfigDict(arbitrary_types_allowed=True, extra="allow")
```
brian-dellabetta, line 149: why is `extra="allow"` needed?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -105,91 +116,100 @@ class AWQModifier(Modifier, QuantizationMixin):
...
+        Attach quantization schemes to modules (needed for grid search simulation)
+        but do NOT finalize quantization - that is left to the subsequent modifier.
 
         :param state: state to run AWQ on
         :return: True on a successful run, False otherwise
         """
-
-        # apply config to model and prepare calibration hooks
-        if QuantizationMixin.has_config(self):
-            QuantizationMixin.initialize_quantization(self, state.model)
```
brian-dellabetta, line 187: removing this logic means apply_quantization_config is never run on the model, so no quantization_scheme is ever set on the targeted modules. I don't think your current implementation will work. Have you run this?

brian-dellabetta (changes requested):
Hi @vishnuprasanth-j , thanks for taking a stab at this. Ultimately we want to move to this, but your diff has a lot of unnecessary changes and I don't think it will work. But we can iterate over this if you'd like. I've left some comments for now

brian-dellabetta:
This refactor was merged in
* #2526

PR #2403, opened 2026-02-24
[Offload] Convert model back to CT offloading for testing

kylesayrs:
## Purpose ##
* Fix testing which requires access to the model after the model has been saved

## Prerequisites ##
* https://github.com/vllm-project/compressed-tensors/pull/601

## Changes ##
* Convert back to CT offloading after converting to accelerate offloading for saving
  * Previously we just "removed dispatch", but this is bad practice as it won't work for disk offloading


dsikka:
One base test failure

kylesayrs:
@dsikka Requires the CT change as prereq

PR #2406, opened 2026-02-25
[Example Testing] Remove and update example test cases

dsikka:
SUMMARY:
- Remove out-dated cases
- Add more up-to-date cases (e.g disk offload, ddp, model free ptq), examples, and models
- Ensure all cases are verified for correct compression format
- Add an optional `qwen` install to enable qwen VL examples which leverage `qwen_vl_utils`
- Will require https://github.com/neuralmagic/llm-compressor-testing/pull/219 for example testing

With these changes, all examples pass: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/22450404023

### DIFF
--- a/tests/examples/test_example_scripts.py
+++ b/tests/examples/test_example_scripts.py
@@ -67,7 +67,7 @@ class TestCase(NamedTuple):
             TestCase(
                 "quantization_w4a16/llama3_ddp_example.py",
                 compressed_format=CompressionFormat.pack_quantized,
-                prefix="W4A16-G128-DDP",
+                prefix="W4A16-G128-DDP2",
             ),
             marks=(requires_gpu(2), pytest.mark.multi_gpu),
         ),


examples/disk_offloading/qwen3_example.py
```diff
@@ -60,6 +60,6 @@
 print("==========================================\n\n")
 
 # Save to disk compressed.
-SAVE_DIR = model_id.rstrip("/").split("/")[-1] + "-NVFP4"
+SAVE_DIR = model_id.rstrip("/").split("/")[-1] + "-NVFP4-Disk-Offload"
```
brian-dellabetta, line 63: do we want this? Disk offloading shouldn't change the checkpoint at all

dsikka: Just following what we did for ddp

PR #2415, opened 2026-02-26
[docs] mkdocs material is EOL, move to a zensical docs build

aireilly:
MkDocs is effectively end-of-life. The underlying MkDocs project has been unmaintained since August 2024. Material for MkDocs is now in maintenance mode — critical bugs and security fixes only for 12 months, no new features. Zensical is the explicit successor from the same creator (squidfunk).

zensical is a drop in replacement in the RTD build, no URL changes etc. Peep the PR RTD build, the search is probably the main difference, zensical uses a new custom client-side search. 

image

This PR:
- Adds [Zensical](https://zensical.org/) as an alternative docs build to replace the existing MkDocs setup
- `zensical.toml` == `mkdocs.yml` with some caveats
- Adds standalone pre-build scripts to replace MkDocs plugins not yet compatible with Zensical (`api-autonav`, `gen-files`)
- Both build systems coexist — `mkdocs.yml` and all MkDocs dependencies remain untouched

Zensical does not support MkDocs plugins. The incompatible plugins are replaced by a build scripts.

```bash
# Zensical build
cd docs && make serve    # dev server
cd docs && make build    # static build

# MkDocs build (unchanged)
mkdocs serve             # dev server
mkdocs build             # static build
```

### DIFF
--- a/zensical.toml
+++ b/zensical.toml
@@ -126,6 +126,7 @@ default_handler = "python"
 
 [project.plugins.mkdocstrings.handlers.python]
 paths = ["src"]
+load_external_modules = false
 
 [project.plugins.mkdocstrings.handlers.python.options]
 docstring_section_style = "list"


docs/README.md
brian-dellabetta, line 1: this looks previously to be a readme on how to build the docs, we no longer need/allow that?

aireilly: Yeah I was pondering whether it was even required. The Makefile is pretty self-explanatory and also zensical picks up README.md as the default file to create the folder index.html in the build.

aireilly: zensical picks up README.md in a folder dir as the root for the output. I moved the docs/README.md to docs/DEVELOPMENT.md, and otherwise left everything else alone. This seems like the best low touch solution.

PR #2422, opened 2026-02-27
Smoothquant bugfixes

HDCharles:
Summary:

smooth quant wasn't actually doing anything since it was only updating the onload, this PR fixes that and adds a test to check the behavior of smoothquant in the future

TEST PLAN:
pytest /home/HDCharles/repos/llm-compressor/tests/llmcompressor/modifiers/transform/smoothquant/test_base.py -k "e2e"

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -146,9 +146,9 @@ setup(
         ),
         ("pillow>=10.4.0,<=12.1.1" if BUILD_TYPE == "release" else "pillow>=10.4.0"),
         (
-            "compressed-tensors==0.13.0"
+            "compressed-tensors==0.14.0"
             if BUILD_TYPE == "release"
-            else "compressed-tensors>=0.13.1a2"
+            else "compressed-tensors>=0.14.1a2"
         ),
     ],
     extras_require={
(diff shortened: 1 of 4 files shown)


tests/llmcompressor/modifiers/transform/smoothquant/test_base.py
```diff
@@ -224,3 +229,54 @@ def test_ignore_behavior():
...
+    model_id = "nm-testing/tinysmokellama-3.2"
+    model = AutoModelForCausalLM.from_pretrained(model_id)
+    tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+    # Record original weights for layers that should be smoothed
+    orig_weights = {
+        name: param.clone()
+        for name, param in model.named_parameters()
+        if "input_layernorm" in name or "q_proj.weight" in name
+    }
```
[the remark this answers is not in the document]
HDCharles: not important

src/llmcompressor/modifiers/transform/smoothquant/base.py
```diff
@@ -355,9 +358,8 @@ def _calculate_smoothing_scales(
         # get the channel-wise dynamic range for each layer to be balanced
         weight_scales = []
         for layer in balance_layers:
-            with align_module_device(layer):
```
brian-dellabetta, line 358: just confirming for my understanding -- we don't need this line anymore because `layer.weight` will be onloaded when accessed, if it needs to be, because it is a parameter in the OffloadCache.

HDCharles: yes

PR #2431, opened 2026-03-03
AWQ smooth layer quantization (v2) [not for land]

HDCharles:
i added different evaluation code along the same lines as what we used previously, didn't want to clutter the other PR so made a new one

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -573,17 +573,15 @@ class AWQModifier(Modifier, QuantizationMixin):
                 smooth_layer_targeted = (
                     self.smooth_layer_quantization
                     and mapping.smooth_name in targeted_names
+                    and hasattr(smooth_layer, "quantization_scheme")
+                    and hasattr(smooth_layer.quantization_scheme, "weights")
                 )
 
                 orig_layer_weights = {
                     balance_layer: balance_layer.weight.clone()
                     for balance_layer in mapping.balance_layers
                 }
-                if (
-                    smooth_layer_targeted
-                    and hasattr(smooth_layer, "quantization_scheme")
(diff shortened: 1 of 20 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -616,11 +670,79 @@ def _run_samples(self, module: Module) -> list[torch.Tensor]:
...
+            should_calculate_gparam=should_calculate_gparam,
+        )
+        quantized = forward_quantize(
+            layer,
+            layer.weight,
+            "weight",
+            w_qscheme,
+        )
+        if is_smooth_layer:
+            layer.weight.data = quantized.to(weight_dtype)
```
ANGDL, line 736: Hi! While reviewing the `also_quantize_smooth_layers` logic in `_rescale_and_fake_quantize_layer`, I noticed a potential math issue. 

Because of this specific check:
```python
if is_smooth_layer:
    layer.weight.data = quantized.to(weight_dtype)
else:
    layer.weight.data = (quantized / scales_view).to(weight_dtype)
```
The smooth layer doesn't divide by its `scales_view` ($1/s$). This seems to break the simulated scaling and cause a **double-scaling ($1/s^2$)** issue during the simulated forward pass (`_run_samples`):

1. **Smooth Layer:** Since the weight is fixed to $Q(W_{sm}/s)$, its physical output activation is scaled down to $\approx s^{-1} \cdot X$.
2. **Balance Layer:** It receives $s^{-1} \cdot X$ from the smooth layer, but it *also* artificially scales down its own weights by $s$ (via the `else` branch).
3. **Combined Effect:** The final output of the block becomes $(s^{-1} \cdot X) \cdot (\frac{Q(W_{bal} \cdot s)}{s})^T = \mathbf{s^{-2}} \cdot \mathbf{X} \cdot Q(W_{bal} \cdot s)^T$.

PR #2449, opened 2026-03-06
[MoE][ddp] Enable distributed MoE calibration replacement

dichn:
SUMMARY:
Extends moe_calibration_context to support PyTorch DDP for parallel MoE module replacement across multiple ranks during calibration.

Key changes:
- Pre-replacement verification: all_gather to ensure consistent module counts across ranks, preventing structure mismatch errors
- Post-replacement barrier: synchronizes all ranks before calibration begins
- Post-restoration barrier: ensures clean context manager exit
- Rank-aware logging: only rank 0 shows progress bars and info logs

TEST PLAN:
Tests include single-rank and DDP integration coverage.

### DIFF
--- a/src/llmcompressor/modeling/moe_context.py
+++ b/src/llmcompressor/modeling/moe_context.py
@@ -15,10 +15,10 @@ import contextlib
 from abc import ABC
 
 import torch
-import torch.distributed as dist
-from compressed_tensors.offload import is_distributed
+from compressed_tensors.offload.dist_utils import is_distributed
 from compressed_tensors.registry import RegistryMixin, standardize_lookup_name
 from loguru import logger
+from torch import distributed as dist
 from tqdm import tqdm
 from transformers import PreTrainedModel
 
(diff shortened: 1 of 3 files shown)


src/llmcompressor/modeling/moe_context.py
```diff
@@ -99,11 +99,33 @@ def moe_calibration_context(
...
+    if is_distributed():
+        rank = dist.get_rank()
+        world_size = dist.get_world_size()
+
+        # Verify all ranks have same number of modules
+        num_modules = torch.tensor([len(modules_to_replace)], dtype=torch.long, device=next(model.parameters()).device)
+        all_counts = [torch.zeros_like(num_modules) for _ in range(world_size)]
+        dist.all_gather(all_counts, num_modules)
+
+        if not all(count.item() == num_modules.item() for count in all_counts):
```
kylesayrs, line 112: Sometimes, the number of modules will be not evenly divisible by the number of ranks, so this check can be harmful.

src/llmcompressor/modeling/moe_context.py
```diff
@@ -99,11 +99,33 @@ def moe_calibration_context(
         if _is_registered(class_name, MoECalibrationModule):
             modules_to_replace.append((name, module, class_name))
 
+    # Step 1.5: Verify all ranks have same number of modules (distributed mode)
```
kylesayrs, line 102: It doesn't look like you actually assign modules to ranks. It seems like right now, all ranks are still doing duplicate work.

kylesayrs (commented):
Seems like you still need to actually implement the distributed workload logic, ie assigning modules to ranks, processing them, then broadcasting the results.

I recommend looking at the following algorithm, as I think this may be the best way to support distributed MoE Calibration Replacement: https://github.com/vllm-project/compressed-tensors/pull/624

src/llmcompressor/modeling/moe_context.py
HDCharles: This is just checking that each rank has the same number of modules total, not assigned I think

src/llmcompressor/modeling/moe_context.py
```diff
@@ -99,11 +99,33 @@ def moe_calibration_context(
...
+                f"Rank {rank} found {num_modules.item()} MoE modules, but other "
+                f"ranks found different counts: {[c.item() for c in all_counts]}. "
+                "All ranks must have identical model structures."
+            )
+
     # Step 2: Replace modules with progress bar
     if modules_to_replace:
-        logger.info(f"Found {len(modules_to_replace)} MoE modules to replace")
+        # Only rank 0 shows progress bar and logs
+        show_progress = not is_distributed() or dist.get_rank() == 0
```
HDCharles, line 122: Can use is_rank0, also probably simpler to just inline  this into the if statement

src/llmcompressor/modeling/moe_context.py
HDCharles, line 102: I also don't think this check is necessary

kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/2647

PR #2453, opened 2026-03-08
Add GLQ (E8 lattice codebook) quantization modifier

cnygaard:
## Summary

- Adds a `GLQModifier` that quantizes linear layer weights using [GLQ](https://github.com/cnygaard/glq) — E8 lattice codebook nearest-neighbor with Randomized Hadamard Transform (RHT) incoherence processing and LDLQ error feedback
- Supports 2, 3, and 4 bits per weight via single-stage and two-stage residual vector quantization
- Follows the same Hessian-based lifecycle as `GPTQModifier` (calibration → compress → finalize)
- Depends on the [`glq`](https://github.com/cnygaard/glq) pip package (added as optional dependency)

### Benchmark (SmolLM3-3B-Base, WikiText-2, NVIDIA A10G)

| Method | Eff. BPW | Perplexity | vs bf16 |
|--------|----------|------------|---------|
| bf16 | 16.00 | 7.90 | 1.00x |
| GLQ 4-bit | 4.00 | 8.11 | 1.03x |
| AWQ 4-bit | 5.60 | 8.15 | 1.03x |
| GLQ 3-bit | 3.00 | 8.91 | 1.13x |
| GLQ 2-bit | 2.00 | 11.35 | 1.44x |

### Usage

```python
from llmcompressor.modifiers.glq import GLQModifier

recipe = [GLQModifier(bits=2, ignore=["lm_head"])]
oneshot(model=model, dataset=ds, recipe=recipe, ...)
```

## Test plan

- [ ] `pytest tests/llmcompressor/modifiers/glq/ -v` — unit tests for registration and parameter handling
- [ ] Run example script on a GPU machine: `python examples/glq/llama3_example.py`
- [ ] Verify `ruff check` and `ruff format` pass on new files

### DIFF
--- a/examples/glq/llama3_example.py
+++ b/examples/glq/llama3_example.py
@@ -11,26 +11,17 @@ model = AutoModelForCausalLM.from_pretrained(
 )
 tokenizer = AutoTokenizer.from_pretrained(model_id)
 
-# Select calibration dataset
+# Select calibration dataset.
+# Use long, high-quality text for better Hessian estimates.
+# Filter short texts (<100 chars) to avoid padding-heavy samples.
 DATASET_ID = "HuggingFaceFW/fineweb-edu-score-2"
 DATASET_SPLIT = "train"
 NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQUENCE_LENGTH = 2048
 
 ds = load_dataset(DATASET_ID, split=DATASET_SPLIT)
+ds = ds.filter(lambda x: len(x["text"]) > 100)
 ds = ds.shuffle(seed=42)
-
-
-def preprocess(example):
-    return {
-        "text": tokenizer.apply_chat_template(
-            [{"role": "user", "content": example["text"]}],
-            tokenize=False,
-        )
-    }
-
-
-ds = ds.map(preprocess)
 
 # Configure GLQ quantization
 # bits=2: 16-bit index per 8 weights (2 bpw)


src/llmcompressor/modifiers/glq/glq_quantize.py
```diff
@@ -0,0 +1,90 @@
...
+__all__ = ["glq_quantize_weight"]
+
+
+def glq_quantize_weight(
+    W: torch.Tensor,
+    H: torch.Tensor,
+    bits: int,
+    codebook: E8ShellCodebook,
+    codebook_small: E8ShellCodebook | None = None,
+    dampening_frac: float = 0.01,
```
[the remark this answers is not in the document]
cnygaard: Fixed in commit e5c0081. Added diagonal dampening before RHT transform.

HDCharles:
hey, my understanding is that AQLM which was the closest format to this, was [removed from vllm](https://github.com/vllm-project/vllm/pull/22943/changes). I don't think this is the direction we're looking to move unless codebooks get a lot more community support? Are these techniques being used in production somewhere you can point to or something like that? 

If not i think we could maybe add this in an experimental folder rather than in top level modifiers

src/llmcompressor/modifiers/glq/base.py
```diff
@@ -0,0 +1,235 @@
...
+from llmcompressor.modifiers.gptq.gptq_quantize import (
+    accumulate_hessian,
+    make_empty_hessian,
+)
+from llmcompressor.utils.metric_logging import CompressionLogger
+
+__all__ = ["GLQModifier"]
+
+
+class GLQModifier(Modifier):
```
brian-dellabetta, line 26: Any reason this doesn't inherit QuantizationMixin? Looks like it is similar to GPTQ.

examples/glq/llama3_example.py
```diff
@@ -0,0 +1,46 @@
...
+
+ds = load_dataset(DATASET_ID, split=DATASET_SPLIT)
+ds = ds.filter(lambda x: len(x["text"]) > 100)
+ds = ds.shuffle(seed=42)
+
+# Configure GLQ quantization
+# bits=2: 16-bit index per 8 weights (2 bpw)
+# bits=3: 16-bit + 8-bit residual index per 8 weights (3 bpw)
+# bits=4: 16-bit + 16-bit residual index per 8 weights (4 bpw)
+recipe = [GLQModifier(bits=2, ignore=["lm_head"])]
```
brian-dellabetta, line 30: Your example recipe doesn't include a QuantizationModifier, and GLQModifier doesn't inherit QuantizationMixin, so i don't think your model would save in a compressed format?

src/llmcompressor/modifiers/glq/__init__.py
brian-dellabetta, line 1: a README would be great, even if just a thin wrapper around glq's docs and some links, and some benchmark results so users can know when this modifier should be used over other techniques

cnygaard:
Thanks for the feedback! You're right that this is work at a very early-stage — an experimental folder   
  makes sense.

  A couple of clarifications:
  - GLQ doesn't depend on codebook kernels at inference time, only during quantization. Models dequantize
  back to bf16 and serve as standard weights in any framework.
  - Currently this means no savings in GPU memory or disk space — the dequantized bf16 weights are stored
  as-is. Compressed serialization (storing codebook indices) and inference kernels are needed to deliver
  actual compression, and that's future work.
  - There's a standalone pip package (pip install glq) for experimentation outside of llm-compressor.
  - Early benchmarks show strong reconstruction quality below 4bpw, which is where I think this approach
  has the most potential.

  I'll close the PR for now and come back when I have compressed serialization working. Thanks for taking
  the time to review and your feedback!

HDCharles:
ok, sounds good @cnygaard we're open to collaboration on experimental techniques and look forward to seeing where this goes!

PR #2455, opened 2026-03-09
feat: defer activation qparam calculation to sequential epoch end

### LINKED ISSUE #2446: [Research] Investigate the effect of calculating activation qparams on sequential epoch end, not for every batch
## Background ##
In the past, the observer abstraction was mostly useful for computing quantization parameters during QAT, where each forward pass of training generated a new set of qparams which affected gradients for future passes. However, this technique might be overkill for techniques like PTQ. From a research philosophy perspective, QAT is performed with the goal of modifying weights to be "aware" of quantization, while PTQ aims to replicate the original unquantized weights as closely as possible.

As of now, each calibration batch triggers a new set of activation qparams to be calculated. For the QuantizationModifier (RTN), these new qparams are used to perform quantized forward passes which affect subsequent modules. However, this is not true for GPTQ, AWQ, and AutoRound (see [DISABLE_QAC_MODIFIERS](https://github.com/vllm-project/llm-compressor/blob/7461d02b9bf9edc35f3be9effdaa97d6639baf1f/src/llmcompressor/utils/helpers.py#L177)).

Either way, the calculation of qparams at every batch step is very expensive. It would be more performant (and I hypothesize, more accurate) to keep running min/max values, then only calculate qparams once calibration has finished. However, there is a risk that this algorithmic change affects accuracy.

## Requested Changes ##
1. Modify `QuantizationMixin` to NOT calculate qparams on each batch in the hook, and instead calculate qparams once on (sequential) calibration epoch end. Remember to [disable quantization in all cases](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/pipeline.py#L125), only during calibration (keep enabled during propagation)
2. Modify observer subclasses to add a `calibrate_module` method which uses its running statistics (min/max) to calculate qparams which are used to update module qparams. After the qparams are updated, the running statistics should be deleted off of the observer to reduce memory usage
3. Run regression tests to see the effect of these changes compared to the original implementation. You can run lm_eval regressions with the following commands:
```bash
CADENCE=weekly TEST_DATA_FILE=tests/lmeval/configs/fp8_static_per_tensor.yaml python3 -m pytest tests/lmeval/test_lmeval.py -sx
CADENCE=weekly TEST_DATA_FILE=tests/lmeval/configs/w4a16_awq_sym.yaml python3 -m pytest tests/lmeval/test_lmeval.py -sx
CADENCE=weekly TEST_DATA_FILE=tests/lmeval/configs/w4a4_nvfp4.yaml python3 -m pytest tests/lmeval/test_lmeval.py -sx
```

dzhengAP:
Fixes #2446
Ready for review. A smoke test on local Mac CPU with all passed attached. Happy to run the lm_eval regression tests (fp8_static_per_tensor, w4a16_awq_sym, w4a4_nvfp4) if access to the test infrastructure can be arranged, or point me to the right hardware setup.

## Summary
Switches `QuantizationModifier` from per-batch activation qparam
calculation to a deferred model where qparams are computed once at
`SEQUENTIAL_EPOCH_END` from accumulated running statistics.

## Changes

**`observers/base.py`**
- `Observer.get_accumulated_min_max()`: returns stored `past_min/max`
  without observing a new tensor. Memoryless observers return `None`.
- `Observer.clear_accumulated_stats()`: deletes `past_*` attrs to free
  memory after qparams have been written.
- `calibrate_module_from_observer()`: module-level helper that flushes
  one observer's accumulated stats into the parent module's scale/zero_point.

**`modifiers/quantization/calibration.py`**
- `calibrate_activations()`: new `stats_only` kwarg; when `True`,
  skips `calculate_qparams`/`gparam` — only accumulates running min/max.
- All activation hooks (`input`, `output`, `q`, `k`, `v`) now pass
  `stats_only=True`.
- `flush_activation_qparams()`: iterates over all activation observer
  base_names for a module and calls `calibrate_module_from_observer`.

**`modifiers/quantization/quantization/base.py`**
- `on_start()`: disables quantization after weight calibration so
  calibration forward passes run in fp32.
- `on_event()`: handles `SEQUENTIAL_EPOCH_END` to call
  `flush_activation_qparams` on all targeted modules.

## Local Validation

Ran a smoke test on `facebook/opt-125m` with explicit static int8 activations (`dynamic=False`) using the sequential pipeline on macOS CPU. Full pipeline ran in ~1 min on CPU (13 subgraphs × 32 batches).

## Notes

- `calibrate_output_hook` previously called `forward_quantize` after
  updating stats. This call is intentionally removed — quantization is
  disabled during calibration batches so there is nothing to quantize.
- Regression tests from the issue still need to be run:
  - `fp8_static_per_tensor`
  - `w4a16_awq_sym`
  - `w4a4_nvfp4`

### DIFF
--- a/src/llmcompressor/modifiers/quantization/calibration.py
+++ b/src/llmcompressor/modifiers/quantization/calibration.py
@@ -191,11 +191,18 @@ def calibrate_activations(
         if quantization_args.strategy == QuantizationStrategy.TENSOR_GROUP:
             calculate_gparam = True
 
-    # In deferred (stats_only) mode, only accumulate running min/max in the
-    # observer — skip writing scale/zero_point until epoch end.
+    # In deferred (stats_only) mode: call the observer to accumulate running
+    # min/max stats but do NOT write scale/zero_point yet.
+    # Qparams are written once at epoch end via flush_activation_qparams.
     if stats_only:
-        calculate_qparams = False
-        calculate_gparam = False
+        # Deferred mode: accumulate global min/max into the observer's
+        # _deferred_min / _deferred_max. Works for ALL observer types,
+        # including MemorylessMinMaxObserver which has no past_min_vals.
+        # Qparams are written once at epoch end via flush_activation_qparams.
+        observer = getattr(module, f"{base_name}_observer", None)
+        if observer is not None:
+            observer.update_deferred_stats(value)
+        return
 
     call_observer(
         module=module,
--- a/src/llmcompressor/modifiers/quantization/calibration.py
+++ b/src/llmcompressor/modifiers/quantization/calibration.py
@@ -17,7 +17,7 @@ from loguru import logger
 from torch.nn import Module
 
 from llmcompressor.observers import Observer
-from llmcompressor.observers.base import calibrate_module_from_observer
+from llmcompressor.observers.base import update_module_qparams_from_observer
 
 __all__ = [
     "initialize_observer",
@@ -31,8 +31,11 @@ __all__ = [
     "calibrate_query_hook",
     "calibrate_key_hook",
     "calibrate_value_hook",
-    "flush_activation_qparams",
+    "write_activation_qparams",
 ]
+
+# Activation observer base names used across calibration and quantization code
+ACTIVATION_BASE_NAMES = ("input", "output", "q", "k", "v")
 
 
 def initialize_observer(
@@ -171,7 +174,7 @@ def calibrate_activations(
     :param stats_only: if True, only update running statistics in the observer
         (accumulate min/max) without computing or writing scale/zero_point.
         Used during deferred qparam calibration — qparams are computed once
-        at epoch end via flush_activation_qparams instead of per batch.
+        at epoch end via write_activation_qparams instead of per batch.
     """
     # If empty tensor, can't update zp/scale
     # Case for MoEs
@@ -193,7 +196,7 @@ def calibrate_activations(
 
     # In deferred (stats_only) mode: call the observer to accumulate running
     # min/max stats but do NOT write scale/zero_point yet.
-    # Qparams are written once at epoch end via flush_activation_qparams.
+    # Qparams are written once at epoch end via write_activation_qparams.
     if stats_only:
         observer = getattr(module, f"{base_name}_observer", None)
         if observer is not None:
@@ -213,7 +216,7 @@ def calibrate_input_hook(module: Module, args: Any):
     """
     Hook to accumulate input activation statistics (min/max) in the observer.
     Scale and zero_point are not written here; they are computed once per subgraph
-    at epoch end via flush_activation_qparams.
+    at epoch end via write_activation_qparams.
     """
     args = args[0] if isinstance(args, tuple) else args
     calibrate_activations(module, value=args, base_name="input", stats_only=True)
@@ -223,7 +226,7 @@ def calibrate_output_hook(module: Module, _args: Any, output: torch.Tensor):
     """
     Hook to accumulate output activation statistics (min/max) in the observer.
     Scale and zero_point are not written here; they are computed once per subgraph
-    at epoch end via flush_activation_qparams.
+    at epoch end via write_activation_qparams.
     Note: forward_quantize is intentionally absent — hooks only collect statistics.
     """
     calibrate_activations(
@@ -287,7 +290,7 @@ def reset_quantization_status(model: Module):
             delattr(module, "quantization_status")
 
 
-def flush_activation_qparams(module: Module):
+def write_activation_qparams(module: Module):
     """
     Compute and write final activation qparams from each observer's accumulated
     running statistics, then free those statistics to reduce memory.
@@ -301,7 +304,7 @@ def flush_activation_qparams(module: Module):
 
     apply to targeted modules with:
         for _, module in match_named_modules(...):
-            flush_activation_qparams(module)
+            write_activation_qparams(module)
 
     :param module: module to flush activation qparams for
     """
@@ -309,5 +312,5 @@ def flush_activation_qparams(module: Module):
     if scheme is None:
         return
 
-    for base_name in ("input", "output", "q", "k", "v"):
-        calibrate_module_from_observer(module, base_name)
+    for base_name in ACTIVATION_BASE_NAMES:
+        update_module_qparams_from_observer(module, base_name)
--- a/src/llmcompressor/modifiers/quantization/calibration.py
+++ b/src/llmcompressor/modifiers/quantization/calibration.py
@@ -17,6 +17,7 @@ from loguru import logger
 from torch.nn import Module
 
 from llmcompressor.observers import Observer
+from llmcompressor.observers.base import calibrate_module_from_observer
 
 __all__ = [
     "initialize_observer",
@@ -30,6 +31,7 @@ __all__ = [
     "calibrate_query_hook",
     "calibrate_key_hook",
     "calibrate_value_hook",
+    "flush_activation_qparams",
 ]
 
 
@@ -156,7 +158,9 @@ def update_weight_zp_scale(module: Module):
     call_observer(module=module, base_name="weight")
 
 
-def calibrate_activations(module: Module, value: torch.Tensor, base_name: str):
+def calibrate_activations(
+    module: Module, value: torch.Tensor, base_name: str, stats_only: bool = False
+):
     """
     Calibrate input or output activations by calling the a module's attached
     observer.
@@ -164,7 +168,10 @@ def calibrate_activations(module: Module, value: torch.Tensor, base_name: str):
     :param module: torch.nn.Module
     :param base_name: substring used to fetch the observer, scales, and zp
     :param value: torch.Tensor to be passed to the observer
-
+    :param stats_only: if True, only update running statistics in the observer
+        (accumulate min/max) without computing or writing scale/zero_point.
+        Used during deferred qparam calibration — qparams are computed once
+        at epoch end via flush_activation_qparams instead of per batch.
     """
     # If empty tensor, can't update zp/scale
     # Case for MoEs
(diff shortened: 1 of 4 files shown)


dzhengAP:
## Local Validation

Ran a smoke test on `facebook/opt-125m` with explicit static int8 
activations (`dynamic=False`) using the sequential pipeline on macOS CPU.

Result: all 72 Linear modules across 12 decoder layers have `input_scale` 
correctly populated after calibration via the deferred flush at 
`SEQUENTIAL_EPOCH_END`.

✅ 72 modules have input_scale — deferred qparam flush working!

Full pipeline ran in ~1 min on CPU (13 subgraphs × 32 batches).
Regression tests (fp8_static, w4a16_awq, w4a4_nvfp4) still pending GPU access.

Result: all 72 Linear modules across 12 decoder layers have `input_scale` 
correctly populated after calibration via the deferred flush at 
`SEQUENTIAL_EPOCH_END`.

dzhengAP:
## Local Validation (updated)

Model: facebook/opt-125m, 32 calibration samples, macOS CPU

| Metric | Value |
|--------|-------|
| FP32 perplexity | 28.86 |
| Deferred INT8 perplexity | 30.78 |
| Perplexity delta | +1.92 (6.7%) |
| Modules with input_scale | 72/72 |
| Observer stats leaked | None |
| Calibration time | 51.3s |

✅ All checks pass. Full lm_eval regression tests pending GPU access. @kylesayrs could you add the `ready` label to trigger CI?

dzhengAP:
Hi @kylesayrs, any more work we wanna continue or we need to run more experiments and ablation study before merging or close this PR?

HDCharles:
Hey, you have fp32, and quant(deferred) we'd primarily compare with quant(main) which is missing

dzhengAP:
## Updated Local Validation

Added quant(main) baseline and Qwen2-0.5B per @HDCharles's feedback.

### facebook/opt-125m (32 calibration samples, macOS CPU)

| Mode | Perplexity | Delta vs FP32 |
|------|-----------|---------------|
| FP32 baseline | 28.86 | — |
| quant(main) per-batch | 30.85 | +1.99 (6.9%) |
| quant(deferred) this PR | 30.38 | +1.52 (5.3%) |

**Deferred vs main: -0.47 ✅ better**

### Qwen/Qwen2-0.5B (32 calibration samples, macOS CPU)

| Mode | Perplexity | Delta vs FP32 |
|------|-----------|---------------|
| FP32 baseline | 12.04 | — |
| quant(main) per-batch | 38.81 | +222% |
| quant(deferred) this PR | 37.30 | +210% |

**Deferred vs main: -1.51 ✅ better**

Note: both methods show large degradation on Qwen2-0.5B due to activation outliers (max scale=15.0). This is pre-existing behavior unrelated to this PR — per-tensor INT8 activation quantization without SmoothQuant is known to degrade on models with outlier activations. Deferred still outperforms main by 1.51 PPL.

All correctness checks pass (168/168 modules have input_scale, no observer leaks).

@kylesayrs @HDCharles could you add the `ready` label to trigger CI? DCO has been fixed.

HDCharles:
ready doesn't run CI, that has to be done manually

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -65,7 +66,10 @@ def on_initialize(self, state: State, **kwargs) -> bool:
 
     def on_start(self, state: State, event: Event, **kwargs):
         """
-        Begin calibrating activations and weights. Calibrate weights only once on start
+        Begin calibrating activations and weights. Calibrate weights only once on start.
+        Quantization is kept DISABLED during calibration batches so that forward passes
+        run in fp32. Activation qparams are computed once per subgraph at
```
HDCharles: i don't understand this docstring, calibration doesn't do quantization and never did

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -90,11 +94,26 @@ def on_start(self, state: State, event: Event, **kwargs):
         for _, module in tqdm.tqdm(named_modules, desc="Calibrating weights"):
             update_weight_zp_scale(module)
 
+        # Disable quantization during calibration batches so that fp32 activations
+        # flow through the model unmodified while hooks accumulate running stats.
+        # Re-enable once after epoch end when qparams have been flushed.
+        from compressed_tensors.quantization import disable_quantization
+
+        state.model.apply(disable_quantization)
```
HDCharles: this is already done as mentioned in the issue https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/pipeline.py#L125

dzhengAP: Addressed in the latest commit — removed the duplicate disable_quantization from on_start and moved the control entirely to the pipeline layer

dzhengAP:
Hi @HDCharles, I've pushed a follow-up commit addressing the review feedback:

- Removed the `disable_qac` / `DISABLE_QAC_MODIFIERS` conditional logic in `pipeline.py` — quantization is now unconditionally disabled during calibration and re-enabled during propagation
- Removed the erroneous `disable_quantization` call from `on_start` in `base.py` — control now lives entirely in the pipeline layer
- Fixed misleading docstrings and moved imports to top-level (Gemini nits)

## Validation Results

Ran local regression tests comparing `main` (per-batch qparam) vs this PR (deferred qparam) with W8A8 static quantization, 32 calibration samples on CPU.

### facebook/opt-125m

| Mode | Perplexity | Delta vs FP32 |
|------|-----------|---------------|
| FP32 baseline | 28.86 | — |
| main (per-batch) | 30.76 | +1.90 (6.6%) |
| this PR (deferred) | 30.79 | +1.94 (6.7%) |

✅ 72/72 modules have `input_scale` · ✅ No observer stats leaked · ✅ Perplexity within 15% of FP32

### Qwen/Qwen2-0.5B

| Mode | Perplexity | Delta vs FP32 |
|------|-----------|---------------|
| FP32 baseline | 12.04 | — |
| main (per-batch) | 48.01 | +35.97 (298.8%) |
| this PR (deferred) | 42.00 | +29.96 (248.9%) |

✅ 168/168 modules have `input_scale` · ✅ No observer stats leaked · ✅ Deferred better than main by 6.01 PPL

### Notes
- Qwen large perplexity gap vs FP32 is expected on CPU with 32 calibration samples; deferred is notably better than main here
- Memory overhead ~400-800MB vs main due to observer stats held until epoch end — expected tradeoff
- Calibration time essentially unchanged

dzhengAP:
Hey @HDCharles , I tried running the lm_eval regression tests on a 4x GPU setup but the tests don't support multi-GPU inference via device_map="auto" — lm_eval requires a single large GPU (e.g. A100 80GB).

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -65,7 +66,9 @@ def on_initialize(self, state: State, **kwargs) -> bool:
 
     def on_start(self, state: State, event: Event, **kwargs):
         """
-        Begin calibrating activations and weights. Calibrate weights only once on start
+        Begin calibrating activations and weights. Calibrate weights only once on start.
+        Activation qparams are computed once per subgraph at SEQUENTIAL_EPOCH_END via
```
HDCharles: don't think you need this here, makes more sense in on_event

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -273,3 +285,29 @@ def reset_quantization_status(model: Module):
...
+        for _, module in match_named_modules(...):
+            flush_activation_qparams(module)
+
+    :param module: module to flush activation qparams for
+    """
+    scheme = getattr(module, "quantization_scheme", None)
+    if scheme is None:
+        return
+
+    for base_name in ("input", "output", "q", "k", "v"):
```
HDCharles: should this be hardcoded? seems like something that should be a list somewhere

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -273,3 +285,29 @@ def reset_quantization_status(model: Module):
...
+        calibrate_module_from_observer(module, base_name)
```
HDCharles: this name is bad, sounds like its doing weight quantization?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -159,10 +154,13 @@ def __call__(
...
 
+                    # flush accumulated stats -> write scale/zero_point once per subgraph
                     LifecycleCallbacks.sequential_epoch_end(subgraph)
 
-                    # this pass does not trigger modifier hooks
-                    # and is only used for capturing outputs of newly compressed modules
+                    # propagation pass: modifier hooks are disabled but quantization is
+                    # re-enabled so that compressed module outputs are quantized.
+                    # This ensures downstream subgraphs receive realistic inputs.
+                    model.apply(enable_quantization)
```
HDCharles: shouldn't this be the same as line 120?

src/llmcompressor/observers/base.py
HDCharles, line 138: shouldn't this be altered as well to behave how we want, in general we should probably continue using this interface but update what it does, i.e. not returning qparams

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -273,3 +285,29 @@ def reset_quantization_status(model: Module):
     for module in model.modules():
         if hasattr(module, "quantization_status"):
             delattr(module, "quantization_status")
+
+
+def flush_activation_qparams(module: Module):
```
HDCharles: the function name 'flush' doesn't really imply compute and write, it implies clearing, should probably be write_activation_qparams or something like that

HDCharles (changes requested):
see comments

dzhengAP:
> see comments

Addressed all review comments @HDCharles 

1. Renamed flush_activation_qparams → write_activation_qparams
2. Renamed calibrate_module_from_observer → update_module_qparams_from_observer
3. Extracted ACTIVATION_BASE_NAMES constant in calibration.py
4. Moved SEQUENTIAL_EPOCH_END note from on_start docstring to on_event
5. Used ExitStack for propagation pass quantization management
6. Updated observer.forward() to accumulate stats alongside computing qparams

kylesayrs:
Now merged via other refactors

PR #2464, opened 2026-03-11
[model_free_ptq] split fused moe experts to ensure quantization

liwei109:
SUMMARY:
This PR adds a `split_fused_moe_experts` function for model-free quantization. This ensures that models with fused MoE layers (e.g., Qwen3.5 and Qwen3-VL) containing fused `gate_up_proj` and `down_proj` weights can be effectively quantized.

TEST PLAN:
We add an example of Qwen3.5 for testing purpose.

## Summary by CodeRabbit

* **New Features**
  * Added an example demonstrating 8-bit model-free PTQ for Qwen 3.5-35B.

* **Enhancements**
  * Improved Mixture-of-Experts handling by automatically splitting fused expert weights prior to processing/quantization.

* **Tests**
  * Added tests validating expert-weight splitting and ensuring non-MoE tensors remain unchanged.


src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -194,3 +195,63 @@ def process_file_microscale_scheme(
...
+                    continue
+                    
+                hidden_size = tensor.shape[1] // 2
+                
+                # Split into individual experts
+                for expert_idx in range(num_experts):
+                    expert_tensor = tensor[expert_idx]  # [2*hidden, intermediate]
+                    
+                    # Split gate and up projections
+                    gate_proj = expert_tensor[:hidden_size, :]
```
HDCharles: might be a little more readable to just do the full indexing at the same time instead of doing it piecemeal, these are all views so performance will be equivalent

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -194,3 +195,63 @@ def process_file_microscale_scheme(
...
+    Find fused MoE experts (with gate_up_proj/down_proj).
+    Split them from 3D tensors into individual 2D expert tensors.
+    
+    Args:
+        tensors: Dictionary of loaded tensors from safetensors file
+        
+    Returns:
+        New dictionary with split expert weights
+    """
+    _tensors = {}
```
HDCharles: rename as fused_tensors

HDCharles (changes requested):
see comments and bot comments, did you run the test successfully, it looks like the x.weight name bug should have produced an error, no?

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -194,3 +195,63 @@ def process_file_microscale_scheme(
...
+                    expert_tensor = tensor[expert_idx]  # [2*hidden, intermediate]
+                    
+                    # Split gate and up projections
+                    gate_proj = expert_tensor[:hidden_size, :]
+                    up_proj = expert_tensor[hidden_size:, :]
+                    
+                    # Create new key names
+                    base_key = name.replace("mlp.experts.gate_up_proj", f"mlp.experts.{expert_idx}")
+                    _tensors[base_key + ".gate_proj.weight"] = gate_proj
+                    _tensors[base_key + ".up_proj.weight"] = up_proj
```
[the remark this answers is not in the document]
liwei109: There is not `.weight` suffix from the original name (e.g. Qwen3.5 and Qwen3-VL).

src/llmcompressor/entrypoints/model_free/process.py
liwei109: Sorry, this should be `splited_tensors`, I have renamed it.

liwei109:
> see comments and bot comments, did you run the test successfully, it looks like the x.weight name bug should have produced an error, no?

Yes, I have test this on Qwen3.5 and Qwen3-VL series models, and all of these works well.

src/llmcompressor/entrypoints/model_free/process.py
liwei109: Done, I have changed it.

liwei109:
@HDCharles This PR is ready to review. Thanks

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -194,3 +196,59 @@ def process_file_microscale_scheme(
...
+    Find fused MoE experts (with gate_up_proj/down_proj).
+    Split them from 3D tensors into individual 2D expert tensors.
+    
+    Args:
+        tensors: Dictionary of loaded tensors from safetensors file
+        
+    Returns:
+        splited_tensors: New dictionary with split expert weights
+    """
+    splited_tensors = {}
```
brian-dellabetta: nit - spelling
```suggestion
    split_tensors = {}
```

brian-dellabetta (commented):
Thanks @liwei109 for preparing this. I agree that this logic is needed, to mimic what is done via the [Qwen3.5 MoE context](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/qwen3_5_moe.py) in llm-compressor used during oneshot. Unfortunately I don't see a super clean way to leverage that directly without having additional code like you've added here.

Will make a note to discuss internally, cc @kylesayrs

src/llmcompressor/entrypoints/model_free/process.py
liwei109: LGTM, modified!

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -253,3 +255,72 @@ def _load_tensors_from_inverse_weights_map(
...
+        if tensor.ndim == 3 and (
+            "experts.gate_up_proj" in name or "experts.down_proj" in name
+        ):
+            # Get number of experts
+            num_experts = tensor.shape[0]
+
+            if "gate_up_proj" in name:
+                # gate_up_proj is typically [num_experts, 2*intermediate, hidden]
+                if tensor.shape[1] % 2 != 0:
+                    logger.info(
```
brian-dellabetta: i think this is worthy of a raise ValueError, to avoid silent pass-throughs. We can reduce severity in future if needed

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -82,6 +83,7 @@ def process_file(
     assert not is_microscale_scheme(scheme), "Use `process_file_microscale_scheme`"
 
     tensors = _load_tensors_from_inverse_weights_map(inverse_weights_map, device)
+    tensors = split_fused_moe_experts(tensors)
```
brian-dellabetta, line 90: we should also add this [here](https://github.com/vllm-project/llm-compressor/pull/2464/changes#diff-a88bd486ce930f2604f136972626972642280e8c21567f197c1ca891c0e5f771L149) (we will probably clean this up a bit more in future)

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -253,3 +255,72 @@ def _load_tensors_from_inverse_weights_map(
...
+    split_tensors = {}
+
+    for name, tensor in tensors.items():
+        # Check if this is a MoE expert weight (3D tensor for experts)
+        if tensor.ndim == 3 and (
+            "experts.gate_up_proj" in name or "experts.down_proj" in name
+        ):
+            # Get number of experts
+            num_experts = tensor.shape[0]
+
```
brian-dellabetta, line 263: can we use a match statement here instead? In case this list has to grow in future

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -253,3 +255,72 @@ def _load_tensors_from_inverse_weights_map(
                     )
                 tensors[tensor_name] = f.get_tensor(tensor_name)
     return tensors
+
+
+def split_fused_moe_experts(
```
brian-dellabetta, line 228: would be good to add a few unit tests for this function, as sanity checks

brian-dellabetta (commented):
Hi @liwei109 , thanks for preparing this. We have discussed supporting quantized 3D tensors directly in vllm, but I think this is looking good as a short-term solution. A few asks.

src/llmcompressor/entrypoints/model_free/process.py
liwei109: fixed.

src/llmcompressor/entrypoints/model_free/process.py
liwei109: Done.

src/llmcompressor/entrypoints/model_free/process.py
liwei109: unit test is added.

src/llmcompressor/entrypoints/model_free/process.py
liwei109: sorry, I don't understand this. what's the "list"?

HDCharles (approved):

HDCharles:
Uhoh quality issues

src/llmcompressor/entrypoints/model_free/process.py
HDCharles: Tensor names,

The if else if of gateupproj, downproj is likely to grow

brian-dellabetta: yeah, so we have cascading match cases instead of a long if/elif chain. Actually it may be better to condense all this logic into a single data model, similar to how vllm does it (example [here](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/qwen3_5.py#L277-L288)). Maybe a data model like this
```python
params_to_split = {
    "gate_up_proj": ["gate_proj", "up_proj"],
    "down_proj": ["down_proj"]
}
```
so all we need to do is add to this when we have to cover new cases. Do you think you can refactor you code to handle a data model like this? That way we don't have to duplicate a lot of the logic that occurs in each conditional.

If this is confusing i can add a commit on top of your branch later today to show this

liwei109: Ok, can you help to add a commit on top of my branch? Thanks

brian-dellabetta: Hi @liwei109 , I pushed [a single commit](https://github.com/vllm-project/llm-compressor/pull/2464/changes/98cf4dfb03e8a04ec52c8305ef8e6202064f0e16) to this branch. Now if we see new 3D tensors that need to split, we just need to add a new line to the data model `params_to_split`. Please take a look, if you are good with this we can look to merge

brian-dellabetta: confirmed that provided example works and runs on vllm with my changes

brian-dellabetta (approved):
Thanks for the contribution! I've updated some of the logic to be more robust to new models, and to reduce some code duplication

liwei109:
> Thanks for the contribution! I've updated some of the logic to be more robust to new models, and to reduce some code duplication

LGTM, thanks for your help.

brian-dellabetta (approved):
Re-approving

HDCharles (approved):

PR #2465, opened 2026-03-11
Add DeepSeekV3 support for spinquant

carrot-o0o:
SUMMARY:
**added support for `DeepseekV3` on `spinquant`**
- Currently only implemented for `R1,R2,R4` and `transform_type="hadamard"`. If this PR seems appealing, i am willing to add support for R3 and other transformation types.
- Temporarily patched `DeepseekV3TopkRouter` as ` nn.Linear `during transform, since `compressed_tensors `only supports `nn.Linear`
- Patched `partial_hadamard` that applies transformation to only a "part" of a weight matrix, since `DeepseekV3` has `kv_proj` weights merged, and `R4` only needs to apply to `v_proj`.
- This is my first PR, so i apologize in advance if this PR lacking in any way. Any kind of feedback is welcome.

TEST PLAN:
I tested this code snippet to ensure spinning R1,R2,R4 results in sane outputs. `kanana-2-30b-a3b-instruct-2601`'s model_type is `deepseek_v3` but has fewer layers, making it faster to test.

```
from compressed_tensors.offload import dispatch_model

from llmcompressor.modifiers.transform import SpinQuantModifier

# Select model and load it.
MODEL_ID = "kakaocorp/kanana-2-30b-a3b-instruct-2601"

# NOTE: currently only rotations R1, R2, and R4 are available
# R3 and learned R1/R2 rotations will be added in a future release.
recipe = [
    SpinQuantModifier(
        rotations=["R1", "R2", "R4"],
        transform_block_size=128,
        transform_type="hadamard",
    ),
    # QuantizationModifier(targets="Linear", scheme="W4A16", ignore=["lm_head"]),
]

# Apply algorithms.
oneshot(model=model, recipe=recipe, pipeline="datafree")

print("\n\n")
dispatch_model(model)
input_ids = tokenizer("Hello my name is", return_tensors="pt").input_ids.to(
    model.device
)

# Save to disk compressed.
SAVE_DIR = MODEL_ID.split("/")[1] + "-spinquantR1R2R4"

```

**output results**
```
<|begin_of_text|>Hello my name is Emily and I am a 3rd year student at the University of Hertfordshire studying BSc (Hons) Forensic Science. I am currently on a placement year and I am working at the University of Hertfordshire as a laboratory demonstrator. I have a passion for science and I am keen to share my knowledge with others. I have previous experience in tutoring and I am able to adapt my teaching style to suit the needs of the student. I am patient and understanding and I am confident
```


brian-dellabetta (commented):
Hi @carrot-o0o , thanks for preparing this. We'd definitely like to build out more spinquant/quip mappings, right now we only have the default ones for llama-style architectures. Rather than adding deepseek next, would you be interested in preparing a diff like this for default MoE models, like Qwen and Llama4 MoE models? Wondering if we would need this PartialHadamardTransform in those cases (I don't think we would). Maybe we can tackle that as an easier problem before looking into deepseek?

carrot-o0o:
@brian-dellabetta 
Thank you for the reply. You're right, MoE models like qwen and llama4 don't use MLA (hence their k_proj, v_proj weights are separated) so partial hadamard wouldn't be needed for those cases.
I will create a new PR for those MoE models in the near future.

brian-dellabetta:
Thanks @carrot-o0o ! We can leave this open for now and revisit after the more basic MoE model mappings land

PR #2466, opened 2026-03-11
[Patch Release] Update compressed-tensors version in setup.py

dsikka:


setup.py
```diff
@@ -146,9 +146,9 @@ def localversion_func(version: ScmVersion) -> str:
         ),
         ("pillow>=10.4.0,<=12.1.1" if BUILD_TYPE == "release" else "pillow>=10.4.0"),
         (
-            "compressed-tensors==0.14.0"
+            "compressed-tensors==0.14.0.1"
             if BUILD_TYPE == "release"
-            else "compressed-tensors>=0.14.1a2"
+            else "compressed-tensors>=0.14.0.1a2"
```
dhuangnm: No need to change the nightly version here, just the version for the release.

setup.py
```diff
@@ -146,7 +146,7 @@ def localversion_func(version: ScmVersion) -> str:
         ),
         ("pillow>=10.4.0,<=12.1.1" if BUILD_TYPE == "release" else "pillow>=10.4.0"),
         (
-            "compressed-tensors==0.14.0"
+            "compressed-tensors==0.14.0.1"
             if BUILD_TYPE == "release"
             else "compressed-tensors>=0.14.1a2"
```
kylesayrs, line 151: This should be fine

PR #2469, opened 2026-03-13
[WIP] [DDP] Refactor quantization lifecycle for performance

kylesayrs:
* Separate observer forward from calculate qparams
* Perform weight and activations updates at the end of sequential epoch only
  * Benefit: Skip activation quantization during calibration
  * Benefit: Hide weight onloading for calibration behind sequential onloading
  * Benefit: TODO: fake quantize weights on epoch end, change subgraph module statuses to frozen to speed up propagation
* Perform weight and activation updates in module-parallel
* Update global scales after updating scales (use invariance trick)


kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/2585

Issue #2456, opened 2026-03-09
[RFC] Add iMatrix weighted MSE observer for weight quantization


Yatimai:
# RFC: Add iMatrix weighted MSE observer for weight quantization

## Background

The importance matrix (iMatrix) technique, introduced in [llama.cpp PR #4861](https://github.com/ggml-org/llama.cpp/pull/4861), uses calibration-time activation statistics to improve low-bit weight quantization. The core idea is to collect `⟨a_i²⟩` ,the mean squared input activation for each logical input channel, and use these as importance weights during quantization parameter selection. Channels that carry more signal get more careful rounding.

This technique is widely used in the GGUF ecosystem but is not currently available in llm-compressor. This RFC adapts the core iMatrix idea, activation-informed weighting of weight reconstruction error, to the existing observer interface.

## Motivation

1. **MSE observer regression ([#2094](https://github.com/vllm-project/llm-compressor/issues/2094)).** `memoryless_mse` currently underperforms `memoryless_minmax` on downstream tasks. Importance weighting closes 92% of the gap on GSM8K Platinum at W4.

2. **Low-bit quantization quality.** At W3, both `memoryless_minmax` and `memoryless_mse` effectively destroy reasoning capability (GSM8K strict-match 0.08 and 0.15 respectively). Importance weighting recovers it to 0.38, still degraded but functional. This is the difference between an unusable model and a usable one.

## Proposal

A new observer `imatrix_mse` in `src/llmcompressor/observers/imatrix.py`. It extends the existing MSE grid search with one change: quantization error is weighted by per-channel importance.
```
err = sum(importance * |Q(w) - w|^p)    # imatrix_mse
err = sum(|Q(w) - w|^p)                 # memoryless_mse
```

When importance is unavailable, the observer falls back to uniform MSE with a warning.

### Scope

**Supported:** `torch.nn.Linear`, weight-only quantization, `CHANNEL` / `GROUP` / `TENSOR_GROUP` strategies.

**Out of scope:** activation quantization, automatic collection inside `oneshot()`, `TENSOR` strategy, non-`Linear` modules.

### Importance data contract

The observer reads `module._imatrix_importance`: a 1D `float` tensor of shape `(in_features,)` representing `E[x²]` per input channel. Importance is collected externally via `register_forward_pre_hook` before calling `oneshot()`. A working collection example will be provided with the PR.

[...]

PR #2473, opened 2026-03-16
feat: add iMatrix weighted MSE observer and IMatrixGatherer

Yatimai:
## SUMMARY

Adds `imatrix_mse`, a new weight observer that uses per-channel activation importance (E[x²]) to weight the quantization error during range selection. Channels that carry more signal get more careful range optimization.

Two components:

- **IMatrixGatherer** (`modifiers/transform/imatrix/base.py`): lightweight lifecycle trigger that orchestrates a calibration pass so the observer can collect E[x²]. Does not quantize, delegates to the subsequent `QuantizationModifier` / `GPTQModifier`.

- **imatrix_mse observer** (`observers/imatrix.py`): extends the MSE grid search with importance weighting: `err = sum(importance * |Q(w) - w|^p)`. Owns the E[x²] collection lifecycle via `Observer.attach(module)` / `Observer.detach(module)`. Supports CHANNEL, GROUP, TENSOR_GROUP, and BLOCK strategies via `flatten_for_calibration`. Falls back to uniform MSE when importance is unavailable (`strict=False` default).

### Usage
```yaml
recipe:
  - IMatrixGatherer:
      ignore: ["lm_head"]
  - QuantizationModifier:
      config_groups:
        group_0:
          targets: ["Linear"]
          weights:
            num_bits: 4
            observer: imatrix_mse
```

Composes with GPTQ:
```yaml
recipe:
  - IMatrixGatherer:
      ignore: ["lm_head"]
  - GPTQModifier:
      config_groups:
        group_0:
          targets: ["Linear"]
          weights:
            observer: imatrix_mse
```

### Results

W4A16, WikiText-2 token-level PPL, 141 chunks × 2048, open_platypus calibration 512 samples.

Llama-3.1-8B, group_size=128:

| Config | PPL |
|---|---|
| FP16 baseline | 6.24 |
| RTN `memoryless_minmax` | 6.96 |
| GPTQ | 6.92 |
| AWQ | 6.89 |
| RTN `imatrix_mse` | 6.85 |
| GPTQ + `imatrix_mse` | 6.83 |

Eval method: GPTQ paper (concatenate WikiText-2 test, non-overlapping segments of 2048 tokens, exp(mean(NLLs))).

RFC #2456

## TEST PLAN

45 tests added (all CPU, no GPU required):

- `test_imatrix.py` (26 tests): grid search with non-uniform importance, CHANNEL/GROUP/TENSOR_GROUP/BLOCK strategies, actorder g_idx reordering, strict vs non-strict fallback, validation (shape, dtype, finite, non-negative)
- `test_imatrix_gatherer.py` (15 tests): E[x²] collection, ignore list, accumulation correctness, hook cleanup, weight immutability, on_event lifecycle
- `test_e2e_integration.py` (4 tests): full pipeline via `oneshot()` on `nm-testing/tinysmokellama-3.2`, gatherer-only, observer fallback without gatherer, regex targets
```bash
python -m pytest tests/llmcompressor/observers/test_imatrix.py tests/llmcompressor/modifiers/transform/imatrix/ -v
```

### DIFF
--- a/examples/quantization_w4a16/README.md
+++ b/examples/quantization_w4a16/README.md
@@ -138,6 +138,3 @@ We can see the resulting scores look good!
 |     |       |strict-match    |     5|exact_match|↑  |0.720|±  |0.0285|
 ```
 
-### Questions or Feature Request?
-
-Please open up an issue on `vllm-project/llm-compressor`
--- a/src/llmcompressor/modifiers/transform/imatrix/base.py
+++ b/src/llmcompressor/modifiers/transform/imatrix/base.py
@@ -1,0 +1,134 @@
+from typing import Dict, List, Union
+
+from compressed_tensors.quantization import QuantizationArgs
+from compressed_tensors.utils import match_named_modules
+from pydantic import Field
+
+from llmcompressor.core import Event, EventType, State
+from llmcompressor.modifiers import Modifier
+from llmcompressor.observers.base import Observer
+
+__all__ = ["IMatrixGatherer"]
+
+
+class IMatrixGatherer(Modifier):
+    """
+    Lifecycle trigger for iMatrix importance collection.
+
+    Triggers a calibration pass so that ``IMatrixMSEObserver`` can collect
+    E[x²] via its ``attach()`` hook.  Does **not** quantize weights — the
+    actual quantization is done by the subsequent
+    ``QuantizationModifier`` / ``GPTQModifier``.
+
+    The observer's ``detach()`` method computes ``_imatrix_importance``
+    from the accumulated statistics and leaves it on the module for the
+    next quantization pass to consume.
+
+    Example recipe::
+
+        recipe:
+          - IMatrixGatherer:
+              ignore: ["lm_head"]
+          - QuantizationModifier:
+              config_groups:
+                group_0:
+                  targets: ["Linear"]
+                  weights:
+                    observer: imatrix_mse
+
+    Or composed with GPTQ::
+
+        recipe:
+          - IMatrixGatherer:
+              ignore: ["lm_head"]
+          - GPTQModifier:
+              config_groups:
+                group_0:
+                  targets: ["Linear"]
+                  weights:
+                    observer: imatrix_mse
+
+    .. note::
+        Auto-prepend (inserting the gatherer automatically when
+        ``imatrix_mse`` is detected in a recipe) is planned for a
+        follow-up PR.
+
+    :param targets: module types to instrument (default: ``["Linear"]``)
+    :param ignore: layer name patterns to skip (default: ``["lm_head"]``)
+    :param weight_observer: observer to attach during calibration.
+        Must be ``"imatrix_mse"`` (default).
+    """
+
+    targets: Union[str, List[str]] = Field(default_factory=lambda: ["Linear"])
+    ignore: List[str] = Field(default_factory=lambda: ["lm_head"])
+    weight_observer: str = "imatrix_mse"
+
+    # ------------------------------------------------------------------ #
+    #  Lifecycle
+    # ------------------------------------------------------------------ #
+
+    def on_initialize(self, state: State, **kwargs) -> bool:
+        """
+        Attach iMatrix observers to target modules for E[x²] collection
+        """
+        self._resolved_targets = (
+            self.targets if isinstance(self.targets, list) else [self.targets]
+        )
+        self._observers: Dict[str, Observer] = {}
+
+        # Minimal QuantizationArgs — only used to instantiate the observer,
+        # no quantization config is applied to the model.
+        observer_args = QuantizationArgs(observer=self.weight_observer)
+
+        for name, module in match_named_modules(
+            state.model, self._resolved_targets, self.ignore
+        ):
+            observer = Observer.load_from_registry(
+                self.weight_observer,
+                base_name="weight",
+                args=observer_args,
+                module=module,
+            )
+            module.register_module("weight_observer", observer)
+            observer.attach(module)
+            self._observers[name] = observer
+
+        return True
+
+    def on_start(self, state: State, event: Event, **kwargs):
+        self.started_ = True
+
+    def on_event(self, state: State, event: Event, **kwargs):
+        if event.type_ == EventType.CALIBRATION_EPOCH_START:
+            if not self.started_:
+                self.on_start(state, None)
+
+        if event.type_ == EventType.CALIBRATION_EPOCH_END:
+            if not self.ended_:
+                self.on_end(state, None)
+
+    def on_end(self, state: State, event: Event, **kwargs):
+        self.ended_ = True
+        for name, observer in self._observers.items():
+            module = observer.module() if observer.module is not None else None
+            if module is not None:
+                observer.detach(module)
+                if hasattr(module, "weight_observer"):
+                    delattr(module, "weight_observer")
+        self._observers.clear()
+
+    def on_finalize(self, state: State, **kwargs) -> bool:
+        """
+        Clean up importance tensors so they don't end up in the checkpoint
+        """
+        if not self.ended_:
+            self.on_end(state, None)
+
+        # Clean up importance tensors so they don't end up in checkpoint
+        for _, module in match_named_modules(
+            state.model, self._resolved_targets, self.ignore
+        ):
+            if hasattr(module, "_imatrix_importance"):
+                del module._imatrix_importance
+
+        return True
--- a/src/llmcompressor/observers/base.py
+++ b/src/llmcompressor/observers/base.py
@@ -194,6 +194,24 @@ class Observer(InternalModule, RegistryMixin):
             global_scale=global_scale,
         )
 
+    def attach(self, module: torch.nn.Module) -> None:
+        """
+        Called when the observer is attached to a module.
+        Subclasses can override to register hooks or initialize state.
+
+        :param module: the module this observer is being attached to
+        """
+        pass
+
+    def detach(self, module: torch.nn.Module) -> None:
+        """
+        Called before the observer is deleted from a module.
+        Subclasses can override to remove hooks and clean up module attributes.
+
+        :param module: the module this observer is being removed from
+        """
+        pass
+
     def _check_has_global_scale(self, global_scale: Optional[torch.nn.Parameter]):
         if (
             self.args.strategy == QuantizationStrategy.TENSOR_GROUP
--- a/src/llmcompressor/observers/imatrix.py
+++ b/src/llmcompressor/observers/imatrix.py
@@ -1,0 +1,364 @@
+import math
+from typing import Optional
+
+import torch
+from compressed_tensors.quantization import QuantizationArgs, QuantizationStrategy
+from compressed_tensors.quantization.lifecycle import fake_quantize
+from compressed_tensors.quantization.utils import calculate_qparams, generate_gparam
+from compressed_tensors.utils import patch_attr
+from loguru import logger
+
+from llmcompressor.observers.base import MinMaxTuple, Observer
+from llmcompressor.observers.helpers import flatten_for_calibration
+
+__all__ = ["IMatrixMSEObserver"]
+
+_GROUP_STRATEGIES = (QuantizationStrategy.GROUP, QuantizationStrategy.TENSOR_GROUP)
+
+IMATRIX_PRECISION = torch.float32
+
+
+@Observer.register("imatrix_mse")
+class IMatrixMSEObserver(Observer):
+    """
+    MSE observer weighted by per-input-channel importance.
+
+    Supports CHANNEL, GROUP, and TENSOR_GROUP for weight-only Linear modules.
+    Falls back to uniform MSE for global_scale search.
+    Extra observer_kwargs: maxshrink, patience, grid, norm, strict.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        kw = self.args.observer_kwargs
+        self.maxshrink = kw.get("maxshrink", 0.95)
+        self.patience = kw.get("patience", 5)
+        self.grid = kw.get("grid", 20)
+        self.norm = kw.get("norm", 3.0)
+        self.strict = kw.get("strict", False)
+
+        if self.grid <= 0:
+            raise ValueError(f"grid must be > 0, got {self.grid}")
+        if self.patience < 0:
+            raise ValueError(f"patience must be >= 0, got {self.patience}")
+        if not (0 <= self.maxshrink <= 1):
+            raise ValueError(f"maxshrink must be in [0, 1], got {self.maxshrink}")
+        if (
+            not isinstance(self.norm, (int, float))
+            or not math.isfinite(self.norm)
+            or self.norm <= 0
+        ):
+            raise ValueError(f"norm must be a finite positive number, got {self.norm}")
+
+    # ------------------------------------------------------------------
+    # Hook lifecycle: collect E[x²] per input channel
+    # ------------------------------------------------------------------
+
+    def attach(self, module: torch.nn.Module) -> None:
+        """Attach a forward-pre hook to accumulate E[x²] per input channel.
+
+        If ``_imatrix_importance`` already exists on the module (second pass,
+        e.g. QuantizationModifier after IMatrixGatherer), skip hook attachment.
+        """
+        if hasattr(module, "_imatrix_importance"):
+            return
+
+        if not hasattr(module, "in_features"):
+            return
+
+        in_features = module.in_features
+        module._imatrix_sum = torch.zeros(in_features, dtype=IMATRIX_PRECISION)
+        module._imatrix_count = 0
+
+        def _hook(mod, args):
+            x = args[0] if isinstance(args, tuple) else args
+            if isinstance(x, tuple):
+                x = x[0]
+            if x is None or not isinstance(x, torch.Tensor):
+                return
+
+            x_f = x.detach().to(IMATRIX_PRECISION)
+            n_tokens = math.prod(x_f.shape[:-1])
+            token_sum = x_f.pow(2).sum(dim=list(range(x_f.dim() - 1)))
+
+            if mod._imatrix_sum.device != token_sum.device:
+                mod._imatrix_sum = mod._imatrix_sum.to(token_sum.device)
+
+            mod._imatrix_sum.add_(token_sum)
+            mod._imatrix_count += n_tokens
+
+        module._imatrix_hook = module.register_forward_pre_hook(_hook)
+
+    def detach(self, module: torch.nn.Module) -> None:
+        """Remove hooks and compute / clean up importance data.
+
+        Case 1 – accumulators present (``_imatrix_sum``): compute importance,
+        remove the hook and accumulators, **leave** ``_imatrix_importance`` on
+        the module so the next quantization pass can use it.
+
+        Case 2 – only ``_imatrix_importance`` present (no accumulators): this
+        is the final cleanup pass — delete it so it doesn't end up in the
+        checkpoint.
+        """
+        if hasattr(module, "_imatrix_sum"):
+            if module._imatrix_count > 0:
+                importance = module._imatrix_sum / module._imatrix_count
+                module._imatrix_importance = importance
+            if hasattr(module, "_imatrix_hook"):
+                module._imatrix_hook.remove()
+                del module._imatrix_hook
+            del module._imatrix_sum
+            del module._imatrix_count
+            return
+
+        if hasattr(module, "_imatrix_importance"):
+            del module._imatrix_importance
+
+    # ------------------------------------------------------------------
+
+    def get_min_max(self, observed: torch.Tensor) -> MinMaxTuple:
+        return _grid_search(
+            observed,
+            self.args,
+            self.maxshrink,
+            self.patience,
+            self.grid,
+            self.norm,
+            global_scale=self._get_module_param("global_scale"),
+            importance_weights=self._prepare_importance(observed),
+        )
+
+    def get_global_min_max(self, observed: torch.Tensor) -> MinMaxTuple:
+        # TODO: support importance weights here by deferring the reshape
+        # to the grid search call. Currently the base class reshapes
+        # observed to (1, 1, -1) which loses the channel layout needed
+        # for importance broadcasting.
+        return _grid_search(
+            observed,
+            self.args,
+            self.maxshrink,
+            self.patience,
+            self.grid,
+            self.norm,
+            optimize_global_scale=True,
+        )
+
+    # ------------------------------------------------------------------
+
+    def _prepare_importance(self, observed: torch.Tensor) -> Optional[torch.Tensor]:
+        """Validate → reorder (g_idx) → normalize → broadcast."""
+        imp = self._get_validated_importance(observed)
+        if imp is None:
+            return None
+
+        imp = imp.to(device=observed.device, dtype=torch.float32)
+        imp = imp / (imp.mean() + torch.finfo(torch.float32).tiny)
+
+        # Expand to weight shape and use flatten_for_calibration
+        # to handle all strategies and g_idx
+        module = self.module() if self.module is not None else None
+        if module is None or not hasattr(module, "weight"):
+            return None
+        out_features = module.weight.shape[0]
+        imp_2d = imp.unsqueeze(0).expand(out_features, -1)
+        g_idx = getattr(module, f"{self.base_name}_g_idx", None)
+        return flatten_for_calibration(imp_2d, self.base_name, self.args, g_idx)
+
+    def _get_validated_importance(
+        self, observed: torch.Tensor
+    ) -> Optional[torch.Tensor]:
+        """Return 1D importance tensor or None (with warning/raise)."""
+        if self.base_name != "weight":
+            if self.strict:
+                raise NotImplementedError(
+                    "imatrix_mse: only supported for weight observers"
+                )
+            logger.warning(
+                "imatrix_mse: only supported for weight observers."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+
+        module = self.module() if self.module is not None else None
+
+        if module is not None and not isinstance(module, torch.nn.Linear):
+            if self.strict:
+                raise TypeError(
+                    "imatrix_mse: only supported for Linear,"
+                    f" got {type(module).__name__}"
+                )
+            logger.warning(
+                "imatrix_mse: only supported for Linear,"
+                f" got {type(module).__name__}."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+
+        if self.args.strategy == QuantizationStrategy.TENSOR:
+            if self.strict:
+                raise NotImplementedError("imatrix_mse: TENSOR strategy not supported")
+            logger.warning(
+                "imatrix_mse: TENSOR strategy not supported."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+
+        imp = getattr(module, "_imatrix_importance", None) if module else None
+        if imp is None:
+            if self.strict:
+                raise ValueError("imatrix_mse: no _imatrix_importance on module")
+            logger.warning(
+                "imatrix_mse: no _imatrix_importance on module."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+        if imp.dim() != 1:
+            if self.strict:
+                raise ValueError(
+                    f"imatrix_mse: expected 1D, got shape {tuple(imp.shape)}"
+                )
+            logger.warning(
+                f"imatrix_mse: expected 1D, got shape {tuple(imp.shape)}."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+        if not torch.is_floating_point(imp):
+            imp = imp.float()
+        if not torch.isfinite(imp).all():
+            if self.strict:
+                raise ValueError("imatrix_mse: contains non-finite values")
+            logger.warning(
+                "imatrix_mse: contains non-finite values."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+        if (imp < 0).any():
+            if self.strict:
+                raise ValueError("imatrix_mse: contains negative values")
+            logger.warning(
+                "imatrix_mse: contains negative values."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+        if torch.all(imp == 0):
+            if self.strict:
+                raise ValueError("imatrix_mse: all zeros")
+            logger.warning(
+                "imatrix_mse: all zeros. Falling back to uniform MSE.", log_once=True
+            )
+            return None
+
+        if self.args.strategy == QuantizationStrategy.CHANNEL:
+            expected = observed.shape[-1]
+        elif self.args.strategy in _GROUP_STRATEGIES:
+            expected = observed.shape[2] * observed.shape[3]
+        else:
+            expected = None
+        if expected is None:
+            if self.strict:
+                raise NotImplementedError(
+                    f"imatrix_mse: unsupported strategy {self.args.strategy}"
+                )
+            logger.warning(
+                f"imatrix_mse: unsupported strategy {self.args.strategy}."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+        if imp.numel() != expected:
+            if self.strict:
+                raise ValueError(
+                    "imatrix_mse: size mismatch:"
+                    f" expected {expected}, got {imp.numel()}"
+                )
+            logger.warning(
+                "imatrix_mse: size mismatch:"
+                f" expected {expected}, got {imp.numel()}."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+        return imp
+
+
+# ---------------------------------------------------------------------------
+# TODO: refactor to replace memoryless_mse's grid search, this function
+# subsumes it when importance_weights=None.
+# ---------------------------------------------------------------------------
+
+
+def _grid_search(
+    observed: torch.Tensor,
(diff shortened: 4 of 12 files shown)


src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,249 @@
...
+            count = self._counts[name]
+            if count == 0:
+                continue
+
+            importance = self._sums[name] / count
+
+            if offload_to_cpu:
+                importance = importance.to("cpu")
+                # also free the accumulator
+                self._sums[name] = torch.zeros(1, dtype=torch.float32)
```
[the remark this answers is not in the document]
Yatimai: `del` is cleaner since the accumulator is no longer needed after offload. Applied.

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,249 @@
...
+                sample_mean = x_f.pow(2).mean(dim=list(range(x_f.dim() - 1)))
+
+                device = self._sums[layer_name].device
+                if device != sample_mean.device:
+                    self._sums[layer_name] = self._sums[layer_name].to(
+                        sample_mean.device
+                    )
+
+                self._sums[layer_name].add_(sample_mean)
+                self._counts[layer_name] += 1
```
HDCharles: does this handle batching?

Yatimai: Yes, `oneshot()` defaults to `batch_size=1` (oneshot.py:270). For `batch_size > 1` the batch is averaged into one contribution, can weight by `x.shape[0]` if needed.

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,248 @@
...
+            self.norm,
+            global_scale=self._get_module_param("global_scale"),
+            importance_weights=self._prepare_importance(observed),
+            maxgrow=self.maxgrow,
+        )
+
+    def get_global_min_max(self, observed: torch.Tensor) -> MinMaxTuple:
+        # Base class reshapes to (1, 1, -1) before calling this, which
+        # destroys the 4D layout needed for importance broadcasting.
+        return _grid_search(
```
HDCharles, line 136: should this also use the importance weights?

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,248 @@
...
+        )
+        return None
+
+
+# ---------------------------------------------------------------------------
+# Grid search (self-contained, mirrors _grid_search_mse with importance)
+# ---------------------------------------------------------------------------
+
+
+def _grid_search(
```
HDCharles, line 297: feels like we should just get rid of the other if this one subsumes the other one

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,248 @@
...
+                    imp = imp.index_select(0, perm)
+
+        imp = imp.to(device=observed.device, dtype=torch.float32)
+
+        imp = imp / (imp.mean() + torch.finfo(torch.float32).tiny)
+
+        # Broadcast to observed layout
+        if self.args.strategy == QuantizationStrategy.CHANNEL:
+            return imp.reshape(1, 1, 1, -1)
+        if self.args.strategy in _GROUP_STRATEGIES:
```
HDCharles: does this work with block strategies?

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,249 @@
...
+                x = args[0] if not isinstance(args, torch.Tensor) else args
+                if isinstance(x, tuple):
+                    x = x[0]
+                if not isinstance(x, torch.Tensor):
+                    return
+
+                # Mean per sample (each sample weighted equally,
+                # regardless of sequence length)
+                x_f = x.detach().float()
+                sample_mean = x_f.pow(2).mean(dim=list(range(x_f.dim() - 1)))
```
HDCharles: should each sample we weighted equally? seems like longer sequences carry more data, also all our other averaging techniques operate once per batch*sequence, Is this how the accumulation is done in other implementations for imatrix?

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,248 @@
...
+                if g_idx is not None and g_idx.numel() == imp.numel():
+                    perm = torch.argsort(g_idx).to(imp.device)
+                    imp = imp.index_select(0, perm)
+
+        imp = imp.to(device=observed.device, dtype=torch.float32)
+
+        imp = imp / (imp.mean() + torch.finfo(torch.float32).tiny)
+
+        # Broadcast to observed layout
+        if self.args.strategy == QuantizationStrategy.CHANNEL:
```
HDCharles: would any of the existing utils like flatten for calibration work here?

it may be more memory intensive but seems like you can expand imp to have the same shape as weight and then do flatten for calibration

src/llmcompressor/observers/imatrix.py
HDCharles: looks like this doesn't support blockwise? probably want to support or explicitly raise an error if it encounters it

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,248 @@
...
+            expected = None
+        if expected is None:
+            return self._warn(
+                f"unsupported strategy {self.args.strategy}", NotImplementedError
+            )
+        if imp.numel() != expected:
+            return self._warn(f"size mismatch: expected {expected}, got {imp.numel()}")
+        return imp
+
+    def _warn(self, msg: str, exc_type: type = ValueError) -> Optional[torch.Tensor]:
```
HDCharles: i'm not a fan of this pattern, better to just use the logger to warn and then return whatever value you'd want to return

src/llmcompressor/observers/imatrix.py
Yatimai: The base class reshapes `observed` to `(1, 1, -1)` before this call, which loses the channel layout needed for importance. Only used for `global_scale` (e.g. NVFP4). Can add a log to clarify.

src/llmcompressor/observers/imatrix.py
Yatimai: Currently raises NotImplementedError for unsupported strategies including block. Will add an explicit check and error message.

src/llmcompressor/observers/imatrix.py
HDCharles: I would at least add a Todo to fix that, should be simple to not reshape until the actual gridsearch call.

Probably another reason to align the importance with the weights for these calculations

src/llmcompressor/observers/imatrix.py
Yatimai: True : when `importance_weights=None`, `_grid_search` is identical to `memoryless_mse`'s grid search. Could refactor to share or replace. Want me to do that in this PR or as a follow-up?

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: Indeed, llama.cpp accumulates per-token. Will align: .sum() instead of .mean(), count += n_tokens.

src/llmcompressor/observers/imatrix.py
Yatimai: Will expand imp to weight shape and use flatten_for_calibration. Handles all strategies and g_idx automatically.

src/llmcompressor/observers/imatrix.py
Yatimai: Using flatten_for_calibration as discussed above will handle block strategies automatically : no separate if/elif needed.

src/llmcompressor/observers/imatrix.py
Yatimai: Will replace _warn with direct logger.warning(), return None at each call site.

HDCharles (changes requested):
looks good, once comments are addressed i think there are a few final pieces

1) add an example to demonstrate the technique
2) either add the feature where iMatrixGatherer is prepended to the recipe if the observer is added or leave a TODO for that feature

Yatimai:
All review comments addressed:
- Per-token accumulation (aligned with llama.cpp)
- `flatten_for_calibration` for importance broadcasting (handles all strategies + `g_idx`)
- Replaced `_warn` with direct `logger.warning()` + `return None`
- TODO for `global_scale` importance support
- TODO for grid search refactor
- Added example in `examples/quantization_w4a16/`
- Auto-prepend TODO in gatherer docstring

src/llmcompressor/observers/imatrix.py
HDCharles: its not necessary now but i'd leave a TODO

Yatimai: Added a TODO at the `_grid_search` definition (line 239) to track the refactor.

HDCharles (changes requested):
can you update the APIs and examples across the PR code and descriptions to be a bit cleaner by using the weight_observer argument in the XModifier and making sure the defaults are reasonable for most users. Could also mutate an existing scheme e.g.

```
scheme = preset_name_to_scheme('W4A16', ['Linear'])
scheme.weights.observer = 'imatrix_mse'
scheme.weights.observer_kwargs = ...

...config_groups={"group_0": scheme}
```

probably want the simplest API for the example and can document the alternatives in a README explaining the technique and usage

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,313 @@
...
+    Extra observer_kwargs: maxshrink, patience, grid, norm, strict.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        kw = self.args.observer_kwargs
+        self.maxshrink = kw.get("maxshrink", 0.20)
+        self.patience = kw.get("patience", 5)
+        self.grid = kw.get("grid", 20)
+        self.norm = kw.get("norm", 2.4)
```
HDCharles: why is this 2.4? is that how its done elsewhere? seems very random

tests/llmcompressor/observers/test_imatrix.py
```diff
@@ -0,0 +1,475 @@
...
+
+# ---------------------------------------------------------------------------
+# maxgrow parameter
+# ---------------------------------------------------------------------------
+
+
+class TestMaxgrow:
+    """Tests for the maxgrow grid search extension."""
+
+    def test_maxgrow_zero_matches_default(self):
```
HDCharles: is this really needed?

tests/llmcompressor/observers/test_imatrix.py
```diff
@@ -0,0 +1,475 @@
...
+        ("value", "valid"),
+        [
+            (-0.1, False),
+            (0.6, False),
+            (0.0, True),
+            (0.5, True),
+            (0.25, True),
+        ],
+    )
+    def test_maxgrow_validation(self, value, valid):
```
HDCharles: do we need this?

tests/llmcompressor/observers/test_imatrix.py
```diff
@@ -0,0 +1,475 @@
...
+    """Sanity checks for the happy path."""
+
+    def test_channel_strategy(self):
+        module = _make_linear_with_importance(in_features=8, out_features=4)
+        observer = _make_observer(module, strategy="channel")
+        scale, zp = observer(module.weight)
+        assert scale.shape == (4, 1)
+        assert torch.isfinite(scale).all()
+
+    def test_group_strategy(self):
```
HDCharles, line 178: should do this for all the different schemes if we're going to have it for one

HDCharles (changes requested):
see remaining comments, i.e. documentation, improve API for better UX, tests, norm?

src/llmcompressor/observers/imatrix.py
Yatimai: Same default as `memoryless_mse` (mse.py line 54).

tests/llmcompressor/observers/test_imatrix.py
Yatimai: Will remove the maxgrow tests to keep the suite focused.

tests/llmcompressor/observers/test_imatrix.py
Yatimai: Will remove.

tests/llmcompressor/observers/test_imatrix.py
Yatimai: Will add tests for all supported strategies: CHANNEL, GROUP, TENSOR_GROUP, and BLOCK.

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,247 @@
+"""
+IMatrixGatherer: collects importance-weighted activation statistics
```
brian-dellabetta: nit -- this is redundant with the class docstring. we typically have the docstring on the class rather than the file. can you merge the two?

tests/llmcompressor/modifiers/transform/imatrix/test_e2e_integration.py
```diff
@@ -0,0 +1,180 @@
...
+    def test_pipeline_produces_quantized_model(self):
+        """Gatherer collects importance, observer consumes it, model is quantized."""
+        model = AutoModelForCausalLM.from_pretrained(MODEL_ID)
+
+        recipe = [
+            IMatrixGatherer(ignore=["lm_head"]),
+            QuantizationModifier(
+                config_groups={
+                    "group_0": {
+                        "targets": ["Linear"],
```
brian-dellabetta, line 57: would be good to add a test for specific targets, like `"re:.*self_attn.(q|k|v|)proj$`

tests/llmcompressor/observers/test_imatrix.py
```diff
@@ -0,0 +1,475 @@
...
+        scale, zp = observer(module.weight)
+        assert torch.isfinite(scale).all()
+
+
+# ---------------------------------------------------------------------------
+# Bug 2: importance must be reordered with g_idx (actorder)
+# ---------------------------------------------------------------------------
+
+
+class TestActorderReordering:
```
brian-dellabetta, line 92: can some of these tests be pruned?

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,313 @@
...
+        if expected is None:
+            if self.strict:
+                raise NotImplementedError(
+                    f"imatrix_mse: unsupported strategy {self.args.strategy}"
+                )
+            logger.warning(
+                f"imatrix_mse: unsupported strategy {self.args.strategy}."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
```
brian-dellabetta, line 273: so this will fall back to uniform MSE using the original strategy?

brian-dellabetta (commented):
Thanks for preparing this @Yatimai , I think things are looking on from a code standpoint. I have a few nits, and it would be good to include a README that helps users understand when and how to use this, potentially including some of the nice results you have in PR summary

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,247 @@
...
+                continue
+
+            importance = self._sums[name] / count
+
+            if offload_to_cpu:
+                importance = importance.to("cpu")
+                # also free the accumulator
+                del self._sums[name]
+
+            module._imatrix_importance = importance
```
kylesayrs: It seems like `_imatrix_importance` is never deleted, which can lead to excessive memory usage.

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,247 @@
...
+        logger.info(f"IMatrixGatherer: targeting {len(self._target_names)}" f" modules")
+
+    # ------------------------------------------------------------------ #
+    #  Hook registration
+    # ------------------------------------------------------------------ #
+
+    def _register_accumulation_hooks(self, model: Module):
+        """Attach a forward-pre hook to every target module."""
+
+        def _create_hook_fn(layer_name: str):
```
kylesayrs: If you key your `_sums` dictionary by `Module` rather than `name`, then you don't need a `_create_hook_fn`. You can just define the hook once.

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,247 @@
...
+            def _hook(module: Module, args):
+                x = args[0] if not isinstance(args, torch.Tensor) else args
+                if isinstance(x, tuple):
+                    x = x[0]
+                if not isinstance(x, torch.Tensor):
+                    return
+
+                # Per-token accumulation
+                x_f = x.detach().float()
+                n_tokens = x_f[..., 0].numel()
```
kylesayrs: Better performance
```suggestion
                n_tokens = x_f.shape[:-1]
```

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,247 @@
...
+
+            def _hook(module: Module, args):
+                x = args[0] if not isinstance(args, torch.Tensor) else args
+                if isinstance(x, tuple):
+                    x = x[0]
+                if not isinstance(x, torch.Tensor):
+                    return
+
+                # Per-token accumulation
+                x_f = x.detach().float()
```
kylesayrs: no gradients. Also use a global var for precision
```suggestion
                x_f = x.to(IMATRIX_PRECISION)
```

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,247 @@
...
+                n_tokens = x_f[..., 0].numel()
+                token_sum = x_f.pow(2).sum(dim=list(range(x_f.dim() - 1)))
+
+                device = self._sums[layer_name].device
```
kylesayrs: Using a `ActivationsCache` or `OffloadCache` helps to implement this logic for you

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,313 @@
...
+                f" expected {expected}, got {imp.numel()}."
+                " Falling back to uniform MSE.",
+                log_once=True,
+            )
+            return None
+        return imp
+
+
+# ---------------------------------------------------------------------------
+# TODO: refactor to replace memoryless_mse's grid search, this function
```
kylesayrs, line 292: Definitely, let's standardize on a shared approach for grid-searching MSE

kylesayrs (commented):
From a functionality perspective, I'm a little confused how imatrix computation is actually occuring:

The `IMatrixMSEObserver` requires that `_compute_and_attach` is called in order to use the importance values. However, `_compute_and_attach` is only called at the sequential epoch end and the observer is never called after `_compute_and_attach` is called, so the importance values are never used?

From a design perspective, the implementation seems split between the `IMatrixMSEObserver` and `IMatrixGatherer`. This design is difficult to work with because it allows users to create footgun recipes where they have an imatrix observer but no imatrix modifier, and vice versa.

I recommend implementing only the `IMatrixQuantizationModifier`. This should help simplify the design a lot. This also prevents someone from using the `IMatrixMSEObserver` for activations, which doesn't really make sense.

HDCharles:
> From a functionality perspective, I'm a little confused how imatrix computation is actually occuring:
> 
> The `IMatrixMSEObserver` requires that `_compute_and_attach` is called in order to use the importance values. However, `_compute_and_attach` is only called at the sequential epoch end and the observer is never called after `_compute_and_attach` is called, so the importance values are never used?
> 
> From a design perspective, the implementation seems split between the `IMatrixMSEObserver` and `IMatrixGatherer`. This design is difficult to work with because it allows users to create footgun recipes where they have an imatrix observer but no imatrix modifier, and vice versa.
> 
> I recommend implementing only the `IMatrixQuantizationModifier`. This should help simplify the design a lot. This also prevents someone from using the `IMatrixMSEObserver` for activations, which doesn't really make sense.

@kylesayrs  This should compose with other modifiers so it needs to be an observer, changing that would make this pointless, it would no longer compose with GPTQ so why use this?

The issue with foot gun recipes is not hard to resolve by validating both exist or adding the gatherer if the imatrix observer is detected.

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: Will merge the file docstring into the class docstring.

tests/llmcompressor/modifiers/transform/imatrix/test_e2e_integration.py
Yatimai: Will add an E2E test with regex targets like `"re:.*self_attn.(q|k|v)_proj$"`.

tests/llmcompressor/observers/test_imatrix.py
Yatimai: Will prune to the essential cases : keep test_reorder_importance_with_g_idx and test_no_g_idx_still_works, remove the others.

src/llmcompressor/observers/imatrix.py
Yatimai: Yes, when importance is unavailable, `_prepare_importance` returns None and the grid search runs without importance weighting (identical to `memoryless_mse`). The strategy is unchanged.

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: Will add cleanup of `_imatrix_importance` in `on_finalize` to free memory after quantization completes.

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: Will key `_sums` by module instead of name to simplify the hook.

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: Will use `math.prod(x_f.shape[:-1])` to get the token count as an int.

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: Will add `IMATRIX_PRECISION = torch.float32` as a module-level constant and use `x.to(IMATRIX_PRECISION)`.

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: Will look into `ActivationsCache` / `OffloadCache` for managing device placement.

src/llmcompressor/observers/imatrix.py
Yatimai: Will tackle in a follow-up PR.

Yatimai:
All review comments addressed:

- Per-token accumulation 
- `flatten_for_calibration` for importance broadcasting (all strategies + `g_idx`)
- Replaced `_warn` with direct `logger.warning()` + `return None`
- TODO for `global_scale` importance support
- TODO for grid search refactor
- Example simplified with `preset_name_to_scheme`
- README added in `examples/quantization_w4a16/`
- Tests for all strategies: CHANNEL, GROUP, TENSOR_GROUP, BLOCK
- Maxgrow tests removed per review
- Actorder tests pruned to essential cases
- E2E test with regex targets added
- `_imatrix_importance` cleanup in `on_finalize`
- `_sums` keyed by module instead of name
- `IMATRIX_PRECISION` constant + `math.prod()`
- Docstrings merged (file -> class)
- Auto-prepend TODO in gatherer docstring

Updated benchmarks and results in the PR description.

43 tests pass. `make style` + `make quality` pass.

examples/quantization_w4a16/README.md
```diff
@@ -138,6 +138,105 @@ We can see the resulting scores look good!
 |     |       |strict-match    |     5|exact_match|↑  |0.720|±  |0.0285|
 ```
 
+---
+
+## iMatrix Importance-Weighted Quantization
```
brian-dellabetta: this documentation and the example above might be better in a separate `examples/imatrix` folder. it looks like this is pretty general for imatrix, and not super specific to W4A16

kylesayrs:
Hey @Yatimai ,

Sorry for the confusion w.r.t. the design of this feature. I’ve synced extensively with @HDCharles and I think I have some suggestions for an approach to this feature.

The confusion and difficulty mainly stems from how this PR interacts with [[WIP] [DDP] Refactor quantization lifecycle for performance](https://github.com/vllm-project/llm-compressor/pull/2469). I would recommend taking the following steps:

1. Move all imatrix functionality to the observer. The observer should not only include the grid search implementation, but the `IMatrixMSEObserver` should also be responsible for adding a calibration hook to the module which is used to calculate the matrix.
    1. You can add a new method `Observer.init(module: Module)` to the `Observer` base class which is called when the observer initializes [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/calibration.py#L84). This gives the `IMatrixMSEObserver` subclass the opportunity to attach a pre_forward hook to the module.
    2. You can add a new method `Observer.detach(module: Module)` which calls `hook.remove()` to remove the hook you added. This method should be called when observers are deleted [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/calibration.py#L267).
2. Make sure that the observer deletes the imatrix off of the module so that it doesn’t end up in the checkpoint. This can be done during the aforementioned `Observer.detach` method.
3. Now that all the functionality is in the observer, the `IMatrixGatherer` does not have any functionality. However, due to how `QuantizationModifier` triggers weight quantization at on start, we need some sort of way to trigger a calibration pass before weight quantization. For this reason, we must keep `IMatrixGatherer` to trigger a forward calibration pass.

```
Recipe: [IMatrixGatherer, QuantizationModifier]

1. IMatrixGatherer.start
	-> IMatrixObservers are attached
2. Sequential Pipeline start
	-> module._imatrix_importance is calibrated
3. IMatrixGatherer.end
	-> Observers are removed
4. QuantizationModifier.start
	-> Observers are attached (_imatrix_importance remains on module)
	-> Weight quantization occurs (with imatrices)
5. DataFree Pipline start
	-> nothing happens
6. QuantizationModifier.end
	-> Observers are removed
```

`IMatrixGatherer` should call `initialize_quantization`  on initialize, `start_calibration` on start, and `end_calibration` on end.

## Changes After Refactor

Once [[WIP] [DDP] Refactor quantization lifecycle for performance](https://github.com/vllm-project/llm-compressor/pull/2469) lands and weight quantization is moved to the end of the sequential epoch, we can do the following simplifications:

1. Move `module._imatrix_importance` from the module to `IMatrixObserver._imatrix_importance`
2. Update examples and docs to remove/deprecate `IMatrixGatherer` , leaving only `IMatrixMSEObserver` to implement the algorithm.
3. Update `QuantizationConfig.requires_calibration_data` to be True if using the `IMatrixMSEObserver`

Yatimai:
Thanks @kylesayrs, this is very clear : moving the hook to the observer keeps the algorithm self-contained while the gatherer handles the lifecycle trigger. Happy to rework along these lines.

Yatimai:
@kylesayrs Quick question on the gatherer: since `initialize_quantization`/`start_calibration`/`end_calibration` are methods of `QuantizationMixin`, should `IMatrixGatherer` inherit from `QuantizationMixin` and build a minimal internal config? Or would you prefer it calls the lower-level functions (`initialize_observer`, `freeze_module_quantization`) directly?

kylesayrs:
@Yatimai I would suggest inheriting from `QuantizationMixin`

Yatimai:
@kylesayrs I checked : the double `apply_quantization_config` is safe (`clear_all_qparams` resets known params first), and `_imatrix_importance` survives since it's not in the cleared params list. I'll proceed with `IMatrixGatherer` inheriting from `QuantizationMixin`.

Yatimai:
Refactor per @kylesayrs design:

- `Observer.init(module)` / `Observer.detach(module)` added to base class
- `IMatrixMSEObserver` now owns the E[x²] hook lifecycle (init attaches, detach computes importance)
- `calibration.py` calls `init` after observer creation and `detach` before deletion
- `IMatrixGatherer` inherits from `QuantizationMixin`, delegates to `initialize_quantization` / `start_calibration` / `end_calibration`
- `_imatrix_importance` persists between gatherer and quantization modifier, cleaned up in `on_finalize`

Benchmarks unchanged, refactor is structural only, no impact on quantization results.

45 tests pass. `make style` + `make quality` clean.

kylesayrs (commented):
I'll give a full review tomorrow, but no notes so far :)

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,110 @@
...
+    :param weight_observer: observer to attach during calibration.
+        Must be ``"imatrix_mse"`` (default).
+    :param ignore: layer name patterns to skip (default: ``["lm_head"]``)
+    :param targets: module types to instrument (default: ``["Linear"]``)
+    """
+
+    scheme: str = "W4A16"
+    weight_observer: str = "imatrix_mse"
+    ignore: List[str] = Field(default_factory=lambda: ["lm_head"])
+    targets: Union[str, List[str]] = Field(default_factory=lambda: ["Linear"])
```
kylesayrs: These fields (with the exception of imatrix_mse) are provided by QuantizationMixin

src/llmcompressor/modifiers/transform/imatrix/base.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # ------------------------------------------------------------------ #
+    #  Lifecycle
+    # ------------------------------------------------------------------ #
+
+    def on_initialize(self, state: State, **kwargs) -> bool:
+        QuantizationMixin.initialize_quantization(self, state.model)
```
kylesayrs: I think you'll want the quantization mixin to be applied by other modifiers (GPTQ/ AWQ/ QMod), not this modifier

kylesayrs (approved):
Awesome tests! Looks great to me

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: @kylesayrs Thanks for the review! I will remove the redundant fields and keep only `weight_observer`.

src/llmcompressor/modifiers/transform/imatrix/base.py
Yatimai: I will refactor so the gatherer creates and manages the observers directly via `Observer.init(module)` / `Observer.detach(module)` without calling `initialize_quantization`. The quantization config stays with the downstream modifier.

src/llmcompressor/observers/base.py
```diff
@@ -133,6 +133,24 @@ def _get_module_param(self, name: str) -> Optional[torch.nn.Parameter]:
         with align_module_device(module):
             return getattr(module, f"{self.base_name}_{name}", None)
 
+    def init(self, module: torch.nn.Module) -> None:
```
brian-dellabetta: if init and detach are inverse operations, i think we should use names indicating as such
```suggestion
    def attach(self, module: torch.nn.Module) -> None:
```

src/llmcompressor/observers/imatrix.py
```diff
@@ -0,0 +1,370 @@
...
+            n_tokens = math.prod(x_f.shape[:-1])
+            token_sum = x_f.pow(2).sum(dim=list(range(x_f.dim() - 1)))
+
+            if mod._imatrix_sum.device != token_sum.device:
+                mod._imatrix_sum = mod._imatrix_sum.to(token_sum.device)
+
+            mod._imatrix_sum.add_(token_sum)
+            mod._imatrix_count += n_tokens
+
+        module._imatrix_hook = module.register_forward_pre_hook(_hook)
```
brian-dellabetta, line 90: i think this is the first case of attaching hooks inside the implementation of an observer? Rather than expanding the api with attach/detach on observers, can't we instead use QuantizationMixin's `_initialize_observers` and `_initialize_hooks` functions? They add observers to modules and hooks to capture input activations. Seems like that same pattern can be used here. 

I'm not sure how well this Observer attach/detach logic would play with the way we use observers elsewhere in the code, for example in AWQ [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L684-L696) (though that's only for weight observers though so maybe it's alright)

brian-dellabetta (commented):
Hi @Yatimai , thanks for all the work on this. I think the top-level API looks good, but i have some questions on the new observer api changes, if they can be resolved another way. Please see comments:

examples/quantization_w4a16/README.md
Yatimai: I will move to `examples/imatrix/`.

src/llmcompressor/observers/base.py
Yatimai: I will rename `init` to `attach`.

src/llmcompressor/observers/imatrix.py
Yatimai: The `Observer.attach/detach` API was designed per @kylesayrs's suggestion to keep the E[x²] lifecycle self-contained in the observer.

HDCharles (approved):
looks good, should be ready to land after the last batch of changes

Yatimai:
All review comments addressed:

- Renamed `init` → `attach`
- Moved example and README to `examples/imatrix/` 
- Updated observer defaults: `norm=3.0`, `maxshrink=0.95` (previously inherited from `memoryless_mse`)
- Removed `maxgrow` parameter (unused with new defaults)
- Updated PR description and results

45 tests pass. `make style` + `make quality` clean.

brian-dellabetta (approved):
Thanks @Yatimai for the contribution!

Yatimai:
Thanks @brian-dellabetta, @kylesayrs, and @HDCharles for the thorough reviews and guidance throughout this PR!

PR #2477, opened 2026-03-17
Fix KV cache calibration for attention modules not named self_attn

changjonathanc:
SUMMARY:
`_apply_kv_cache_scheme` (in compressed-tensors) discovers attention modules via `is_attention_module()`, which is name-agnostic. However, `start_calibration` only iterates modules matching `resolved_targets`, which includes `"re:.*self_attn$"` for KV cache. This regex misses attention modules with different names (e.g. `"attention"`, `"self_attention"`), leaving their observers uninitialized and KV cache scales as garbage values.

Add a fallback pass in `start_calibration` and `end_calibration` that uses `is_attention_module()` to catch any attention modules missed by the regex. Gated by `kv_cache_scheme is not None` so there is zero cost when KV cache quantization is not used. This addresses the existing TODO: "decouple reliance on this regex for matching attention".

TEST PLAN:
- Added unit test with a stub model whose attention modules are named `attention` (not `self_attn`). Verifies observers are initialized, hooks are registered, and modules are frozen correctly.
- All 51 existing quantization modifier tests pass.

### DIFF
--- a/examples/imatrix/README.md
+++ b/examples/imatrix/README.md
@@ -1,0 +1,70 @@
+# iMatrix Importance-Weighted Quantization
+
+`imatrix_mse` is an observer that uses per-channel activation importance (E[x²]) to weight quantization error during range selection. Channels that carry more signal get more careful range optimization.
+
+Two components work together:
+- **`IMatrixGatherer`**: triggers a calibration pass so the observer can collect importance data
+- **`imatrix_mse` observer**: collects E[x²] per input channel via forward pre-hooks and uses importance weighting in the MSE grid search: `err = sum(importance * |Q(w) - w|^p)`
+
+> See [RFC #2456](https://github.com/vllm-project/llm-compressor/discussions/2456) for the full design discussion.
+
+## Quickstart
+
+```bash
+python3 llama3_imatrix_example.py
+```
+
+The simplest setup uses `preset_name_to_scheme` to configure W4A16 and swaps in the `imatrix_mse` observer:
+
+```python
+from compressed_tensors.quantization import preset_name_to_scheme
+from llmcompressor.modifiers.quantization import QuantizationModifier
+from llmcompressor.modifiers.transform.imatrix import IMatrixGatherer
+
+scheme = preset_name_to_scheme("W4A16", ["Linear"])
+scheme.weights.observer = "imatrix_mse"
+
+recipe = [
+    IMatrixGatherer(ignore=["lm_head"]),
+    QuantizationModifier(
+        config_groups={"group_0": scheme},
+        ignore=["lm_head"],
+    ),
+]
+```
+
+## Composing with GPTQ
+
+iMatrix composes with GPTQ by providing importance-weighted ranges for the Hessian-based rounding:
+
+```python
+from llmcompressor.modifiers.gptq import GPTQModifier
+
+scheme = preset_name_to_scheme("W4A16", ["Linear"])
+scheme.weights.observer = "imatrix_mse"
+
+recipe = [
+    IMatrixGatherer(ignore=["lm_head"]),
+    GPTQModifier(
+        config_groups={"group_0": scheme},
+        ignore=["lm_head"],
+    ),
+]
+```
+
+## Results
+
+W4A16, Llama-3.1-8B, group_size=128, WikiText-2 token-level perplexity (141 chunks x 2048):
+
+| Config | PPL |
+|---|---|
+| FP16 baseline | 6.24 |
+| RTN `memoryless_minmax` | 6.96 |
+| GPTQ | 6.92 |
+| AWQ | 6.89 |
+| RTN `imatrix_mse` | 6.85 |
+| GPTQ + `imatrix_mse` | 6.83 |
+
+## Questions or Feature Request?
+
+Please open up an issue on `vllm-project/llm-compressor`
--- a/examples/imatrix/llama3_imatrix_example.py
+++ b/examples/imatrix/llama3_imatrix_example.py
@@ -1,0 +1,61 @@
+from compressed_tensors.offload import dispatch_model
+from compressed_tensors.quantization import preset_name_to_scheme
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+from llmcompressor.modifiers.transform.imatrix import IMatrixGatherer
+
+# Select model and load it.
+model_id = "meta-llama/Meta-Llama-3.1-8B"
+
+model = AutoModelForCausalLM.from_pretrained(model_id, dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+# Select calibration dataset.
+DATASET_ID = "open_platypus"
+
+# Select number of samples. 512 samples is a good place to start.
+# Increasing the number of samples can improve accuracy.
+NUM_CALIBRATION_SAMPLES = 512
+MAX_SEQUENCE_LENGTH = 2048
+
+# Configure the quantization algorithm to run.
+#   * trigger a calibration pass with IMatrixGatherer so the observer can collect E[x²]
+#   * quantize the weights to 4 bit with group size 128
+#   * use imatrix_mse observer to weight quantization error by channel importance
+scheme = preset_name_to_scheme("W4A16", ["Linear"])
+scheme.weights.observer = "imatrix_mse"
+
+recipe = [
+    IMatrixGatherer(ignore=["lm_head"]),
+    QuantizationModifier(
+        config_groups={"group_0": scheme},
+        ignore=["lm_head"],
+    ),
+]
+
+# Apply algorithms.
+oneshot(
+    model=model,
+    dataset=DATASET_ID,
+    splits={"calibration": "train[:5%]"},
+    recipe=recipe,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+)
+
+# Confirm generations of the quantized model look sane.
+print("\n\n")
+print("========== SAMPLE GENERATION ==============")
+dispatch_model(model)
+sample = tokenizer("Hello my name is", return_tensors="pt")
+sample = {key: value.to(model.device) for key, value in sample.items()}
+output = model.generate(**sample, max_new_tokens=100)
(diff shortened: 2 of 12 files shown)


brian-dellabetta:
> However, start_calibration only iterates modules matching resolved_targets, which includes "re:.*self_attn$" for KV cache

Hi @changjonathanc , can you point me to where you're seeing this? [These lines](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/lifecycle/apply.py#L160-L163) in compressed-tensors? Is it just a matter of updating the targets list to include `"re:.*(self_attn|attention)$"`?

changjonathanc:
https://github.com/vllm-project/llm-compressor/blob/370c04c34191c0e5fb88cf889e16578bf8c7fe20/src/llmcompressor/modifiers/quantization/quantization/mixin.py#L204-L206

brian-dellabetta:
> https://github.com/vllm-project/llm-compressor/blob/370c04c34191c0e5fb88cf889e16578bf8c7fe20/src/llmcompressor/modifiers/quantization/quantization/mixin.py#L204-L206

Hi @changjonathanc , can't we just update the regex to `"re:.*(self_attn|attention)$"`?

changjonathanc:
Hi @brian-dellabetta , 

honestly i'm not 100% sure, so here is what claude told me and i think it makes sense:

---

  When you do KV cache quantization, the library needs to:
  1. Find all attention modules in the model
  2. Attach a quantization scheme to them
  3. Initialize calibration observers/hooks on them

  The bug is that steps 1-2 and step 3 use different methods to find the attention modules, so they don't always agree on which modules to process.

  ---
  Step 1 & 2: How modules get their scheme attached

  This happens in compressed-tensors, in _apply_kv_cache_scheme:
```
  for submodule in model.modules():         # iterate ALL modules
      if is_attention_module(submodule):    # check: is this an attention module?
          submodule.quantization_scheme = scheme

  is_attention_module checks the class name of the module:
  "attention" in module.__class__.__name__.lower()
  # AND has k_proj, v_proj, qkv_proj, or kv_b_proj
```
  So a module whose Python class is named LlamaAttention passes this check — regardless of what it's called inside the model tree (self_attn, attention, mha, whatever).

  After this step, any attention module has quantization_scheme set on it.

  ---
  Step 3: How calibration observers get initialized

  This happens in llm-compressor, in start_calibration:
```
  for _, module in match_named_modules(model, self.resolved_targets, ...):
      self._initialize_observers(module)
      ...
```
  match_named_modules uses self.resolved_targets, which contains:
  "re:.*self_attn$"   # a regex matching the module's PATH in the tree

  This regex matches module paths like model.layers.0.self_attn — it checks the name you'd use to look up the module, not the class name.

  ---
  The Bug

  Imagine a model where the attention module is stored at path layers.0.attention (not self_attn), but the class is still LlamaAttention.

  ---
  What the PR does

  It adds a fallback in start_calibration: after the regex loop, do a second pass using is_attention_module() — the exact same check that _apply_kv_cache_scheme uses — to catch anything the regex missed:
```
  # First pass: regex-based (existing)
  for _, module in match_named_modules(model, self.resolved_targets, ...):
      self._start_calibrating_module(module)
      initialized_ids.add(id(module))

  # Second pass: class-name-based fallback (new)
  if self.resolved_config.kv_cache_scheme is not None:
      for _, module in model.named_modules():
          if id(module) not in initialized_ids and is_attention_module(module) ...:
              self._start_calibrating_module(module)
```
  Now both sides use is_attention_module() as the source of truth, so they always agree.`

brian-dellabetta:
> It adds a fallback in start_calibration: after the regex loop, do a second pass using is_attention_module() — the exact same check that _apply_kv_cache_scheme uses — to catch anything the regex missed

@changjonathanc rather than re-applying after the fact, we just need to update the regex used to match on both `self_attn` and `attention`. The one i proposed above should be able to handle that without needing custom logic after-the-fact. 

Unfortunately there's no guaranteed way to determine if a module is an attention module and we have to rely on these fuzzy matches, so we'll have to update when necessary as naming conventions drift. Can you try with that one-line change to the regex, rather than the changes you have in `src` in this PR, and see if your tests still pass?

changjonathanc:
Hi @brian-dellabetta, I think your suggestion passes the test, but the confusion still exists, where one part of the code patches by the class name of a module, while the other matches the path. wdyt?

brian-dellabetta:
> Hi @brian-dellabetta, I think your suggestion passes the test, but the confusion still exists, where one part of the code patches by the class name of a module, while the other matches the path. wdyt?

Hi @changjonathanc , these are two separate checks:
1. given a pointer to a `torch.nn.Module`, how can I know if it is an attention module? We have some fuzzy matching logic to check the class name of the module and if it has attributes like `k_proj` or `qkv_proj`. If we see new class names, like I know deepseek v3.2 uses `MLA` as the class name for [multi-latent attention](https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/inference/model.py#L498) module, we will need to update this check.
2. given that i want to apply kv_cache quantization to a model, what should the regex be to catch all possible naming conventions that are used, like `model.layers.0.self_attn` or in the case of deepseek v3.2 it will be `model.layers.0.attn`, and you hit a case where it's just `*.attention`. 

In case (1) we will have to update the conditional logic to capture any drift in module class naming conventions, 
in case (2) we will have to update the regex to capture any drift in model submodule naming conventions.

changjonathanc:
Hi @brian-dellabetta, that makes sense. But since compressed-tensors already decided which modules get `quantization_scheme` via `is_attention_module()`, llm-compressor is re-deriving that same decision with a regex that can drift independently. We actually hit this exact bug in our codebase and had to monkey-patch it for the same reason. Happy to go with the simpler regex fix if you prefer consistency with the rest of the codebase though.

brian-dellabetta:
> Hi @brian-dellabetta, that makes sense. But since compressed-tensors already decided which modules get `quantization_scheme` via `is_attention_module()`, llm-compressor is re-deriving that same decision with a regex that can drift independently. We actually hit this exact bug in our codebase and had to monkey-patch it for the same reason. Happy to go with the simpler regex fix if you prefer consistency with the rest of the codebase though.

Thanks @changjonathanc  for the detail, good to know. I almost replied that a better solution would be to try to consolidate the logic, but that might be better as an RFC. I will raise this issue with the team next week and discuss internally

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -202,7 +202,7 @@ def resolved_targets(self) -> Set[str]:
                 targets.add(target)
 
         if self.resolved_config.kv_cache_scheme is not None:
-            # TODO: decouple reliance on this regex for matching attention
+            # TODO: also apply is_attention_module() fallback in initialize_quantization
             targets.add("re:.*self_attn$")
```
brian-dellabetta:
```suggestion
            targets.add("re:.*(self_attn|attention)$")
```

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -238,9 +244,18 @@ def start_calibration(self, model: torch.nn.Module):
             untie_word_embeddings(model)
 
         for _, module in match_named_modules(model, self.resolved_targets, self.ignore):
-            self._initialize_observers(module)
-            self._calibration_hooks |= self._initialize_hooks(module)
-            apply_calibration_status(module)
+            self._start_calibrating_module(module)
+
+        # Fallback: catch attention modules missed by the "re:.*self_attn$" regex.
+        if self.resolved_config.kv_cache_scheme is not None:
```
brian-dellabetta: with the update to targets, ideally we can remove this fallback

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -255,6 +270,16 @@ def end_calibration(self, model: torch.nn.Module):
         for _, module in match_named_modules(model, self.resolved_targets, self.ignore):
             freeze_module_quantization(module)  # remove observers
 
+        # Also freeze attention modules missed by the regex fallback in start_calibration.
+        if self.resolved_config.kv_cache_scheme is not None:
```
brian-dellabetta: with update to targets, ideally we can remove this fallback

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -255,6 +270,16 @@ def end_calibration(self, model: torch.nn.Module):
...
+            for _, module in model.named_modules():
+                if (
+                    is_attention_module(module)
```
brian-dellabetta: note if we need to update `is_attention_module` we can do that in a compressed-tensors PR

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -202,8 +202,7 @@ def resolved_targets(self) -> Set[str]:
                 targets.add(target)
 
         if self.resolved_config.kv_cache_scheme is not None:
-            # TODO: decouple reliance on this regex for matching attention
-            targets.add("re:.*self_attn$")
+            targets.add("re:.*(self_attn|attention)$")
```
brian-dellabetta: can we keep the TODO? It is still something we'd like to address in future, as it stands we will have to continue updating this regex over time
```suggestion
            # TODO: decouple reliance on this regex for matching attention
            targets.add("re:.*(self_attn|attention)$")
```

tests/llmcompressor/modifiers/quantization/test_kv_cache_calibration.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+        self.layers = nn.ModuleList([_StubBlock(dim) for _ in range(num_layers)])
+
+    def forward(self, x):
+        for layer in self.layers:
+            x = layer(x)
+        return x
+
+
+def test_attention_module_not_named_self_attn_gets_calibrated():
```
kylesayrs, line 70: These tests are really verbose. Do you think they can be simplified/shortened?

brian-dellabetta: I don't see a clear way to reduce a ton, but now that this PR is limited to a single line change, we could probably remove it entirely. wdyt?

brian-dellabetta: Nothing heard, i will just merge this in. I think it's fine as is.

PR #2478, opened 2026-03-17
Add sequential_weight_offload_device to skip unnecessary weight offloading

changjonathanc:
SUMMARY:
Separate weight offloading from activation offloading in the sequential pipeline.
Today `dispatch_for_sequential` always offloads weights to CPU, even when the model
fits on the GPU. The new `sequential_weight_offload_device` parameter (default `"cpu"`,
backward-compatible) lets users set `"none"` to disable weight offloading, eliminating
unnecessary CPU↔GPU transfers while still benefiting from sequential activation caching.

Benchmark (TinyLlama-1.1B, W8A8, 32 samples, seq_len=512):
- `sequential_weight_offload_device="cpu"`:  17.1s
- `sequential_weight_offload_device="none"`: 14.5s  (~17% faster)

TEST PLAN:
- Added parametrized unit test for `dispatch_for_sequential` with `None`, `"none"`, and `"None"`.
- Verified quantized output is identical between `cpu` and `none` modes.
- All existing tests pass.

### DIFF
--- a/src/llmcompressor/pipelines/sequential/helpers.py
+++ b/src/llmcompressor/pipelines/sequential/helpers.py
@@ -534,7 +534,7 @@ def dispatch_for_sequential(
     """
     if onload_device is None:
         onload_device = get_main_device()
-    if isinstance(offload_device, str) and offload_device.lower() == "none":
+    if isinstance(offload_device, str) and offload_device.strip().lower() == "none":
         offload_device = None
     if offload_device is None:
         remove_hook_from_module(model, recurse=True)


HDCharles:
@kylesayrs i don't think thats what this PR is getting at, we currently don't support enabling activation offloading but disable weight offloading in oneshot which is what i think this is attempting to enable.

@changjonathanc 
i think its reasonable to want to do this, however the sequential pipeline is fundamentally about sequentially onloading/offloading the model and in general the way to enable running the whole model on gpu is using the the basic pipeline. The basic pipeline doesn't have activation offloading though I think it would make sense to add it as an option there, rather than removing sequential offloading from the sequential offloading pipeline. while it may be easier in some cases to just delete the hooks and don't offload in the sequential pipeline, in general sequential pipeline offloading/onloading is under heavy development between DDP work and various offloading context (e.g. the big model stuff kyle mentioned earlier) so adding more responsibilities for this path seems like a bad idea.

edit: talking with kyle actually this isn't the case, if the model is loaded on with device_map="auto" then the offloading will maintain that so this is already doable. I also see another PR in compressed tensors so i'm not sure what you're trying to do exactly.

kylesayrs:
Hi @changjonathanc, feel free to join the [LLM Compressor Slack](https://communityinviter.com/apps/vllm-dev/join-vllm-developers-slack) and we can sync about what you're trying to achieve!

changjonathanc:
Thanks @kylesayrs 
I see the issue now. I updated my code to put model on the correct devices before it touches llmcompressor code and it works. i think this pr is not needed anymore.

Issue #2479, opened 2026-03-17
AWQ various enhancements and research


HDCharles:
Figure out whether the following techniques would be useful/profitable to enable

1) AWQ + MSE observer where AWQ jointly optimizes scale + shrinkage, i.e. if you AWQ with minmax -> quantization with MSE observer how much worse does that perform than including shrinkage factor in the grid search?

2) in general AWQ's scale search is brute force, there exist myriad black box optimization techniques e.g. bayesian optimization that could potentially massively reduce the search time, especially if (1) above is fruitful.

HDCharles:
assigned to @dzhengAP

dzhengAP:
Sure @HDCharles, I will take it

dzhengAP:
@HDCharles — I am attaching some findings summary for #2479.

##  AWQ + MSE Observer Alignment
I open a tentative PR and benchmarked all 4 combinations of (search, final)x(mse, minmax)on Llama-3.1-8B-Instruct W4A16, open-platypus calibration, WikiText-2 eval:

| Recipe | PPL | Δ vs baseline |
|---|---|---|
| FP16 (reference) | 9.459 | — |
| minmax search + minmax final (baseline) | 10.023 | +0.000 |
| minmax search + mse final | 9.953 | **-0.070** |
| mse search + minmax final | 9.966 | -0.057 |
| mse search + mse final | 10.029 | +0.006 |

It appears that `memoryless_mse` as the final quantization observer consistently outperforms `memoryless_minmax` regardless of the search observer used (+0.057–0.070 PPL improvement). However, using MSE in the grid search itself appears counterproductive — `mse search + mse final` is the worst performing MSE combination, suggesting that MSE shrinkage during search finds scales optimized for aggressive clipping that do not generalize well at deployment time. _Or maybe the research direction can go the other way like, why misalignment perform better?_

These findings are based on a single model (Llama-3.1-8B-Instruct) and dataset (open-platypus). Results may vary across different architectures and calibration data, also the delta across the combinations are fairly small— would appreciate your thoughts on whether broader evaluation is warranted before drawing conclusions.

## PR #2492 Contributions

While the benchmark shows alignment alone does not yield consistent gains, the investigation surfaced several concrete improvements to the AWQ implementation:

- **`search_observer` parameter** on `AWQModifier` — aligns grid search observer with final quantization observer. Default `"memoryless_minmax"` is fully backward compatible
[...]

PR #2492, opened 2026-03-20
[AWQ] Add output_mse_shrinkage: per-group clipping via activation projection

dzhengAP:
## Problem
 
AWQ's grid search optimizes the per-channel scale (α) against **output MSE** via a full forward pass. However, the clipping range (shrinkage factor p) is determined independently by the observer using **weight MSE**. These two objectives are misaligned — the clipping range is never evaluated against the same output MSE objective that drives the scale search.
 
## Solution
 
`output_mse_shrinkage`: for each scale candidate α, find the best clipping factor `p` per quantization group by minimizing the activation-projected weight quantization error:
 
```
w_err   = W_quant - W_scaled              # (out_ch, G, group_size)
out_err = einsum('ogs,ngs->ogn',
                 w_err, X_grouped)        # (out_ch, G, n_tokens)
err     = out_err.pow(2).sum(n)           # (out_ch, G)
```
 
`X` is collected from real calibration samples via a forward hook on each balance layer, giving true output-space error per group. Each group independently selects its optimal clipping factor `p` — groups with outlier weights but small activations can be clipped aggressively, while groups with large activations remain conservative.
 
## Usage
 
```python
recipe = [
    AWQModifier(
        ignore=["lm_head"],
        scheme="W4A16_ASYM",
        targets=["Linear"],
        n_shrink_grid=10,   # number of shrink candidates (1 = disabled)
        maxshrink=0.20,     # search range: p in [1-maxshrink, 1.0]
    ),
    QuantizationModifier(...),
]
```
 
## Ablation results
 
Model: `meta-llama/Llama-3.1-8B-Instruct`, W4A16 ASYM group=128, open-platypus 128 samples, WikiText-2 eval
 
| Recipe | PPL | Δ vs RTN+minmax | Time |
|---|---|---|---|
| RTN + minmax *(lower bound)* | 10.165 | +0.000 | 0.7m |
| GPTQ only | 10.038 | -0.127 ✓ | 14.3m |
| AWQ + minmax *(vanilla AWQ)* | 9.982 | -0.182 ✓ | 8.5m |
| **AWQ + output MSE *(ours)*** | **9.963** | **-0.202** ✓ | 34.8m |
 
Our method achieves the best PPL, improving -0.020 over vanilla AWQ and -0.202 over the RTN baseline.
 
## Implementation notes
 
- Chunked einsum over `out_ch` (~256MB per chunk) to bound peak GPU memory
- Activation samples capped at 2048 tokens to prevent OOM on large models
- `n_shrink_grid=1` (default) disables shrinkage entirely — fully backward compatible, no existing recipes affected
- 5 new unit tests covering defaults, valid inputs, and backward compatibility
 
## Changes
 
- Add `n_shrink_grid: int = 1` and `maxshrink: float = 0.20` fields to `AWQModifier`
- Implement `_apply_output_mse_shrinkage` — per-group clipping optimization via activation projection
- Chunked einsum + 2048-token cap to prevent OOM on large models
- 5 new unit tests
- All testes passed

Part of #2479
cc @HDCharles @brian-dellabetta

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -1,4 +1,5 @@
 import inspect
+import math
 from itertools import product
 from typing import Iterator, Literal
 
@@ -84,6 +85,10 @@ class AWQModifier(Modifier, QuantizationMixin):
         # This change is only useful for MoE models with parallel transformer blocks,
         # and one should use the default value (None) in most cases.
       ignore: ["lm_head"]
+      # Set search_observer to match the observer used in the final QuantizationModifier
+      # for best scale alignment. "memoryless_mse" is recommended when pairing with
+      # MSE-based weight quantization. Defaults to "memoryless_minmax".
+      search_observer: "memoryless_minmax"
       config_groups:
         group_0:
           targets:
@@ -151,6 +156,21 @@ class AWQModifier(Modifier, QuantizationMixin):
         this specifies how many grid points should be used. To decrease the runtime,
         at the possible cost of slightly worse scales, this can be decreased.
         Defaults to 20
+    :param search_observer: name of the observer used to simulate quantization during
+        the grid search. For best accuracy, this should match the observer used in the
+        final QuantizationModifier (e.g. "memoryless_mse" when using MSE-based weight
+        quantization). Defaults to "memoryless_minmax" for backward compatibility.
+        Valid options: "memoryless_minmax", "memoryless_mse"
+    :param n_shrink_grid: number of shrinkage grid points to jointly search over
+        alongside the scale grid. When > 1, for each scale candidate (α) the grid
+        search also sweeps over shrink factors p ∈ (1 - maxshrink, 1], selecting the
+        (α, p) pair that minimizes output MSE jointly. This implements the joint
+        scale + shrinkage optimization described in issue #2479.
+        Defaults to 1 (disabled, shrinkage determined solely by the observer).
+        Recommended value: 5-10 for a lightweight joint search.
+    :param maxshrink: maximum shrinkage factor used when n_shrink_grid > 1.
+        Shrink factors are swept from 1.0 down to (1 - maxshrink).
+        Defaults to 0.20 (matching the memoryless_mse observer default).
     """
 
     # Allow arbitrary types because AWQMapping has fields of type torch.nn.Module
@@ -162,6 +182,9 @@ class AWQModifier(Modifier, QuantizationMixin):
     offload_device: torch.device | None | Sentinel = Sentinel("not_provided")
     duo_scaling: bool | Literal["both"] = True
     n_grid: int = 20
+    search_observer: str = "memoryless_minmax"
+    n_shrink_grid: int = 1
+    maxshrink: float = 0.20
 
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: list[ResolvedMapping] = PrivateAttr(default_factory=list)
@@ -673,8 +696,9 @@ class AWQModifier(Modifier, QuantizationMixin):
                 n_grid = self.n_grid
                 duo_scalings = [self.duo_scaling]
 
-        # Where appropriate, replace observers with memoryless_minmax
-        # for duration of grid search
+        # Where appropriate, replace observers with search_observer
+        # for duration of grid search. This should match the observer
+        # used in the final QuantizationModifier for best scale alignment.
         balance_layers_to_patch = [
             balance_layer
             for balance_layer in mapping.balance_layers
@@ -686,7 +710,7 @@ class AWQModifier(Modifier, QuantizationMixin):
             "weight_observer",
             [
                 Observer.load_from_registry(
-                    "memoryless_minmax",
+                    self.search_observer,
                     base_name="weight",
                     args=balance_layer.quantization_scheme.weights,
                     module=balance_layer,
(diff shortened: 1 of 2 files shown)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -686,7 +700,7 @@ def _compute_best_scale(
             "weight_observer",
             [
                 Observer.load_from_registry(
-                    "memoryless_minmax",
+                    self.search_observer,
```
brian-dellabetta, line 733: do we ever want this to be different than the observer defined in the quantization config? Wondering if we can save users another configuration option on AWQModifier, while also being more flexible to use different observers for heterogeneous quantization. Maybe something like this:
```suggestion
                    balance_layer.quantization_scheme.observer or "memoryless_minmax",
```
cc @HDCharles , wdyt?

HDCharles:
hmm, pretty insignificant relative to AWQ itself, probably we can cross this one off our list, its a bunch of runtime/complexity for almost no gain.

HDCharles:
Oh i realize the problem: the shrinkage factor needs to be distinct for each different quantization group, right now its hardcoding the shrinkage for all groups.

This does seem workable, with some finesseing

normally MSE observer reshapes the weight to be num_groups x groupsize and then picks the best shrinkage for each group since we can calculate the MSE for each group separately

doing this for the module output is possible too for per channel quantization but becomes a little more complicated for group/block quantization

what we could do:

take weight (out_channels x in_channels) and activations (n x in_channels),
dot them together for each sample: (n x out_channels x in_channels)
subtract from original dotted with activations: (n x out_channels x in_channels)
square and sum over n: (out_channels x in_channels)
 reshape for quantization: (num_groups x group_size)
 sum each group: (num_groups) then you have an error value for each group and could pick the best one for each group similar to how its done in the MSE observer

this would reduce the O(num_groups*num_shrinkage_factors) optimization problem to just O(num_shrinkage_factors)

dzhengAP:
Hi @HDCharles, I am attaching some new results following our discussion, please check below. Key Takeaway is the new methods works, and it doesn't need to go monotonously, as n = 10 shows the best results (probably due to not overfitting).

## Motivation
AWQ's grid search optimizes scale (alpha) against output MSE, but the quantization clipping range (shrinkage) is determined independently by the observer using weight MSE. These objectives are misaligned.

## Methods & Algo
This commit adds output_mse_shrinkage: per-group shrinkage optimization using the same output MSE objective as the scale search. For each scale candidate, the best clipping factor p is selected per quantization group by minimizing the activation-projected weight quantization error:

  w_err  = W_quant - W_scaled            (out_ch, G, group_size)
  out_err = einsum('ogs,ngs->ogn', w_err, X_grouped)     (out_ch, G, n_tokens)
  err     = out_err^2.sum(n)             (out_ch, G)

X is collected from real calibration samples via a forward hook on the balance layer, so the error reflects actual token distributions rather than a proxy. Each group independently selects its optimal p, allowing aggressive clipping where activations are small and conservative clipping where they are large.

New parameters:
  n_shrink_grid: int = 1    (1 = disabled, backward compatible)
  maxshrink: float = 0.20   (search range: p in [1-maxshrink, 1.0])

## Results
Benchmarked on Llama-3.1-8B-Instruct W4A16 ASYM group=128, open-platypus calibration, WikiText-2 eval:

  Baseline (n_shrink_grid=1):   PPL 9.995
  output_mse_shrinkage (n=10):  PPL 9.890  (-0.105) ← best
  output_mse_shrinkage (n=50):  PPL 9.953  (-0.042)
  output_mse_shrinkage (n=100): PPL 9.941  (-0.054)

All n values improve over baseline. n=10 gives best result; improvement is not strictly monotonic, suggesting diminishing returns from finer shrinkage resolution on calibration data.

##Notes
Implementation notes:
  - Chunked einsum over out_ch to bound peak memory (~256MB per chunk)
  - Activation samples capped at 2048 tokens to prevent OOM on large models
  - 5 new unit tests, all passed

Part of #2479
CC @brian-dellabetta

src/llmcompressor/modifiers/awq/base.py
HDCharles: Yeah agree

dzhengAP:
# [AWQ] Add output_mse_shrinkage: per-group clipping via activation projection
 
Part of #2479
cc @HDCharles @brian-dellabetta
 
## Problem
 
AWQ's grid search optimizes the per-channel scale (α) against **output MSE** via a full forward pass. However, the clipping range (shrinkage factor p) is determined independently by the observer using **weight MSE**. These two objectives are misaligned — the clipping range is never evaluated against the same output MSE objective that drives the scale search.
 
## Solution
 
`output_mse_shrinkage`: for each scale candidate α, find the best clipping factor `p` per quantization group by minimizing the activation-projected weight quantization error:
 
```
w_err   = W_quant - W_scaled              # (out_ch, G, group_size)
out_err = einsum('ogs,ngs->ogn',
                 w_err, X_grouped)        # (out_ch, G, n_tokens)
err     = out_err.pow(2).sum(n)           # (out_ch, G)
```
 
`X` is collected from real calibration samples via a forward hook on each balance layer, giving true output-space error per group. Each group independently selects its optimal clipping factor `p` — groups with outlier weights but small activations can be clipped aggressively, while groups with large activations remain conservative.
 
## Usage
 
```python
recipe = [
    AWQModifier(
        ignore=["lm_head"],
        scheme="W4A16_ASYM",
        targets=["Linear"],
        n_shrink_grid=10,   # number of shrink candidates (1 = disabled)
        maxshrink=0.20,     # search range: p in [1-maxshrink, 1.0]
    ),
    QuantizationModifier(...),
]
```
 
## Ablation results
 
Model: `meta-llama/Llama-3.1-8B-Instruct`, W4A16 ASYM group=128, open-platypus 128 samples, WikiText-2 eval
 
| Recipe | PPL | Δ vs RTN+minmax | Time |
|---|---|---|---|
| RTN + minmax *(lower bound)* | 10.165 | +0.000 | 0.7m |
| GPTQ only | 10.038 | -0.127 ✓ | 14.3m |
| AWQ + minmax *(vanilla AWQ)* | 9.982 | -0.182 ✓ | 8.5m |
| **AWQ + output MSE *(ours)*** | **9.963** | **-0.202** ✓ | 34.8m |
 
Our method achieves the best PPL, improving -0.020 over vanilla AWQ and -0.202 over the RTN baseline.
 
## Implementation notes
 
- Chunked einsum over `out_ch` (~256MB per chunk) to bound peak GPU memory
- Activation samples capped at 2048 tokens to prevent OOM on large models
- `n_shrink_grid=1` (default) disables shrinkage entirely — fully backward compatible, no existing recipes affected
- 5 new unit tests covering defaults, valid inputs, and backward compatibility
 
## Changes
 
- Add `n_shrink_grid: int = 1` and `maxshrink: float = 0.20` fields to `AWQModifier`
- Implement `_apply_output_mse_shrinkage` — per-group clipping optimization via activation projection
- Chunked einsum + 2048-token cap to prevent OOM on large models
- 5 new unit tests; all units test passed
- Updated to the main PR description

PR #2493, opened 2026-03-20
[Feature] Allow targeting multiples of sequential targets

### LINKED ISSUE #2481: [Feature] Allow targeting multiples of sequential targets
## Background ##
LLM Compressor uses [sequential onloading](https://docs.vllm.ai/projects/llm-compressor/en/latest/guides/big_models_and_distributed/sequential_onloading/) to onload subgraphs of a model. How and which subgraphs are created is determined by the `sequential_targets` argument, either passed by the user or inferred from the model definition (typically one DecoderLayer).

The choice of sequential targets comes with tradeoffs. Larger sequential targets are more runtime efficient, but use more memory. Smaller sequential targets use less memory, but are less runtime efficeint.

Currently, a user can only specify the `sequential_targets` argument. However, there are cases where a user will want to pack multiple sequential targets into a single subgraph (for example, onload two decoder layers per subgraph, rather than just one).

The logic for creating subgraphs from targets is implemented by the [`topological_partition`](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/helpers.py#L263-L273) function.

## Requested Changes ##
1. Design an interface for allow users to target multiple sequential targets per subgraph (for example, if your GPUs can fit 2-3 decoder layers)
2. Modify `topological_partition` to allow multiple targets to be assigned to the same subgraph
3. Test the feature. You should see higher vram usage and lower runtime

aayush7511:
SUMMARY:
Closes #2481

TEST PLAN:

### DIFF
--- a/src/llmcompressor/pipelines/sequential/helpers.py
+++ b/src/llmcompressor/pipelines/sequential/helpers.py
@@ -84,6 +84,7 @@ def trace_subgraphs(
     sample_input: dict[str, Any],
     sequential_targets: list[str],
     ignore: list[str],
+    targets_per_subgraph: int = 1
 ) -> list[Subgraph]:
     """
     Trace a model to produce subgraphs, where each sequential target belongs to exactly
@@ -95,6 +96,8 @@ def trace_subgraphs(
         __len__, __bool__, and __contains__ values are assumed constant across batches
     :param sequential_targets: list of patterns matching sequential targets
     :param ignore: function and method names to skip during tracing
+    :param targets_per_subgraph: number of targets to include per subgraph
+    
     :return: a list of Subgraphs in order of execution
     """
     # find modules
@@ -149,7 +152,7 @@ def trace_subgraphs(
     graph.device = model.device
 
     # perform subgraph partition
-    partitions = topological_partition(graph, targets)
+    partitions = topological_partition(graph, targets, targets_per_subgraph)
     subgraphs = partition_graph(model, partitions)
     trace_consumed_names(subgraphs)
 
@@ -257,7 +260,7 @@ def find_target_nodes(graph: GraphModule, targets: set[Module]) -> set[Node]:
     )
 
 
-def topological_partition(graph: GraphModule, targets: set[Module]) -> list[list[Node]]:
+def topological_partition(graph: GraphModule, targets: set[Module], targets_per_subgraph: int = 1) -> list[list[Node]]:
     """
     Partition the graph into partitions such that each `target` belongs to exactly one
     partition and executing each partition depends only on intermediate values produced
@@ -265,6 +268,7 @@ def topological_partition(graph: GraphModule, targets: set[Module]) -> list[list
 
     :param graph: graph being partitioned
     :param targets: target modules which will be assigned to disjoint partitions
+    :param targets_per_subgraph: number of targets to include per subgraph
     :return: list of partitions, where each partition is a list of nodes belonging to
         that partition
     """
@@ -277,6 +281,7 @@ def topological_partition(graph: GraphModule, targets: set[Module]) -> list[list
         for node in graph.graph.nodes
     }
     partition_index = 0  # global counter
+    targets_seen = 0 # global counter
 
     # start with graph input nodes,
     # but delay the `get_attr` nodes as long as possible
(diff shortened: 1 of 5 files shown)


src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -260,27 +262,32 @@ def find_target_nodes(graph: GraphModule, targets: set[Module]) -> set[Node]:
...
+        raise ValueError("targets_per_subgraph is required to be greater than or equal to one")
 
     partitions: list[list[Node]] = [[]]
     remaining_indegrees = {
         node: len([node for node in node.all_input_nodes if node.op != "get_attr"])
         for node in graph.graph.nodes
     }
     partition_index = 0  # global counter
-
+    targets_seen = 0 # number of targets encountered so far
```
kylesayrs:
```suggestion
    targets_seen = 0  # number of targets encountered so far
```

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -260,27 +262,32 @@ def find_target_nodes(graph: GraphModule, targets: set[Module]) -> set[Node]:
...
     :param graph: graph being partitioned
     :param targets: target modules which will be assigned to disjoint partitions
+    :param targets_per_subgraph: number of targets to include per subgraph
     :return: list of partitions, where each partition is a list of nodes belonging to
         that partition
     """
     assert graph_is_well_formed(graph.graph)
     target_nodes = find_target_nodes(graph, targets)
+    
+    if(targets_per_subgraph <= 0):
```
kylesayrs:
```suggestion
    if targets_per_subgraph <= 0:
```

kylesayrs (approved):
Looks great! Please consider adding some tests to verify that each subgraph has the submodules you expect (see `Subgraph.submodules`).

aayush7511:
Hi @kylesayrs , 

To test if each submodule has the necessary targets, I'm generating a subgraph using trace_subgraph. Currently trace_subgraph requires model.config in one of its lines to run. 

`src/llmcompressor/pipelines/sequential/helpers.py`
```python
        # calibration context
        stack.enter_context(HooksMixin.disable_hooks())

        # flags useful for tracing
        stack.enter_context(patch_attr(model.config, "_attn_implementation", "eager"))
        stack.enter_context(patch_attr(torch.compiler, "_is_compiling_flag", True))

        # autowrap forwards
        stack.enter_context(autowrap_forwards(ancestors, ignore))
```


With the way the current tests are setup, the models are simple torch.nn.module layers with no config. **What model would you recommend using or should I create a model with a dummy config file.** 

`tests/llmcompressor/pipelines/sequential/test_helpers.py`
```python
class DummyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.seq = torch.nn.Sequential(torch.nn.Linear(10, 20), torch.nn.ReLU())
        self.fc = torch.nn.Linear(20, 5)

    def forward(self, x):
        x = self.seq(x)
        return self.fc(x)

class DummyModelMultipleSequentialLayers(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.layer1 = torch.nn.Linear(10, 10)
        self.layer2 = torch.nn.Linear(10, 10)
        self.layer3 = torch.nn.Linear(10, 10)
        self.layer4 = torch.nn.Linear(10, 10)
        self.layer5 = torch.nn.Linear(10, 10)
        self.layer6 = torch.nn.Linear(10, 10)
        
    
    def forward(self, x):
        x = self.layer1(x)
        x = self.layer2(x)
        x = self.layer3(x)
        x = self.layer4(x)
        x = self.layer5(x)
        x = self.layer6(x)
        return x
```

kylesayrs:
@aayush7511 I recommend using `Qwen/Qwen3-0.6B` on the meta device, similar to [these tests](https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/transformers/tracing/test_models.py).

brian-dellabetta (approved):
Thanks for the contribution! lgtm pending open comments

aayush7511:
Hi, 
I'm just working on bugs coming up with the Qwen model. Will have the tests done soon!

kylesayrs:
@aayush7511 Please make sure to run `make style; make quality` as well

kylesayrs:
@aayush7511 Looks like there's some test failures

Issue #2202, opened 2026-01-08
Regarding the Quantization Scheme for the Qwen-Next 80B Model W8A8


vonchenplus:
Hi there,

I’m planning to quantize the Qwen-Next 80B model to W8A8 using a combination of SmoothQuant and GPTQ. However, I noticed that SmoothQuant does not currently support this model.

In this situation, would it be acceptable to use the default mapping method for SmoothQuant? If not, are there any recommended alternatives or best practices for achieving stable and accurate W8A8 quantization on this model?

Thanks in advance for your help!

dsikka:
@rahul-tuli please take a look

rahul-tuli:
Hi @vonchenplus thanks for opening this issue, The SmoothQuant mapping format and how to choose the right modules is documented here in our SmoothQuant Modifier Mapping Tutorial: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/smoothquant/README.md.
​
This walks through the rationale for targeting q/k/v and feed-forward blocks, explains why we match leaf modules, and shows concrete regex-based mapping examples you can adapt to Qwen-Next 80B.
​
Please take a look and try setting up your mappings based on that guide, and if anything is still unclear or you run into issues with Qwen-Next 80B specifically, let us know here and we’ll be happy to help further.

vonchenplus:
> Hi [@vonchenplus](https://github.com/vonchenplus) thanks for opening this issue, The SmoothQuant mapping format and how to choose the right modules is documented here in our SmoothQuant Modifier Mapping Tutorial: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/smoothquant/README.md.​This walks through the rationale for targeting q/k/v and feed-forward blocks, explains why we match leaf modules, and shows concrete regex-based mapping examples you can adapt to Qwen-Next 80B.​Please take a look and try setting up your mappings based on that guide, and if anything is still unclear or you run into issues with Qwen-Next 80B specifically, let us know here and we’ll be happy to help further.

Hi, thanks for the detailed explanation and for pointing me to the SmoothQuant mapping tutorial — this is very helpful.

I’ll go through the guide and try setting up the mappings for Qwen-Next 80B accordingly. If I run into any issues or have questions specific to this model, I’ll follow up here.

Thanks again for your help!

anmarques:
[...]

PR #2500, opened 2026-03-21
fix: add norm calibration context for unit-offset RMSNorm (Gemma/Qwen3Next)

### LINKED ISSUE #2102: [Bug]: AWQ for Gemma 3 fails to quantize, or fails to produce a viable model

<details>

```text
Operating System: `Linux-6.14.0-1013-nvidia-aarch64-with-glibc2.39`
Python Version: `3.12.3 (main, Nov  6 2025, 13:44:16) [GCC 13.3.0]`
llm-compressor Version: `0.8.1`
compressed-tensors Version: `0.12.2`
transformers Version: `4.56.2`
torch Version: `2.8.0+cu129`
CUDA Devices: `['NVIDIA GB10']`
AMD Devices: `None`
```

</details>

### 🐛 Describe the bug

Trying to quantize `google/gemma-3-12b-it` with AWQ:

```python
from transformers import AutoModelForCausalLM, AutoTokenizer, AutoProcessor
import os
import torch

# from llmcompressor.modifiers.quantization import GPTQModifier
from llmcompressor.entrypoints import oneshot

MODEL_ID = "google/gemma-3-12b-it"
OUTPUT_DIR = "model_gemma3-12b-it-4bit"

    MODEL_ID,
    device_map="auto",
    torch_dtype="auto",
)

recipe = [
    AWQModifier(ignore=["lm_head"], scheme="W4A16_ASYM", targets=["Linear"]),
]

"""
recipe = AWQModifier(
    ignore=["lm_head"],
    config_groups={
        "group_0": {
            "targets": ["Linear"],
            "weights": {
                "num_bits": 4,
                "type": "int",
                "symmetric": False,
                "strategy": "group",
                # Changed from 128 to 16 to be divisible by 4304
                "group_size": 16,
            },
        }
    },
)
"""

# Calibration dataset - use a representative sample
CALIBRATION_DATASET = "ultrachat-200k"
# quick test
NUM_CALIBRATION_SAMPLES = 16
# full calibration
# NUM_CALIBRATION_SAMPLES = 128
MAX_SEQ_LENGTH = 2048

# Run quantization
oneshot(
    model=model,
    tokenizer=tokenizer,
    dataset=CALIBRATION_DATASET,
    splits="train",
    recipe=recipe,
    max_seq_length=MAX_SEQ_LENGTH,
    preprocessing_num_workers=os.cpu_count(),
    output_dir=OUTPUT_DIR,
)

# Save processor for completeness
processor.save_pretrained(OUTPUT_DIR)

print(f"Quantized model saved to {OUTPUT_DIR}")
```

If I try the first recipe (the one not commented out) then at the very end of the quantization code I get this error:

https://gist.github.com/FlorinAndrei/22a40707756318a5f7e23ec60daf4d2f

If I try the other recipe (the one commented out) it completes the quantization process just fine, but if I try to load the quantized model in MMLU-Pro https://github.com/FlorinAndrei/MMLU-Pro then I get this error:

https://gist.github.com/FlorinAndrei/8c28c26b8f0c5305dbcee5b75b13ac5d

MMLU-Pro has no problem loading the original Gemma 3 or fine-tuned versions of it.

### 🛠️ Steps to reproduce

Just run the code.

### LINKED ISSUE #2365: [Bug]: Gemma3ForCausalLM(Medgemma-27-text-it) model produces garbage output

import torch

# Select calibration dataset
DATASET_ID = "FreedomIntelligence/medical-o1-reasoning-SFT"
DATASET_SPLIT = "train"
language = "en"

# Increasing the number of calib samples to 256 or higher can improve accuracy.
NUM_CALIBRATION_SAMPLES = 256
MAX_SEQUENCE_LENGTH = 4096

ds = load_dataset(DATASET_ID, language, split=f"{DATASET_SPLIT}[:{NUM_CALIBRATION_SAMPLES}]")
ds = ds.shuffle(seed=42)
print(f"Dataset size: {len(ds)} samples")

# Preprocess: apply chat template and tokenize properly
def preprocess(example):
    text = tokenizer.apply_chat_template(
        [{"role": "user", "content": example["Question"]}],
        tokenize=False,
        add_generation_prompt=True,
    )
    return tokenizer(
        text,
        padding=False,
        truncation=True,
    )

ds = ds.map(preprocess, remove_columns=ds.column_names)

# Gemma 3 specific mappings (GQA-aware: skip v_proj -> o_proj due to dimension mismatch)
gemma3_mappings = [
    # 1. Smooth Input Norm against Q, K, V
    AWQMapping(
        smooth_layer="re:.*input_layernorm$",
        balance_layers=["re:.*q_proj$", "re:.*k_proj$", "re:.*v_proj$"]
    ),
    # 2. Smooth the MLP using the pre-feedforward norm
    AWQMapping(
        smooth_layer="re:.*pre_feedforward_layernorm$",
        balance_layers=["re:.*gate_proj$", "re:.*up_proj$"]
    ),
    # 3. Smooth Up-proj against Down-proj
    AWQMapping(
        smooth_layer="re:.*up_proj$",
        balance_layers=["re:.*down_proj$"]
    ),
]

# Initialize the modifier with these mappings
recipe = [
    AWQModifier(
        targets=["Linear"],
        scheme="W4A16_ASYM",
        ignore=["lm_head"],
        mappings=gemma3_mappings,
    )
]

# Run quantization algorithm
oneshot(
    model=model,
    dataset=ds,
    recipe=recipe,
)

### 🐛 Describe the bug

testing time model produce garbage results

### 🛠️ Steps to reproduce

_No response_

Yatimai:
## SUMMARY

Some architectures (Gemma, Gemma2, Gemma3, Qwen3Next) use an offset normalization where the forward computes `output * (1 + weight)` instead of `output * weight`. This breaks any modifier that smooths norm weights (AWQ, SmoothQuant, SpinQuant, QuIP) because dividing a `(1+weight)` parameter by scales produces `1 + weight/scales` instead of `(1 + weight)/scales`.

Following @brian-dellabetta's suggestion, this adds a `norm_calibration_context` that temporarily replaces offset-norm modules with standard-norm equivalents during calibration, following the same pattern as `moe_calibration_context`. On entry, offset norms are replaced with `CalibrationOffsetNorm` modules (`weight = 1 + original`). On exit, modules are restored with updated weights (`weight = smoothed - 1`).

Only norms operating on `hidden_size` are converted. Norms operating on `head_dim` (e.g. `q_norm`/`k_norm` in Gemma3 attention) are skipped since no modifier smooths them.

## TEST PLAN

Unit tests (8/8 passing):
- Weight conversion and dtype preservation
- Forward equivalence with original norm
- Restore roundtrip (with and without smoothing)
- Registry detection (positive and negative)
- `hidden_size` filter: `q_norm`/`k_norm` correctly skipped

E2E validation:

| Model | Modifier | Norms converted | Output |
|---|---|---|---|
| `google/gemma-2-2b-it` | AWQ W4A16 | 105 | Coherent |
| `google/medgemma-27b-text-it` | AWQ W4A16 | 249 (373 total, 124 q/k skipped) | Coherent |
| upstream (no fix) on medgemma | AWQ W4A16 | 0 | Garbage |

Qwen3-Next architecture verified structurally: `hidden_size=2048`, `head_dim=256`, `Qwen3NextRMSNorm` uses same `(1+weight)` pattern. No smaller Qwen3-Next model exists for e2e testing (80B MoE only).

Fixes #2365
Fixes #2102
Related to #2202
Related to #2059


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -986,6 +1000,21 @@ def validate_duo_scaling(cls, v):
         return v
 
 
+_UNIT_OFFSET_NORM_NAMES = {
```
brian-dellabetta: This is also an issue with [Qwen 3 Next](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_next/modeling_qwen3_next.py#L254)
```suggestion
_UNIT_OFFSET_NORM_NAMES = {
    "Qwen3NextRMSNorm",
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -579,11 +579,25 @@ def _smooth(
                         )
                     elif module == smooth_layer:
                         if module.weight.ndim == 1:
-                            update_offload_parameter(
-                                module,
-                                "weight",
-                                module.weight.div_(scales),
-                            )
+                            if _uses_unit_offset_norm(module):
```
brian-dellabetta: This would also be an issue in SmoothQuantModifie, QuipModifier and SpinQuantModifier. Can we move this logic instead to a calibration context so it is generally applicable to any modifier expecting conventional norm forward implementations? See the MoE calibration context [here](https://github.com/vllm-project/llm-compressor/blob/026c917c7eace05b1ebb26e4ddb356c9803bc2a3/src/llmcompressor/modeling/moe_context.py#L31) for an example

brian-dellabetta (commented):
Hi @Yatimai , thanks for taking an initial stab at this, this has been on the docket but haven't had a chance to open a PR yet. I've left some comments, main concern is that this fix is specific to AWQ whereas this custom norm forward implementation would break other modifiers as well. A general solution would be preferred

src/llmcompressor/modifiers/awq/base.py
Yatimai: Will be covered by the calibration context approach.

src/llmcompressor/modifiers/awq/base.py
Yatimai: Makes sense to generalize this as a calibration context rather than patching individual modifiers. I'll follow the MoE calibration context pattern : convert `(1+weight)` norms to standard convention on entry, restore on exit. That covers AWQ, SmoothQuant, SpinQuant, Quip and includes Qwen3Next. Will update the PR.

src/llmcompressor/modifiers/awq/base.py
Yatimai: Covered : `Qwen3NextRMSNorm` is registered as an alias in the `NormCalibrationModule` registry.

src/llmcompressor/modifiers/awq/base.py
Yatimai: Done : implemented as `norm_calibration_context` in `src/llmcompressor/modeling/offset_norm.py`, following the `moe_calibration_context` pattern. Nested inside `moe_calibration_context` in `oneshot.py`. Covers AWQ, SmoothQuant, SpinQuant, QuIP for any model with offset norms.

Yatimai:
Update: reworked the fix based on @brian-dellabetta's review, instead of patching AWQ only, it's now a `norm_calibration_context` that handles the `(1 + weight)` conversion for all modifiers (AWQ, SmoothQuant, SpinQuant, QuIP). Tested e2e on `medgemma-27b-text-it`, output is coherent after AWQ W4A16.

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -217,30 +218,31 @@ def apply_recipe_modifiers(
...
-            )
-            user_pipeline = self.dataset_args.pipeline
-            pipeline = CalibrationPipeline.from_modifiers(
-                session.lifecycle.recipe.modifiers, user=user_pipeline
-            )
-
-            pipeline(
+        # Apply calibration contexts for the entire calibration process
+        with norm_calibration_context(self.model):
+            with moe_calibration_context(
```
brian-dellabetta: You should be able to do this to save some tabs
```suggestion
        with norm_calibration_context(self.model), moe_calibration_context(
```

brian-dellabetta (commented):
Hi @Yatimai , thanks for this, things are looking really good. I have one nit, if we can patch that and I'll run it on my side before approving and adding reviewers

src/llmcompressor/entrypoints/oneshot.py
Yatimai: Hi @brian-dellabetta, pushed.

src/llmcompressor/modeling/offset_norm.py
```diff
@@ -0,0 +1,159 @@
...
+class NormCalibrationModule(ABC, torch.nn.Module, RegistryMixin):
+    """
+    Abstract base class for norm calibration modules.
+
+    Calibration modules replace original norm modules during the calibration
+    phase so that modifiers see standard ``output * weight`` semantics.
+
+    Subclasses must:
+    1. Implement ``__init__()`` with signature: ``(self, original, config)``
+    2. Implement ``restore()`` to convert back to the original norm convention
```
brian-dellabetta: Rather than listing requirements in docstring, you can make these `@abstractmethod`s defined below

tests/llmcompressor/modeling/test_calib_offset_norm.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+        saved = original.weight.data.clone()
+
+        calib = CalibrationOffsetNorm(original, config=None)
+        calib.restore(original)
+
+        assert original.weight.dtype == torch.bfloat16
+        assert torch.allclose(original.weight.data.float(), saved.float(), atol=1e-3)
+
+    def test_restore_after_smoothing(self):
```
brian-dellabetta, line 98: can we also add a unit test here for `norm_calibration_context`, repeating these assertions in the context, and does the assertions outside of the context to ensure they've restored back successfully?

src/llmcompressor/modeling/offset_norm.py
```diff
@@ -0,0 +1,159 @@
...
+            run_calibration(model)
+        # Norms restored to offset convention with smoothed weights
+    """
+
+    replaced = {}
+
+    # Step 1: Collect all offset-norm modules that need replacement
+    logger.debug("Entering norm calibration context")
+    modules_to_replace = []
+    hidden_size = getattr(model.config, "hidden_size", None)
```
brian-dellabetta: sorry, i missed this in first pass -- can we remove this restriction? I don't think there's any harm in setting them all to be the same in the context, as long as they are restored after context. This saves some config checking and i'm not sure this is guaranteed in cases like spinquant, see fuse norms logic [here](https://github.com/vllm-project/llm-compressor/blob/ed2d7f6093fade12a80a23eeb4dc4dc2292ef4a7/src/llmcompressor/modifiers/transform/spinquant/base.py#L201-L209).

tests/llmcompressor/modeling/test_calib_offset_norm.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def test_standard_norm_not_detected(self):
+        """Standard LayerNorm should not be in the registry."""
+        registered = NormCalibrationModule.registered_names()
+        assert "layernorm" not in registered
+        assert "rmsnorm" not in registered
+
+
+@pytest.mark.unit
+class TestHiddenSizeFilter:
```
brian-dellabetta: if we are in agreement with message above, we can prune this test

brian-dellabetta (commented):
Hi @Yatimai , sorry a few more nits after 2nd pass. Will try this out now

src/llmcompressor/modeling/offset_norm.py
Yatimai: Will do.

tests/llmcompressor/modeling/test_calib_offset_norm.py
Yatimai: Will add.

src/llmcompressor/modeling/offset_norm.py
Yatimai: Retested on medgemma-27b-text-it converting all 373 norms (no filter) : output is coherent after storing the converted weight in the original dtype (bfloat16) instead of float32. Removing the filter.

tests/llmcompressor/modeling/test_calib_offset_norm.py
Yatimai: Will remove with the filter.

brian-dellabetta (approved):
Thanks @Yatimai  for the contribution! This looks great to me. We may ultimately want to merge the moe and norm calibration contexts, as they share the same interface, but we can leave that off for now. Approving and will ping team for a second reviewer. 

I was able to run AWQ on `google/gemma-2-2b-it` on this branch and get reasonable output, whereas on main it's gibberish

Yatimai:
Thanks @brian-dellabetta for your help!

Jeevi10:
hey,
I ran an experiment with GPTQ and AWQ for google/medgemma-27b-text-it

google/medgemma-27b-text-it 
|  Tasks  |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|---------|------:|------|-----:|------|---|----:|---|-----:|
|careqa_en|      1|none  |     5|acc   |↑  |0.856|±  |0.0222|

I used W4A16 for both GPTQ and AWQ

AWQ result's in fix/awq-gemma-rmsnorm in this branch 
|  Tasks  |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|---------|------:|------|-----:|------|---|----:|---|-----:|
|careqa_en|      1|none  |     5|acc   |↑  |0.692|±  |0.0293|

GPTQ results in main branch 
|  Tasks  |Version|Filter|n-shot|Metric|   |Value|   |Stderr|
|---------|------:|------|-----:|------|---|----:|---|-----:|
|careqa_en|      1|none  |     5|acc   |↑  |0.836|±  |0.0235|


@Yatimai Please let me know if you want me to do more testing, but its producing coherent output @fix/awq-gemma-rmsnorm.

Yatimai:
Thanks @Jeevi10 for testing, appreciate the validation!

HDCharles (approved):

PR #2502, opened 2026-03-23
[Docs] Add Qwen3.5 to Key Models

dsikka:
SUMMARY:
- Example walkthrough 
- Notes about transformers version
- Should land https://github.com/vllm-project/llm-compressor/pulls/dsikka to make these examples make sense


docs/key-models/qwen3.5/nvfp4-moe-example.md
```diff
@@ -0,0 +1,123 @@
...
+    dataset=ds,
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+    moe_calibrate_all_experts=True,
+    data_collator=data_collator,
+)
+```
+
+### 5. Save to Disk in Compressed-Tensors Format
+
```
kylesayrs, line 114: Consider adding an explanation for the resaving

dsikka: There's a comment before the last function call - do you think more detail is required?

dsikka:
@claude review

docs/key-models/qwen3.5/nvfp4-moe-example.md
kylesayrs: I still don't really understand why MTP layers need to be re-saved separately? I assume this is because they're not loaded as part of `Qwen3_5MoeForConditionalGeneration`?

kylesayrs: It's my opinion that, if I'm guessing, then it's not going to be obvious to the average user.

dsikka: Yes there is a comment already that says this exactly on lines 120-121

I am not sure why you’re confused

kylesayrs: Neither am I

dsikka:
@claude hello - update this PR

PR #2511, opened 2026-03-24
[AWQ] Restructure AWQModifier as smoothing-only, decouple from Quanti…

### LINKED ISSUE #2327: Restructure and simplify the AWQModifier similar to SmoothQuant

colldata79:
…zationMixin

Remove QuantizationMixin inheritance from AWQModifier so it becomes a pre-quantization transform (like SmoothQuant) that only applies smoothing scales. Final quantization is now handled by a downstream quantizer (QuantizationModifier, GPTQModifier) stacked after AWQ in the recipe.

Key changes:
- Drop QuantizationMixin inheritance, keep quant config locally for grid search pseudo-quantization only
- Add _temporary_quant_schemes context manager that snapshots/restores all quant-related module state (schemes, observers, scales, zero-points, forward overrides) with full exception safety
- Decompose _compute_best_scale into pure helpers: _collect_activation_stats, _generate_scale_candidates, _evaluate_candidate, _select_best_scale, _apply_best_scales
- Add recipe validation: errors on mismatched scheme config, warns on missing downstream quantizer or reversed ordering
- Remove all quantization lifecycle calls from on_end (no more update_weight_zp_scale, update_weight_global_scale, end_calibration)
- Update all examples and e2e recipes to stacked pattern [AWQModifier, QuantizationModifier]

Closes #2327

SUMMARY:

TEST PLAN:

### DIFF
--- a/examples/awq/fp8_block_llama_example.py
+++ b/examples/awq/fp8_block_llama_example.py
@@ -51,11 +51,12 @@ def tokenize(sample):
 
 # Configure the quantization algorithm to run.
 # AWQModifier applies smoothing, then QuantizationModifier finalizes quantization.
+_ignore = ["lm_head"]
+_scheme = "FP8_BLOCK"
+_targets = ["Linear"]
 recipe = [
-    AWQModifier(
-        ignore=["lm_head"], scheme="FP8_BLOCK", targets=["Linear"], duo_scaling="both"
-    ),
-    QuantizationModifier(ignore=["lm_head"], scheme="FP8_BLOCK", targets=["Linear"]),
+    AWQModifier(ignore=_ignore, scheme=_scheme, targets=_targets, duo_scaling="both"),
+    QuantizationModifier(ignore=_ignore, scheme=_scheme, targets=_targets),
 ]
 
 # Apply algorithms.
--- a/examples/awq/fp8_block_llama_example.py
+++ b/examples/awq/fp8_block_llama_example.py
@@ -4,6 +4,7 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
 from llmcompressor.modifiers.awq import AWQModifier
+from llmcompressor.modifiers.quantization import QuantizationModifier
 
 # Select model and load it.
 MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
@@ -49,10 +50,12 @@ def tokenize(sample):
 
 
 # Configure the quantization algorithm to run.
+# AWQModifier applies smoothing, then QuantizationModifier finalizes quantization.
 recipe = [
     AWQModifier(
         ignore=["lm_head"], scheme="FP8_BLOCK", targets=["Linear"], duo_scaling="both"
     ),
+    QuantizationModifier(ignore=["lm_head"], scheme="FP8_BLOCK", targets=["Linear"]),
 ]
 
 # Apply algorithms.
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -1,13 +1,21 @@
 import inspect
+from contextlib import contextmanager
+from dataclasses import dataclass, field
 from itertools import product
-from typing import Iterator, Literal
+from typing import Any, Dict, Iterator, List, Literal, Optional, Set, Union
 
 import torch
 from compressed_tensors.offload.dist_utils import as_broadcastable, is_distributed
 from compressed_tensors.quantization import (
+    QuantizationConfig,
+    QuantizationScheme,
+    QuantizationStatus,
     QuantizationStrategy,
+    apply_quantization_config,
     disable_quantization,
     forward_quantize,
+    is_preset_scheme,
+    preset_name_to_scheme,
 )
 from compressed_tensors.utils import (
     align_modules,
@@ -20,7 +28,7 @@ from compressed_tensors.utils import (
     update_offload_parameter,
 )
 from loguru import logger
-from pydantic import ConfigDict, PrivateAttr, field_validator
+from pydantic import ConfigDict, Field, PrivateAttr, field_validator
 from torch import distributed as dist
 from torch.nn import Module
 from torch.utils._pytree import tree_leaves
@@ -35,10 +43,9 @@ from llmcompressor.modifiers.awq.mappings import (
 )
 from llmcompressor.modifiers.quantization.calibration import (
     call_observer,
-    update_weight_global_scale,
-    update_weight_zp_scale,
+    initialize_observer,
+    reset_quantization_status,
 )
-from llmcompressor.modifiers.quantization.quantization import QuantizationMixin
 from llmcompressor.modifiers.utils import update_fused_layer_weight_global_scales
 from llmcompressor.modifiers.utils.hooks import HooksMixin
 from llmcompressor.modifiers.utils.pytorch_helpers import is_moe_model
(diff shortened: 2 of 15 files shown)


examples/awq/fp8_block_llama_example.py
```diff
@@ -49,10 +50,13 @@ def tokenize(sample):
...
+# AWQModifier applies smoothing, then QuantizationModifier finalizes quantization.
+_ignore = ["lm_head"]
+_scheme = "FP8_BLOCK"
+_targets = ["Linear"]
 recipe = [
-    AWQModifier(
-        ignore=["lm_head"], scheme="FP8_BLOCK", targets=["Linear"], duo_scaling="both"
-    ),
+    AWQModifier(ignore=_ignore, scheme=_scheme, targets=_targets, duo_scaling="both"),
+    QuantizationModifier(ignore=_ignore, scheme=_scheme, targets=_targets),
```
brian-dellabetta, line 59: Rather than requiring user to set this explicitly, and to retain backward compatibility, i propose we append this when the recipe is parsed. if a user provides AWQ without a follow-on modifier that quantizes, we should append it with the same ignore,scheme,targets,config_groups

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -53,25 +60,32 @@
 
 __all__ = ["AWQModifier"]
 
+_MISSING = object()  # sentinel for snapshot/restore of temp quant schemes
 
-class AWQModifier(Modifier, QuantizationMixin):
+
+@dataclass
+class AWQSearchResult:
```
brian-dellabetta, line 67: there's quite a bit in here that is beyond the scope of this PR.  Please prune, there are already a lot of changes in this file

Image

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -84,128 +98,459 @@ class AWQModifier(Modifier, QuantizationMixin):
...
     _parent_args_cache: dict[Module, IntermediatesCache] = PrivateAttr(
         default_factory=dict
     )
     # Dict[smooth layer name, [activation sums, activation counts]]
     _smooth_activation_stats: dict[str, list[torch.Tensor]] = PrivateAttr(
         default_factory=dict
     )
-    # List to store error metrics for each layer
     _error_metrics: list[dict] = PrivateAttr(default_factory=list)
+    _fp16_baseline_cache: dict[Module, IntermediatesCache] = PrivateAttr(
```
brian-dellabetta, line 173: what is this? i don't see it used anywhere

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -84,128 +98,459 @@ class AWQModifier(Modifier, QuantizationMixin):
...
+            quantization_status=QuantizationStatus.INITIALIZED,
+            ignore=ignore,
+        )
+
+    # ------------------------------------------------------------------ #
+    #  Temporary quant scheme management (context-managed)               #
+    # ------------------------------------------------------------------ #
+
+    @contextmanager
+    def _temporary_quant_schemes(self, model: Module, with_observers: bool = False):
```
brian-dellabetta, line 276: rather than maintaining this, we are better off applying quantization config on start and removing it on end, as I show in my PR [here](https://github.com/vllm-project/llm-compressor/pull/2206/changes#diff-a79ed86c2a25b361e101f22c2cdcc549cf2dbc09b48de91c752f5edc4cf925a7R314-R317)

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -84,128 +98,459 @@ class AWQModifier(Modifier, QuantizationMixin):
...
+                # Restore prior forward override
+                module.__dict__["forward"] = prev_forward
+
+        self._temp_quant_snapshot.clear()
+
+    # ------------------------------------------------------------------ #
+    #  Recipe validation                                                  #
+    # ------------------------------------------------------------------ #
+
+    def _validate_recipe(self, state: State):
```
brian-dellabetta, line 395: recipe validation would need to occur higher up in scope, like during oneshot

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -476,8 +807,6 @@ def cache_smooth_activations_hook(
             return cache_smooth_activations_hook
 
         for mapping in self._resolved_mappings:
-            # parent kwargs needed for future forward passes
-            # same parent may appear multiple times in resolved mappings
```
brian-dellabetta, line 480: why remove these comments?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -914,11 +1311,7 @@ def _compute_layer_means(layers: list[Module]) -> torch.Tensor:
         """
         Compute per-channel/group/block/tensor mean of normalised weights
         for all passed in layers taking into account the quantization_scheme.
-
-        To minimize memory requirements, layers are reduced to a running total
-            of sums and counts when calculating mean
```
brian-dellabetta, line 919: why are lines here and below removed?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -84,128 +98,459 @@ class AWQModifier(Modifier, QuantizationMixin):
...
+
+    def _has_quant_config(self) -> bool:
+        return not (
+            self.config_groups is None
+            and self.targets == ["Linear"]
+            and self.ignore == []
+            and self.scheme is None
+        )
+
+    def _resolve_quantization_config(self) -> QuantizationConfig:
```
brian-dellabetta, line 238: having to copy this over means to me that we need another abstraction that can provide this functionality to both AWQModifier and QuantizationMixin classes, which can subclass it.

brian-dellabetta (commented):
Hi @colldata79 , thanks for taking a stab at this. Please see comments below. This is proving to be a trickier thing to implement, I will message you over slack to discuss how to proceed

brian-dellabetta:
We have merged this refactor in #2526

PR #2525, opened 2026-03-26
feat: add ActivationOrdering support for per-channel GPTQ quantization

### LINKED ISSUE #2524: GPTQ ActivationOrdering support for per-channel quantization
there's no reason for this to be disabled except for groupwise quantization as far as i can tell, it should be straightforward to implement this (also maybe per-tensor as well?)

https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L114-L138

1) add support
2) test that it works

matdou:
Fixes #2524

GPTQ's `actorder` parameter was being silently ignored for per-channel quantization. If you set `actorder=weight` with `strategy=channel`, nothing would happen. This PR wires it up properly.

Changes:
- `resolve_quantization_config` now applies actorder resolution for `CHANNEL` strategy, not just `GROUP`
- `quantize_weight` has a new `CHANNEL` branch: `WEIGHT` ordering works as expected, and `GROUP` ordering raises a clear `ValueError` (because it doesn't make sense without groups)
- Added `test_actorder_resolution_channel` and channel cases in `test_config_resolution`

SUMMARY:
`actorder` was silently ignored for per-channel quantization. This PR extends support to `CHANNEL` strategy. `WEIGHT` ordering now works correctly, and `GROUP` ordering raises a clear `ValueError`.

TEST PLAN:
30 tests passing before changes, 42 after (12 new). New tests cover: `channel + actorder=weight` succeeds, `channel + actorder=group` raises `ValueError`, and channel cases in `test_config_resolution`. Run locally on macOS (M2, Python 3.14, CPU only).

NOTE:
Per-tensor support (`strategy=tensor`) could be a natural follow-up.

### DIFF
--- a/src/llmcompressor/modifiers/gptq/gptq_quantize.py
+++ b/src/llmcompressor/modifiers/gptq/gptq_quantize.py
@@ -136,6 +136,19 @@ def quantize_weight(
 
         else:
             scale, zero_point = observer(W)
+    elif strategy == QuantizationStrategy.CHANNEL:
+        if actorder == ActivationOrdering.GROUP:
+            raise ValueError(
+                "ActivationOrdering.GROUP is not supported for per-channel "
+                "quantization (strategy=CHANNEL). Use ActivationOrdering.WEIGHT "
+                "instead."
+            )
+        elif actorder == ActivationOrdering.WEIGHT:
+            # compute per-channel scale first, then permute columns
+            scale, zero_point = observer(W)
+            W, H, perm = _apply_activation_ordering(W, H)
+        else:
+            scale, zero_point = observer(W)
     else:
         scale, zero_point = observer(W)
 
@@ -268,6 +281,11 @@ def quantize_weight(
 
             # only save g_idx if mapping is not identity
             has_gidx = True
+    elif strategy == QuantizationStrategy.CHANNEL:
+        if actorder == ActivationOrdering.WEIGHT:
+            # restore original permutation
+            invperm = torch.argsort(perm)
+            W = W[:, invperm]
 
     if not has_gidx:
         g_idx = None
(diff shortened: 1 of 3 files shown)


matdou:
PR is ready for review (tests passing locally). Let me know if anything should be adjusted. Could someone add the ready label?

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -136,6 +136,19 @@ def quantize_weight(
 
         else:
             scale, zero_point = observer(W)
+    elif strategy == QuantizationStrategy.CHANNEL:
+        if actorder == ActivationOrdering.GROUP:
+            raise ValueError(
+                "ActivationOrdering.GROUP is not supported for per-channel "
+                "quantization (strategy=CHANNEL). Use ActivationOrdering.WEIGHT "
```
HDCharles: or disable ActivationOrdering

src/llmcompressor/modifiers/gptq/gptq_quantize.py
HDCharles: this is the same as what you have, all weight activation ordering should be handled at the same time

HDCharles: this is the same as what you have, lets unify them if possible, g_idx can be specific to group quantization, the permutation of W,H is the same for all activation ordering.

HDCharles:
this is mildly out of date, see the most recent, i cleaned up a bit of this code

tests/llmcompressor/modifiers/quantization/test_base.py
```diff
@@ -109,12 +109,60 @@ def test_actorder_resolution(
         assert resolved.config_groups["group_1"].weights.actorder == expected_1
 
 
+_CHANNEL_Q_CONFIG_KWARGS = dict(
```
HDCharles, line 112: lets get rid of the test changes here and do it here instead:

tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py

add a test for actorder + per-channel quant

matdou:
Closing as the fix was implemented in #2670. Thanks for the review.

PR #2526, opened 2026-03-26
[AWQ] AWQ as transform

brian-dellabetta:
SUMMARY:
This PR refactors `AWQModifier` to be a transform, i.e. a modifier that changes the values of weights without altering qparams or quantization schemes. 

AWQModifier now needs to be run in conjunction with QuantizationModifier or another modifier that subclasses QuantizationMixin. 

This is to be viewed in contrast to #2206, which updates the modifier lifecycle to apply quant config on start, and remove it on end.

- [x] No longer subclasses QuantizationMixin. Quantization Config is applied on_init by additional modifier in Recipe, so it will already be set by the time AWQ starts.
- [x] Add recipe validation to make sure AWQ/SmoothQuant are run with a quantization modifier
- [x] Move to `src/llmcompressor/modifiers/transform`, deprecate old location with warning explaining API change.
- [x] Retain `llmcompressor.modifiers.awq.AWQModifier` for backward-compatibility, along with updates to recipe parsing logic for nested lists of modifiers that this can cause.
- [x] Port incompatible layer logging cleanup from #2206  
- [x] e2e/lm_eval tests -- https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24687948562. All green except for lm-eval failures for `tests/lmeval/configs/w4a16_actorder_group.yaml`, `tests/lmeval/configs/w4a16_actorder_none.yaml`, `tests/lmeval/configs/w4a16_grouped_quant.yaml`, seen previously and unrelated to AWQ

TEST PLAN
- [x] AWQ + NVFP4 improves over what's currently on main, in wikitext PPL and I can reproduce the improvements in gsm8k_eval.py first reported in #2206 
- [x] Negligible difference in scores for AWQ + W4A16, as expected

lm_eval gsm8k results are inconsistent, but this is part of an issue being explored in #2606. gsm8k_platinum has a narrower gap

| Scheme | Commit | Wikitext PPL | vllm gsm8k_eval.py | gsm8k strict | gsm8k_platinum strict |
|---------|--------|---------|---|------|-------|
| NVFP4 | this | 11.502 | .672 | .5694 | .5633 |
| NVFP4 | main | 11.5210 | .655 | .5762 | .5666 |
| W4A16 | this | 10.6120 | .672 | .5785 | .6096 |
| W4A16 | main | 10.6103 | .673 | .5800 | .5972 |


src/llmcompressor/recipe/recipe.py
```diff
@@ -67,12 +67,23 @@ def from_modifiers(
...
+        # backward-compatibility for llmcompressor.modifiers.awq.AWQModifier, which
+        # returns [AWQTransformModifier, QuantizationModifier]
+        def flatten(nested):
+            for item in nested:
+                if isinstance(item, list):
+                    yield from flatten(item)
+                else:
+                    yield item
+
+        modifiers = list(flatten(modifiers))
```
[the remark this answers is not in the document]
brian-dellabetta: thanks, resolved

examples/awq/README.md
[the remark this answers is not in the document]
brian-dellabetta: updated, thanks

src/llmcompressor/modifiers/awq/__init__.py
```diff
@@ -1,5 +1,75 @@
 # ruff: noqa
+"""
+Backwards compatibility shim for SmoothQuantModifier.
 
-from .mappings import *
-from .dynamic_mappings import *
-from .base import *
+This module has been moved to llmcompressor.modifiers.transform.smoothquant.
+This shim will be removed in a future version.
```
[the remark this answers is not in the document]
brian-dellabetta: fixed

src/llmcompressor/modifiers/awq/__init__.py
```diff
@@ -1,5 +1,79 @@
...
+    "AWQModifier",
+    "AWQMapping",
+    "AWQ_MAPPING_REGISTRY",
+    "AWQ_DYNAMIC_MAPPING_REGISTRY",
+    "get_layer_mappings_from_model",
+    "default_mappings",
+]
+
+
+def AWQModifier(**kwargs):
```
kylesayrs, line 60: It seems like bad practice to silently change types from `Type` to `Function` (and from `Modifier` to `list[Modifier]`, this makes things hard to debug.

Would it be better to instead spawn a `QuantizationModifier` instance when the recipe is ingested?

src/llmcompressor/modifiers/factory.py
```diff
@@ -46,6 +46,7 @@ def load_from_package(package_path: str) -> dict[str, type[Modifier]]:
         # exclude deprecated packages from registry so
         # their new location is used instead
         deprecated_packages = [
+            "llmcompressor.modifiers.awq",
```
kylesayrs, line 49: Maybe we can keep the original AWQ path and use case, but instead spawn a QuantizationModifier? This way we don't break any existing scripts, and we still get to maintain `llmcompressor.modifiers.awq`.

`awq/base.py`
```python
AWQModifier = LegacyAWQModifier  # export under original alias

class LegacyAWQModifier(AWQModifier, QuantizationMixin):
  def split(self) -> tuple[AWQModifier, QuantizationMixin]:
    # use QuantizationMixin.model_fields.keys()
    awq_args = {key: self.model_fields[key] for key in AWQModifier.model_fields.keys()}
    qmod_args = {key: self.model_fields[key] for key in AWQModifier.model_fields.keys()}
    return (AWQModifier(**awq_args), QuantizationMixin(**qmod_args))
```

`oneshot.py`
```python
if any(isinstance(mod, LegacyAWQModifier) for mod in recipe.modifiers):
  ...
```

src/llmcompressor/recipe/recipe.py
```diff
@@ -34,17 +34,17 @@ class Recipe(BaseModel):
...
+    modifiers: list[Modifier] = Field(default_factory=list)
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     @classmethod
     def from_modifiers(
         cls,
-        modifiers: Union[Modifier, List[Modifier]],
-        modifier_group_name: Optional[str] = None,
+        modifiers: Modifier | list[Modifier | list[Modifier]],
```
kylesayrs, line 47: I really don't like this expansion of acceptable types. Also mitigated by the proposed design

kylesayrs (approved):
Spoke offline, this design works

src/llmcompressor/modifiers/awq/__init__.py
```diff
@@ -1,5 +1,79 @@
...
+        "weight_observer",
+        "input_observer",
+        "output_observer",
+        "observer",
+        "bypass_divisibility_checks",
+    )
+    quant_kwargs = {k: v for k, v in kwargs.items() if k in quant_keys}
+    awq_kwargs = {k: v for k, v in kwargs.items() if k not in quant_keys}
+
+    return [AWQTransformModifier(**awq_kwargs), QuantizationModifier(**quant_kwargs)]
```
HDCharles, line 79: Does t this make AWQ ->GPTQ impossible?

src/llmcompressor/recipe/recipe.py
```diff
@@ -67,26 +67,37 @@ def from_modifiers(
         """
         logger.info("Creating recipe from modifiers")
 
-        if isinstance(modifiers, Modifier):
-            modifiers = [modifiers]
+        # workaround for modifiers that can be nested lists of Modifiers
+        # backward-compatibility for llmcompressor.modifiers.awq.AWQModifier, which
+        # returns [AWQTransformModifier, QuantizationModifier]
+        def flatten(nested):
```
HDCharles: Probably cleaner to use pytree leaves than implement a new one

https://github.com/vllm-project/llm-compressor/blob/d44643bdb5966351cee0ea9a41fdaf6a588c1977/src/llmcompressor/modifiers/awq/base.py#L363

HDCharles (approved):
Looks good, see comments

src/llmcompressor/modifiers/awq/__init__.py
brian-dellabetta: In the new API with `llmcompressor.modifiers.transform.AWQModifier`, you just stack them together. If someone had `llmcompressor.modifiers.awq.AWQModifier` followed by `GPTQModifier`, it will still work in the same way as before, by resolving to the stack. see changes to [examples/quantization_non_uniform/quantization_multiple_modifiers.py](https://github.com/vllm-project/llm-compressor/pull/2526/changes#diff-6b164b62eabbffac12dce3bf78a160a40eab9df0c481a98bdc19cb8113482d7f)

src/llmcompressor/modifiers/awq/__init__.py
brian-dellabetta: we discussed offline, marking resolved

src/llmcompressor/modifiers/factory.py
brian-dellabetta: we discussed offline, marking resolved

src/llmcompressor/recipe/recipe.py
brian-dellabetta: we discussed offline, marking resolved

src/llmcompressor/recipe/recipe.py
brian-dellabetta: Updated, thanks

src/llmcompressor/modifiers/awq/__init__.py
brian-dellabetta: validated that example works with independent and sequential pipelines

PR #2531, opened 2026-03-27
[bugfix][AWQ] include smooth layer when resolving parent

### LINKED ISSUE #2522: [Bug]: Evaluate AWQ and GPTQ for Gemma
Operating System: `Linux-6.8.0-100-generic-x86_64-with-glibc2.39`
Python Version: `3.12.11 | packaged by Anaconda, Inc. | (main, Jun  5 2025, 13:09:17) [GCC 11.2.0]`
llm-compressor Version: `0.10.0.1`
compressed-tensors Version: `0.14.0.1`
transformers Version: `4.57.6`
torch Version: `2.10.0`
CUDA Devices: `['NVIDIA A100-SXM4-80GB', 'NVIDIA A100-SXM4-80GB']`
AMD Devices: `None`
NPU Devices: `None`

### 🐛 Describe the bug

AWQ W4A16 quantization via llm-compressor produces catastrophically degraded output quality on Gemma 3-based models (tested on `google/medgemma-27b-it`). Accuracy drops from **85.6%** (FP16 baseline) to **69.2%** on the CareQA medical benchmark — a ~16 point degradation. In contrast, GPTQ W4A16 on the same model retains **83.6%**, demonstrating that the issue is specific to AWQ's scaling mechanism, not a general quantization sensitivity.

https://github.com/vllm-project/llm-compressor/pull/2500#issuecomment-4128173409

### 🛠️ Steps to reproduce

```python
import torch

# Select calibration dataset
DATASET_ID = "FreedomIntelligence/medical-o1-reasoning-SFT"
DATASET_SPLIT = "train"
language = "en"



# Select model and load it.
MODEL_ID = "google/medgemma-27b-text-it"

# Increasing the number of calib samples to 256 or higher can improve accuracy.
NUM_CALIBRATION_SAMPLES = 512
MAX_SEQUENCE_LENGTH = 2048

ds = load_dataset(DATASET_ID, language, split=f"{DATASET_SPLIT}[:{NUM_CALIBRATION_SAMPLES}]")
ds = ds.shuffle(seed=42)
print(f"Dataset size: {len(ds)} samples")


# Preprocess: apply chat template and tokenize properly
def preprocess(example):
    text = tokenizer.apply_chat_template(
        [{"role": "user", "content": example["Question"]}],
        tokenize=False,
        #add_generation_prompt=True,
    )
    return tokenizer(
        text,
        padding=False,
        truncation=True,
        add_special_tokens=False,
    )


ds = ds.map(preprocess, remove_columns=ds.column_names)

# Gemma 3 specific mappings (GQA-aware: skip v_proj -> o_proj due to dimension mismatch)
gemma3_mappings = [
    # 1. Smooth Input Norm against Q, K, V
    AWQMapping(
        smooth_layer="re:.*input_layernorm$",
        balance_layers=["re:.*q_proj$", "re:.*k_proj$", "re:.*v_proj$"]
    ),


    AWQMapping("re:.*v_proj$", ["re:.*o_proj$"]),
    # 2. Smooth the MLP using the pre-feedforward norm
    AWQMapping(
        smooth_layer="re:.*pre_feedforward_layernorm$",
        balance_layers=["re:.*gate_proj$", "re:.*up_proj$"]
    ),
    # 3. Smooth Up-proj against Down-proj
    AWQMapping(
        smooth_layer="re:.*up_proj$",
        balance_layers=["re:.*down_proj$"]
    ),
]

# Initialize the modifier with these mappings
recipe = [
    AWQModifier(
        targets=["Linear"],
        scheme="W4A16_ASYM",
        ignore=["lm_head"],
        #mappings=gemma3_mappings,
        duo_scaling="both"
    )
]

"""
     targets="Linear",
     scheme="FP8_DYNAMIC",
     ignore=["lm_head"],
     duo_scaling="both"
 )
"""


#Run quantization algorithm
oneshot(
    model=model,
    dataset=ds,
    recipe=recipe,
)
```

brian-dellabetta:
SUMMARY:

When finding the parent for a given mapping, we previously resolved the parent to be the lowest common ancestor that contained all the balance layers for the mapping. This excludes the smooth layer, causing the parent to be resolved incorrectly for the case of `Gemma3ForCausalLM` mapping 
```
AWQMapping(
        "re:.*pre_feedforward_layernorm$",
        ["re:.*gate_proj$", "re:.*up_proj$"],
    ),
```

On main, the parent is resolved to be the MLP layer for this mapping, but the pre_feedforward_layernorm lives outside the MLP (see [here](https://github.com/huggingface/transformers/blob/v5.4.0/src/transformers/models/gemma3/modeling_gemma3.py#L429-L430)). This is unintended, and is causing the regression reported in #2522 

With these changes, the parent is resolved to be the entire Decoder layer.

Resolves #2522 

TEST PLAN:
- [x] unit tests updated
- [ ] Confirm with user this branch resolves their issue #2522

### DIFF
--- a/tests/llmcompressor/modifiers/awq/test_base.py
+++ b/tests/llmcompressor/modifiers/awq/test_base.py
@@ -96,17 +96,18 @@ def test_set_resolved_mappings():
                 self_attn.k_proj,
                 self_attn.v_proj,
             }
-            assert mapping.parent_name == "decoder.self_attn"
-            assert mapping.parent == self_attn
+            assert mapping.parent_name == "decoder"
+            assert mapping.parent == model["decoder"]
         if "self_attn.v_proj" in mapping.smooth_name:
             assert set(mapping.balance_names) == {"decoder.self_attn.o_proj"}
-            assert mapping.parent_name == "decoder.self_attn.o_proj"
+            assert mapping.parent_name == "decoder.self_attn"
         if "mlp.experts" in mapping.smooth_name and "up_proj" in mapping.smooth_name:
             expert_idx = mapping.smooth_name.split(".")[-2]
-            expected_down_proj = f"decoder.mlp.experts.{expert_idx}.down_proj"
-            assert set(mapping.balance_names) == {expected_down_proj}
-            assert mapping.parent_name == expected_down_proj
-            assert mapping.parent == mlp["experts"][int(expert_idx)]["down_proj"]
+            assert set(mapping.balance_names) == {
+                f"decoder.mlp.experts.{expert_idx}.down_proj"
+            }
+            assert mapping.parent_name == f"decoder.mlp.experts.{expert_idx}"
+            assert mapping.parent == mlp["experts"][int(expert_idx)]
 
     awq = AWQModifier(
         mappings=[


brian-dellabetta:
This isn't needed, the parent only needs to contain the balance layers. The smooth layer activations are scaled as part of the forward passes through the balance layer parent. Considerations for quantizing the smooth layer are covered already in a separate PR:
* #2323

Issue #2520, opened 2026-03-25
[GPTQ] Add support for block-wise quantization


brian-dellabetta:
**Is your feature request related to a problem? Please describe.**
GPTQ currently does not support `QuantizationStrategy.BLOCK`. This is most likely because it predates the addition of block-wise quantization strategy to compressed-tensors/llm-compressor and its continued usage in DeepSeek models. I don't see any reason why it would be incompatible with GPTQ, and I'd like to see how it would GPTQ would perform vs. round-to-nearest on a DeepSeek3.2 NVFP4+FP8_BLOCK model

**Describe the solution you'd like**
Add case for block strategy to GPTQ algorithm [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L196-L232)

**Additional context**
PR should include some basic benchmarks comparing GPTQ FP8_BLOCK to round-to-nearest, hopefully showing advantage

zeel2104:
I would like to work on it. Could you assign it to me?

brian-dellabetta:
You got it @zeel2104 !

dsikka:
Tracked in JIRA: https://issues.redhat.com/browse/INFERENG-5601

**JIRA Details:**
- Issue Type: Task
- Priority: Normal
- Component: LLM Compressor
- Team: INFERENG Model Optimizations
- Status: Backlog
- Labels: good-first-issue

HDCharles:
if you're interested, there's a small followup here to add weight activation ordering. 

The only change would be
g_idx = range / groupsize -> g_idx = range / block_width

this would actually consolidate the existing code because as of now we have 2 ways to map columns, g_idx and [this line](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L221) which is doing the same thing as g_idx but as a one off. If we change it to use g_idx then we get actorder=weight for free

PR #2533, opened 2026-03-28
Add GPTQ support for block quantization

zeel2104:
## Summary

Adds GPTQ support for `QuantizationStrategy.BLOCK`.

Previously, GPTQ only handled tensor, channel, group, and tensor-group strategies in the weight quantization loop. This change adds block-wise handling by selecting the correct block-column quantization parameters for each GPTQ column update, while keeping the existing Hessian-based error propagation flow unchanged.

## Changes Made

- added `QuantizationStrategy.BLOCK` support in `src/llmcompressor/modifiers/gptq/gptq_quantize.py`
- quantized each GPTQ column as a 2D block slice using the matching block qparams
- added a focused unit test in `tests/llmcompressor/modifiers/gptq/test_gptq_quantize.py`
- verified existing GPTQ quantization config parsing tests still pass

## Test Plan

Tested locally in the repo development environment.

Commands run:

```bash
$env:PYTHONPATH="src"
pytest tests\llmcompressor\modifiers\gptq\test_gptq_quantize.py tests\llmcompressor\modifiers\quantization\test_base.py -q

### DIFF
--- a/src/llmcompressor/entrypoints/model_free/__init__.py
+++ b/src/llmcompressor/entrypoints/model_free/__init__.py
@@ -21,7 +21,7 @@ from llmcompressor.entrypoints.model_free.helpers import (
     gpu_if_available,
 )
 from llmcompressor.entrypoints.model_free.microscale import (
-    build_inverse_weights_map,
+    build_microscale_inverse_weights_map,
     is_microscale_scheme,
 )
 from llmcompressor.entrypoints.model_free.process import (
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by zeel2104** [comment]
This PR is code complete and ready for review.

Local validation run:
```bash
$env:PYTHONPATH="src"
pytest -p no:inline_snapshot tests\llmcompressor\modifiers\gptq\test_gptq_quantize.py tests\llmcompressor\modifiers\quantization\test_base.py -q

File: src/llmcompressor/modifiers/gptq/gptq_quantize.py [hunk 1]
```diff
@@ -174,6 +174,11 @@ def quantize_weight(
         Hinv = H = torch.eye(num_columns, dtype=H.dtype, device=H.device)
 
     # See section 3.4 of https://arxiv.org/abs/2203.07259
+    block_width = (
```
dsikka: nit: move down to line 235 so we can use it when we're sure this is the correct strategy

brian-dellabetta: yeah, since we're only addressing `quant_args.block_structure[1]`, it's fine to have in the scope. I think you moved it based on gemini's suggestion, but your original location is better for readability

zeel2104:
Updated per review feedback to keep `block_width` scoped inside the `QuantizationStrategy.BLOCK` branch for readability. Re-ran the targeted local tests successfully.

@brian-dellabetta 
I don’t currently have access to suitable GPU hardware in this environment for a meaningful FP8_BLOCK end-to-end benchmark.

If there’s a recommended internal machine, cluster, or standard benchmark command I should use, I’m happy to run it if I can get access.

brian-dellabetta:
> @brian-dellabetta I don’t currently have access to suitable GPU hardware in this environment for a meaningful FP8_BLOCK end-to-end benchmark.
> 
> If there’s a recommended internal machine, cluster, or standard benchmark command I should use, I’m happy to run it if I can get access.

Hi @zeel2104 , thanks for updating. Unfortunately we don't expose pathways to test, but i will validate and confirm tomorrow and we can merge this in

HDCharles:
closes https://github.com/vllm-project/llm-compressor/issues/2520

Issue #2523, opened 2026-03-26
Refactor Activation Ordering across LLMC and CT


HDCharles:
ActivationOrdering is curerntly used to do 2 different things

in [GPTQ there is an option](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L121-L135) to change the order in which we iterate through the columns which prioritises high activation columns first

but there's also a [format in compressed tensors](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_args.py#L146) which requires a different packing format and quantization strategy, the scales have to be calculated on permuted weights and the kernel needs a group index.

currently these both exist in CT (which should only specify the quantization format as activationOrdering=True/False) and further you can pass ActivationOrdering into GPTQ which seems to overwrites the quantization scheme.

These need to be separated, this refactor is looking for 3 changes

1) GPTQ should have functionality to change the column iteration order based on activation ordering, not change the quantization scheme
2) CT shouldn't have anything about GPTQ in it (remove the static/weight option and just have it be true or false) 
note: handling for the case where CT activation ordering is enabled but GPTQ activation ordering is off, it is handled as though both are enabled

nrmlthms:
Hi @HDCharles , I would like to take this up

HDCharles:
Sounds good, also you can reach out to me on vllm slack if you want to discuss anything.

I did a little bit of cleaning up of the actorder stuff just to make the existing code and docs cleaner, see: https://github.com/vllm-project/llm-compressor/pull/2541 and https://github.com/vllm-project/compressed-tensors/pull/654

in 2541 there are some comments about other parts of the actorder code that should be cleaned up once the main changes for the refactor are made.

HDCharles:
so internally there's some push back on this refactor so i'm going to close this.

PR #2541, opened 2026-03-30
GPTQ Actorder Refactor

HDCharles:
Summary:

this code was really spaghetti before, tried to clean it up and make it a little more readable.
clean up actorder if-tower
no longer saving random values we don't need (g_idx that hasn't been permuted)
update tests to actually test actorder
make tests faster (smaller model + don't tokenize whole dataset)
note: smaller model required me to change groupsize to 32

TEST PLAN:
pytest -vs -rs /home/HDCharles/repos/llm-compressor/tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py


src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -110,7 +109,8 @@ def quantize_weight(
     num_rows = W.shape[0]
     num_columns = W.shape[1]
 
-    # generate scale, should include tensor group / use global scale
+    scale, zero_point = observer(W)
+    # handle g_idx and activation ordering
     if strategy in (QuantizationStrategy.GROUP, QuantizationStrategy.TENSOR_GROUP):
```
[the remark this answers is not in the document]
HDCharles: no it doesn't hit that, thats unchanged from before

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -119,26 +119,16 @@ def quantize_weight(
...
+            # actually need scale/zp for permuted weight for this format
             scale, zero_point = observer(W)
-
             # use identity g_idx (invert permutation later)
 
         elif actorder == ActivationOrdering.WEIGHT:
-            # update groups first, then permute by activation order
-            scale, zero_point = observer(W)
+            # permute weights and g_idx
             W, H, perm = _apply_activation_ordering(W, H)
```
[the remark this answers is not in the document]
HDCharles: no, weight is an alias for static

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -119,26 +119,16 @@ def quantize_weight(
         )
 
         if actorder == ActivationOrdering.GROUP:
-            # permute by activation order first, then update groups
             W, H, perm = _apply_activation_ordering(W, H)
-            update_offload_parameter(module, "weight_g_idx", g_idx)
```
brian-dellabetta, line 124: was this just wrong before? It looks like we were setting it before g_idx was ever permuted

HDCharles: not quite this whole thing is complicated and dumb but:

1) group activation ordering is kind of like saying 'ok lets permute the whole weight matrix and then feed it into GPTQ and pretend that was the original weight' this is why g_idx is left as an identity mapping and we previously were 'saving it before it was ever permuted'

2) a non identity g_idx IS needed (the actual format for the weight is the original format even though the scales/zp are from the permuted matrix so we do need the mapping) for the final save, but its always set at the end, no reason it was ever set here which is why i removed it.

3) the complication comes from the fact that we are using g_idx to do 2 mostly-separate but kind-of-related things, it handles the mid-gptq column to column mapping for weight actordering AND for group actordering, we use it at the very end for the g_idx we need to save. Theoretically these functions should be decoupled in https://github.com/vllm-project/llm-compressor/issues/2523

tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py
```diff
@@ -58,20 +105,38 @@
         recipe_modifier_full_group,
         recipe_modifier_shorthand_a,
         recipe_modifier_shorthand_b,
+        recipe_modifier_group_actorder_weight,
```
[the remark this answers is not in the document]
HDCharles: thats an alias, static=weight

tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py
```diff
@@ -58,20 +105,38 @@
...
+        recipe_modifier_group_actorder_group,
     ],
 )
 def test_oneshot_application(recipe, tmp_path):
     output = tmp_path / "oneshot_output"
-    model = "TinyLlama/TinyLlama-1.1B-Chat-v1.0"
+    model_id = "nm-testing/tinysmokellama-3.2"
```
HDCharles, line 114: made model smaller

tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py
```diff
@@ -58,20 +105,38 @@
...
+
+    # Quantize model
     oneshot(
-        model=model,
+        model=model_id,
         dataset=dataset,
         output_dir=output,
         recipe=recipe,
         num_calibration_samples=9,
+        splits={"calibration": "train[:9]"},
```
HDCharles, line 139: don't tokenize the whole dataset anymore

tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py
```diff
@@ -98,3 +163,32 @@ def test_oneshot_application(recipe, tmp_path):
...
+        ), "WEIGHT actorder should not have g_idx"
+
+    # Numerical validation: check MSE
+    with torch.no_grad():
+        quantized_output = model_loaded(**inputs).logits
+
+    mse = torch.nn.functional.mse_loss(quantized_output, original_output).item()
+
+    # MSE threshold - quantization should not degrade quality too much
+    mse_threshold = 0.015
```
HDCharles, line 186: small MSE check to make sure its all working as expected numerically

src/llmcompressor/modifiers/gptq/gptq_quantize.py
brian-dellabetta: Thanks for the details 👍

src/llmcompressor/modifiers/gptq/gptq_quantize.py
kylesayrs, line 260: I'd argue that this variable is also kind of spaghet

kylesayrs, line 263: If you have any ideas on how to clean this up too

HDCharles: I'm saving some spaghetti for later

HDCharles: Yeah I figure the refactor will handle it

PR #2547, opened 2026-03-30
Add MoE calibration wrapper for GLM-4.7-Flash (Glm4MoeLiteMoE)

Nottlespike:
## Summary

GLM-4.7-Flash uses a separate MoE class (`Glm4MoeLiteMoE`) that is not covered by the existing `Glm4MoeMoE` wrapper. Without this fix, MoE calibration is **silently skipped** for GLM-4.7-Flash models, resulting in quantization that doesn't properly calibrate expert weights.

## Problem

The model `zai-org/GLM-4.7-Flash` (31B MoE) uses `Glm4MoeLiteMoE` with a different architecture than `Glm4MoeMoE`:

- Has `shared_experts` attribute (not `shared_expert`)
- Uses `Glm4MoeLiteNaiveMoe` experts interface: `(hidden_states, topk_indices, topk_weights)`
- Has group-based routing with `n_group`, `topk_group` parameters
- 64 routed experts + shared experts

When quantizing with NVFP4, the MoE calibration context manager checks for registered wrappers but `Glm4MoeLiteMoE` doesn't match `Glm4MoeMoE`, so calibration silently falls back to standard forward passes without collecting expert activation statistics.

## Solution

Add `CalibrationGlm4MoeLiteMoE` wrapper class that:
- Registers for `Glm4MoeLiteMoE` class specifically
- Implements `route_tokens_to_experts()` for proper group-based routing
- Collects activation statistics for all 64 experts during calibration
- Provides proper `forward()` that routes through all experts

## Changes

- `src/llmcompressor/modeling/glm4_moe_lite.py` - New 119-line wrapper
- `src/llmcompressor/modeling/__init__.py` - Import the new wrapper

## Testing

Verified that quantization now shows:
```
Found 46 MoE modules to replace
Replaced 46 MoE modules for calibration
```

Instead of the previous behavior where no MoE modules were detected for `Glm4MoeLiteMoE`.

## Hardware Tested

- 2× RTX PRO 6000 Blackwell (96GB each)
- NVFP4 quantization with sequential offload

## Summary by CodeRabbit

* **New Features**
  * Added GLM4 MoE Lite model calibration support
  * Introduced `CalibrationGlm4MoeLiteMoE` and `CalibrationOffsetNorm` modules to public exports
  * Expanded calibration module availability

* **Tests**
  * Added comprehensive test coverage for GLM4 MoE Lite calibration including expert routing and output validation


brian-dellabetta (commented):
Hi @Nottlespike , thanks for preparing this. It looks like a lot of the code is shared with what is in `src/llmcompressor/modeling/glm_moe_dsa.py`. Have you explored what it would look like to import and subclass the classes from that file directly? I know transformers sticks to the approach of no shared code across model definitions, but given we want to apply the same operation to the 3D expert tensors in both, maybe we won't have to repeat our code

Nottlespike:
Thanks for the feedback @brian-dellabetta — great call on the shared code.

Refactored `CalibrationGlm4MoeLiteMoE` to subclass `CalibrationGlmMoeDsaMoE` directly. The routing logic (`route_tokens_to_experts`) and sequential expert forward pass were identical between the two, so the Lite wrapper now inherits both and overrides only two factory methods:

- `_get_num_experts(config)` — returns `config.n_routed_experts` (vs `config.num_local_experts` in DSA)
- `_make_experts(config, original_experts)` — constructs `SequentialGlm4MoeLiteExperts` (unpacks `Glm4MoeLiteMLP` instead of `GlmMoeDsaMLP`)

`is_permanent = True` and `restore()` are inherited from the base class. Net: **-69 lines**.

Also addressed the automated review feedback:
- **Gemini** (duplicated branch) and **Copilot** (`torch.no_grad` on calibration pass) — both referred to the v1 batch-dispatch `forward()` which was replaced with sequential expert iteration in v2. No longer applicable.
- **Copilot** (missing unit test) — `test_calib_glm4_moe_lite.py` has 3 tests covering all-experts-triggered, output-matches, and experts-are-linear.

`ruff check` + `ruff format` pass.

Nottlespike:
Worth noting: the `torch.no_grad()` + `copy_()` fix in this PR also lands in the **base class** `SequentialGlmMoeDsaExperts` (`glm_moe_dsa.py`), so it benefits all `GlmMoeDsaMoE` models — not just GLM-4.7-Flash Lite.

This is especially timely given the recent release of **GLM-5.1** which uses the DSA MoE architecture. Any future calibration runs on GLM-5.1 (or other DSA-family models) will get the safer parameter transfer out of the box.

src/llmcompressor/modeling/__init__.py
```diff
@@ -12,15 +12,16 @@
 # trigger registration
 from .afmoe import CalibrationAfmoeMoE  # noqa: F401
 from .deepseek_v3 import CalibrationDeepseekV3MoE  # noqa: F401
+from .fuse import *
```
brian-dellabetta: does the order of imports in this file need to change? If not, would be good to change so it is scoped purely to your glm changes

brian-dellabetta (commented):
Thanks @Nottlespike for the updates, this is looking good. just one nit on the changing of import ordering in `__init__.py`

Nottlespike:
Fixed — restored upstream import ordering. The diff for `__init__.py` is now a single `+1` line (our `glm4_moe_lite` import). Also squashed all commits into one clean commit with DCO.

Nottlespike:
Could a maintainer add the `ready` label when this looks good? Needed for the full CI test suite to run.

brian-dellabetta (approved):
Thanks for the contribution! lgtm

kylesayrs (approved):

PR #2566, opened 2026-04-06
[Bugfix] Fix model_free_ptq for models with non-contiguous fused attention layers

RobTand:
## Summary

Fix `model_free_ptq` for models with non-contiguous fused attention layers (Qwen3.5), and add a working example for full NVFP4 quantization including DeltaNet linear attention layers.

## Problem

Qwen3.5 uses a hybrid attention architecture: 3 DeltaNet (`linear_attn`) layers followed by 1 full (`self_attn`) layer, repeating. This causes `model_free_ptq` to crash with a `ValueError` when fused weights (q/k/v) are split across shards.

### Bug 1: `match_names_set_eager` crashes on cross-shard fused weights

In Qwen3.5-27B, all `q_proj` weights land in shard 8 while `k_proj`/`v_proj` are in shard 11. When the validation processes shard 8, `match_names_set_eager` matches layer 3's `q_proj`, then layer 7's `q_proj` hits the same target slot before the set completes.

**Fix:** When a duplicate target match is encountered, flush the incomplete set and start a new one rather than raising `ValueError`. Incomplete sets are returned for the existing cross-shard resolution logic.

### Bug 2: Tokenizer double-batching in calibration example

The processor returns nested lists `[[tok1, tok2, ...]]` with a batch dimension. When stored in an HF dataset and loaded by a DataLoader with `batch_size=1`, this creates `(1, 1, seq_len)` tensors instead of `(1, seq_len)`, causing 4D inputs during calibration.

**Fix:** Unwrap the processor's batch dimension in the tokenize function.

## Results

With these fixes, full NVFP4 W4A4 quantization of Qwen3.5-27B (including DeltaNet layers) produces:
- **29% faster decode** (117ms → 82.5ms per step) on DGX Spark (SM121)
- **26% smaller model** (27GB → 20GB)
- **Coherent output** — FP4 activation quantization on DeltaNet layers works correctly
- Native CUTLASS W4A4 compute (not Marlin W4A16 fallback)

## Test plan

- [x] `model_free_ptq` with `NVFP4A16` on `Qwen/Qwen3.5-27B` completes successfully
- [x] `oneshot` with `NVFP4` on `Qwen/Qwen3.5-27B` completes with DeltaNet layers included
- [x] Calibrated checkpoint loads and serves correctly in vLLM with CUTLASS W4A4
- [x] Output quality validated (coherent reasoning, no "!" artifacts)
- [x] Profiling confirms 29% decode speedup from quantizing DeltaNet layers
- [x] No regression for standard models (co-located q/k/v still work)

### DIFF
--- a/src/llmcompressor/entrypoints/model_free/process.py
+++ b/src/llmcompressor/entrypoints/model_free/process.py
@@ -82,6 +82,7 @@ def process_file(
     assert not is_microscale_scheme(scheme), "Use `process_file_microscale_scheme`"
 
     tensors = _load_tensors_from_inverse_weights_map(inverse_weights_map, device)
+    tensors = split_fused_moe_experts(tensors)
 
     if converter is not None:
         converter.process(tensors)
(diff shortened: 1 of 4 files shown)


src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -253,3 +254,51 @@ def _load_tensors_from_inverse_weights_map(
                     )
                 tensors[tensor_name] = f.get_tensor(tensor_name)
     return tensors
+
+
+def split_fused_moe_experts(
```
brian-dellabetta, line 259: Looks like a lot of this code is duplicated from another community user PR, see [here](https://github.com/vllm-project/llm-compressor/pull/2464/changes#diff-a88bd486ce930f2604f136972626972642280e8c21567f197c1ca891c0e5f771R260). Should we get that in first and then cover the rest of the changes you have in your PR?

brian-dellabetta (commented):
Hi @RobTand , thanks for preparing this. See comment below

brian-dellabetta:
Hi @RobTand , this has been resolved by 
* #2464 
If you are still encountering issues or would like us to review other parts of this PR not handled by #2464, please re-open or open an issue. Thanks!

PR #2581, opened 2026-04-07
Disable code rabbit PR approval workflow

dsikka:
Changed request_changes_workflow to false to prevent code rabbit from approving PRs.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -18,6 +18,19 @@ pull_request_rules:
         add:
           - documentation
 
+  - name: auto-rebase if approved, ready, and 5 commits behind main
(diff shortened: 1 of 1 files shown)


HDCharles (approved):
is that the only thing "request_changes_workflow" does? or will it also make it not request specific changes?

dsikka:
> request_changes_workflow

Based on the docs: https://docs.coderabbit.ai/reference/configuration

> Automatically approve once CodeRabbit’s comments are resolved and no pre-merge checks are failing. Note: In GitLab, all discussions must be resolved.

So I think so? Might have to test and see

PR #2585, opened 2026-04-08
Observers refactor

### LINKED ISSUE #2665: [Bug]: AWQ DDP run on Qwen3.6-27B fails at layer 31 (8th full-attention block) with safetensors "header too small"
### Summary

DDP AWQ run on Qwen3.6-27B (BF16 source, hybrid linear/full attention arch) fails reproducibly during `_apply_smoothing` at the 8th full-attention block (layer index 31 with `full_attention_interval=4`). Raises `safetensors_rust.SafetensorError: Error while deserializing header: header too small` from `compressed_tensors/offload/cache/disk.py:72` while reading a `_scale` parameter through `OffloadCache.__getitem__`.

### Environment

- Hardware: 3× NVIDIA RTX 3090 (24 GiB each, sm_86, PCIe-only, no NVLink/P2P)
- Driver/CUDA: NVIDIA driver 580.126.20, CUDA runtime 12.9 (per torch build)
- Container base: `vllm/vllm-openai:v0.19.0`
- llm-compressor: `0.10.1.dev121+g7a52e9e2` (pip install `git+https://github.com/vllm-project/llm-compressor.git@main` on 2026-04-28)
- transformers: 5.7.0
- torch: 2.10.0+cu129
- compressed-tensors: latest tip pulled as a transitive dependency on the same date
- Launch: `torchrun --nproc_per_node=3` inside a Kubernetes Job pod

NCCL env (PCIe-only Ampere):
```
NCCL_P2P_DISABLE=1
NCCL_IB_DISABLE=1
PYTORCH_ALLOC_CONF=expandable_segments:True
```

### Model

`Qwen/Qwen3.6-27B` (BF16, dense Qwen3-Next family). Per its `config.json`:

- `architectures: ["Qwen3_5ForConditionalGeneration"]`
- `model_type: "qwen3_5"`
- `full_attention_interval: 4`
- `hidden_size: 5120`, `head_dim: 256`, `num_hidden_layers: 64`
- `layer_types`: pattern `[linear_attention, linear_attention, linear_attention, full_attention, ...]` repeating, giving 16 full-attention layers and 48 linear-attention layers

### Reproduction

Adapted from `examples/awq/qwen3_moe_example_ddp.py`:

```python
import os, torch
from compressed_tensors.offload import dispatch_model, init_dist, load_offloaded_model
from transformers import AutoModelForCausalLM, AutoProcessor

from llmcompressor.datasets.utils import get_rank_partition
from llmcompressor.modifiers.transform.awq import AWQModifier

MODEL_ID = "Qwen/Qwen3.6-27B"
NUM_CALIBRATION_SAMPLES = 256
MAX_SEQUENCE_LENGTH = 512

init_dist()
rank = torch.distributed.get_rank()
OFFLOAD_DIR = f"/scratch/awq_offload_rank{rank}"
os.makedirs(OFFLOAD_DIR, exist_ok=True)

with load_offloaded_model():
        MODEL_ID, dtype="auto",
        device_map="auto_offload",
        offload_folder=OFFLOAD_DIR,
        trust_remote_code=True,
    )
processor = AutoProcessor.from_pretrained(MODEL_ID, trust_remote_code=True)

ds = load_dataset(
    "HuggingFaceH4/ultrachat_200k",
    split=get_rank_partition("train_sft", NUM_CALIBRATION_SAMPLES),
)
ds = ds.select_columns(["messages"]).shuffle(seed=42)

def preprocess(example):
    msgs = [
        {"role": m["role"], "content": [{"type": "text", "text": m["content"]}]}
        for m in example["messages"]
    ]
    return processor.apply_chat_template(
        msgs, return_tensors="pt", padding=False, truncation=True,
        max_length=MAX_SEQUENCE_LENGTH, tokenize=True,
        add_special_tokens=False, return_dict=True,
        add_generation_prompt=False,
    )

ds = ds.map(preprocess, batched=False, remove_columns=ds.column_names)

def collator(batch):
    assert len(batch) == 1
    return {k: torch.tensor(v) for k, v in batch[0].items()}

recipe = [
    AWQModifier(duo_scaling="both", offload_device=torch.device("cpu")),
    QuantizationModifier(ignore=["lm_head"], scheme="W8A16", targets=["Linear"]),
]

oneshot(
    model=model, dataset=ds, recipe=recipe,
    data_collator=collator,
)
```

Launch:
```
torchrun --nproc_per_node=3 quant.py
```

### Expected

`oneshot` completes; quantized W8A16 checkpoint is saved.

### Actual

Process advances through the sequential pipeline. AWQ `_apply_smoothing` succeeds on full-attention blocks at layer indexes 3, 7, 11, 15, 19, 23, 27 (the first 7 full-attention layers). On layer index 31 (the 8th full-attention layer), rank 0 raises:

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/pipelines/sequential/helpers.py", line 467, in wrapper
[rank0]:     return func(*args, **kwargs)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/pipelines/sequential/pipeline.py", line 162, in __call__
[rank0]:     LifecycleCallbacks.sequential_epoch_end(subgraph)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/core/session_functions.py", line 165, in sequential_epoch_end
[rank0]:     return cls.event(EventType.SEQUENTIAL_EPOCH_END, subgraph=subgraph, **kwargs)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/core/session_functions.py", line 91, in event
[rank0]:     return active_session().event(event_type, **kwargs)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/core/session.py", line 181, in event
[rank0]:     mod_data = self._lifecycle.event(...)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/core/lifecycle.py", line 204, in event
[rank0]:     data = mod.update_event(state=self.state, event=event, **kwargs)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/modifiers/modifier.py", line 122, in update_event
[rank0]:     self.on_event(state, event, **kwargs)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/modifiers/transform/awq/base.py", line 254, in on_event
[rank0]:     self._apply_smoothing(state.model)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/modifiers/transform/awq/base.py", line 558, in _apply_smoothing
[rank0]:     best_scales = self._compute_best_scale(...)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/llmcompressor/modifiers/transform/awq/base.py", line 722, in _compute_best_scale
[rank0]:     forward_quantize(...)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/compressed_tensors/quantization/lifecycle/forward.py", line 319, in forward_quantize
[rank0]:     scale = getattr(module, f"{base_name}_scale")
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1956, in __getattr__
[rank0]:     return _parameters[name]
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/compressed_tensors/offload/cache/base.py", line 187, in __getitem__
[rank0]:     onloaded = self.onload(offloaded)
[rank0]:   File "/usr/local/lib/python3.12/dist-packages/compressed_tensors/offload/cache/disk.py", line 72, in onload
[rank0]:     with safe_open(weight_info["safetensors_file"], framework="pt", device=device) as file:
[rank0]: safetensors_rust.SafetensorError: Error while deserializing header: header too small
```

Ranks 1 and 2 exit with SIGTERM after rank 0's failure.

### Additional observations

1. **Reproducibility:** Failure occurs at layer index 31 in two consecutive runs, on independent restarts of the pod, with different `offload_folder` paths (one on container `/tmp`, one on a Ceph RBD-backed PVC). Layer index 31 is reached after AWQ has successfully processed seven prior full-attention blocks.

2. **Per-rank offload directory occupancy:** With `offload_folder` set to `/scratch/awq_offload_rank{rank}` (distinct path per rank), inspection after failure shows:
   - `awq_offload_rank0/`: 947 `ct_disk_cache*.safetensors` files (smallest 113 bytes, total ~9 MB)
   - `awq_offload_rank1/`: empty
   - `awq_offload_rank2/`: empty

   Only rank 0 writes any offloaded data. No warning or error is emitted regarding the empty offload directories on ranks 1 and 2.

3. **Disk space is not the cause:** Ceph PVC had 336 GB free at start of the failed run (well over 3× model weights + workspace).

4. **Architectural correlation:** All eight full-attention layers (indexes 3, 7, …, 31) share the same module shape under `full_attention_interval=4`; only the 8th triggers the failure. AWQ's hybrid attention mapping discovery logged at startup:
   ```
   build_hybrid_attention_mappings | INFO - Built dynamic hybrid attention AWQ mappings:
     16 full-attention layers, 48 linear-attention layers,
     linear projections: ['in_proj_qkv', 'in_proj_z', 'in_proj_b', 'in_proj_a'], MoE: False
   ```

5. **Single-process control:** Running the same model with `device_map="cpu"`, no `init_dist`, and no `load_offloaded_model()` context (i.e., the non-DDP, non-`auto_offload` code path) advances past layer 31. The bug observed here is specific to the `device_map="auto_offload"` + DDP path.

6. **Class-level `DiskCache.index`:** `compressed_tensors.offload.cache.disk.DiskCache.index` is defined as a class attribute (not assigned in `__init__`). All instances of `DiskCache` within a single process share the same `dict`.

7. **File path not surfaced:** The traceback does not include the path of the file that raised. At the time of post-failure inspection, all extant `ct_disk_cache*.safetensors` files in rank 0's offload directory open successfully via `safetensors.safe_open` (smallest valid file is 113 bytes; no zero-byte or sub-8-byte files present).

### Notes for repro by others

The failure has been observed only on Qwen3.6-27B (Qwen3-Next dense hybrid). Whether `Qwen/Qwen3-Next-80B-A3B-*` exhibits the same behavior with the upstream `examples/awq/qwen3_moe_example_ddp.py` (modified to use offload_folder per rank, since that argument is required) is not tested in this report.

HDCharles:
# Background

problems this PR attempts to solve:

## Large Problems

1) observers always calculate qparams when they are run, even if you need to do synchronization across ranks on an intermediate value
2) all the qparam calculation functionality was duplicated because we had qparam calculation happen on run and again after synchronize
3) global scale was treated partially like a qparam and partially, not. Each modifier handled global scale differently
4) there was a modifier dependent lifecycle for enabling/disabling quantization. Quantization enabled during calibration for quantization modifier but not for AWQ,GPTQ...etc

## Smaller Problems 

5) observers were complicated, they needed information from the module (that wasn't passed in) to do their job so they needed to directly access module attributes
6) MSE calculations used global scale even though numerically the global scale could be ignored until later.
7) sync_activation_observers did both syncing and updating of qparams
8) imatrix observer wasn't set up to be synchronized in DDP. 
9) g_idx was treated as kind of like a qparam in several places but then never actually used, i.e. there was functionality within observers to handle this, but then in reality g_idx was set as identity when observer was using it. 
10) sequential pipeline was doing duplicated work in sequential epoch end and calibration epoch end

# Summary

1,2) observers now have update_statistics_from_observed which runs on forward and get_qparams, other functionality removed.
3) observers have references to observers of fused weights for global_scale calculation so when they calculate qparams, global_scale will be set correctly. global_scale logic is now entirely contained in observers
4) quantization disabled across the board during calibration, quantization now always happens at the end of the lifecycle
5) removed references to module from observers
6) removed references to global_scale from MSE calculations
7) removed observer functionality for g_idx (was never used)
8) cleaned up a lot of observer utils
9) consolidated synchronization logic, there is a notion of activation_statistics that all activation observers and one weight observer has that allows for synchronization
10) removed g_idx from observers, handled externally in GPTQ
11) all work done on sequential epoch end, added sequential epoch end to basic and datafree pipelines eplicating a separation of concerns where sequential epoch end does compression while calibration epoch end does cleanup

# Test Plan

duplicate: https://gist.github.com/HDCharles/a8ece53d76ce89ce81ddeed4fba3aa28
Results: https://gist.github.com/HDCharles/ec03e247f964afa91b10ec679e74c262

overall results look reasonable, largely stable. There was an imatrix+global_scale bug in main that is now fixed by this simpler approach


src/llmcompressor/observers/base.py
```diff
@@ -50,83 +61,89 @@ def __init__(
...
+                    # Parameter already exists, update it
+                    from compressed_tensors.utils import update_offload_parameter
+                    update_offload_parameter(module, param_name, global_scale)
+                else:
+                    # Parameter doesn't exist yet, create it
+                    setattr(module, param_name, global_scale)
+        else:
+            global_scale = None
+
+        qparams = self._compute_qparams_from_statistics()
```
brian-dellabetta: i like the `QParamsDict`, but it seems weird that `self._compute_qparams_from_statistics()` returns a QParamsDict where global_scale will always be overwritten by the output of `self._compute_gparams_from_statistics()` on the next line. Can the API be reduced to just having each observer satisfy a single `self._compute_qparams_from_statistics()` abstractmethod, which returns a QParamsDict that has global_scale set if the quant strategy requires one? It looks like they're always called in conjunction

src/llmcompressor/observers/moving_base.py
```diff
@@ -110,13 +123,8 @@ def synchronize(self) -> List[dist.Work]:
...
-            "past_min_vals",
-            "past_max_vals",
-            "past_global_min_vals",
-            "past_global_max_vals",
-        ):
-            val = getattr(self, attr, None)
+        for key in ("min_vals", "max_vals", "global_min_vals", "global_max_vals"):
+            val = self.statistics.get(key)
             if val is not None:
                 val.div_(world_size)
```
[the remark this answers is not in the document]
brian-dellabetta: hmm, is this a bug?

src/llmcompressor/observers/min_max.py
```diff
@@ -18,11 +20,55 @@ class MemorylessMinMaxObserver(Observer):
     :param **observer_kwargs: keyword arguments for observer initialization
     """
 
-    def get_min_max(self, observed: torch.Tensor) -> MinMaxTuple:
-        return _get_min_max(observed)
+    accumulates_statistics = False
```
brian-dellabetta: nit -- if it's the case that `accumulates_statistics` is False if and only if it is a "Memoryless" observer, consider renaming this to `is_memoryless`

src/llmcompressor/observers/base.py
```diff
@@ -50,83 +61,89 @@ def __init__(
         self.module = ref(module) if module is not None else None
         self.base_name = base_name
         self.args = args
+        self.statistics = {}  # Dictionary to store observer statistics
```
brian-dellabetta: nit -- consider a typed dict like you have for QParamsDict. It looks like there are only ever 4 keys -- min_vals, max_vals, global_min_vals, global_max_vals

brian-dellabetta (commented):
Things look good to me, the side effects that occurred in observers always made me nervous. I have one comment on the new observer API, it looks like the abstractmethods could be cleaned up a bit, and a few nits.

src/llmcompressor/observers/base.py
HDCharles: currently there's only 4 values but it can theoretically expand.

src/llmcompressor/observers/base.py
HDCharles: yeah

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -190,17 +190,18 @@ def process_file_microscale_scheme(
...
+        # 2.1. fuse global scale
         global_scales = [m.weight_global_scale for m in named_modules.values()]
         fused_global_scale = torch.min(torch.cat(global_scales, dim=0))
 
+        # 2.2. adjust scales to account for updated global_scale
         for name, module in named_modules.items():
             module_name, _ = name.rsplit(".", 1)
+            old_global_scale = module.weight_global_scale
+            scale_adjustment = fused_global_scale / old_global_scale
+            module.weight_scale.data.mul_(scale_adjustment)
```
brian-dellabetta: Discussed with @HDCharles that we believe this is an oversight both in model_free_ptq and in `update_fused_layer_weight_global_scales`. If we want to change in this PR, we should probably propagate the changes for that function as well

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -282,27 +282,19 @@ def sync_activation_observers(self, model: torch.nn.Module):
                 observer = getattr(module, f"{base_name}_observer", None)
                 if observer is None:
                     continue
-                pending_comms.extend(observer.synchronize())
+                pending_comms.extend(observer.synchronize_observer())
```
brian-dellabetta: nit -- any reason for the rename here? `observer.synchronize()` is a little more elegant, it's implicit what we're synchronizing

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -89,32 +89,34 @@ def call_observer(
     module: Module,
     base_name: str,
     value: Optional[torch.Tensor] = None,
-    should_calculate_gparam: bool = False,
-    should_calculate_qparams: bool = True,
+    should_calculate_gparam: bool = False,  # Kept for backward compatibility, ignored
+    should_calculate_qparams: bool = True,  # Kept for backward compatibility, ignored
```
brian-dellabetta: nit -- consider raising deprecation warning if these are set by user to something other than default value, or by using a sentinel

brian-dellabetta (approved):
Thanks for updates, things LGTM. i just have some nits, biggest thing being the re-scaling of weight_scale you are doing after fusing weight_global_scale. we should either split out to a separate PR or add the changes to `update_fused_layer_weight_global_scales`

src/llmcompressor/observers/base.py
```diff
@@ -1,198 +1,161 @@
...
-        module: torch.nn.Module | None = None,
         **observer_kwargs,
     ):
         super().__init__()
-        self.module = ref(module) if module is not None else None
         self.base_name = base_name
         self.args = args
+        # weight observers don't need to sync
+        if base_name == "weight":
+            self._act_sync_dict = {}
```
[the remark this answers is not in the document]
HDCharles: yeah

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -263,48 +267,40 @@ def end_calibration(self, model: torch.nn.Module):
...
         """
         if not is_distributed():
             return
 
         pending_comms = []
-        modules_to_update = []
-
         for _, module in match_named_modules(model, self.resolved_targets, self.ignore):
-            for base_name in ("input", "output", "q", "k", "v"):
+            for base_name in ("weight", "input", "output", "q", "k", "v"):
```
brian-dellabetta, line 302: why is weight added here? I thought this was scoped to just activation quant?

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -267,16 +265,13 @@ def on_end(self, state: State, event: Event, **kwargs):
...
                 if hasattr(layer, "weight_observer"):
-                    update_weight_global_scale(layer)
-        for mapping in self._resolved_mappings:
-            update_fused_layer_weight_global_scales(mapping.parent)
-        for mapping in self._resolved_mappings:
-            for layer in mapping.balance_layers + [mapping.smooth_layer]:
-                if hasattr(layer, "weight_observer"):
-                    update_weight_zp_scale(layer)
+                    smoothed_layers.append(layer)
+        observe(smoothed_layers, base_name="weight")
```
brian-dellabetta: the goal here is to make sure after scaling that qparams are updated. After your other PR, where that gets deferred, will this no longer be necessary?

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -267,16 +265,13 @@ def on_end(self, state: State, event: Event, **kwargs):
...
             for layer in mapping.balance_layers + [mapping.smooth_layer]:
                 if hasattr(layer, "weight_observer"):
-                    update_weight_global_scale(layer)
-        for mapping in self._resolved_mappings:
-            update_fused_layer_weight_global_scales(mapping.parent)
-        for mapping in self._resolved_mappings:
-            for layer in mapping.balance_layers + [mapping.smooth_layer]:
-                if hasattr(layer, "weight_observer"):
-                    update_weight_zp_scale(layer)
+                    smoothed_layers.append(layer)
```
brian-dellabetta: there's no functional difference between appending all modules to a list and running observe/update_qparams collectively on them vs. just running observe/update_qparams on each right?

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -728,15 +714,6 @@ def _compute_best_scale(
                         / _scalesview
                     ).to(balance_layer.weight.dtype)
 
-                # Apply fused global scales for TENSOR_GROUP during grid search
```
brian-dellabetta, line 731: awesome that we can remove all this by setting up fused observers from the get-go

src/llmcompressor/modifiers/transform/awq/base.py
brian-dellabetta, line 657: do we have to worry about fusing here, if we're patching the observers? If we're patching q/k/v proj here, I would think the patching of the attr means we need to fuse them after patching

src/llmcompressor/modifiers/utils/helpers.py
```diff
@@ -1,78 +0,0 @@
...
-    # and their weights are all TENSOR_GROUP quantized,
-    # ensure the global scale is fused
-    layers_to_fuse_list = [
-        # mlp / expert layers have fused gate_up_proj
-        ("gate_proj", "up_proj"),
-        # attention layers have fused qkv_proj
-        ("q_proj", "k_proj", "v_proj"),
-        # DeepSeek multi-latent attention has fused_qkv_a_proj
-        ("q_a_proj", "kv_a_proj_with_mqa"),
-    ]
```
brian-dellabetta, line 40: where does the data model for this live now? somewhere in the observer fusing logic? I didn't see it but i would think this still needs to be retained somewhere

src/llmcompressor/observers/base.py
```diff
@@ -1,198 +1,161 @@
...
 
-        with align_module_device(module):
-            return getattr(module, f"{self.base_name}_{name}", None)
+        :param observers: list of observers to fuse together
+        """
+        observers = list(observers)
+        for obs in observers:
+            for other in observers:
+                if other is not obs:
+                    obs._fused_observers.append(other)
```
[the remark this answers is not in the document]
brian-dellabetta: Yeah might want to check if other is not in obs._fused_observers already? Also wonder if a WeakRef would be desired here to prevent observers from being garbage collected?

src/llmcompressor/observers/helpers.py
```diff
@@ -13,15 +13,15 @@
...
 
-__all__ = ["flatten_for_calibration"]
+__all__ = ["flatten_for_calibration", "fuse_weight_observers", "FUSED_LAYER_NAMES"]
 
 
 def flatten_for_calibration(
     value: torch.Tensor,
     base_name: str,
     args: QuantizationArgs,
-    g_idx: torch.Tensor | None = None,
```
brian-dellabetta, line 24: this is no longer needed?

src/llmcompressor/modifiers/utils/helpers.py
brian-dellabetta: Nm I see it below now in src/llmcompressor/observers/helpers.py

src/llmcompressor/observers/imatrix.py
```diff
@@ -291,16 +252,19 @@ def _get_validated_importance(self, observed: torch.Tensor) -> torch.Tensor | No
 
 def _grid_search(
     observed: torch.Tensor,
-    args: QuantizationArgs,
+    args,
```
brian-dellabetta: is this no longer type QuantizationArgs?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -104,19 +100,9 @@ def __call__(
...
-            for mod in session.lifecycle.recipe.modifiers
-        )
-
         with contextlib.ExitStack() as stack:
             stack.enter_context(calibration_forward_context(model))
-            # Optionally disable quantization
-            if not dataset_args.quantization_aware_calibration or disable_qac:
-                stack.enter_context(DisableQuantization(model))
-
+            stack.enter_context(DisableQuantization(model))
```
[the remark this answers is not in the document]
brian-dellabetta: Yeah i think we need this in basic/sequential pipelines? quant needs to be enabled for quantization aware calibration, like for QuantizationModifier with static act quant

brian-dellabetta (commented):
Things look good to me, but lots of comments since my last review. One particular concern is the change to disable quant in the pipelines. I think we need that, otherwise we'll get silent regressions from improperly calibration static activation qparams

HDCharles:
@coderabbitai re-review PR

src/llmcompressor/observers/base.py
```diff
@@ -50,83 +61,89 @@ def __init__(
...
+                    from compressed_tensors.utils import update_offload_parameter
+                    update_offload_parameter(module, param_name, global_scale)
+                else:
+                    # Parameter doesn't exist yet, create it
+                    setattr(module, param_name, global_scale)
+        else:
+            global_scale = None
+
+        qparams = self._compute_qparams_from_statistics()
+        qparams["global_scale"] = global_scale
```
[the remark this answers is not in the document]
HDCharles: old

src/llmcompressor/observers/imatrix.py
```diff
@@ -116,32 +116,77 @@ def detach(self, module: torch.nn.Module) -> None:
...
+            global_observed,
             self.args,
             self.maxshrink,
             self.patience,
             self.grid,
             self.norm,
             optimize_global_scale=True,
         )
+        self.statistics['global_min_vals'] = global_min
+        self.statistics['global_max_vals'] = global_max
```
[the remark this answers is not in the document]
HDCharles: old

src/llmcompressor/observers/base.py
```diff
@@ -39,6 +46,10 @@ class Observer(InternalModule, RegistryMixin):
     :param **observer_kwargs: keyword arguments for observer initialization
     """
 
+    # Class attribute indicating whether this observer accumulates statistics across calls
+    # Memoryless observers should override this to False
+    accumulates_statistics: bool = True
```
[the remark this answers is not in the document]
HDCharles: old

src/llmcompressor/observers/min_max.py
HDCharles: changed the whole thing

src/llmcompressor/entrypoints/model_free/process.py
HDCharles: updated

src/llmcompressor/modifiers/quantization/quantization/mixin.py
HDCharles: wanted it to be more descriptive, now is: sync_activation_stats

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: entirely got rid of this interface

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles: no

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles: yeah this whole section should be deleted, let me do that

HDCharles: ok done

src/llmcompressor/modifiers/quantization/quantization/mixin.py
HDCharles: weight observers can have activation statistics (iMatrix) that need to be synced

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles: this is how its been done, see how the fusion was ignored before? 

to do this properly would be crazily expensive, we'd have to do all fused modules together at the same time and solve an n dimensional optimization problem, rather than n 1 dimensional optimization problems. this is because the weight scales chosen for any module could alter the fused global scale and affect all the other error values.

Since we're just solving for a scale factor that alters the weights, we can say: that scale factor (based on the AWQ theory) should be largely insensitive to the specific rounding decisions that the correct global_scale will cause (fp8 rounding of the quantization scales) and so we can ignore it and only take it into account later when we're actually calculating quantization scales. So for MSE (calculate min/max)/AWQ (calculate weight scaling) we can ignore it while GPTQ requires it.

src/llmcompressor/observers/helpers.py
HDCharles: we never used it before, it was only ever set in GPTQ and was only set after qparams had been calculated. Just explicated this behavior now.

src/llmcompressor/observers/imatrix.py
HDCharles: fixed

HDCharles:
@brian-dellabetta that's why we're trying this lifecycle change, ideally the evals will tell us whether this is safe or not

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -110,19 +106,9 @@ def __call__(
...
```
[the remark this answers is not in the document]
HDCharles: this is an expected change

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -68,43 +66,34 @@ def on_initialize(self, state: State, **kwargs) -> bool:
...
+                m
+                for _, m in match_named_modules(
+                    state.model, self.resolved_targets, self.ignore
+                )
+            ]
+
+            self.sync_obs_act_stats(state.model)
+            self.update_activation_qparams(state.model)
+            observe(modules, base_name="weight")
+            update_qparams(modules, base_name="weight")
```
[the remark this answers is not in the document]
HDCharles: fixed this

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles: its more that we shouldn't be fusing anything in here since we don't have all the fused modules scaled yet.

src/llmcompressor/modifiers/transform/awq/base.py
brian-dellabetta: Fusion is done below after call_observer, there's an explicit call to `update_fused_layer_weight_global_scales(mapping.parent)` that you've deleted it on this diff. That was added in [this PR](https://github.com/vllm-project/llm-compressor/pull/2189). This is possible because the fused layers always occur in the same mapping.parent

src/llmcompressor/modifiers/transform/smoothquant/base.py
```diff
@@ -154,8 +154,6 @@ def on_event(self, state: State, event: Event, **kwargs):
             self._apply_smoothing(state.model)
 
         if event.type_ == EventType.CALIBRATION_EPOCH_END:
-            self._apply_smoothing(state.model)
```
brian-dellabetta, line 157: this is needed for basic pipeline

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -254,30 +252,16 @@ def on_event(self, state: State, event: Event, **kwargs):
             self._apply_smoothing(state.model)
 
         elif event.type_ == EventType.CALIBRATION_EPOCH_END:
-            self._apply_smoothing(state.model)
```
brian-dellabetta, line 257: same as smoothquant, this is needed for basic pipeline

src/llmcompressor/observers/helpers.py
brian-dellabetta: we never used `g_idx`? i see it in imatrix too on main. seems like it's important unless i'm missing that the logic is captured elsewhere in the refactor

src/llmcompressor/pipelines/data_free/pipeline.py
```diff
@@ -6,6 +6,7 @@
 
 from llmcompressor.core.session_functions import LifecycleCallbacks
 from llmcompressor.pipelines.registry import CalibrationPipeline
+from llmcompressor.pipelines.sequential.helpers import whole_model_subgraph
```
brian-dellabetta: i don't like the idea of importing from a complex pipeline like sequential into data-free or basic just so sequential epoch end can be called on the whole subgraph. This is all just to save the on_calibration_epoch_end extra calls to apply_smoothing in AWQ and equivalent in smoothquant?

src/llmcompressor/observers/helpers.py
HDCharles: this g_idx functionality in observers is so we can do: 'given weight W, if we permuted W's columns with g_idx, what would the scale and zero_point be'

in practice we never actually used this functionality, we would permute W and then pass it into the observer with g_idx = None or identity. We would do it manually.

src/llmcompressor/modifiers/transform/smoothquant/base.py
HDCharles: not anymore, I added calibration_epoch_end to the basic and datafree pipelines since otherwise we'd duplicate work in the sequential pipeline when we call both epoch_ends

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -195,22 +204,20 @@ def process_file_microscale_scheme(
...
+        observe(named_modules.values(), base_name="weight")
+        update_qparams(named_modules.values(), base_name="weight")
 
         for name, module in named_modules.items():
-            module_name, _ = name.rsplit(".", 1)
-            module.weight_global_scale.data.copy_(fused_global_scale)
-
-            # 2.2. finish calibration with fused global scale
-            calibrate_scale_zp(module)
+            freeze_module_quantization(module)
```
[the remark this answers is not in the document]
HDCharles: this is probably still correct, we'd want to keep the observers around until their fusion statistics are no longer needed.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -85,31 +84,17 @@ def quantize_weight(
...
+    observer = module.weight_observer
 
-    # standardize shape and dtype
-    match module:
-        case torch.nn.Conv2d():
-            W = W.flatten(1)
-        case transformers.Conv1D():
-            W.transpose_(0, 1)
     W = W.to(dtype=GPTQ_PRECISION)
     num_rows = W.shape[0]
```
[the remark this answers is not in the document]
HDCharles: it was intentional to get rid of those, we don't actually use that pathway so the mistake was not getting rid of it at the end.

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles: nah, its a sneaky change to the lifecycle where the basic pipeline calls sequentail_epoch_end

brian-dellabetta: Yeah, I left a comment on that below

src/llmcompressor/pipelines/data_free/pipeline.py
HDCharles: it was reobserving everything which is bad, no?

I can add a util file and import from there, it doesn't have to be from sequential pipeline.

src/llmcompressor/modifiers/transform/smoothquant/base.py
brian-dellabetta: we spoke offline about these changes

src/llmcompressor/modifiers/transform/awq/base.py
brian-dellabetta: spoke offline about these changes

src/llmcompressor/pipelines/data_free/pipeline.py
brian-dellabetta: spoke offline about these changes

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles: you're right! i fixed it, fusion is back on the menu. I didn't realize AWQ mapping corresponded with fusion groups.
I'm not confident fusion was correctly implemented before though, i think it would have needed something along the same lines as what is going on here where you do quantization after global_scale is fused

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -79,142 +78,86 @@ def initialize_observer(
...
-    :param value: torch.Tensor to be passed to the observer
-
     """
-    # If empty tensor, can't update zp/scale
-    # Case for MoEs
-    if value.numel() == 0:
+    if isinstance(module, Iterable):
+        for m in module:
+            update_qparams(m, base_name)
         return
```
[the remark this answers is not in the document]
kylesayrs: This is fine and documented behavior by the typehint

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -78,143 +77,87 @@ def initialize_observer(
...
-            calculate_gparam = True
-
-    call_observer(
-        module=module,
-        base_name=base_name,
-        value=value,
-        should_calculate_gparam=calculate_gparam,
-        should_calculate_qparams=calculate_qparams,
-    )
+    with align_module_device(module):
```
kylesayrs: `align_module_device` is deprecated. There is no benefit from using this context (or its equivalent, `disable_offloading`.

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -195,22 +204,20 @@ def process_file_microscale_scheme(
 
     # Compress fused modules with shared global scale
     for named_modules in fused_modules.values():
-        # 2.1. compute fused global scale across all members of the fused set
-        global_scales = [m.weight_global_scale for m in named_modules.values()]
-        fused_global_scale = torch.min(torch.cat(global_scales, dim=0))
+        # 2. fuse observers, observe weights, and get qparams
+        Observer.fuse([mod.weight_observer for mod in named_modules.values()])
```
kylesayrs, line 208: Should we use `fuse_weight_observers` here?

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -78,143 +77,87 @@ def initialize_observer(
...
-            update_offload_parameter(module, f"{base_name}_scale", scale)
-            if hasattr(module, f"{base_name}_zero_point"):
-                update_offload_parameter(module, f"{base_name}_zero_point", zero_point)
-
-
-def update_weight_global_scale(module: Module):
-    if getattr_chain(module, "quantization_scheme.weights", None) is None:
+    if isinstance(module, Iterable):
+        for m in module:
+            observe(m, base_name)
```
kylesayrs, line 97: Consider memoizing to avoid double observation in the shared modules case and infinite recursion in the cycle case

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -109,19 +105,9 @@ def __call__(
...
```
kylesayrs, line 110: In theory, we should enter `DisableQuantization` during calibration, and exit (ie turn on quantization) during propagation.

src/llmcompressor/pipelines/data_free/pipeline.py
```diff
@@ -33,4 +34,5 @@ def __call__(
         dispatch_model(model)
 
         LifecycleCallbacks.calibration_epoch_start()
+        LifecycleCallbacks.sequential_epoch_end(subgraph=whole_model_subgraph(model))
```
kylesayrs: FYI this has been changed to modules list on main. Hopefully this means we can kill `whole_model_subgraph`

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -51,17 +52,9 @@ def __call__(
 
         LifecycleCallbacks.calibration_epoch_start()
 
-        disable_qac = any(
-            type(mod).__name__ in DISABLE_QAC_MODIFIERS
-            for mod in session.lifecycle.recipe.modifiers
-        ) or getattr(dataset_args, "quantization_aware_calibration", True)
```
kylesayrs, line 57: We should deprecate `dataset_args.quantization_aware_calibration`, right? There should be a way to add a high level deprecation warning.

src/llmcompressor/observers/moving_base.py
```diff
@@ -1,130 +1,35 @@
...
+from compressed_tensors.quantization import QuantizationArgs
 from torch import distributed as dist
 
-from llmcompressor.observers.base import MinMaxTuple, Observer
+from llmcompressor.observers.base import Observer
 
 __all__ = ["MovingAverageObserverBase"]
 
 
 class MovingAverageObserverBase(Observer):
```
kylesayrs: Should this class not provide an implementation of `update_statistics`?

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -221,13 +213,12 @@ def on_event(self, state: State, event: Event, **kwargs):
                 self.on_start(state, None)
 
         if event.type_ == EventType.SEQUENTIAL_EPOCH_END:
-            QuantizationMixin.sync_activation_observers(self, state.model)
+            self.sync_obs_act_stats(state.model)
+            self.update_activation_qparams(state.model)
```
kylesayrs, line 240: Maybe `sync_obs_act_stats` should be folded into `update_activation_qparams`? Since you should always sync stats before updating activation qparams?

src/llmcompressor/observers/imatrix.py
```diff
@@ -56,18 +69,25 @@ def __init__(self, *args, **kwargs):
...
         """Attach a forward-pre hook to accumulate E[x²] per input channel.
 
-        If ``_imatrix_importance`` already exists on the module (second pass,
-        e.g. QuantizationModifier after IMatrixGatherer), skip hook attachment.
+        If raw accumulators (``_imatrix_sum`` / ``_imatrix_count``) already
+        exist on the module (second pass after IMatrixGatherer), copy them
+        to the observer and skip hook registration.
         """
-        if hasattr(module, "_imatrix_importance"):
+        self._module_ref = weakref.ref(module)
```
kylesayrs: This is no longer used

src/llmcompressor/observers/helpers.py
```diff
@@ -154,3 +148,48 @@ def _flatten_attention(value: torch.Tensor, args: QuantizationArgs):
...
+            observers = []
+            for layer in layers:
+                obs = getattr(layer, "weight_observer", None)
+                if obs is None:
+                    break
+                if obs.args.strategy != QuantizationStrategy.TENSOR_GROUP:
+                    break
+                observers.append(obs)
+            else:
+                Observer.fuse(observers)
```
kylesayrs, line 195: This function along with `Observer.fuse` is slightly dangerous. If `Observer.fuse` is called multiple times, it will continue to append to `_fused_observers`. Consider making `Observer.fuse` overwrite existing fusions, or adding validation to make it as safe as possible.

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -78,143 +77,87 @@ def initialize_observer(
...
+            return
+        if not observer.has_statistics:
+            return
+
+        # Dynamic (activation) quantization: only store global_scale, not scale/zp
+        args = observer.args
+        is_dynamic = getattr(args, "dynamic", False) in (True, DynamicType.LOCAL)
+
+        qparams = observer.get_qparams()
+        for param_name, param_val in qparams.items():
```
kylesayrs: You might need to be slightly careful about iterating over a non-ordered structure like `dict`. I've run into DDP desynchronization issues in the past where different ranks would iterate in different orders, leading to deadlocks.

kylesayrs (commented):
Overall looks good, I would make a suggestion w.r.t. to observer structure which might help dramatically decrease lifecycle complexity.

## Global Scale Handling ##
I think the approach to global scale handling is good, I like the upfront fusion of global scales such that future observers don't have to worry about it.

## Observers/ Lifecycle ##

I see what you're saying about separating weight and activation observers. Treating activation observers and weight observers as the same class feels kind of silly.

If we separate them into separate classes, then I think the implementation and lifecycle becomes a lot cleaner:

Observers:
```python3
@Observer.register(("weight", "minmax"))
class WeightObserver:
  def update_statistics():
    # update minmax

  def update_qparams():
    # calculate qparams
    # delete statistics

@Observer.register(("activation", "minmax"))
class ActivationObserver:
  def update_statistics():
    # moving average logic

  def update_qparams():
    # sync across ranks
    # only rank0 updates offload
    # delete statistics if not qat
```

Lifecycle:
```python3
def on_start():
  fuse_weight_observers(model)

def on_sequential_epoch_end():
    calculate_qparams(module)

def calculate_qparams(module):
  if hasattr(module, "weight_observer"):
    ... = module.weight_observer.update_params()
  if hasattr(module, "input_observer"):
    ... = module.input_observer.update_params()
```

src/llmcompressor/observers/imatrix.py
HDCharles: imatrix has to get the nonsense off the observer so its still needed until the imatrix refactor

src/llmcompressor/modifiers/gptq/base.py
HDCharles: activation stats can exist on weight observers so they're doing slightly different things

1) sync all activation information on all observers weight and activation alike
2) update qparams for activation observers only

src/llmcompressor/observers/moving_base.py
HDCharles: the two moving_base observers, MovingAverageMSEObserver and MinMaxObserver (OMG this naming)

each update statistics differently, they can't really do the same thing. They could both implement a lerp function maybe

src/llmcompressor/pipelines/data_free/pipeline.py
HDCharles: yes, fixed

src/llmcompressor/entrypoints/model_free/process.py
HDCharles: that's intended for use on a model, i don't think it would work

HDCharles:
@kylesayrs about lifecycle stuff, i think we discussed this offline right?

for posterity, the issue is activation stats aren't limited to activation observers. Imatrix obs takes activation info into account when it does quantization

so if we put the sync into get_qparams we'd run into issues when we try to do async compression like GPTQ we can't have a sync op, we need to sync those stats ahead of time.

src/llmcompressor/observers/helpers.py
HDCharles: nah it checks for existing obs and only adds new ones

```python
    @staticmethod
    def fuse(observers: Iterable["Observer"]) -> None:
        """
        Link all observers in the list with each other for shared global_scale.

        :param observers: list of observers to fuse together
        """
        observers = list(observers)
        for obs in observers:
            for other in observers:   
               if other is not obs and other not in obs._fused_observers: ######<<<<<
                    obs._fused_observers.append(other)
```

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: fixed

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: wait whats the failure mode here? an infinite iterable?

we either have a module or an iterable of modules, not children of hte module

src/llmcompressor/pipelines/sequential/pipeline.py
HDCharles: didn't we not want to do this? and just ignore quantization propagation?

src/llmcompressor/args/utils.py
```diff
@@ -57,13 +57,19 @@ def parse_args(
...
-    # raise depreciation warnings
+    # raise deprecation warnings
     if dataset_args.remove_columns is not None:
         logger.warning(
-            "`remove_columns` argument is depreciated. When tokenizing datasets, all "
+            "`remove_columns` argument is deprecated. When tokenizing datasets, all "
             "columns which are invalid inputs the tokenizer will be removed",
             DeprecationWarning,
         )
+    if dataset_args.quantization_aware_calibration is not True:
```
kylesayrs:
```suggestion
    if not dataset_args.quantization_aware_calibration:
```

src/llmcompressor/observers/helpers.py
kylesayrs: Got it, thanks. We should consider making `_fused_observers` a `set` then.

src/llmcompressor/observers/base.py
```diff
@@ -1,198 +1,162 @@
...
-        (weight, activation, or attention state).
+        Compute quantization parameters from accumulated statistics.
 
-        :param observed: value being observed
-        :return: calibrated scale and zero point
+        :return: dict with keys "scale", "zero_point", and "global_scale"
         """
-        scales, zero_points, _min, _max = self._forward_with_minmax(observed)
-        return (scales, zero_points)
+        return self.compute_qparams_from_statistics()
```
kylesayrs: Seems like this is a very thin wrapper?

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs: Spoke offline, I'm fine releasing with this configuration but one of the theoretical benefits of propagation is that future layers can adapt to the effects of quantization of the previous layers

src/llmcompressor/pipelines/data_free/pipeline.py
```diff
@@ -33,4 +33,6 @@ def __call__(
         dispatch_model(model)
 
         LifecycleCallbacks.calibration_epoch_start()
+        modules = list(model.modules())
+        LifecycleCallbacks.sequential_epoch_end(modules)
```
kylesayrs: copy style from basic pipeline
```suggestion
        LifecycleCallbacks.sequential_epoch_end(list(model.modules()))
```

src/llmcompressor/observers/moving_base.py
kylesayrs: I disagree? Both implementations copy exactly this same logic

```python3
if hasattr(self, "min_vals") and self.avg_constant != 1.0:
    min_vals = self._lerp(self.min_vals, min_vals, self.avg_constant)
    max_vals = self._lerp(self.max_vals, max_vals, self.avg_constant)
  
  self.min_vals = min_vals
  self.max_vals = max_vals
```

src/llmcompressor/observers/imatrix.py
kylesayrs: I don't see where in the codebase where this variable is used?

src/llmcompressor/modifiers/quantization/calibration.py
kylesayrs: If you call `observe(model.modules())`, then shared modules will be observed twice.

kylesayrs (commented):
Looks good to me, last comment on observers:

Seems like (`get_qparams`, `compute_qparams_from_statistics`) and (`update_statistics`, `forward`) are all very similar, but all subtly different. Maybe it would be possible to consolidate to just `get_qparams` and `forward`?

src/llmcompressor/observers/base.py
HDCharles: I think it's a bit clearer like this than if we had each observer implement get qparams, no?

HDCharles:
> Looks good to me, last comment on observers:
> 
> Seems like (`get_qparams`, `compute_qparams_from_statistics`) and (`update_statistics`, `forward`) are all very similar, but all subtly different. Maybe it would be possible to consolidate to just `get_qparams` and `forward`?

We could do that, in general I'd rather have thin wrappers that give good ux and clear functions that explicate responsibilities, than the tradeoff of picking one or the other.

src/llmcompressor/observers/moving_base.py
HDCharles: yeah thats the lerp function i mentioned. but again thats not an update statistics function because update_statistics takes an input of observed and both of the actual observers do something different with observed.

the abstraction here is that each observer has a specific way to take observed and update statistics and another to take statistics and calculate qparams. moving average specifies how to update statistics from other statistics, not from observed.

src/llmcompressor/observers/imatrix.py
HDCharles: oh you're right, sorry i spent some time trying to get rid of all observer logic needing to be aware of the module but i couldn't do it until the refactor. You're right that this is not needed. I misremembered what was/wasn't needed.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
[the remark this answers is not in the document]
HDCharles: It was cleaned up

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: Ok, can do

src/llmcompressor/observers/base.py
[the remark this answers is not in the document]
HDCharles: We want explicit strong refs

src/llmcompressor/observers/helpers.py
```diff
@@ -156,3 +148,48 @@ def _flatten_attention(value: torch.Tensor, args: QuantizationArgs):
...
```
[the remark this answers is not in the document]
HDCharles: We have reduplicated fused observers

src/llmcompressor/observers/helpers.py
HDCharles: Ya will do

src/llmcompressor/observers/imatrix.py
```diff
@@ -20,13 +23,21 @@
...
+    Falls back to uniform MSE when importance data is unavailable.
+
+    Importance is accumulated as raw ``_imatrix_sum`` / ``_imatrix_count``
+    and synced across DDP ranks via ``_sync_dict`` before observation.
     """
 
+    _act_sync_dict = {
+        "_imatrix_sum": dist.ReduceOp.SUM,
+        "_imatrix_count": dist.ReduceOp.SUM,
+    }
```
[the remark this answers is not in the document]
HDCharles: Fixed

src/llmcompressor/observers/base.py
HDCharles: compute_qparams_from_statistics is no more

HDCharles:
ok i think i've addressed everything

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -68,43 +67,25 @@ def on_initialize(self, state: State, **kwargs) -> bool:
...
-
     def on_event(self, state: State, event: Event, **kwargs):
         if event.type_ == EventType.CALIBRATION_EPOCH_START:
             if not self.started_:
                 self.on_start(state, None)
 
         if event.type_ == EventType.SEQUENTIAL_EPOCH_END:
-            QuantizationMixin.sync_activation_observers(self, state.model)
+            tmp = match_named_modules(state.model, self.resolved_targets, self.ignore)
+            modules = [m for _, m in tmp]
```
brian-dellabetta, line 82: nit
```suggestion
            modules = [m for _, m in match_named_modules(state.model, self.resolved_targets, self.ignore)]
```

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -78,143 +76,99 @@ def initialize_observer(
...
-
-
-def update_weight_global_scale(module: Module):
-    if getattr_chain(module, "quantization_scheme.weights", None) is None:
+    if isinstance(module, Iterable):
+        # iterate through but avoid duplicates
+        seen = set()
+        for m in module:
+            if id(m) not in seen:
+                seen.add(id(m))
```
brian-dellabetta: nit -- set can take an iterable.
```suggestion
        unique_modules = set(module)
        for m in unique_modules:
```

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -78,143 +76,99 @@ def initialize_observer(
...
-
+    :only_update_onload: option to only update the onloaded value, useful
+        when we want to do a temporary update or in DDP situations where
+        we want only want one rank to update the offload+onload to avoid
+        multiple writes to the offload (rest just update onload)
     """
-    # If empty tensor, can't update zp/scale
-    # Case for MoEs
-    if value.numel() == 0:
+    if isinstance(module, Iterable):
```
brian-dellabetta, line 126: you have to use a set in `def observe` above to only call once on each module, that's not needed here as well?

src/llmcompressor/modifiers/transform/awq/base.py
brian-dellabetta: awesome, looks good! will resolve this

src/llmcompressor/modifiers/gptq/base.py
kylesayrs: I see, I guess this is specifically for the imatrix observer. It sucks that we need to separate them, as it makes the lifecycle/ usage more complicated

src/llmcompressor/pipelines/basic/pipeline.py
brian-dellabetta: are we deprecating because we're sure it doesn't affect anything, or because it's hard to maintain this flexibility? I just wonder if it holds for all cases, especially long-context, which most of our evals/regression tests aren't covering

kylesayrs (approved):
Awesome, great work

brian-dellabetta (approved):
Super nice that we can remove so much of the complexity in the modifier level, really helps with readability. some nits, and a slight lingering concern about the move to always disable quant. it seems like the deprecation of quantization_aware_calibration could be done separate to these changes. But if we're confident in the evals then I think it's good to land

src/llmcompressor/modifiers/quantization/quantization/base.py
HDCharles: Nah I find the before to be more readable

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: fixed

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -267,16 +265,13 @@ def on_end(self, state: State, event: Event, **kwargs):
...
-                    update_weight_global_scale(layer)
-        for mapping in self._resolved_mappings:
-            update_fused_layer_weight_global_scales(mapping.parent)
-        for mapping in self._resolved_mappings:
-            for layer in mapping.balance_layers + [mapping.smooth_layer]:
-                if hasattr(layer, "weight_observer"):
-                    update_weight_zp_scale(layer)
+                    smoothed_layers.append(layer)
+        observe(smoothed_layers, base_name="weight")
+        update_qparams(smoothed_layers, base_name="weight")
```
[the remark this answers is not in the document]
HDCharles: fixed

tests/llmcompressor/observers/test_distributed.py
```diff
@@ -19,107 +20,114 @@ def _make_observer(cls, **kwargs):
...
-    observer.past_global_max_vals = torch.tensor([2.0])
+    observer.min_vals = torch.tensor([-1.0])
+    observer.max_vals = torch.tensor([1.0])
 
-    comms = observer.synchronize()
-    assert len(comms) == 4
-    assert mock_dist.all_reduce.call_count == 4
+    comms = observer.sync_activation_stats()
+    assert len(comms) == 2
+    assert mock_dist.all_reduce.call_count == 2
```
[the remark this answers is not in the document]
HDCharles: fixed

kylesayrs (approved):

Issue #2584, opened 2026-04-08
[Bug]: Using `AWQModifier` with `kv_cache_scheme` fails.


siahuat0727:
### ⚙️ Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
### Environment Information ###
Operating System: `Linux-5.15.0-119-generic-x86_64-with-glibc2.35`
Python Version: `3.13.9 | packaged by Anaconda, Inc. | (main, Oct 21 2025, 19:16:10) [GCC 11.2.0]`
llm-compressor Version: `0.10.1.dev72+g6f796f08`
compressed-tensors Version: `0.14.1a20260326`
transformers Version: `5.5.0`
torch Version: `2.10.0`
CUDA Devices: `['NVIDIA H200', 'NVIDIA H200', 'NVIDIA H200', 'NVIDIA H200', 'NVIDIA H200', 'NVIDIA H200', 'NVIDIA H200', 'NVIDIA H200']`
AMD Devices: `None`
NPU Devices: `None````

</details>


### 🐛 Describe the bug



Using `AWQModifier` with `kv_cache_scheme` fails.

The first observed error is:

```text
AttributeError: 'NoneType' object has no attribute 'strategy'
```

from the `duo_scaling` validation in `AWQModifier.on_initialize()`, where attention modules introduced by `kv_cache_scheme` can have `quantization_scheme.weights = None`.

While debugging locally, I modified the guard in `AWQModifier.on_initialize()` to skip modules whose `quantization_scheme.weights` is `None`. That avoided the initial `AttributeError`, but the run then failed later during calibration with a `RecursionError`:
```diff
-                    and hasattr(module.quantization_scheme, "weights")
+                    and module.quantization_scheme.weights is not None
```

```text
RecursionError: maximum recursion depth exceeded
```

This appears to happen because AWQ smoothing replay reuses `past_key_values` after it has been replaced with `QuantizedKVCache`, so the same attention module can receive its own `QuantizedKVCache` and recurse.

### 🛠️ Steps to reproduce

```python
from datasets import Dataset
from llmcompressor import oneshot
from llmcompressor.modifiers.awq import AWQModifier
from compressed_tensors.quantization import QuantizationArgs, QuantizationType, QuantizationStrategy
from transformers import AutoModelForCausalLM, AutoTokenizer

[...]

PR #2591, opened 2026-04-09
[bugfix][awq] fix kv_cache + awq bug

HDCharles:
intermediatescache would get handed a quantizedkv_cache and misbehave.

added functionality to ignore any quantized kv cache that the hook would pick up.

partial fix for https://github.com/vllm-project/llm-compressor/issues/2584

added smoke test for AWQ + kv cache quantization

test plan:

see unit test

### DIFF
--- a/tests/llmcompressor/modifiers/awq/test_base.py
+++ b/tests/llmcompressor/modifiers/awq/test_base.py
@@ -15,6 +15,7 @@ from llmcompressor.modifiers.awq import AWQMapping, AWQModifier
 from llmcompressor.modifiers.awq.base import get_lowest_common_ancestor_with_avoid
 from llmcompressor.modifiers.factory import ModifierFactory
 from llmcompressor.utils import get_high_precision
+from tests.testing_utils import requires_gpu
 
 
 @pytest.mark.unit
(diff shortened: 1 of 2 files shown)


kylesayrs (commented):
I think that caching the `Cache` at all is non-ideal, as you're potentially caching a very large amount of memory for no reason (infact, I would expect this to change the behavior of grid search forward passes). I think you should instead consider filtering out any values which are instances of `Cache`.

tests/llmcompressor/modifiers/awq/test_base.py
```diff
@@ -674,3 +674,47 @@ def test_block_strategy_compute_layer_means(rows, cols, block_height, block_widt
     # check
     assert_close(llmc_awq_means, ref_means, atol=1e-5, rtol=1e-5)
     assert_close(llmc_awq_means, auto_awq_means, atol=1e-5, rtol=1e-5)
+
+
+@pytest.mark.smoke
+def test_awq_with_kv_cache_quantization():
```
kylesayrs, line 682: This is a pretty lengthy test for a simple change, do you want to consider making this a unit test against `_setup_activation_cache_hooks`?

HDCharles: regardless of the size of the gap found when someone actually tried kv+awq this seems like something we should have a test for.

HDCharles: do you want it elsewhere?

brian-dellabetta (approved):

tests/llmcompressor/modifiers/awq/test_base.py
kylesayrs: Ugh yeah, ideally in a refactored oneshot tests but here is fine for now

kylesayrs (approved):
Technically you should never see a base class instance since we disable this during calibration, so it’s safe in use

PR #2593, opened 2026-04-09
Refactor torch dependency condition in setup.py

dsikka:
SUMMARY:

TEST PLAN:


brian-dellabetta (approved):

dsikka:
@coderabbitai why is torch 2.10 being installed and not 2.11?

dsikka:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> `@dsikka` Great question! The reason torch 2.10 was being installed instead of 2.11 is because of the **upper-bound version caps** that existed in `setup.py` **before** this PR:
> 
> * For **release** builds: `torch>=2.9.0,<=2.10.0` — explicitly capped at 2.10.0
> * For **non-release** builds: `torch>=2.9.0,<2.11` — excluded 2.11 and above
> 
> PyTorch 2.11.0 was released on March 23, 2026, and 2.10.0 was released on January 21, 2026. So when the old constraints were in place, pip would resolve to the highest allowed version — torch 2.10.x — and would **not** install 2.11 since it was explicitly excluded.
> 
> This PR removes those upper bounds, changing both constraints to simply `torch>=2.9.0`, which will now allow pip to install **2.11.0** (or any future version). That's the purpose of the refactor — to stop pinning torch to a fixed maximum and allow newer releases to be picked up automatically.



> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> `@dsikka` Great question! The reason torch 2.10 was being installed instead of 2.11 is because of the **upper-bound version caps** that existed in `setup.py` **before** this PR:
> 
> * For **release** builds: `torch>=2.9.0,<=2.10.0` — explicitly capped at 2.10.0
> * For **non-release** builds: `torch>=2.9.0,<2.11` — excluded 2.11 and above
> 
> PyTorch 2.11.0 was released on March 23, 2026, and 2.10.0 was released on January 21, 2026. So when the old constraints were in place, pip would resolve to the highest allowed version — torch 2.10.x — and would **not** install 2.11 since it was explicitly excluded.
> 
> This PR removes those upper bounds, changing both constraints to simply `torch>=2.9.0`, which will now allow pip to install **2.11.0** (or any future version). That's the purpose of the refactor — to stop pinning torch to a fixed maximum and allow newer releases to be picked up automatically.

@coderabbitai yes but with the removal of the cap, the workflows on this PR should also install torch 2.11 but they are not. Why is this happening?

kylesayrs (approved):
I really don't see any point to leaving some of these library versions unbounded, such as torch and torchcodec

Every time torch releases they do it in a way that breaks our tests because we weren't bounded. It's not like we depend on the latest torch features all the time, I think manual/ (automated + checks) are both better solutions than unbounded libs

PR #2594, opened 2026-04-09
[linkspector] only run action if there are changes to markdown files

brian-dellabetta:
SUMMARY:
Update the linkspector github action to only run if markdown files have been changed in the PR. This action can be flaky and go red when github is unresponsive, and really only needs to be checked if markdown files have been altered.

TEST PLAN:
n/a


HDCharles:
Does this catch changed docstrings?

HDCharles (approved):

brian-dellabetta:
> Does this catch changed docstrings?

@HDCharles just took a look at docs [here](https://github.com/UmbrellaDocs/linkspector?tab=readme-ov-file#configuration), linkspector only checks markdown files

dsikka (commented):
Have we tested this still works as expected for cases where the file does change?

brian-dellabetta:
> Have we tested this still works as expected for cases where the file does change?

@dsikka [created a PR with broken link](https://github.com/vllm-project/llm-compressor/pull/2626) to confirm, test is failing [here](https://github.com/vllm-project/llm-compressor/actions/runs/24532204184/job/71718065090?pr=2626)

kylesayrs (approved):

PR #2602, opened 2026-04-10
[docs] deepseek v3.2 docs

brian-dellabetta:
SUMMARY:
Follow-up from #2491 

Add docs for DeepSeek v3.2 in the README.md what's new section, and in key models, highlighting unique steps needed for DeepSeek v3.2 support

TEST PLAN:
n/a


README.md
```diff
@@ -37,6 +37,7 @@ Big updates have landed in LLM Compressor! To get a more in-depth look, check ou
 
 Some of the exciting new features include:
 
+* **DeepSeek V3.2 Support**: DeepSeek V3.2 can now be quantized using LLM Compressor with support for disk offloading to handle the 671B parameter model. The example demonstrates mixed-precision quantization with NVFP4 for MLP layers and FP8-BLOCK for attention layers. See the [DeepSeek V3.2 docs](llm-compressor/docs/key-models/deepseek-v32) for details. A pre-quantized checkpoint is available at [RedHatAI/DeepSeek-V3.2-NVFP4-FP8-BLOCK](https://huggingface.co/RedHatAI/DeepSeek-V3.2-NVFP4-FP8-BLOCK).
```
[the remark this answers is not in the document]
brian-dellabetta: woops, typo, fixed

docs/key-models/deepseek-v32/nvfp4-fp8-example.md
```diff
@@ -0,0 +1,154 @@
...
+DATASET_ID = "ultrachat-200k"
+DATASET_SPLIT = "train_sft"
+NUM_CALIBRATION_SAMPLES = 20
+MAX_SEQUENCE_LENGTH = 2048
+
+recipe = QuantizationModifier(
+    config_groups={
+        "config_group_0": QuantizationScheme(
+            targets=[
+                r"re:model.*mlp.*(gate|up|down|gate_up)_proj$",
```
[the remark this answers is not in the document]
brian-dellabetta: needed for vllm

docs/key-models/deepseek-v32/nvfp4-fp8-example.md
```diff
@@ -0,0 +1,154 @@
...
+### 2. Merge Config Files
+
+DeepSeek splits important config info into a separate file that will break loading in transformers if not merged into `config.json`.
+
+```python
+import json
+
+with open(f"{BFLOAT16_SAVE_DIR}/config.json", "r") as f:
+    orig_config = json.load(f)
+with open(f"{BFLOAT16_SAVE_DIR}/inference/config_671B_v3.2.json", "r") as f:
```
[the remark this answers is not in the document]
brian-dellabetta: custom hackery required for deepseek v3.2, as called out in docs

docs/key-models/deepseek-v32/nvfp4-fp8-example.md
```diff
@@ -0,0 +1,154 @@
...
+1. Convert checkpoint to bfloat16
+2. Merge config files
+3. Load model with disk offloading
+4. Configure quantization algorithm and scheme
+5. Apply quantization
+6. Save to disk in compressed-tensors format
+
+### 1. Convert Checkpoint to bfloat16
+
+The original `deepseek-ai/DeepSeek-V3.2` checkpoint has layers quantized in the FP8_BLOCK scheme. We first convert it to bfloat16 so that we can compress it in compressed-tensors format with a different configuration.
```
dsikka, line 15: why did we need to convert to bfloat16? and not just rename the scales?

dsikka (commented):
One question

docs/key-models/deepseek-v32/nvfp4-fp8-example.md
brian-dellabetta: FP8_BLOCK has a weight and a weight_scale_inv. we need to multiply weight by weight_scaled_inv, for each block, in high precision, then cast to bfloat16. that logic is [here](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/entrypoints/convert/converters/fp8block_dequantizer.py#L111-L158) in the FP8BlockDequantizer

dsikka: I don't follow - why can't we just keep the original quantization and just rename the scales for the CT format, if we're maintaining the same scheme?

brian-dellabetta: oh i see, you mean to convert from their FP8_BLOCK to our FP8_BLOCK. We can do that, we have a ModelOptNVFP4Converter that does something like that, but our desired output format here is NVFP4, so to get their we upconvert first to bfloat16, then run NVFP4 with calibration dataset in oneshot

dsikka (changes requested):
Note - the goal is to make custom def support reproducible this qaurter - it would be good to also talk about the points listed here: https://github.com/vllm-project/llm-compressor/tree/main/src/llmcompressor/modeling/deepseekv32 and maybe provide a bit more detail / context as opposed to just listing a ton of point which can be difficult to parse through

PR #2604, opened 2026-04-10
Add FP8 quantization example for Qwen3.5 MoE

dsikka:
## Summary

- Adds a data-free `FP8_DYNAMIC` quantization example for the Qwen3.5-122B-A10B sparse MoE model
- Includes example script at `examples/quantization_w8a8_fp8/qwen3_5_moe_example.py`
- Adds corresponding docs page at `docs/key-models/qwen3.5/fp8-example.md`
- Links the new FP8 example from the Qwen3.5 key-models index

## Test plan

- [ ] Script runs end-to-end on Qwen3.5-122B-A10B
- [ ] MTP layers are correctly preserved via `save_mtp_tensors_to_checkpoint`
- [ ] Docs page renders correctly and is reachable from the index

### DIFF
--- a/docs/key-models/qwen3.5/fp8-example.md
+++ b/docs/key-models/qwen3.5/fp8-example.md
@@ -1,0 +1,68 @@
+## Qwen3.5 FP8 Example
+
+This example quantizes the Qwen3.5-122B-A10B sparse MoE model to FP8 (weights and activations quantized to FP8) using data-free PTQ.
+
+NOTE: This example requires `transformers >= v5`.
+
+### Code Walkthrough
+
+Let's walk through the main steps of the quantization process:
+1. Load model
+2. Configure quantization algorithm and scheme
+3. Apply quantization
+4. Save to disk in compressed-tensors format
+
+### 1. Load Model
+
+```python
+from compressed_tensors.utils import save_mtp_tensors_to_checkpoint
+from transformers import AutoProcessor, Qwen3_5MoeForConditionalGeneration
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+
+MODEL_ID = "Qwen/Qwen3.5-122B-A10B"
+
+model = Qwen3_5MoeForConditionalGeneration.from_pretrained(MODEL_ID, dtype="auto")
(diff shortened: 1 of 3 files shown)


dsikka:
@coderabbitai why was model_free_ptq added as label

dsikka:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> # Check PR labels and events
> gh pr view 2604 --json labels,timeline
> 
> # Check for labeler config
> cat .github/labeler.yml 2>/dev/null || echo "No labeler.yml found"
> fd -i "labeler" .github/ --exec cat {}
> ```
> 
> Repository: vllm-project/llm-compressor Length of output: 96
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> # Check workflows for any labeling logic
> fd -e yml -e yaml . .github/workflows/ --exec grep -l "label" {} 2>/dev/null
> echo "---"
> # Search for model_free_ptq anywhere in the repo config
> rg "model.free.ptq" .github/ 2>/dev/null || echo "Not found in .github/"
> rg "model.free.ptq" 2>/dev/null | head -30
> echo "---"
> # List all workflow files
> ls .github/workflows/ 2>/dev/null
> echo "---"
> # Check any labeling workflows
> fd -e yml -e yaml . .github/ --exec cat {}
> ```
> 
> Repository: vllm-project/llm-compressor Length of output: 25837
> 
> `@dsikka` The `model_free_ptq` label was applied by **CodeRabbit's auto-labeling** feature. Since this PR adds a data-free `FP8_DYNAMIC` quantization example, and the `model_free_ptq` pathway is specifically designed for data-free schemes like FP8 Dynamic/Block, CodeRabbit semantically associated the PR with that label.
> 
> However, looking at the actual example script (`examples/quantization_w8a8_fp8/qwen3_5_moe_example.py`), it uses the `oneshot` entrypoint rather than `model_free_ptq` — so the label is likely a **false positive**. I can remove it if you'd like, or if you'd prefer I stop applying the `model_free_ptq` label automatically in the future, I can record that as a guideline.

@coderabbitai yes remove the label and only apply it when the model_free_ptq pathway is applied or updated

PR #2608, opened 2026-04-12
fix: support transformers >= 5.0 (TORCH_INIT_FUNCTIONS fallback)

quivent:
## Summary

- Fix `TORCH_INIT_FUNCTIONS` import for transformers >= 5.0 — the existing fallback to `transformers.initialization` doesn't exist in 5.5.x. Added a third fallback that rebuilds from `torch.nn.init`.
- Relax dev version pin from `<=4.57.6` to allow transformers 5.x

## Motivation

New model architectures like Qwen3.5 (`qwen3_5`) only exist in transformers >= 5.0. The current pin blocks quantization of these models entirely.

## Testing

- All 126 llmcompressor Python modules import cleanly with transformers 5.5.3
- Full W4A16 quantization of Qwen3.5-27B completed successfully (496 layers, 18.6GB output)
- `compressed-tensors` 0.14.0.1 is fully compatible with transformers 5.5.3 (verified all 11 import sites)

## Details

The only two breaking imports between llm-compressor and transformers 5.x are:

1. **`TORCH_INIT_FUNCTIONS`** — removed from `transformers.modeling_utils`, not in `transformers.initialization` either. Fixed with `torch.nn.init` rebuild.
2. **`Conv1D`** — moved to `transformers.pytorch_utils`. Already handled by existing try/except in the installed package (the `wanda_sparsify.py` uses `transformers.Conv1D` which still works).

Reference implementation and testing details: https://github.com/quivent/llmcompressor-transformers5

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -125,7 +125,7 @@ setup(
         (
             "transformers>=4.56.1,<=4.57.6"
             if BUILD_TYPE == "release"
-            else "transformers>=4.56.1,<=4.57.6"
+            else "transformers>=4.56.1"
         ),
         ("datasets>=4.0.0,<=4.6.0" if BUILD_TYPE == "release" else "datasets>=4.0.0"),
         (
--- a/src/llmcompressor/utils/dev.py
+++ b/src/llmcompressor/utils/dev.py
@@ -14,11 +14,23 @@ from safetensors.torch import save_file
 from transformers import AutoModelForCausalLM, PreTrainedModel
 
 try:
-    # Transformers < v5 support
     from transformers.modeling_utils import TORCH_INIT_FUNCTIONS
 except ImportError:
-    # Transformers v5 support
-    from transformers.initialization import TORCH_INIT_FUNCTIONS
+    try:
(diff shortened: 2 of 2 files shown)


dsikka:
I still see it? 
https://github.com/huggingface/transformers/blob/75d3bdcd4b3ba70cee21287218d9764f33da41f0/src/transformers/initialization.py#L24

setup.py
```diff
@@ -125,7 +125,7 @@ def localversion_func(version: ScmVersion) -> str:
         (
             "transformers>=4.56.1,<=4.57.6"
             if BUILD_TYPE == "release"
-            else "transformers>=4.56.1,<=4.57.6"
+            else "transformers>=4.56.1"
```
dsikka, line 128: We can't uncap transformers support as many of the existing models that are supported have to be migrated

gkumbhat:
Hi @quivent @dsikka is the work on this effort still ongoing,  seems like it was may be paused?

kylesayrs:
No longer relevant as of https://github.com/vllm-project/llm-compressor/pull/2647

PR #2610, opened 2026-04-12
[CLOSED] Add actorder support for GPTQ block quantization

### LINKED ISSUE #2587: Add weight activation ordering for fp8 block
Followup for https://github.com/vllm-project/llm-compressor/issues/2520

 to add weight activation ordering since fp8 block is now supported.

The only change would be
g_idx = range / groupsize -> g_idx = range / block_width

this would actually consolidate the existing code because as of now we have 2 ways to map columns, g_idx and the calculation for fp8 block which is doing the same thing as g_idx but as a one off. If we change it to use g_idx then we get actorder almost for free

1) use g_idx for fp8 block indexing
2) add fp8 block + act order

rk119:
SUMMARY:

- Closes:  #2587 

- Replaced the separate `block_column_idx` calculation in block quantization by reusing `g_idx`, first getting the divisor from either the group size value or the block width value. This should automatically ensure the use of actorder.

- Remove `has_gidx` from and simplified the logic to save g_idx if `ActivationOrdering.GROUP` is opted for.

TEST PLAN:

- Added an additional block quantization recipe in `tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py`.
- Ran the relevant tests to ensure that the changes do not introduce inconsistencies or errors.

CHANGES PENDING:

- Address errors.
- Add more relevant recipes in the tests and ensure that all tests pass.

### DIFF
--- a/examples/quantization_w8a8_fp8/qwen2vl_example.py
+++ b/examples/quantization_w8a8_fp8/qwen2vl_example.py
@@ -17,7 +17,7 @@ processor = AutoProcessor.from_pretrained(MODEL_ID)
 recipe = QuantizationModifier(
     targets="Linear",
     scheme="FP8_DYNAMIC",
-    ignore=["re:.*lm_head", "re:visual.*"],
+    ignore=["re:.*lm_head", "re:.*visual.*"],
 )
 
 # Apply quantization and save to disk in compressed-tensors format.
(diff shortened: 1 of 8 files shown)


rk119:
Closed this due to a tangled git history during a rebase. I have recreated this draft PR cleanly here: #2616

PR #2614, opened 2026-04-13
[Tests] Add transformers v5 modeling tests and clean up import guards

dsikka:
## Summary

- Removes `pytest.importorskip` / try-except import guards from `test_calib_glm4_moe`, `test_calib_qwen3_next`, and `test_calib_qwen3_vl_moe` as they are all supported 
- Adds Gemma4 and Qwen3.5 example script test cases to `test_example_scripts.py`
- Adds Gemma4 and Qwen3.5 modeling tests
- Fixes `qwen3_5_moe.py` gate weight initialization to propagate `device` alongside `dtype`
- In order to enable proper testing of the examples and modeling tests with the valid transformers versions, add a requires_v5/v4 marker which skips the test if the incorrect version of transformers is not installed 
     - v5 models do not exist in transformers <v5 and our current v4 models have not been updated to be compatible with v5 
     - Our CI/CD tests both latest transformers and our pinned transformers - the marker will allow proper skipping and can be removed once all modeling definitions have been updated


brian-dellabetta (approved):
lgtm, bummer that we have to handle each model calibration on a case-by-case basis to migrate fully. hopefully not a huge lift

PR #2616, opened 2026-04-14
Add actorder support for GPTQ block quantization

### LINKED ISSUE #2587: Add weight activation ordering for fp8 block

rk119:
SUMMARY:

- Depends on: https://github.com/vllm-project/compressed-tensors/pull/682

- Closes:  #2587 

- Replace the separate `block_column_idx` calculation in block quantization by reusing `g_idx`, first getting the divisor from either the group size value or the block width value. This should automatically ensure the use of actorder.

- Remove `has_gidx` from and simplified the logic to save g_idx if `ActivationOrdering.GROUP` is opted for.

TEST PLAN:

https://gist.github.com/HDCharles/dbe9b89ce5e44cd38c0790ab82c73af7
```

model                                scheme       task                  no-actorder         weight   wt vs base
---------------------------------------------------------------------------------------------------------------
Meta-Llama-3-8B-Instruct             FP8_BLOCK    gsm8k                      67.40%         67.17%       -0.34%
Meta-Llama-3-8B-Instruct             FP8_BLOCK    gsm8k_platinum             68.73%         70.64%       +2.78%
Meta-Llama-3-8B-Instruct             FP8_BLOCK    mmlu                       67.02%         67.24%       +0.33%
Meta-Llama-3-8B-Instruct             FP8_BLOCK    wikitext                    10.09          10.10       -0.10%
Mixtral-8x7B-Instruct-v0.1           FP8_BLOCK    gsm8k                      63.84%         64.22%       +0.60%
Mixtral-8x7B-Instruct-v0.1           FP8_BLOCK    gsm8k_platinum             64.43%         65.34%       +1.41%
Mixtral-8x7B-Instruct-v0.1           FP8_BLOCK    mmlu                          N/A            N/A          N/A
Mixtral-8x7B-Instruct-v0.1           FP8_BLOCK    wikitext                      N/A            N/A          N/A
Qwen2.5-32B-Instruct                 FP8_BLOCK    gsm8k                      11.30%         10.99%       -2.74%
Qwen2.5-32B-Instruct                 FP8_BLOCK    gsm8k_platinum              9.59%          9.18%       -4.28%
Qwen2.5-32B-Instruct                 FP8_BLOCK    mmlu                       82.54%         82.60%       +0.07%
Qwen2.5-32B-Instruct                 FP8_BLOCK    wikitext                     7.02           6.60       +5.98%
Qwen3-VL-8B-Instruct                 FP8_BLOCK    gsm8k                      56.94%         57.47%       +0.93%
Qwen3-VL-8B-Instruct                 FP8_BLOCK    gsm8k_platinum             58.97%         59.47%       +0.85%
Qwen3-VL-8B-Instruct                 FP8_BLOCK    mmlu                       75.94%         75.85%       -0.12%
Qwen3-VL-8B-Instruct                 FP8_BLOCK    wikitext                      N/A            N/A          N/A
```

seems reasonable, i may be missing a chat template for the qwen 2.5 32B but overall seems indicative that this is working


src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -253,24 +260,18 @@ def quantize_weight(
...
             invperm = torch.argsort(perm)
             W = W[:, invperm]
-            g_idx = g_idx[invperm]
-
-            # only save g_idx if mapping is not identity
-            has_gidx = True
 
-    if not has_gidx:
-        g_idx = None
+            if actorder == ActivationOrdering.GROUP:
```
HDCharles: We can simplify this, only group act order saves g_idx, can just check for that
No need to check twice, just do this line on line 287 and remove g_idx_to_save

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -253,24 +260,18 @@ def quantize_weight(
...
 
-    has_gidx = False
-    if strategy in (QuantizationStrategy.GROUP, QuantizationStrategy.TENSOR_GROUP):
-        if actorder == ActivationOrdering.WEIGHT:
-            # restore original permutation
-            invperm = torch.argsort(perm)
-            W = W[:, invperm]
-
-        elif actorder == ActivationOrdering.GROUP:
+    if strategy in (
```
HDCharles: I think we can remove the 'if strategy' stuff too right?

We just check for act order and undo the weight permutation, right?

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -217,8 +224,8 @@ def quantize_weight(
                     global_scale=global_scale,
                 )
             elif strategy == QuantizationStrategy.BLOCK:
-                block_width = quant_args.block_structure[1]
-                block_column_idx = (i1 + i) // block_width
+                column_idx = i1 + i
```
HDCharles, line 228: Much nicer

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -111,12 +111,19 @@ def quantize_weight(
...
+        QuantizationStrategy.GROUP,
+        QuantizationStrategy.TENSOR_GROUP,
+        QuantizationStrategy.BLOCK,
+    ):
         # mapping from column index to group index
-        g_idx = (
-            torch.arange(num_columns, device=W.device, dtype=torch.int)
-            // quant_args.group_size
+        divisor = (
+            quant_args.group_size
```
HDCharles, line 131: Nice

src/llmcompressor/modifiers/gptq/gptq_quantize.py
rk119: Oh yes, since compressed-tensors library is ensuring that enabling actorder is bounded to group quantization strategy hence we do not need to repeat the check here. I shall make the change.

HDCharles:
See quality checks instructions and bot comments, otherwise looks pretty good just some minor fixes.

We should also run some evals for this, once you have the changes in I can help you with that. You can reach out to me in vLLM slack for easier coordination

HDCharles (changes requested):

src/llmcompressor/modifiers/gptq/gptq_quantize.py
rk119: Yep, done!

HDCharles (approved):

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -110,23 +110,31 @@ def quantize_weight(
...
+    if strategy in (
+        QuantizationStrategy.GROUP,
+        QuantizationStrategy.TENSOR_GROUP,
+        QuantizationStrategy.BLOCK,
+    ):
+        # mapping from column index to group index
+        divisor = (
+            quant_args.group_size
+            if strategy != QuantizationStrategy.BLOCK
+            else quant_args.block_structure[1]
```
brian-dellabetta, line 133: why `quant_args.block_structure[1]` here? because the weight.data matrix is transposed for linear?

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -217,8 +225,8 @@ def quantize_weight(
                     global_scale=global_scale,
                 )
             elif strategy == QuantizationStrategy.BLOCK:
-                block_width = quant_args.block_structure[1]
-                block_column_idx = (i1 + i) // block_width
+                column_idx = i1 + i
+                block_column_idx = g_idx[column_idx]
```
brian-dellabetta, line 229: For block, you may want to wrap with the [maybe pad logic](https://github.com/vllm-project/compressed-tensors/blob/09423bf77de257844bc8efdaede3068d52315bae/src/compressed_tensors/quantization/utils/helpers.py#L401), in case the weight matrix is not evenly divisible by block size

brian-dellabetta (commented):
Thanks for the contribution! Couple questions from my side

src/llmcompressor/modifiers/gptq/gptq_quantize.py
kylesayrs: Yes, weight.shape[1] is the input (hidden) size.
Linear := x @ W.T

kylesayrs (approved):
Nice job cleaning up the actorder logic, lgtm!

src/llmcompressor/modifiers/gptq/gptq_quantize.py
kylesayrs: Perhaps this should be saved for a follow up? Currently it will lead to a diagnosable error.

brian-dellabetta: yeah good point

PR #2618, opened 2026-04-14
fix CI failure

HDCharles:
Summary:

CI failure:
https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24376049082/job/71189804970

E         (EngineCore pid=6388)   File "/home/runner/.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/subprocess.py", line 1026, in __init__
E         (EngineCore pid=6388)     self._execute_child(args, executable, preexec_fn, close_fds,
E         (EngineCore pid=6388)   File "/home/runner/.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/subprocess.py", line 1955, in _execute_child
E         (EngineCore pid=6388)     raise child_exception_type(errno_num, err_msg, err_filename)
E         (EngineCore pid=6388) FileNotFoundError: [Errno 2] No such file or directory: 'ninja'

last passed on April 7
(https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24112855626/job/70351011358)

on April 8 https://github.com/vllm-project/vllm/pull/33892 was landed to vllm which requires JIT for this path (which requires ninja).

Ninja is installed but the test runs in a subprocess that doesn't activate the venv, adding the venv should fix this

running test to see if this fixes it: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24570535490


brian-dellabetta (approved):

dsikka (commented):
Can you ask Dan / Domenic to review - I thought we installed ninja in our test flows

dsikka (dismissed):
If this required by vLLM, not llmcompressor, I don’t think it makes sense to add to our setup.py?

dhuangnm:
I agree with @dsikka that if ninja is needed from the vLLM side it should be from the vLLM installation.

setup.py
```diff
@@ -158,6 +158,7 @@ def localversion_func(version: ScmVersion) -> str:
             "pytest-rerunfailures>=13.0",
             "lm_eval==0.4.9.2",
             # test dependencies
+            "ninja",
```
dhuangnm: Can we find out what needs ninja in vLLM? It should be in vLLM installation if it's needed now.

dhuangnm: Actually ninja does look installed along with vLLM in the nightly run: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24431720186
 
image

dbarbuzzi: While vLLM installs `ninja`, it can’t ‘see’ it because we don’t actually activate that virtual environment.

When `ninja` gets installed, it creates an “executable” that would normally be included in your `PATH` through the activation of the virtual environment. Because we don't activate the environment, just run the `python` binary from it directly, it cannot ‘see’ other executables in that (not-activated) virtual environment.

dbarbuzzi: A better solution would be:
* Wrap the `run_vllm.py` script in a Bash script that first activates the virtual environment and then executes the `run_vllm.py` script the same way it gets executed now
* Update the test code to subprocess out to the Bash script instead of the Python script

New args/etc. may be required to let the Bash script know the location of the environment to be activated and so on.

dhuangnm: Oh right, good point. I guess we can add the path where ninja is installed to the PATH env var when running run_vllm.py. This might be the easist?

HDCharles: how does the current change look?

dhuangnm (approved):
Cool, looks good!

PR #2623, opened 2026-04-16
Adding test_group to lm-eval configs

debroy-rh:
Adding test_group to the following lm-eval configs:
[fp8_dynamic_per_token.yaml](https://github.com/vllm-project/llm-compressor/blob/main/tests/lmeval/configs/fp8_dynamic_per_token.yaml)

[w4a4_nvfp4.yaml](https://github.com/vllm-project/llm-compressor/blob/main/tests/lmeval/configs/w4a4_nvfp4.yaml)

[w4a16_actorder_none.yaml](https://github.com/vllm-project/llm-compressor/blob/main/tests/lmeval/configs/w4a16_actorder_none.yaml)

[vl_w4a16_actorder_weight.yaml](https://github.com/vllm-project/llm-compressor/blob/main/tests/lmeval/configs/vl_w4a16_actorder_weight.yaml)

This is to test rhaiis model-opt image for lm-eval accuracy.


dsikka (changes requested):
I think RHAIIS testing should use vLLM lm-eval testing, not HF as is currently used
This is something we are going to update in upstream soon but definitely what we should use for product release

tests/lmeval/configs/w4a16_actorder_none.yaml
```diff
@@ -1,10 +1,14 @@
 cadence: "weekly"
+test_group: "rhaiis"
 model: meta-llama/Meta-Llama-3-8B-Instruct
 scheme: W4A16_actorder_none
 recipe: tests/e2e/vLLM/recipes/actorder/recipe_w4a16_actorder_none.yaml
 dataset_id: HuggingFaceH4/ultrachat_200k
 dataset_split: train_sft
 lmeval:
+  recovery_threshold:
```
dhuangnm: Why do we need to add this here?

dhuangnm (commented):
It LGTM in general, just one question about a model config.

I think this PR needs to land along with the corresponding changes in the llm-compressor-testing repo. @debroy-rh can you please create the PR for the llm-compressor-testing repo as well so we can test the two PRs together before landing? Thanks.

tests/lmeval/configs/w4a16_actorder_none.yaml
debroy-rh: removed.

debroy-rh:
@dhuangnm , @dsikka - I was trying to run the the lm-eval for vl_configs - [vl_w4a16_actorder_weight.yaml]. It was failing. Two reasons -
1. Parsing error - two "model" tag in the config file, the 2nd one under lm-eval. So, needed to update the tests/e2e/vLLM/run_tests_in_rhaiis.sh file.
2. Add torchvision to install_requires, so the same default install that already pulls in torch also pulls in torchvision, because the VLM / processor path needs it. Also, add an lmeval conftest bootstrap so pytest loads torchvision before Transformers.

Please take a look.
Run - https://github.com/neuralmagic/llm-compressor-testing/actions/runs/24686238622/job/72196254120

tests/lmeval/conftest.py
dhuangnm, line 1: This script is not needed as torchvision is in the extra requirement of llmcompressor and we shouldn't handle its installation here separately.

setup.py
```diff
@@ -122,6 +122,11 @@ def localversion_func(version: ScmVersion) -> str:
         ),
         ("tqdm>=4.66.3,<=4.67.3" if BUILD_TYPE == "release" else "tqdm>=4.66.3"),
         ("torch>=2.10.0,<=2.11.0" if BUILD_TYPE == "release" else "torch>=2.10.0"),
+        (
```
dhuangnm: torchvision is already in the extra requirement. This is not needed here and we don't want to limit its version too since it should be compatible with torch.

dhuangnm (changes requested):

dhuangnm (approved):
Looks good, thanks!

debroy-rh:
> I think RHAIIS testing should use vLLM lm-eval testing, not HF as is currently used This is something we are going to update in upstream soon but definitely what we should use for product release

@dsikka , as I understand this is going to be handled in another PR. Can you please approve and label this one.

PR #2633, opened 2026-04-20
[Deprecation] [Offload] [Tracing] Remove legacy offloading logic in tracing

kylesayrs:
## Purpose ##
* Remove logic which used to handle onloading sequential ancestors which were offloaded
  * These modules used to have to be force onloaded because accelerate did not support tracing. Now, [we always trace as offloaded](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/helpers.py#L132-L133), so there is no conflict and this logic is irrelevant
  * Remove `dispatch_for_sequential`, as this is now just an alias for `set_onload_device`


brian-dellabetta (approved):

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -180,23 +172,12 @@ class SequentialTracer(HFTracer):
     :param offloaded: modules which have offloaded params and should not be traced
```
HDCharles: ^

HDCharles (approved):

HDCharles (approved):

HDCharles (approved):

PR #2635, opened 2026-04-21
[AWQ] Seed grid search with identity baseline + fail fast on non-finite loss

juju812:
## Summary

Fix two correctness bugs in the AWQ grid search that allowed `_compute_best_scale` to either pick a worse-than-no-smoothing candidate or to silently label a numerically broken layer as a clean identity fallback.

### Bugs fixed

1. **Identity baseline mis-measured.** `initial_error` was the loss of "the first surviving grid candidate", which under `duo_scaling=True` is **not** identity. The recorded reduction therefore compared two smoothed candidates against each other, and the grid loop was allowed to pick a worse-than-no-smoothing scale set whenever the first candidate happened to be worse than later ones.

2. **Non-finite loss silently labelled as identity fallback.** If the identity baseline forward produced NaN/Inf, every grid candidate's `loss < non_finite` was `False`, no candidate ever won, and the layer was silently labelled `fell_back_to_identity=True`. The artifact then shipped with poison metrics. The same outcome occurred if every executed grid candidate produced a non-finite loss (e.g. fp16 overflow on heavy outliers) while identity was finite.

### Implementation

- Extract per-candidate `apply scales → quantize → forward → MSE` into `_eval_scales(scales, ..., update_fused)`. Call it once with `scales = ones` to seed both `initial_error` and `best_error` from the real identity baseline. The grid loop only keeps a candidate if it strictly beats baseline; otherwise the layer stays at identity by construction.
- Track whether any candidate improved with an explicit `grid_improved_over_identity` boolean instead of a float sentinel. Each layer is recorded with a `fell_back_to_identity` flag in `_error_metrics` so post-hoc inspection can flag layers that didn't benefit from AWQ.
- Raise `RuntimeError` in two narrow correctness cases: identity baseline non-finite, or identity finite but every grid candidate non-finite. Layers genuinely stuck at identity (all candidate losses finite, none beat baseline) are recorded for inspection, not raised on.
- Hoist the `update_fused` (TENSOR_GROUP strategy) check and the host→device transfer of `orig_layer_weights` out of the per-candidate hot path. Relevant when the model is offloaded.

### What this PR does NOT include

The Qwen3.5 `attn_output_gate` per-output-slice mapping fix is in **#2636** (stacked on this PR).

## Test plan

- [x] `pytest tests/llmcompressor/modifiers/awq/` — full AWQ suite passes (38 tests)
- [x] New end-to-end regression tests in `test_compute_best_scale_baseline.py` drive `_compute_best_scale` through the real code path on a tiny W4A16 module:
  - `test_initial_error_equals_identity_loss`
  - `test_best_error_never_exceeds_initial_error`
  - `test_compute_best_scale_raises_when_identity_baseline_is_non_finite`
  - `test_compute_best_scale_raises_when_all_grid_candidates_are_non_finite`
- [ ] Re-run Qwen3.5 27B AWQ calibration end-to-end (to be done on real hardware)


src/llmcompressor/modifiers/awq/base.py
```diff
@@ -698,6 +801,43 @@ def _compute_best_scale(
                 for balance_layer in balance_layers_to_patch
             ],
         ):
+            initial_error = self._eval_scales(
```
brian-dellabetta: rather than splitting this out into 2 separate flows, one where scales=[1,1,1,1,1...,1] is evaluated and another where we do the grid search, we should just ensure that `scales=[1,1,1,1,1...,1]` is included in the grid search. if that is not already the case, that is a bug. Are you only seeing this for `duo_scaling=True` and not `duo_scaling=False`, or `duo_scaling="both"`?

brian-dellabetta: Did some looking, you are right that `duo_scaling=True` means that the case where `scales=[1,1,1,1,1]` [is not automatically checked](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L713-L718). it is only checked when duo_scaling=False or duo_scaling="both". We should update our sweep to always include it

brian-dellabetta: I set up a PR to ensure identity scales are always included in the grid search -- https://github.com/vllm-project/llm-compressor/pull/2640/changes

juju812: Thanks @brian-dellabetta — I agree that #2640 is the right direction, and I’m happy to defer to that.

The one narrow gap I think #2640 plus the existing `best_ratio == -1` guard still leaves is when the identity baseline forward returns a non-finite value (`NaN/Inf` under fp16). In that case, `NaN < inf` is false, so identity never wins, but the first finite scaled candidate does. That makes best_ratio become >= 0, which means the existing guard never trips, and we can end up shipping smoothed scales chosen against a broken baseline.

Once #2640 lands, would you prefer that I trim this PR down to just a one-line isfinite(initial_error) check and drop the rest of the diff?

brian-dellabetta: Thanks @juju812 , yes let's revisit once #2640 lands. i want to get #2526 in first, then that one, and we can see what your diff for NaN/Inf issue looks like once those land

brian-dellabetta: @juju812 we landed those 2 PRs, can you rebase to include your isfinite check?

juju812: @brian-dellabetta I haven't observed this NaN path in our Qwen3.5-27B AWQ runs; the test patches _compute_loss to demonstrate the control-flow gap, not a real failure.

The motivation is symmetric to the existing `best_ratio == -1` guard: post-#2640, identity is candidate 0 and seeds `initial_error`. Under IEEE 754, `NaN < anything` is always false, so an identity-NaN never wins, but a later finite candidate can still flip `best_ratio` and bypass the guard. A single `math.isfinite(initial_error)` check closes that gap.

The check is a one-line change and a no-op on healthy runs. This PR can be dropped now that #2640 has landed if you consider it overly defensive.

juju812:
> @brian-dellabetta I haven't observed this NaN path in our Qwen3.5-27B AWQ runs; the test patches _compute_loss to demonstrate the control-flow gap, not a real failure.

> The motivation is symmetric to the existing best_ratio == -1 guard: post-https://github.com/vllm-project/llm-compressor/pull/2640, identity is candidate 0 and seeds initial_error. Under IEEE 754, NaN < anything is always false, so an identity-NaN never wins, but a later finite candidate can still flip best_ratio and bypass the guard. A single math.isfinite(initial_error) check closes that gap.

> The check is a one-line change and a no-op on healthy runs. This PR can be dropped now that https://github.com/vllm-project/llm-compressor/pull/2640 has landed if you consider it overly defensive.

Hi, @brian-dellabetta, the branch is rebased, please share your comment with the new diff?

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -756,13 +757,20 @@ def _compute_best_scale(
                     best_scales = scales.clone()
                 pbar.set_postfix({"best_error": f"{best_error:.3e}"})
 
-        if best_ratio == -1:
+        # Symmetric to ``best_ratio == -1`` under IEEE 754: post-#2640 identity
```
brian-dellabetta: can you clean up this comment a bit?

brian-dellabetta (commented):
Hi @juju812 , thanks for rebasing. This looks good, one nit on the comment still in place, if we can clean it up and shorten it a bit. Then we can merge this in. Thanks!

src/llmcompressor/modifiers/transform/awq/base.py
juju812: Cleaned up and shortened the comment.

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -756,13 +757,16 @@ def _compute_best_scale(
                     best_scales = scales.clone()
                 pbar.set_postfix({"best_error": f"{best_error:.3e}"})
 
-        if best_ratio == -1:
+        # The first grid point is identity. If its loss is NaN/Inf, fail fast
+        # instead of allowing a later finite candidate to hide a broken baseline.
+        if best_ratio == -1 or not math.isfinite(initial_error):
```
brian-dellabetta: i think we should move this check to above on line 749, if initial_error is not finite then raise the exception. then we don't have to worry about the `best_ratio == -1` check because if initial_error is finite, it will be set

tests/llmcompressor/modifiers/transform/awq/test_compute_best_scale_isfinite.py
brian-dellabetta, line 1: We should be able to clean this up by replicating `tests/llmcompressor/modifiers/transform/awq/test_base.py:test_awq_with_kv_cache_quantization`. Set up AWQ on the tiny smoke model, update some of the weights to be NaNs, confirm we see the right exception during oneshot. Should be doable in <50 lines of code

juju812:
Hi, @brian-dellabetta,

Thanks for the detailed review comments. 
I moved the finiteness check to the initial_error assignment, and replaced the private helper test with a tiny-model oneshot smoke test.

tests/llmcompressor/modifiers/transform/awq/test_base.py
```diff
@@ -687,6 +680,44 @@ def test_awq_with_kv_cache_quantization():
...
+    dataset = Dataset.from_dict({"text": ["Hello world " * 10]})
+
+    self_attn = model.model.layers[0].self_attn
+    forward_count = 0
+
+    def corrupt_weights_after_baseline(module, _inputs, _output):
+        nonlocal forward_count
+        forward_count += 1
+        if forward_count == 2:
+            module.q_proj.weight.data.fill_(float("nan"))
```
brian-dellabetta, line 703: why was this needed?

brian-dellabetta (commented):
Thanks @juju812 , things look good. Just one question on the nonlocal var in the test

tests/llmcompressor/modifiers/transform/awq/test_base.py
juju812: This is used to corrupt the weights only after the fp16 baseline forward has completed. If the weights are NaN before oneshot starts, AWQ sees non-finite fp16_outputs and exits before reaching `_compute_best_scale`, so the test would not exercise the `math.isfinite(initial_error)` guard.

The hook runs after `self_attn` forward; on the second call the baseline output has already been produced, then we set `q_proj.weight` to NaN so the following grid-search identity candidate produces the non-finite loss.

brian-dellabetta: thanks, i left a comment to clarify

brian-dellabetta (approved):
thanks for the discussion to get to the bottom of this, and the contribution!

HDCharles (approved):

PR #2636, opened 2026-04-21
[AWQ] Per-output-slice grid search for fused q_proj (Qwen3.5 attn_output_gate)

juju812:
> **Stacked on top of https://github.com/vllm-project/llm-compressor/pull/2635** (`fix/awq-identity-baseline`). 

## Summary

Qwen3.5 fuses the query and output-gate projections into a single `q_proj` of shape `[2*H*D, hidden]` (gate rows are sigmoid'd downstream). Gate rows have wildly different weight magnitude than query rows, so any AWQ grid search that minimises MSE over the *full* output gets pulled toward scales that minimise gate error and degrade query error. End-to-end this pushed Qwen3.5-27B AWQ to `bpb=3.8` versus baseline `bpb=0.5` — a complete failure.

This PR teaches the grid search to score MSE only against the rows that actually participate in the AWQ algebra (the query rows, not the sigmoid'd gate rows), wires that up automatically for `attn_output_gate=True` configs, and preserves the algebraic invariant of `_apply_smoothing` so the persisted weights remain valid.

### Changes

1. **`AWQMapping.balance_output_slices: dict[str, SliceLike] | None`** (new). `SliceLike = slice | Sequence[int] | dict` is YAML-friendly via a `_coerce_slice` parser. Resolved into `dict[Module, slice]` on `ResolvedMapping` by `_set_resolved_mappings`.

2. **`_eval_scales` becomes slice-aware.** When a balance layer has a slice configured, the grid candidate's smoothing transform is applied only to the slice rows (in-place via `copy_(orig); [sl].mul_(_scalesview)` and `copy_(orig); [sl].copy_(quantized[sl] / _scalesview)` — no per-candidate `orig.clone()`, and `weight.data.copy_` preserves offload hooks). Rows outside the slice stay at the un-smoothed FP weight, so the per-candidate MSE is computed against just the slice rows.

3. **`_apply_smoothing._smooth` is intentionally NOT slice-aware.** AWQ correctness rests on `Y = balance(LN(x)) = (W*s)((LN with γ/s)(x))` holding element-wise across **all** input channels. The persisted balance-layer weight must be scaled on every row even though only the slice rows participated in the search; otherwise the un-scaled rows silently see `x / s` from the layernorm without the compensating `* s` on the weight, and the forward pass on those rows drifts from the FP baseline. A NOTE comment in `_smooth` pins this invariant.

4. **Auto-wiring for Qwen3.5 in `dynamic_mappings`.** `_build_q_proj_output_slice` detects `attn_output_gate=True` (also looking under `text_config` for VLM variants like `Qwen3_5ForConditionalGeneration`) and emits `balance_output_slices = {"re:.*self_attn.q_proj$": slice(0, num_attention_heads * head_dim)}` so the existing `q_proj` mapping just works without any user-facing recipe change.

### Tests

* `test_compute_best_scale_baseline.py`:
  * `test_balance_output_slices_changes_grid_outcome_for_fused_q_proj` — production-path regression. Constructs a model where gate rows dominate the un-sliced MSE (10x weight magnitude vs query rows) and asserts that scoping the search to the query-row slice produces a strictly smaller `initial_error` AND a different `best_error` than the un-sliced run. If the slice were silently ignored, both runs would compute MSE over the same rows and produce identical baselines — guaranteeing the field is actually consulted by the real grid loop.
  * `test_apply_smoothing_preserves_forward_equivalence` — production-path regression for the algebraic invariant. Runs the full `_apply_smoothing` path (including `_compute_best_scale` with a non-trivial slice) and asserts the parent module's FP forward output is preserved bit-close before vs after smoothing. If `_smooth` ever regresses to a slice-aware update, this test fails with O(weight magnitude) divergence on the un-scaled rows.
* `test_dynamic_mappings.py`:
  * `TestQProjOutputSlice::test_slice_added_for_qwen3_5_dense` — top-level config path.
  * `TestQProjOutputSlice::test_slice_reads_from_text_config` — nested `config.text_config` path (VLM variants).
  * Combined with the existing `build_hybrid_attention_mappings` coverage, these cover both production code paths through `_build_q_proj_output_slice`; no-slice/no-q_proj branches are implicit in every other `build_hybrid_attention_mappings` test in the file.

`pytest tests/llmcompressor/modifiers/awq/` → 46 passed.

### Reproduction

End-to-end Qwen3.5-27B AWQ with this PR + #PR_B applied: `bpb=0.5123` (baseline FP `bpb=0.5089`); without these PRs: `bpb=3.8`.

### Backward compatibility

`balance_output_slices` defaults to `None` on both `AWQMapping` and `ResolvedMapping`, and `_eval_scales` falls through to the existing full-output behaviour when no slice is configured. Models that don't set `attn_output_gate=True` (every model the dynamic registry handles today except Qwen3.5) get no behavioural change.

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -1,4 +1,5 @@
 import inspect
+import math
 from itertools import product
 from typing import Iterator, Literal
 
@@ -629,6 +630,77 @@ class AWQModifier(Modifier, QuantizationMixin):
             for output in outputs
         ]
 
+    @torch.no_grad()
+    def _eval_scales(
+        self,
+        scales: torch.Tensor,
+        mapping: ResolvedMapping,
+        fp16_outputs: list[torch.Tensor],
+        orig_layer_weights: dict[torch.nn.Module, torch.Tensor],
+        balance_layers_to_patch: list[Module],
+        device: torch.device,
+        update_fused: bool = False,
+    ) -> float:
+        """
+        Apply ``scales`` to the balance layers via the AWQ smoothing transform
+        ``W' = Q(W * s) / s``, run the parent module on the cached calibration
+        batches, and return the MSE between the smoothed/quantized output and
+        the fp16 reference.
+
+        Used both for the identity-baseline measurement (``scales = ones``)
+        and for every grid-search candidate. Pulling this out of the grid
+        loop keeps the baseline measurement and the grid measurements on
+        exactly the same code path.
+
+        ``update_fused`` is hoisted out of the per-iteration check at the
+        call site (whether every balance layer is on TENSOR_GROUP strategy
+        is static during the search), so this method just dispatches on
+        the precomputed flag instead of re-evaluating ``getattr`` chains
+        per grid candidate. ``scales`` and the ``orig_layer_weights``
+        tensors are also assumed to already live on ``device`` (the
+        caller pre-moves both before the grid loop) so this method
+        contains no per-iteration host-to-device transfers.
+        """
+        _scalesview = scales.view(1, -1)
+
+        # Q(W * s)
+        for balance_layer in balance_layers_to_patch:
+            w_qscheme = balance_layer.quantization_scheme.weights
+            orig = orig_layer_weights[balance_layer]
+            balance_layer.weight.data.copy_(orig * _scalesview)
+
+            should_calculate_gparam = (
+                w_qscheme.strategy == QuantizationStrategy.TENSOR_GROUP
+            )
+            call_observer(
+                balance_layer,
+                "weight",
+                balance_layer.weight,
+                should_calculate_gparam=should_calculate_gparam,
+            )
+            quantized = forward_quantize(
+                balance_layer,
+                balance_layer.weight,
+                "weight",
+                w_qscheme,
+            )
+            balance_layer.weight.data.copy_(quantized / _scalesview)
+
+        # Apply fused global scales for TENSOR_GROUP during grid search
+        # to match inference behavior. ``update_fused`` is precomputed
+        # by ``_compute_best_scale`` so we don't re-walk the layer list
+        # for every grid candidate.
+        if update_fused:
+            update_fused_layer_weight_global_scales(mapping.parent)
+
+        # W * X
+        int_w_outputs = self._run_samples(mapping.parent)
+
+        # compute mean squared error (L2 norm)
+        loss = self._compute_loss(fp16_outputs, int_w_outputs)
+        del int_w_outputs
+        return loss
+
     def _compute_best_scale(
         self,
         mapping: ResolvedMapping,
@@ -652,10 +724,6 @@ class AWQModifier(Modifier, QuantizationMixin):
         :return: tensor of best scales, one for each channel
         """
         history = []
-        best_ratio = -1
-        best_scales = None
-        best_error = float("inf")
-        initial_error = None
 
         device = get_execution_device(mapping.parent)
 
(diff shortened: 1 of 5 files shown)


brian-dellabetta:
Hi @juju812 , thanks for raising this, I wasn't aware of the Qwen3.5 gated attention logic merged in with the q_proj layer. What a pain. But I don't understand your reasoning -- if the input_layernorm is fed into the gate, we have to include it when doing the best scales search, because it will also be scaled and quantized. The MSE of the entire parent layer (in this case the entire attention block) is used when finding scales that minimize MSE. If we choose the best scales for the query states, disregarding the gate, then i would expect that to increase MSE when comparing the output of the entire attention block. Does that make sense?

juju812:
@brian-dellabetta 
Yeah, that makes sense, and I think you're right at the proxy level — restricting the search to the query rows can absolutely make whole-block MSE go up. This PR is intentionally giving up that objective, because on Qwen3.5's fused `q_proj` it correlates badly with end-to-end quality.

Worth separating implementation from objective:

- Gate rows are still smoothed and quantized in the persisted artifact. `_smooth` deliberately applies the chosen `s` to every row of `q_proj` (gate included), precisely to keep the AWQ balancing invariant `Y = (W·s)·((LN/s)x)` valid on every input channel — there's a NOTE pinning this so a future refactor doesn't accidentally make `_smooth` slice-aware too.
- Only the search objective (the MSE that scores grid candidates) is slice-restricted.

Why give up whole-block MSE as the search objective on this architecture: gate and query rows live inside the same `q_proj` and share one input-channel `s`, but their weight magnitude statistics are very different. In practice the unsliced search picks an `s` whose span across input channels lands around 25× on Qwen3.5-27B — that's wide enough that group-wise W4 group scales saturate and the query rows lose a meaningful chunk of effective precision. The calibration loss looks fine, but the W4 grid is no longer fine enough to represent the rows that actually drive the answer.

End-to-end on Qwen3.5-27B: unsliced AWQ lands at `bpb=3.8150` (FP16 reference `0.5081`); the W4A16 no-smoothing fallback is `0.5154`; the sliced search recovers to `0.5121`, i.e. slightly better than no-smoothing — AWQ is once again contributing positively on this model, which it can't do under the unsliced objective.

So I agree the unsliced objective wins on whole-block MSE. The case for the slice is just that the resulting `s` is a bad pick for what actually goes into the W4 artifact on this architecture. And I can pull per-mapping `initial_error`/`best_error`/scale-span numbers from a re-run if you want to see the proxy-vs-end-to-end gap quantified directly.

brian-dellabetta:
> The case for the slice is just that the resulting s is a bad pick for what actually goes into the W4 artifact on this architecture

Hi @juju812 , I see, the issue is that `duo_scaling=True` is choosing terrible options for scales, because the massive gate weights are distorting them away from scales that would actually help. This, combined with the fact that you're not checking identity scales (bug we discussed in other PR) is causing AWQ to break. Can you test my hypothesis that using `duo_scaling="both"` will resolve your issue on main? duo_scaling="both" will try both the duo_scaling=False and duo_scaling=True cases. when False, **only** the activations are used when choosing best scales candidates, and weights are disregarded. so it shouldn't be an issue. plus identity scales will be tried, so performance shouldn't degrade.

> unsliced AWQ lands at bpb=3.8150 (FP16 reference 0.5081); the W4A16 no-smoothing fallback is 0.5154; the sliced search recovers to 0.5121, i.e. slightly better than no-smoothing

Note that this could be due to other mappings as well, beyond the input_layernorm -> q/k/v proj mapping.

juju812:
Hi, @brian-dellabetta 
Verified empirically. Upstream/main with #2640 merged, no slice changes from this PR, on Qwen3.5-27B (250 samples):

| variant | bpb | vs BF16 | vs no-smoothing |
| --- | ---: | ---: | ---: |
| BF16 baseline | 0.5081 | — | -0.0073 |
| W4A16 no-smoothing (W4A16 with no AWQ scales applied) | 0.5154 | +0.0073 | — |
| **upstream main + `duo_scaling=True` (default)** | **0.5136** | **+0.0055** | **-0.0018** |
| upstream main + `duo_scaling="both"` | 0.5148 | +0.0067 | -0.0006 |
| this PR (sliced search, `duo_scaling="both"`) | 0.5121 | +0.0040 | -0.0033 |
| pre-#2640 AWQ (the original catastrophe) | 3.8150 | +3.31 | +3.30 |

Your read is largely right — #2640 alone removes the catastrophic 3.8 bpb collapse on this fused-gate architecture. The sliced search still recovers ~0.0015 bpb more on top of that, by keeping AWQ smoothing meaningful on the fused `q_proj` instead of having the gate rows dominate the search and effectively neuter it.

But 0.0015 bpb is small enough that I don't think it justifies adding a Qwen3.5-specific mapping path upstream. Going to close this one. Thanks for the push-back — the duo-both verification was useful either way.

brian-dellabetta:
Hi @juju812 , thanks for the details! Good to know that helped close the gap. I agree that the decrease is not big enough to introduce the slicing complexity. 

aside -- the grid search has always bothered me about AWQ, I don't see anything that prevents us from adding learnable scales that can be trained via backprop, which is almost always preferred to a grid search. If you are interested in this as a research spike, let me know. I hope to devote some time to it in future

PR #2640, opened 2026-04-21
[AWQ] explicitly include identity scales for duo_scaling=True grid search

brian-dellabetta:
SUMMARY:
First reported in #2635 , identity scales were not being included in the best scales grid search when `duo_scaling=True`. This explicitly adds the case `(ratio=0.0, duo_scaling=False)` so the identity scale case is always included in the search.

TEST PLAN:
unit tests added


HDCharles (approved):
looks good to me

PR #2647, opened 2026-04-24
Transformers v5

### LINKED ISSUE #2159: [GPT-OSS] Expanded Support for Activation Quantization
Weight-only quantization support for GPT-OSS has recently been approved via [[CPU] Linearize gpt_oss model and add example to quantize it to w4a8](https://github.com/vllm-project/llm-compressor/pull/2113) from @isharif168 🔥.

There are a couple follow ups that will make this work even better and more expansive
1. Refactor `convert_model_for_quantization_gptoss` to use the `MoECalibrationModule` interface
2. Add support for the `calibrate_all_experts` option in the forward definition
3. Test the forward function with modifiers such as GPTQ or AWQ
4. Add an example of how to run the quantized model in vLLM (with reference to any necessary patches)
5. Potential code cleanup/simplification, described in comments [here](https://github.com/vllm-project/llm-compressor/pull/2113#pullrequestreview-3599079869)

kylesayrs:
# Transformers v5 #
* Resolves #2159

## Examples and Model Loading ##

* `dtype` now defaults to `"auto"`, so we can remove this explicit argument to reduce confusion
* `from_pretrained` no longer supports `use_auth_token`
* Model structures differ between v4 and v5. This means that our example regexes need to be updated. This was done manually and with assistance from [AI-generated tests](https://github.com/vllm-project/llm-compressor/pull/new/kylesayrs/example_regex_tests)

## Tied Embeddings ##
* Transformers now has a slightly different policy for handling checkpoints with tied embeddings, but where the values differ in the checkpoint. Previously, transformers would throw away the lm_head. Now, this conflict is detected, and the embeddings are untied. In practice, these models are essentially malformed, but `nm-testing/tinysmokellama-3.2` is one such malformed model and we need to update our tests to use different models/ account for this change in behavior.

## MoE ##

There are two default pathways to adding MoE support for quantization, both of which are specified in `linearize.py`. I've pulled out the MoE linearization logic from `moe_calibration_context` into dedicated methods `load_quantizable_moe`, and `linearize_moe`. This allow us to do the super-fast load pathway, and also means that we can toggle the calibrate_experts flag without necessarily having to unfold the model for saving.

### `load_quantizable_moe` ###
This context allows users to load moe models with efficient linearization and integration with the `calibrate_all_experts` flag.

1. In the best case, the model architecture has a pre-baked conversion mapping, which allows the model to be converted into 2d linearization as it's loaded. This case is the best for load runtime.

```python3
experts_cls, forward_mapping, backward_mapping = _get_2d_mappings(model_type)
linear_experts_2d_cls = LinearExperts2D.create_linear_experts_cls(experts_cls)
register_patch_mapping({experts_cls.__name__: linear_experts_2d_cls})
register_checkpoint_conversion_mapping(
    model_type, forward_mapping, overwrite=True
)
```

2. In the fallback case, the model is loaded into the standard transformers v5 `Experts` form, then converted to a linearization. This is often inefficient, as models are sometimes undergo 2d -> 3d -> 2d, but is guaranteed to work for all transformers-v5-compliant model definitions.

```python3
linear_experts_cls = LinearExperts2D.create_linear_experts_cls(original_experts_cls)
for name, module in non_linearized_moes.items():
    linear_moe = linear_experts_cls.from_experts_module(module, model.config)
    model.set_submodule(name, linear_moe)
```

3. In the final case, if a model does not have the standard transformers v5 `Experts` form, then we must write a custom replacement. This involves registering a new subclass of `LinearExperts2D`.

```python3
class GraniteMoeLinearExperts(LinearExperts2D):
  def from_experts_module(...):
    ...

  def __init__(...):
    ...

  def forward(...):
    ...
```

## Adjust LM Eval Thresholds ##
This PR makes two adjustments to lm_eval test thresholds

1. `w4a16_awq_sym: flexible-extract` (`0.91 -> 0.90`): The base model's score increased, while the compressed model's score stayed the same
2. `w4a16_actorder_group`: strict-match` (`0.95 -> 0.94`): All scores for this config decreased by ~0.025 points, the strict match happened to just go under the threshold.

From the charts below, you can see that there are minimal changes in lm-eval scores before and after v5 upgrade, any there is no evidence of a correlated trend downwards.

<details><summary>Δ scores</summary>
v5-scores
</details>

<details><summary>Δ recovery</summary>
v5-recovery
</details>

## Follow Ups ##
* Implement a load context like `load_compat` to simplify UX for loading models
* Add documentation and tutorials for supporting MoE linearization for new models

## Thoughts on Future Model Support ##
* A high priority for LLM Compressor moving forward should be to improve zero day support by supporting model definitions which are not Transformers compliant. This means
  * Supporting hooks/templates for MoE linearization
  * Reducing reliance on the `PreTrainedModel` definition, potentially allowing users to pass a config template
  * Adding guides for all of the above

## Testing ##
* Integration, E2E, lm_eval: [passing](https://github.com/neuralmagic/llm-compressor-testing/actions/runs/26935762392); Examples: [passing](https://github.com/neuralmagic/llm-compressor-testing/actions/runs/26959652272)
* Added tests which replace the previously individual MoE replacement tests
* Added tests for MoE replacement's compatibility with offloading and distributed offloading
* Ran AI-generated tests for testing that regexes still match after v5 upgrade https://github.com/vllm-project/llm-compressor/pull/new/kylesayrs/example_regex_tests


src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,241 @@
...
+                expert_output = expert(hidden_states[token_indices])
+
+            # apply weighting to outputs
+            expert_weights = top_k_weights[token_indices, top_k_pos, None]
+            weighted_output = expert_output * expert_weights
+
+            # accumulate the selected tokens
+            final_hidden_states.index_add_(0, token_indices, weighted_output)
+
+        return final_hidden_states
```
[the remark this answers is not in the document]
kylesayrs: Hidden states is expected to be already flattened
https://github.com/huggingface/transformers/blob/a66638d854ae536e0ca31e8bcfa480adfaf58284/src/transformers/models/qwen3_vl_moe/modeling_qwen3_vl_moe.py#L141

src/llmcompressor/modeling/moe/helpers.py
```diff
@@ -0,0 +1,81 @@
...
+            elif isinstance(decorator, ast.Call) and isinstance(
+                decorator.func, ast.Name
+            ):
+                name = decorator.func.id
+            else:
+                continue
+            if name == "use_experts_implementation":
+                return True
+
+    return False
```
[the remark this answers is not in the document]
kylesayrs: Yep, will think more on this. I think this is the most robust solution, but should probably be lru cached against the module class.

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,241 @@
...
+from . import context
+from .helpers import (
+    FusedExpertsModule,
+    _get_moe_shapes,
+    _is_moe_experts_converter,
+    _is_moe_experts_module,
+)
+
+
+def linearize_moe_model(model: PreTrainedModel):
```
kylesayrs: In the base case, we need to do a conversion. However, we can optimize this on a per-model basis by writing our own weight converters which are used directly during loading

maxdebayser:
Thanks for working on this @kylesayrs. The IBM Spyre stack for vLLM depends on llm-compressor because of this library: https://github.com/foundation-model-stack/fms-model-optimizer. Currently we're unable to upgrade to transformers 5 due to this. Do you need any help?

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -0,0 +1,230 @@
...
+            expert_mask = expert_mask.permute(2, 1, 0)
+
+        for expert_index in range(self.num_experts):
+            # select tokens for this expert
+            top_k_pos, token_indices = torch.where(expert_mask[expert_index])
+
+            # apply expert
+            expert = self[expert_index]
+            if get_calibrate_all_experts_flag():
+                expert_output = expert(hidden_states)[token_indices]
```
orestis-z, line 258: What is the specific use case where an expert needs to calibrate on every token?
I'm thinking that evaluating expert(hidden_states) all at once is going to trigger a VRAM spike on normal batch sizes. Maybe worth chunking if you agree it's of concern?

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,100 @@
...
+            model_type, forward_mapping, overwrite=True
+        )
+
+        # load model
+        model: PreTrainedModel = original_from_pretrained(*args, **kwargs)
+
+        # prepare for saving to be called later
+        clear_patch_mapping()
+        model._conversion_mapping = backward_mapping
+        conversion_mapping._checkpoint_conversion_mapping_cache = None
```
orestis-z: Do they expose a public method to achieve the same? They could change their private API in a minor version bump.

If not I would make it explicit:


```python
def _unsafe_clear_hf_conversion_cache():
    """
    WARNING: Mutating internal HF state. 
    """
    from transformers import conversion_mapping
    conversion_mapping._checkpoint_conversion_mapping_cache = None

# ... later in the code ...
_unsafe_clear_hf_conversion_cache()
```

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,100 @@
...
+    conversion_mapping,
+)
+from transformers.conversion_mapping import (
+    register_checkpoint_conversion_mapping,
+)
+from transformers.monkey_patching import clear_patch_mapping, register_patch_mapping
+
+from llmcompressor.modeling.moe.helpers import FusedExpertsProtocol
+
+from .conversion_mappings import _get_2d_mappings, _has_2d_mappings
```
orestis-z: given that we're importing them now they're not module-level private anymore. Suggesting to remove the `_` -prefix.

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,100 @@
...
+        linear_moe = linear_experts_cls.from_experts_module(module, model.config)
+        model.set_submodule(name, linear_moe)
+
+
+def requires_linearize_moe(model: torch.nn.Module):
+    for module in model.modules():
+        if isinstance(module, FusedExpertsProtocol):
+            return True
+
+    return False
```
orestis-z: Minor syntactic sugar
```suggestion
    return any(isinstance(module, FusedExpertsProtocol) for module in model.modules())
```

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,100 @@
...
+        for name, module in model.named_modules()
+        if isinstance(module, FusedExpertsProtocol)
+    }
+
+    if len(non_linearized_moes) <= 0:
+        logger.warning("TODO could not find experts to replace")
+        return model
+
+    logger.warning("TODO")
+    original_experts_cls = next(iter(non_linearized_moes.values())).__class__
```
orestis-z: this assumes that every MoE module in the model shares the exact same class type as the very first one we find. I understand that this is might be true for most models right now, but I would recommend to iterate and check, or explicitly assert that all values in non_linearized_moes share the same `__class__`

orestis-z (commented):
Hey @kylesayrs, I'll continue my review tomorrow. In the meantime I'm submitting my comments so far.

examples/awq/qwen3_moe_example_ddp.py
```diff
@@ -71,9 +70,9 @@ def tokenize(sample):
 recipe = [
     AWQModifier(),
     QuantizationModifier(
-        ignore=["lm_head", "re:.*mlp.gate$", "re:.*mlp.shared_expert_gate$"],
         scheme="W4A16",
         targets=["Linear"],
+        ignore=["lm_head", "re:.*mlp.gate$"],
```
brian-dellabetta, line 75: why are you removing shared expert gate from these?

src/llmcompressor/modeling/moe/conversion_mappings.py
```diff
@@ -0,0 +1,73 @@
...
+    WeightConverter,
+    WeightRenaming,
+    WeightTransform,
+)
+from transformers.models.deepseek_v4.modeling_deepseek_v4 import DeepseekV4Experts
+
+# TODO: in the future, we can potentially grep the source code for this
+ARCH_TO_EXPERTS_MODULE_CLS = {"deepseek_v4": DeepseekV4Experts}
+
+ARCH_TO_2D_MAPPINGS = {
```
brian-dellabetta, line 28: we've discussed this a couple times. would be worth having a design meeting about what this means going forward. seems like lots of manual maintenance, just like the dang AWQ/SmoothQuant mappings

examples/quantization_w8a8_fp8/granite4_example.py
brian-dellabetta, line 1: we should make sure to run all examples as part of ci checks for this PR. is this cleanup or all these changes a transformers v5 necessity?

examples/quantizing_moe/deepseek_v4_example.py
```diff
@@ -0,0 +1,138 @@
...
+)
+
+from llmcompressor import oneshot
+from llmcompressor.modeling.moe.linearize import load_quantizable_moe
+from llmcompressor.modifiers.quantization import QuantizationModifier
+
+# Upstream BUG: norms should be loaded in float32, but usually aren't due to the base
+# model having a quant_config which overrides this. Loading in float32 actually
+# breaks the model definition (it expects bfloat16). Let's force load in bfloat16.
+DeepseekV4PreTrainedModel._keep_in_fp32_modules_strict = set()
```
brian-dellabetta, line 19: oh boy

src/llmcompressor/modeling/moe/context.py
brian-dellabetta, line 1: there's a lot going on in `src/llmcompressor/modeling/moe`. to the point where i'm wondering if we should revive the conversation about swapping transformers for vllm directly

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,100 @@
...
+from transformers.monkey_patching import clear_patch_mapping, register_patch_mapping
+
+from llmcompressor.modeling.moe.helpers import FusedExpertsProtocol
+
+from .conversion_mappings import _get_2d_mappings, _has_2d_mappings
+from .linear_experts import LinearExperts2D
+
+
+@contextlib.contextmanager
+def load_quantizable_moe(model_cls: Type[PreTrainedModel] = AutoModelForCausalLM):
```
brian-dellabetta, line 30: to avoid UX headache, maybe we should just have users always wrap `with load_compat` context that always loads it in an llm-compressor compatible way? it seems we can determine what is needed based on model architecture.

src/llmcompressor/typing.py
```diff
@@ -23,3 +23,40 @@
 
 # Torch types
 NamedModules = Iterable[tuple[str, torch.nn.Module]]
+
+
+class TorchModuleProtocolMeta(_ProtocolMeta):
```
brian-dellabetta, line 28: we should discuss this and src/llmcompressor/modeling/moe in person

tests/llmcompressor/modifiers/quantization/test_handling_shared_embeddings.py
```diff
@@ -22,12 +23,12 @@ def test_quantization_with_automatic_untie():
     _untie_if_target_shared is called during start_calibration to automatically
     handle shared input/output embeddings when they are targeted for quantization.
     """
-    model_id = "nm-testing/tinysmokellama-3.2"
+    model_id = "Qwen/Qwen3-0.6B"
```
brian-dellabetta: why do we need to change? tiny smoke is 354k params

brian-dellabetta (commented):
I've looked through about half, it would be best to just cover this in a team meeting. Leaving my comments so far

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,100 @@
...
+
+        config = AutoConfig.from_pretrained(*args, **kwargs)
+        model_type = config.model_type
+
+        # model is 3d (or otherwise doesn't have mappings)
+        # fall back to post-load conversion
+        if not _has_2d_mappings(model_type):
+            model = original_from_pretrained(*args, **kwargs)
+            linearize_moe(model)
+            return model
```
orestis-z: Maybe worth logging a warning here? Since this path converts the weights in-memory

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -0,0 +1,230 @@
...
+                    ExpertMLPWithoutGate(
+                        moe_config.hidden_dim,
+                        moe_config.intermediate_size,
+                        moe_config.use_bias,
+                        act_fn,
+                        moe_config.dtype,
+                    )
+                    for _ in range(moe_config.num_experts)
+                ]
+            )
```
orestis-z:
```suggestion
        if self.has_gate:
            expert_cls = ExpertMLPWithGate
            apply_gate_func = self._apply_gate
        else:
            expert_cls = ExpertMLPWithoutGate
            apply_gate_func = act_fn
        super().__init__(
            [
                expert_cls(
                    moe_config.hidden_dim,
                    moe_config.intermediate_size,
                    moe_config.use_bias,
                    apply_gate_func,
                    moe_config.dtype,
                )
                for _ in range(moe_config.num_experts)
            ]
        )
```

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -0,0 +1,230 @@
...
+        with skip_weights_initialize():
+            self = cls(config)
+
+        # TODO: experiment with copying views rather than data
+        for index in range(self.num_experts):
+            expert: ExpertMLPWithGate = self[index]
+
+            # load weights
+            gate_weight = experts.gate_up_proj[index, : self.intermediate_size]
+            up_weight = experts.gate_up_proj[index, self.intermediate_size :]
```
orestis-z: If weights are transposed aren't we slicing along the wrong dimension here? 

standard layout: weights stored as E x 2I x H
transposed layout: weights stored as E x H x 2I

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: There is already a warning [here](https://github.com/vllm-project/llm-compressor/pull/2647/changes#diff-181ab587b987dc5455e588e3c938b2b2e2ffa2647510f44745ee61d2a51616bcR107). Putting to here means that we also catch when oneshot is the one performing the conversion

tests/llmcompressor/modifiers/quantization/test_handling_shared_embeddings.py
kylesayrs: Spoke offline, tinysmokellama is a (malformed) checkpoint which has differing lm_head and embed_tokens weights, which breaks the behavior of this test in v5.

examples/awq/qwen3_moe_example_ddp.py
kylesayrs: Because this model does not have `shared_expert_gate`

src/llmcompressor/modeling/moe/conversion_mappings.py
kylesayrs: Spoke offline, no maintainance required because the fallback pathway (almost always) works

src/llmcompressor/modeling/moe/context.py
brian-dellabetta: discussed in code review that this is hopefully pretty robust, we will see

examples/quantizing_moe/deepseek_v4_example.py
brian-dellabetta: This is specific to deepseek right? not something users will have to do frequently outside of this model?

src/llmcompressor/modeling/moe/granitemoe.py
```diff
@@ -0,0 +1,76 @@
...
+
+        for i in range(self.num_experts):
+            expert_out = self[i](input_list[i].unsqueeze(0))[0]
+            output_list.append(expert_out)
+        results = torch.cat(output_list, dim=0)
+        return results
+
+
+# register in registry
+LinearExperts2D._registry[GraniteMoeParallelExperts] = GraniteMoeLinearExperts
```
brian-dellabetta, line 84: any reason you're not using the RegistryMixin + decorator pattern for these?

src/llmcompressor/modeling/moe/linear_experts.py
kylesayrs: Through research and experimentation we've found that calibrating each expert on every token yields the best results. You could theoretically get the same results by calibrating with 100x as many tokens so as to make sure that each expert gets enough activations, but we've found empirically that routing this way yields a valid calibration sample.

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: I agree! I've changed it to create a new class for each class seen, reusing classes where applicable

examples/quantization_w8a8_fp8/granite4_example.py
kylesayrs: Any changes made are necessary for v5. I've written and run extensive tests to check that all regexes are valid matches against the new model structures
https://github.com/vllm-project/llm-compressor/compare/kylesayrs/example_regex_tests?expand=1

kylesayrs: Our example tests should be able to cover the rest

examples/quantizing_moe/deepseek_v4_example.py
kylesayrs: Yeah, we can upstream a bug fix later

src/llmcompressor/modeling/moe/context.py
kylesayrs: My goal is to leverage transformers where applicable, and support pytorch models generically.

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: Yep, adding this for now and we can add a `load_compat` context in a followup

src/llmcompressor/modeling/moe/linear_experts.py
kylesayrs: Good call!

src/llmcompressor/modeling/moe/granitemoe.py
kylesayrs: Because the registery pattern is bad/ incomplete 🙃

The registry pattern is missing 2 things:
1. A refresh method to dynamically import subclasses
2. The ability to register with non-string keys. This registry needs to have `type[torch.nn.Module]` keys. The string name won't work, since the class itself is needed in order to use `get_use_experts_implementation_args`

src/llmcompressor/modeling/moe/helpers.py
```diff
@@ -0,0 +1,143 @@
...
+            # Handle @use_experts_implementation(...) with arguments
+            if isinstance(decorator, ast.Call) and isinstance(decorator.func, ast.Name):
+                if decorator.func.id == "use_experts_implementation":
+                    args = default_args
+
+                    # Extract keyword arguments
+                    for keyword in decorator.keywords:
+                        if keyword.arg in args:
+                            # Evaluate the constant value
+                            if isinstance(keyword.value, ast.Constant):
```
HDCharles, line 91: Why do we check ast.Constamt 3 times?

src/llmcompressor/typing.py
brian-dellabetta: discussed in design

src/llmcompressor/modeling/moe/linear_experts.py
kylesayrs: Really good catch, will fix this

src/llmcompressor/modeling/moe/helpers.py
kylesayrs: Ask claude

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: Not really, but I made an explicit function breakout as you suggested

src/llmcompressor/modeling/moe/granitemoe.py
```diff
@@ -0,0 +1,76 @@
...
+
+        torch.nn.ModuleList.__init__(
+            self,
+            [
+                torch.nn.Linear(input_size, output_size, bias=False, dtype=config.dtype)
+                for _ in range(num_experts)
+            ],
+        )
+
+    def forward(self, inputs: torch.Tensor, expert_size: list[int]):
```
HDCharles, line 60: i think it'd be ideal if we could explicate the difference between the default implementation and the specific implementation of the forward that we're adding for the model type.

kylesayrs: FS, I will better document this. The differences are
1. The weight is linearized
2. The forward supports the `calibrate_all_experts_flag`

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -0,0 +1,249 @@
...
+            else _default_apply_gate
+        )
+        return type("LinearExperts2D", (cls,), experts_cls_args)
+
+    @classmethod
+    @torch.no_grad()
+    def from_experts_module(
+        cls, experts: FusedExpertsProtocol, config: PreTrainedConfig
+    ):
+        if not cls.has_gate:
```
HDCharles: so we can only support modules that have a gate? i'm not sure i'm understanding the logic here,

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,125 @@
...
+    def load_compat():
+        stack = ...
+        if is_moe:
+            stack.enter_context(load_quantizable_moe)
+
+
+
+    """
+    original_from_pretrained = model_cls.from_pretrained
+    patched_fn_called = False
```
HDCharles, line 50: may want to attach model cls to this so that weird models with multiple moe's can work?

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -0,0 +1,125 @@
...
+    original_experts_cls = next(iter(non_linearized_moes.values())).__class__
+    linear_experts_cls = LinearExperts2D.create_linear_experts_cls(original_experts_cls)
+    for name, module in non_linearized_moes.items():
+        linear_moe = linear_experts_cls.from_experts_module(module, model.config)
+        model.set_submodule(name, linear_moe)
+
+
+def requires_linearize_moe(model: torch.nn.Module):
+    for module in model.modules():
+        if isinstance(module, FusedExpertsProtocol):
```
HDCharles: don't you need to check the registry?

setup.py
```diff
@@ -122,11 +122,7 @@ def localversion_func(version: ScmVersion) -> str:
         ),
         ("tqdm>=4.66.3,<=4.67.3" if BUILD_TYPE == "release" else "tqdm>=4.66.3"),
         ("torch>=2.10.0,<=2.11.0" if BUILD_TYPE == "release" else "torch>=2.10.0"),
-        (
-            "transformers>=4.56.1,<=4.57.6"
-            if BUILD_TYPE == "release"
-            else "transformers>=4.56.1,<=4.57.6"
-        ),
+        ("transformers>=5.9.0" if BUILD_TYPE == "release" else "transformers>=5.9.0"),
```
HDCharles: wtf?

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -127,7 +135,11 @@ def create_linear_experts_cls(
             if experts_cls_args["has_gate"]
             else _default_apply_gate
```
HDCharles: why not use apply_gate directly?
```suggestion
experts_cls_args["_apply_gate"] = experts_cls.getattr("_apply_gate", _default_apply_gate)
```

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -89,37 +89,34 @@ def patched(cls, *args, **kwargs):
...
         "checkpoint is already linearized (2D -> 3D -> 2D). If your checkpoint has 2D "
-        "experts, consider registering your own mappings. See TODO docs"
+        "experts, consider registering a load converter."  # TODO: add docs
     )
-    original_experts_cls = next(iter(non_linearized_moes.values())).__class__
-    linear_experts_cls = LinearExperts2D.create_linear_experts_cls(original_experts_cls)
-    for name, module in non_linearized_moes.items():
+
+    for name, module in non_linearized_moes:
+        linear_experts_cls = LinearExperts2D.create_linear_experts_cls(module.__class__)
```
HDCharles: should create_linear_expert_cls just be get_X? it looks ike you're creating a new class every time but i know this is actually checking if its already created and just reusing that.

src/llmcompressor/modeling/moe/helpers.py
```diff
@@ -34,8 +34,8 @@ class FusedExpertsProtocol(TorchModuleProtocol):
     gate_up_proj_bias: torch.nn.Parameter
     down_proj_bias: torch.nn.Parameter
 
-    @staticmethod
-    def __validate__(object: object):
+    @classmethod
```
HDCharles, line 37: porque?

src/llmcompressor/typing.py
```diff
@@ -26,39 +26,48 @@
...
+        """
+        Validate that an object satisfies this protocol's requirements.
+
+        Args:
+            obj: The object to validate
+
+        Returns:
+            True if the object satisfies the protocol, False otherwise
+        """
+        return True
```
HDCharles, line 73: ?

examples/autoround/ddp/ddp_qwen3_example.py
```diff
@@ -79,9 +79,7 @@ def config_deterministic():
 ###### DDP MODEL LOAD CHANGE #####
 init_dist()
 with load_offloaded_model():
```
HDCharles, line 81: If we're going to do the model_compat universal thing, we should probably do it in this PR

examples/awq/qwen3_moe_example_ddp.py
```diff
@@ -14,18 +14,17 @@
...
 # Select model and load it.
 MODEL_ID = "Qwen/Qwen3-30B-A3B"
 
 init_dist()
-with load_offloaded_model():
-    model = AutoModelForCausalLM.from_pretrained(
-        MODEL_ID, dtype="auto", device_map="auto_offload"
-    )
-tokenizer = AutoTokenizer.from_pretrained(MODEL_ID, trust_remote_code=True)
+with load_offloaded_model(), load_quantizable_moe():
```
HDCharles, line 25: can we just do load_compatible_model() always instead of sometimes having both, sometimes having one, sometimes having the other...etc to simplify usage?

HDCharles (changes requested):
i think we should unify the context managers for model loading now rather than later.

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: Fixed in a later commit

src/llmcompressor/modeling/moe/linear_experts.py
kylesayrs: I've just added support for non-gated experts

setup.py
kylesayrs: I've now upper bounded the release version

src/llmcompressor/modeling/moe/linear_experts.py
kylesayrs: Sure

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: Renamed

src/llmcompressor/modeling/moe/helpers.py
kylesayrs: Vibes. It seems right that `__validate__` should have access to the class it's validating with

src/llmcompressor/typing.py
kylesayrs: I don't see the problem, this is just the default implementation.

examples/autoround/ddp/ddp_qwen3_example.py
kylesayrs: I'd much rather do it in a follow up, but it's reasonable to ask for this before release. I'd still rather not though.

examples/awq/qwen3_moe_example_ddp.py
kylesayrs: See other comment

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: Spoke offline; the context only wraps one class at a time, so it's safe.

HDCharles (approved):

brian-dellabetta (approved):
Thanks for covering all this in a live review. I think we should allow a few biz days to test locally before releasing, or do a release candidate .rc1 if we are looking to release this right away.

examples/autoround/ddp/ddp_qwen3_example.py
kylesayrs: Spoke offline, will do in follow-up

kylesayrs:
There was a 5+ hour hang on `tests/llmcompressor/transformers/tracing/test_models.py` with this branch. I was never able to get a full stack trace, but I suspect the hang was a result of either (trying to execute weight conversion logic on cpu) or (machine slowdowns from 1.5T of cpu memory for large models, maybe v5 models are slightly larger).

Either way, I've resolved this by performing the tracing tests on `meta` device. I originally avoided this because I wasn't sure if any models relied on having a real device, but it seems safe now.

brian-dellabetta (approved):

PR #2655, opened 2026-04-26
[Model] DeepSeekV4

kylesayrs:


kylesayrs:
Accomplished by https://github.com/vllm-project/llm-compressor/pull/2647

Issue #2646, opened 2026-04-24
cheap KLD metric


HDCharles:
# Background

[KL Divergence](https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence) is a useful metric for measuring the similarity between two model's output distributions (i.e. to see how similar a quantized model is compared to a non quantized model). It requires extracting the output probabilities for two models and then comparing them.

In vllm its hard to extract logprobs. As an example, extracting full vocab logprobs for 10 tokens with llama-3-8B takes 1.1 tokens/sec, making a wikitext kl-divergence calculation take ~64 hours

https://gist.github.com/HDCharles/030a3c5d7439680afdbebd935758366d

Most of this time is spent transferring the massive tensor to disk. you could get around this by taking fewer logprobs, but the tail tends to dominate the logprobs calculation unless you're using a temperature way below the common .7-1.0 temperature used for most generation. 

so approximation won't work and the normal method won't work, we need a different approach.

# Approach

the primary issue is the amount of data that we need to extract from within vllm, one thing we can do to reduce this is extract a step before the logprobs have been generated. While the vocab size is 120k, the hidden dimension is just 4096 and can be used to generate the full logprobs. Using this would theoretically reduce our transfer time and storage size by a factor of ~30x.

Getting this data is also tricky but
in https://vllm.ai/blog/extract-hidden-states a method for extracting hidden states was developed for training speculative decoding models that require hidden state information.

From here we can theoretically do the following, run both models on vllm, extract the hidden state information before the lm_head, then to calculate KLDivergence, populate the lm_head (which is usually the same for both models, but could be 2 lm_heads if they differ), load the tokens in batches, uses lm_head to calculate the logprobs, and from there calculates kl-divergence.

This has the benefit of using existing vllm compoenents, saving on storage, and being fast.

# Task

Create a tool/script to quickly and efficiently calculate kl divergence using the approach above.
Test it, figure out how easy it is to generalize to different model architectures....etc

jwpark33:
@HDCharles  I'd like to contribute to this! Can I be assigned to work on this ?

HDCharles:
yes @jwpark33 it looks like there are a few open PRs after you requested to be assigned, feel free to use them as a jumping off point or collaborate those submitters.

jayakumarpujar:
@HDCharles Moving the design discussion here as suggested in #2659.

I want to add few points into the PR thread:

[...]

PR #2659, opened 2026-04-27
feat: add KL Divergence evaluation module using vLLM hidden states (#2646)

jayakumarpujar:
## Summary

Implements #2646 - a cheap KLD metric for measuring quantization quality by comparing output distributions between a baseline and quantized model.

**Problem**: Extracting full vocab logprobs from vLLM is too slow (~64hrs for WikiText on Llama-3-8B) due to the massive `[seq_len × 120k]` tensor transfer.

**Approach** (as described in the issue):
- Hook vLLM's `LogitsProcessor` via `LLM.apply_model` to capture **pre-lm_head hidden states** (`hidden_dim ~4096`) - ~30x smaller than full logprobs
- `enforce_eager=True` is always forced - CUDA graph replay silently skips Python hook callbacks
- Run prompts one at a time to avoid prompt-batching alignment issues
- Copy baseline `lm_head.weight` to CPU and apply offline to reconstruct log-probabilities for both models
- Compute token-weighted `KL(P_base || P_quant)` averaged over all token positions

## New files

| File | Description |
|------|-------------|
| `src/llmcompressor/evaluation/kld.py` | Core module - `KLDivergenceEvaluator`, `evaluate_kl_divergence` |
| `src/llmcompressor/evaluation/__init__.py` | Package init |
| `tests/unit/evaluation/test_kld.py` | 26 unit tests (all passing, CPU-only) |
| `examples/kld_evaluation/kld_metric.ipynb` | Colab notebook (validated on T4 GPU) |
| `examples/kld_evaluation/README.md` | Usage guide |

## Usage

```python
from llmcompressor.evaluation import evaluate_kl_divergence

result = evaluate_kl_divergence(
    base_model_id="meta-llama/Meta-Llama-3-8B",
    quantized_model_id="./Meta-Llama-3-8B-W4A16",
)
print(f"Mean KLD: {result.mean_kld:.6f}")
```

## Validated results (Colab T4, `facebook/opt-125m` W4A16)

**Base vs W4A16 (8 prompts):**
```
KLDivergenceResult(mean_kld=0.202843, num_prompts=8, num_tokens=95, skipped=0)

Per-prompt KLD:
  [0] 0.197243  |  The quick brown fox jumps over the lazy dog.
  [1] 0.170719  |  Machine learning models require large amounts of training data.
  [2] 0.206847  |  Quantization reduces model size by lowering the precision of weights.
  [3] 0.211776  |  Natural language processing has advanced significantly in recent years.
  [4] 0.202980  |  Efficient inference is critical for deploying large language models.
  [5] 0.228462  |  The transformer architecture revolutionized sequence modeling tasks.
  [6] 0.218196  |  Gradient descent optimizes model parameters by minimizing a loss function.
  [7] 0.186520  |  Attention mechanisms allow models to focus on relevant input tokens.
```

**Sanity check (same model → KLD = 0):**
```
Sanity check KLD (same model): 0.00000000  ✓
```

**WikiText-2 (64 samples, 6106 tokens):**
```
WikiText Mean KLD: 0.154366
Tokens evaluated:  6106
```
## Test plan

- [x] 26 unit tests passing (CPU-only, covers math, helpers, error handling, nan alignment, token weighting)
- [x] GPU end-to-end validated on Colab T4 with `facebook/opt-125m` W4A16
- [x] Sanity check (same model → KLD = 0.0) confirmed
- [x] WikiText-2 evaluation (64 samples) confirmed


jayakumarpujar:
@kylesayrs @brian-dellabetta 
Could you add the ready label to unblock the remaining CI checks?

HDCharles:
this looks different from the ask in the issue, i think it would make sense to establish what changes to that design you've made as well as the pros and cons in a discussion form. Being forced to use eager here seems like a massive cost that the other design can avoid.

jayakumarpujar:
> this looks different from the ask in the issue, i think it would make sense to establish what changes to that design you've made as well as the pros and cons in a discussion form. Being forced to use eager here seems like a massive cost that the other design can avoid.

Thanks for the feedback! You're right - I diverged from the spec-decode based 
design in the issue and should have mentioned that upfront.

Design differences:
- Issue asked for vLLM's ExtractHiddenStatesProposer / spec-decode pipeline 
  to capture hidden states inside the graph
- This PR uses LLM.apply_model + register_forward_hook on LogitsProcessor 
  instead - simpler to implement, but requires enforce_eager=True because 
  CUDA graph replay skips Python callbacks

Pros of this approach:
- No changes to vLLM internals, no speculative_config setup needed
- Works on any decoder model without model-specific config

Cons vs the original design:
- eager is a real cost - disables CUDA graph optimization for 
  the entire inference run
- The spec-decode path in the issue avoids this by running inside the graph

Happy to rework toward the original design using ExtractHiddenStatesProposer 
if that's the preferred direction, or discuss further here. What's your 
take?

HDCharles:
why don't we move the discussion to the issue if you think this approach merits consideration.

jayakumarpujar:
Closing in favour of the architectural direction discussed in #2646. The hook-based approach (enforce_eager + apply_model) diverges from what the maintainer wants - the hidden_states_extractor + disk-based two-phase flow is the right path, and PR #2663 is already heading there. The KLD math, _teardown_llm, and unit tests from this PR are available to reuse if useful.

PR #2663, opened 2026-04-28
Feat/issue 2646

rpathade:
**Summary:**

Implements #2646 — a tool to efficiently compute KL divergence between a base and quantized model using vLLM's hidden state extraction API.

**Problem:** Extracting full-vocab logprobs from vLLM is too slow (~64hrs for WikiText on Llama-3-8B) due to serializing [seq_len × 128k] tensors.

**Approach** (as described in the issue):
- Use vLLM's [hidden state extraction API](https://vllm.ai/blog/extract-hidden-states) to save post-norm hidden states (dim ~4096 vs vocab ~128k = ~30x smaller)
- Two-phase workflow: extract hidden states to disk, then compute KL offline
- Load only the lm_head weight from the checkpoint (no full model load needed)
- Reconstruct logprobs offline and compute token-weighted KL(P_base || P_quant)

**New files:**

| File | Description |
|------|-------------|
| `tools/kl_divergence/extract_hidden_states.py` | Phase 1 — vLLM hidden state extraction to safetensors |
| `tools/kl_divergence/compute_kl.py` | Phase 2 — offline KL divergence computation |
| `tools/kl_divergence/lm_head_utils.py` | Efficient lm_head weight loading from checkpoints |
| `tools/kl_divergence/test_local.py` | 10 unit tests (CPU-only, torch + safetensors) |
| `tools/kl_divergence/README.md` | Usage guide |

**Test plan:**
- 10 local unit tests passing (self-KL=0, KL>0 for different inputs, non-negativity, temperature scaling, E2E with safetensors I/O, token ID mismatch detection, dimension validation)
- End-to-end on Colab T4 with TinyLlama-1.1B: hidden state extraction via vLLM (16 samples) + self-comparison KL = 0.000000


tools/kl_divergence/lm_head_utils.py
```diff
@@ -0,0 +1,204 @@
...
+}
+
+_GPT2_WEIGHT_NAMES = {
+    "lm_head_weight": "lm_head.weight",
+    "lm_head_bias": None,
+    "embed_weight": "transformer.wte.weight",
+}
+
+ARCHITECTURE_WEIGHT_NAMES = {
+    "LlamaForCausalLM": _DEFAULT_WEIGHT_NAMES,
```
HDCharles: so i'm looking at model definitions and seeing that there's often a norm right after the last hidden state but before the lm_head, you may have to extract both of these

https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L410-L425

HDCharles:
do you have any benchmarks for how long it took?

tools/kl_divergence/lm_head_utils.py
```diff
@@ -0,0 +1,204 @@
...
+    "GraniteForCausalLM": _DEFAULT_WEIGHT_NAMES,
+    "InternLM2ForCausalLM": _DEFAULT_WEIGHT_NAMES,
+    "CohereForCausalLM": _DEFAULT_WEIGHT_NAMES,
+    "DeepseekV3ForCausalLM": _DEFAULT_WEIGHT_NAMES,
+    "GPTNeoXForCausalLM": _GPTNEOX_WEIGHT_NAMES,
+    "GPT2LMHeadModel": _GPT2_WEIGHT_NAMES,
+}
+
+
+def infer_weight_names(
```
HDCharles: i think most models include a

model.get_output_embeddings() interface that could make things easier.

 https://github.com/vllm-project/llm-compressor/blob/7a52e9e24558bbb4cf8c3a47c0eda630ea6e64be/src/llmcompressor/utils/transformers.py#L86

tools/kl_divergence/lm_head_utils.py
```diff
@@ -0,0 +1,204 @@
...
+import torch
+from safetensors import safe_open
+from transformers import AutoConfig
+
+# Architecture -> weight name mappings
+_DEFAULT_WEIGHT_NAMES = {
+    "lm_head_weight": "lm_head.weight",
+    "lm_head_bias": None,
+    "embed_weight": "model.embed_tokens.weight",
+}
```
HDCharles: i'm a bit unclear on why we need these mappings, it looks like the data_generation script in speculators that uses the same extraction framework doesn't need them, is this because the model support is narrower or something more nuanced?

https://github.com/vllm-project/speculators/blob/main/scripts/data_generation_offline.py

tools/kl_divergence/lm_head_utils.py
rpathade: Done.

tools/kl_divergence/lm_head_utils.py
rpathade: Right, checked the vLLM source - num_hidden_layers captures pre-norm. Now loading and applying the final norm before lm_head.

tools/kl_divergence/lm_head_utils.py
rpathade: mappings were only needed for loading `lm_head` offline from safetensors. Replaced with `init_empty_weights()` + `get_output_embeddings()`, so no manual config is needed now.

rpathade:
> do you have any benchmarks for how long it took?

Ran on Colab with Llama-3.1-8B, A100-40GB, 64 x 2048 tokens:

Extract:  89.5s  (1,464 tok/s)
Compute:  21.3s  (6,151 tok/s)
Total:    110.9s (1,182 tok/s)
Storage:  1,025 MB (63x smaller than full logprobs)
Self-KL:  0.0

[Colab notebook](https://colab.research.google.com/drive/17rJttZhVlFBbBlAxqJwIChoXiSqlP2_s?usp=sharing)

tools/kl_divergence/lm_head_utils.py
HDCharles: can you clarify how/where these comments are resolved? without explanation its unclear what changes correspond to which fixes..etc

HDCharles:
can you also comment on the issue so i can assign you, also if you need additional feedback feel free to reach out to me on the vllm slack where i'm also HDCharles

tools/kl_divergence/lm_head_utils.py
rpathade: so for the mappings I removed the whole dict and replaced it with a function that creates an empty model and calls `get_output_embeddings()` to figure out the weight names automatically. that's what the speculators script does too, just didn't need per-architecture mappings at all.

for the norm — yeah you were right about that, looked into it and the hidden states at `num_hidden_layers` are actually pre-norm. so now it also loads the norm weight from the checkpoint and applies it in `compute_kl.py` before passing through lm_head.

HDCharles:
It looks like you resolved the comments but can you indicate whether/where/how this was fixed for clarity?

rpathade:
> Just submitted my pending replies — sorry about that, didn't realize they were stuck in draft. Here's where each concern was addressed:
> 
> 1. **Architecture mappings removed** (`lm_head_utils.py`): Removed the entire per-architecture dict. Now using `init_empty_weights()` + `get_output_embeddings()` to discover lm_head weight names dynamically — same approach as the speculators script. No manual config needed.
> 
> 2. **Norm layer** (`compute_kl.py`, `lm_head_utils.py`): You were right — `num_hidden_layers` gives pre-norm hidden states. Now loading the norm weight from the checkpoint in `lm_head_utils.py` and applying it in `compute_kl.py` before the lm_head projection.

HDCharles:
Ok cool, I'll try to run some tests and see how it does

Issue #2490, opened 2026-03-19
Consolidate Intermediate Offloading


HDCharles:
we have a class [IntermediatesCache](https://github.com/vllm-project/llm-compressor/blob/cf3bd6463e8d471ad6c8cc20a6a9b053c178e555/src/llmcompressor/pipelines/cache.py#L33) which has a bunch of features for on/offloading intermeidates needed in various modifiers/transforms but then we have a few places doing adhoc offloading logic too

AWQ, [smooth_activation_stats](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/awq/base.py#L658-L661) is manually moved to cpu when not being used and device when it is

GPTQ, [hessian offloading](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/base.py#L253-L269) has its own offloading stuff

the reason for this is the IntermediatesCache does both on/off loading and list-like access logic

for this task i'm looking for someone to

1) change intermediatesCache to be for cacheing a single tensor onloading/offloading -> IntermediateCache
2) change existing usage of intermediatesCache to e.g. use a list of IntermediateCache rather than the cache itself behaving like a list `list.append(IntermediateCache(Tensor))`
3) update existing cached tensors to use IntermediatesCache
4) take the sequential prefetch code and move it to a function that checks the state to do/not do prefetch and the update usage to something like:    

for x in maybe_prefetch(batches_of_stuff)

menogrey:
Hi @HDCharles , i would love to work on this. I have checked the code and make some changes locally, and i have some confusion about the point 1. The class `IntermediatesCache` has an attr `batch_intermediates: list[IntermediateValues]`, you want to change it to `batch_intermediate: IntermediateValues` or just delete the dict of `IntermediateValue` to use `intermediate: tensor`?

HDCharles:
Yeah the intermediate value abstraction is kinda pointless with this refactor, feel free to remove it

HDCharles:
The PR revealed a major issue, the abstraction that i outlined loses the nice UX with automatic onloading since we can no longer steal the `__iter__` as the onloading point.

We need to go back to the drawing board: (@kylesayrs  @brian-dellabetta )

i guess high level the question is a design that can do all of...

A) automatically handle on/offloading
B) work well with prefetch
C) can support singletons without just using a batch with size 0.

Maybe this is workable in the following way:

# option 0 (current main)
[...]

PR #2664, opened 2026-04-28
[Refactor] Consolidate Intermediate Offloading

menogrey:
SUMMARY:
According to the discuss in this issue https://github.com/vllm-project/llm-compressor/issues/2490 and slack thread, the PR refactors IntermediatesCache to use a flat key-value store with generic keys instead of a batch-indexed structure, enabling more flexible caching patterns, and replace the adhoc offloading logic with this IntermediatesCache.

Key changes:
* **Cache refactor**: `IntermediatesCache` changed from a batch-indexed nested structure (`list[dict[str, IntermediateValue]]`) with string keys to a flat generic key-value store (`dict[TKey, Any]`) supporting arbitrary key types. `IntermediateValue` simplified from recursive wrapping of nested containers to only wrapping tensor leaves, using pytree utilities for handling nested structures.
* pipelines/sequential/pipeline.py: Batch data keys changed from implicit list indexing (batch_intermediates[batch_idx]) to explicit integer keys (batch_idx). Loss mask, update, and delete operations now use tuple keys (batch_idx, name) for individual item access instead of batch-level dictionary operations.
* modifiers/awq/base.py: _parent_args_cache replaced dict[Module, IntermediatesCache] (per-module separate caches) with a single IntermediatesCache using composite tuple keys (module, batch_idx, arg_name). Added _parent_args_batch_idx counter per module. _smooth_activation_stats replaced dict[str, list[torch.Tensor]] with IntermediatesCache using string keys.
* modifiers/gptq/base.py: _hessians_cache replaced dict[torch.nn.Module, torch.Tensor] with IntermediatesCache using module keys. Removed _maybe_onload_hessian context manager since offloading is now handled automatically by the cache.
* modifiers/pruning/sparsegpt/base.py and sgpt_sparsify.py: _hessians_cache replaced dict[torch.nn.Module, torch.Tensor] with IntermediatesCache using module keys. sparsify_weight function changed from accepting hessians_dict and manually deleting entries to accepting a single hessian tensor. Removed _maybe_onload_hessian context manager.

|Feature|	Old IntermediatesCache|	New IntermediatesCache|
|-|-|-|
|Storage|	list[dict[str, IntermediateValue]]|	dict[TKey, Any]|
|Type|	Non-generic|	Generic[TKey]|
|Key Type|	str only (nested by batch index)|	Any hashable type (int, tuple, str, etc.)|
|Constructor|	empty(num_batches, offload_device) or __init__(batch_intermediates)|	__init__(offload_device)|
|Fetch|	fetch(batch_index, input_names=None) -> dict|	__getitem__(key) / fetch(key) -> Any|
|Update|	update(batch_index, values: dict)|	update(key, value)|
|Delete|	delete(batch_index, consumed_names=None)|	delete(key)|
|Append|	append(values: dict) (auto batch index)|	append(key_prefix, value) -> (key, index)|
|Clear|	N/A|	clear()|
|Iterate|	iter(input_names=None)/iter_prefetch(input_names=None)|	iter(keys=None) / iter_prefetch(keys=None)|
|Nested handling|	Custom recursive match|	Pytree tree_flatten/tree_unflatten|

TEST PLAN:
make style
make quality
make test passed.

I noticed the local testing didn't fully cover all the changes like GPTQ, sparsegpt, any more tests suggested?

### DIFF
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -432,7 +432,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                 if isinstance(v, QuantizedKVCache):
                     values.arguments[k] = None
 
-            self._parent_args_cache.append((module, None), values.arguments)
+            self._parent_args_cache.append(module, values.arguments)
 
         def create_cache_smooth_activations_hook_fn(smooth_name):
             def cache_smooth_activations_hook(
@@ -461,14 +461,14 @@ class AWQModifier(Modifier, QuantizationMixin):
                     masked_activations = activations.flatten(0, -2)
 
                 # accumulate activation sum&count
-                new_sum = masked_activations.float().sum(dim=0).cpu()
-                new_count = torch.tensor(masked_activations.size(0)).cpu()
+                new_sum = masked_activations.float().sum(dim=0)
+                new_count = torch.tensor(masked_activations.size(0))
                 if smooth_name not in self._smooth_activation_stats:
                     self._smooth_activation_stats.update(
                         smooth_name,
                         [torch.zeros_like(new_sum), torch.zeros_like(new_count)],
                     )
-                x_sum, count = self._smooth_activation_stats.fetch(smooth_name)
+                x_sum, count = self._smooth_activation_stats[smooth_name]
                 x_sum += new_sum
                 count += new_count
                 self._smooth_activation_stats.update(smooth_name, [x_sum, count])
@@ -614,9 +614,9 @@ class AWQModifier(Modifier, QuantizationMixin):
         use_prefetch = active_session().state.sequential_prefetch
 
         if use_prefetch:
-            batch_iter = cache.iter_prefetch([(module, None)])
+            batch_iter = cache.iter_prefetch([(module)])
         else:
-            batch_iter = cache.iter([(module, None)])
+            batch_iter = cache.iter([(module)])
 
         outputs = [module(**batch_kwargs) for batch_kwargs in batch_iter]
         return [
@@ -655,10 +655,10 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         device = get_execution_device(mapping.parent)
 
-        x_sum, count = self._smooth_activation_stats.fetch(mapping.smooth_name)
+        x_sum, count = self._smooth_activation_stats[mapping.smooth_name]
         if is_distributed():
             x_sum, count = _allreduce_data_sum([x_sum, count])
-        x_mean = x_sum.to(device) / count.to(device)
+        x_mean = x_sum / count
 
         if self.duo_scaling:
             w_mean = self._compute_layer_means(mapping.balance_layers).to(device)
@@ -1041,13 +1041,6 @@ def get_lowest_common_ancestor_with_avoid(
 
 
 def _allreduce_data_sum(data: list[torch.Tensor]) -> list[torch.Tensor]:
-    # needs to be on device to broadcast
-    device = torch.device(
-        torch.accelerator.current_accelerator().type,
-        torch.accelerator.current_device_index(),
-    )
-    data = [datum.to(device) for datum in data]
-
     pending_comms = []
     for datum in data:
         pending_comms.append(
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -162,15 +162,11 @@ class AWQModifier(Modifier, QuantizationMixin):
 
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: list[ResolvedMapping] = PrivateAttr(default_factory=list)
-    # Cache list of forward input args for each parent module, one dict for each batch
-    _parent_args_cache: dict[Module, IntermediatesCache] = PrivateAttr(
-        default_factory=dict
-    )
-    # Dict[smooth layer name, [activation sums, activation counts]]
+    _parent_args_cache: IntermediatesCache = PrivateAttr(default=None)
+    _parent_args_batch_idx: dict[Module, int] = PrivateAttr(default_factory=dict)
     _smooth_activation_stats: dict[str, list[torch.Tensor]] = PrivateAttr(
         default_factory=dict
     )
-    # List to store error metrics for each layer
     _error_metrics: list[dict] = PrivateAttr(default_factory=list)
 
     def on_initialize(self, state: State, **kwargs) -> bool:
@@ -314,7 +310,9 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         self._log_error_metrics()
 
-        self._parent_args_cache.clear()
+        if self._parent_args_cache is not None:
+            self._parent_args_cache.clear()
+        self._parent_args_batch_idx.clear()
         self._smooth_activation_stats.clear()
         self._resolved_mappings.clear()
         self._error_metrics.clear()
@@ -438,7 +436,12 @@ class AWQModifier(Modifier, QuantizationMixin):
                 if isinstance(v, QuantizedKVCache):
                     values.arguments[k] = None
 
-            self._parent_args_cache[module].append(values.arguments)
+            batch_idx = self._parent_args_batch_idx.get(module, 0)
+            for arg_name, arg_value in values.arguments.items():
+                self._parent_args_cache.update(
+                    (module, batch_idx, arg_name), arg_value
+                )
+            self._parent_args_batch_idx[module] = batch_idx + 1
 
         def create_cache_smooth_activations_hook_fn(smooth_name):
             def cache_smooth_activations_hook(
@@ -480,13 +483,10 @@ class AWQModifier(Modifier, QuantizationMixin):
             return cache_smooth_activations_hook
 
         for mapping in self._resolved_mappings:
-            # parent kwargs needed for future forward passes
-            # same parent may appear multiple times in resolved mappings
-            if mapping.parent not in self._parent_args_cache:
-                self._parent_args_cache[mapping.parent] = IntermediatesCache(
-                    None,
-                    self.offload_device,
-                )
+            if self._parent_args_cache is None:
+                self._parent_args_cache = IntermediatesCache(self.offload_device)
+            if mapping.parent not in self._parent_args_batch_idx:
+                self._parent_args_batch_idx[mapping.parent] = 0
                 self.register_hook(
                     mapping.parent,
                     cache_parent_kwargs_hook,
@@ -613,21 +613,33 @@ class AWQModifier(Modifier, QuantizationMixin):
                 del self._smooth_activation_stats[mapping.smooth_name]
                 del orig_layer_weights
 
-        for v in self._parent_args_cache.values():
-            v.batch_intermediates.clear()
+        self._parent_args_cache.clear()
         self._assert_all_activations_consumed()
 
     @torch.no_grad()
     def _run_samples(self, module: Module) -> list[torch.Tensor]:
-        cache = self._parent_args_cache[module]
+        cache = self._parent_args_cache
         use_prefetch = active_session().state.sequential_prefetch
-        batch_iter = cache.iter_prefetch() if use_prefetch else cache
-        outputs = [module(**batch_kwargs) for batch_kwargs in batch_iter]
-        return [
-            # If tuple, assume that first argument is the input
-            output[0] if isinstance(output, tuple) else output
-            for output in outputs
-        ]
+
+        module_keys = [k for k in cache._store.keys() if k[0] == module]
+        if not module_keys:
+            return []
+
+        def group_by_fn(k):
+            return (k[0], k[1])
+
+        if use_prefetch:
+            batch_iter = cache.iter_prefetch_grouped(module_keys, group_by=group_by_fn)
+        else:
+            batch_iter = cache.iter_prefetch_grouped(module_keys, group_by=group_by_fn)
+
+        outputs = []
+        for group_dict in batch_iter:
+            batch_kwargs = {k[2]: v for k, v in group_dict.items()}
+            output = module(**batch_kwargs)
+            outputs.append(output[0] if isinstance(output, tuple) else output)
+
+        return outputs
 
     def _compute_best_scale(
         self,
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -163,9 +163,7 @@ class AWQModifier(Modifier, QuantizationMixin):
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: list[ResolvedMapping] = PrivateAttr(default_factory=list)
     _parent_args_cache: IntermediatesCache[tuple[Module, int]] = PrivateAttr(default=None)
-    _smooth_activation_stats: dict[str, list[torch.Tensor]] = PrivateAttr(
-        default_factory=dict
-    )
+    _smooth_activation_stats: IntermediatesCache[str] = PrivateAttr(default=None)
     _error_metrics: list[dict] = PrivateAttr(default_factory=list)
 
     def on_initialize(self, state: State, **kwargs) -> bool:
@@ -466,18 +464,21 @@ class AWQModifier(Modifier, QuantizationMixin):
                 new_sum = masked_activations.float().sum(dim=0).cpu()
                 new_count = torch.tensor(masked_activations.size(0)).cpu()
                 if smooth_name not in self._smooth_activation_stats:
-                    self._smooth_activation_stats[smooth_name] = [
-                        torch.zeros_like(new_sum),
-                        torch.zeros_like(new_count),
-                    ]
-                self._smooth_activation_stats[smooth_name][0] += new_sum
-                self._smooth_activation_stats[smooth_name][1] += new_count
+                    self._smooth_activation_stats.update(
+                        smooth_name,
+                        [torch.zeros_like(new_sum), torch.zeros_like(new_count)],
+                    )
+                x_sum, count = self._smooth_activation_stats.fetch(smooth_name)
+                x_sum += new_sum
+                count += new_count
+                self._smooth_activation_stats.update(smooth_name, [x_sum, count])
 
             return cache_smooth_activations_hook
 
         for mapping in self._resolved_mappings:
             if self._parent_args_cache is None:
                 self._parent_args_cache = IntermediatesCache(self.offload_device)
+                self._smooth_activation_stats = IntermediatesCache(self.offload_device)
 
                 self.register_hook(
                     mapping.parent,
@@ -535,7 +536,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                         "found to scale. This can occasionally occur in MoE models "
                         "when certain experts are not activated by calibration samples."
                     )
-                    del self._smooth_activation_stats[mapping.smooth_name]
+                    self._smooth_activation_stats.delete(mapping.smooth_name)
                     continue
                 if not all(
                     [fp16_output.isfinite().all() for fp16_output in fp16_outputs]
@@ -549,7 +550,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                         "ways. If you encounter this consistently, raise an issue at "
                         "https://github.com/vllm-project/llm-compressor/issues"
                     )
-                    del self._smooth_activation_stats[mapping.smooth_name]
+                    self._smooth_activation_stats.delete(mapping.smooth_name)
                     continue
 
                 orig_layer_weights = {
@@ -601,8 +602,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                     _smooth(layer, orig_layer_weights)
                 _smooth(smooth_layer, orig_layer_weights)
 
-                # remove caches needed to smooth this mapping
-                del self._smooth_activation_stats[mapping.smooth_name]
+                self._smooth_activation_stats.delete(mapping.smooth_name)
                 del orig_layer_weights
 
         self._parent_args_cache.clear()
@@ -655,7 +655,7 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         device = get_execution_device(mapping.parent)
 
-        x_sum, count = self._smooth_activation_stats[mapping.smooth_name]
+        x_sum, count = self._smooth_activation_stats.fetch(mapping.smooth_name)
         if is_distributed():
             x_sum, count = _allreduce_data_sum([x_sum, count])
         x_mean = x_sum.to(device) / count.to(device)
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -162,8 +162,7 @@ class AWQModifier(Modifier, QuantizationMixin):
 
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: list[ResolvedMapping] = PrivateAttr(default_factory=list)
-    _parent_args_cache: IntermediatesCache = PrivateAttr(default=None)
-    _parent_args_batch_idx: dict[Module, int] = PrivateAttr(default_factory=dict)
+    _parent_args_cache: IntermediatesCache[tuple[Module, int]] = PrivateAttr(default=None)
     _smooth_activation_stats: dict[str, list[torch.Tensor]] = PrivateAttr(
         default_factory=dict
     )
@@ -312,7 +311,6 @@ class AWQModifier(Modifier, QuantizationMixin):
 
         if self._parent_args_cache is not None:
             self._parent_args_cache.clear()
-        self._parent_args_batch_idx.clear()
         self._smooth_activation_stats.clear()
         self._resolved_mappings.clear()
         self._error_metrics.clear()
@@ -436,12 +434,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                 if isinstance(v, QuantizedKVCache):
                     values.arguments[k] = None
 
-            batch_idx = self._parent_args_batch_idx.get(module, 0)
-            for arg_name, arg_value in values.arguments.items():
-                self._parent_args_cache.update(
-                    (module, batch_idx, arg_name), arg_value
-                )
-            self._parent_args_batch_idx[module] = batch_idx + 1
+            self._parent_args_cache.append((module, None), values.arguments)
 
         def create_cache_smooth_activations_hook_fn(smooth_name):
             def cache_smooth_activations_hook(
@@ -485,8 +478,7 @@ class AWQModifier(Modifier, QuantizationMixin):
         for mapping in self._resolved_mappings:
             if self._parent_args_cache is None:
                 self._parent_args_cache = IntermediatesCache(self.offload_device)
-            if mapping.parent not in self._parent_args_batch_idx:
-                self._parent_args_batch_idx[mapping.parent] = 0
+
                 self.register_hook(
                     mapping.parent,
                     cache_parent_kwargs_hook,
@@ -621,25 +613,17 @@ class AWQModifier(Modifier, QuantizationMixin):
         cache = self._parent_args_cache
         use_prefetch = active_session().state.sequential_prefetch
 
-        module_keys = [k for k in cache._store.keys() if k[0] == module]
-        if not module_keys:
-            return []
-
-        def group_by_fn(k):
-            return (k[0], k[1])
-
         if use_prefetch:
-            batch_iter = cache.iter_prefetch_grouped(module_keys, group_by=group_by_fn)
+            batch_iter = cache.iter_prefetch([(module, None)])
         else:
-            batch_iter = cache.iter_prefetch_grouped(module_keys, group_by=group_by_fn)
+            batch_iter = cache.iter([(module, None)])
 
-        outputs = []
-        for group_dict in batch_iter:
-            batch_kwargs = {k[2]: v for k, v in group_dict.items()}
-            output = module(**batch_kwargs)
-            outputs.append(output[0] if isinstance(output, tuple) else output)
-
-        return outputs
+        outputs = [module(**batch_kwargs) for batch_kwargs in batch_iter]
+        return [
+            # If tuple, assume that first argument is the input
+            output[0] if isinstance(output, tuple) else output
+            for output in outputs
+        ]
 
     def _compute_best_scale(
         self,
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -162,8 +162,11 @@ class AWQModifier(Modifier, QuantizationMixin):
 
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: list[ResolvedMapping] = PrivateAttr(default_factory=list)
+    # Cache of forward input args for each parent module, one dict for each batch
     _parent_args_cache: IntermediatesCache[tuple[Module, int]] = PrivateAttr(default=None)
+    # Cache dict[smooth layer name, [activation sums, activation counts]]
     _smooth_activation_stats: IntermediatesCache[str] = PrivateAttr(default=None)
+    # List to store error metrics for each layer
     _error_metrics: list[dict] = PrivateAttr(default_factory=list)
 
     def on_initialize(self, state: State, **kwargs) -> bool:
@@ -476,6 +479,8 @@ class AWQModifier(Modifier, QuantizationMixin):
             return cache_smooth_activations_hook
 
         for mapping in self._resolved_mappings:
+            # parent kwargs needed for future forward passes
+            # same parent may appear multiple times in resolved mappings
             if self._parent_args_cache is None:
                 self._parent_args_cache = IntermediatesCache(self.offload_device)
                 self._smooth_activation_stats = IntermediatesCache(self.offload_device)
@@ -602,6 +607,7 @@ class AWQModifier(Modifier, QuantizationMixin):
                     _smooth(layer, orig_layer_weights)
                 _smooth(smooth_layer, orig_layer_weights)
 
+                # remove caches needed to smooth this mapping
                 self._smooth_activation_stats.delete(mapping.smooth_name)
                 del orig_layer_weights
 
--- a/src/llmcompressor/modifiers/awq/base.py
+++ b/src/llmcompressor/modifiers/awq/base.py
@@ -163,7 +163,9 @@ class AWQModifier(Modifier, QuantizationMixin):
     # Private vars set during initialization, cleared during finalization
     _resolved_mappings: list[ResolvedMapping] = PrivateAttr(default_factory=list)
     # Cache of forward input args for each parent module, one dict for each batch
-    _parent_args_cache: IntermediatesCache[tuple[Module, int]] = PrivateAttr(default=None)
+    _parent_args_cache: IntermediatesCache[tuple[Module, int]] = PrivateAttr(
+        default=None
+    )
     # Cache dict[smooth layer name, [activation sums, activation counts]]
     _smooth_activation_stats: IntermediatesCache[str] = PrivateAttr(default=None)
     # List to store error metrics for each layer
--- a/src/llmcompressor/modifiers/gptq/base.py
+++ b/src/llmcompressor/modifiers/gptq/base.py
@@ -261,7 +261,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
                 tuple(), device=get_execution_device(module)
             )
 
-        # Fetch hessian (onloads to original device automatically), accumulate, then update
+        # Fetch hessian (onloads to original device automatically)
         hessian = self._hessians_cache[module]
         hessian, self._num_samples[module] = accumulate_hessian(
             inp, module, hessian, self._num_samples[module]
--- a/src/llmcompressor/modifiers/gptq/base.py
+++ b/src/llmcompressor/modifiers/gptq/base.py
@@ -282,7 +282,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
         world_size = dist.get_world_size()
 
         # Get module list from cache
-        module_list = list(self._hessians_cache._store.keys())
+        module_list = list(self._hessians_cache.iter_keys())
 
         # Assign modules to ranks
         module_list, rank_to_modules, module_to_rank = greedy_bin_packing(
--- a/src/llmcompressor/modifiers/gptq/base.py
+++ b/src/llmcompressor/modifiers/gptq/base.py
@@ -262,7 +262,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
             )
 
         # Fetch hessian (onloads to original device automatically), accumulate, then update
-        hessian = self._hessians_cache.fetch(module)
+        hessian = self._hessians_cache[module]
         hessian, self._num_samples[module] = accumulate_hessian(
             inp, module, hessian, self._num_samples[module]
         )
@@ -288,7 +288,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
         module_list, rank_to_modules, module_to_rank = greedy_bin_packing(
             module_list,
             world_size,
-            item_weight_fn=lambda mod: self._hessians_cache.fetch(mod).shape[0],
+            item_weight_fn=lambda mod: self._hessians_cache[mod].shape[0],
         )
 
         # send hessians to assigned ranks
@@ -311,7 +311,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
                 align_module_device(module),
                 CompressionLogger(module) as comp_logger,
             ):
-                hessian = self._hessians_cache.fetch(module) / num_samples
+                hessian = self._hessians_cache[module] / num_samples
                 self._hessians_cache.delete(module)
                 self._num_samples.pop(module)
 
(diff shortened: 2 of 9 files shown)


src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -124,7 +124,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
 
     # private variables
     _module_names: Dict[torch.nn.Module, str] = PrivateAttr(default_factory=dict)
-    _hessians: Dict[torch.nn.Module, torch.Tensor] = PrivateAttr(default_factory=dict)
+    _hessians_cache: IntermediatesCache[torch.nn.Module] = PrivateAttr(default=None)
```
HDCharles: can we default to the IntermediatesCache itself? otherwise the type hint is wrong

can change the offload_device from None in on_start?

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -150,28 +162,47 @@ class AWQModifier(Modifier):
...
-    )
-    # Dict[smooth layer name, [activation sums, activation counts]]
-    _smooth_activation_stats: dict[str, list[torch.Tensor]] = PrivateAttr(
-        default_factory=dict
+    # Cache of forward input args for each parent module, one dict for each batch
+    _parent_args_cache: IntermediatesCache[tuple[Module, int]] = PrivateAttr(
+        default=None
     )
+    # Cache dict[smooth layer name, [activation sums, activation counts]]
+    _smooth_activation_stats: IntermediatesCache[str] = PrivateAttr(default=None)
```
HDCharles: same issue

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -295,7 +312,8 @@ def on_finalize(self, state: State, **kwargs) -> bool:
 
         self._log_error_metrics()
 
-        self._parent_args_cache.clear()
+        if self._parent_args_cache is not None:
```
HDCharles: don't need if correct the default

src/llmcompressor/modifiers/awq/base.py
HDCharles, line 180: can set offload_device for caches after this maybe?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -174,8 +171,19 @@ def __call__(
...
+                                    # Merge output into existing batch dict
+                                    existing_batch = activations[batch_idx]
+                                    merged = {**existing_batch, **output}
+                                    activations.update(batch_idx, merged)
+                                if subgraph.consumed_names:
+                                    # Remove consumed names from batch dict
+                                    existing_batch = activations[batch_idx]
+                                    for name in subgraph.consumed_names:
+                                        existing_batch.pop(name, None)
+                                    activations.update(batch_idx, existing_batch)
```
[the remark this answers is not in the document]
HDCharles: this seems like a real problem i see 2 solutions

1) we could use more granular tokens so its easier to update/delete

```
for batch_idx, batch in enumerate(dataloader):
   for dict_key, value in batch:
       cache.append([batch_idx, dict_key], value)

for key in subgraph.consumed_names:
    cache.delete([batch_idx, key])
for key, value in output
    cache.update([batch_idx, key], value])
```

then update _get_batches a bit

pros: 
would simplfy update/delete
we already need to construct the batch in _get_batches so this would be a minor change
no changes to intermediates cache functionality in the PR
cons:
changes the way we handle batches compared to other situations

2) we could add functionality for keeping things offloaded and interacting with them

```
     existing_batch = activation.keep_offloaded(batch_idx)
     merged = {**existing_batch, **output}
     activations.update(batch_idx, merged, onload_device)
```

pros:
keeps the way we handle batches compared to other situations
likely need this functionality regardless
cons:
also need to add functionality to update, keep_offloaded
more complicated

src/llmcompressor/pipelines/cache.py
```diff
@@ -76,144 +58,171 @@ def from_dataloader(
...
+                k
+                for k in self._store.keys()
+                if isinstance(k, tuple)
+                and len(k) == len(prefix) + 1
+                and k[:-1] == prefix
+                and isinstance(k[-1], int)
+            ]
+
+            next_idx = max((k[-1] for k in matching_keys), default=-1) + 1
+            full_key = prefix + (next_idx,)
```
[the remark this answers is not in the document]
HDCharles: agree, this is probably a good idea to track, 

```
self.cur_index[key_prefix]=default_dict(int)

...

def append(key_prefix, value)
   self.cur_index[key_prefix]+=1
   # handle situations where self.cur_index hadn't been populated yet because of update
   # or maybe should just do an assert to not mix storage methods?
   while self._store[[key_prefix, cur_index[key_prefix]]:
       self.cur_index[key_prefix]+=1
   self._store[[key_prefix, cur_index[key_prefix]]=value
```

src/llmcompressor/modifiers/awq/base.py
```diff
@@ -150,28 +162,47 @@ class AWQModifier(Modifier):
...
     _resolved_mappings: list[ResolvedMapping] = PrivateAttr(default_factory=list)
-    # Cache list of forward input args for each parent module, one dict for each batch
-    _parent_args_cache: dict[Module, IntermediatesCache] = PrivateAttr(
-        default_factory=dict
-    )
-    # Dict[smooth layer name, [activation sums, activation counts]]
-    _smooth_activation_stats: dict[str, list[torch.Tensor]] = PrivateAttr(
-        default_factory=dict
+    # Cache of forward input args for each parent module, one dict for each batch
+    _parent_args_cache: IntermediatesCache[tuple[Module, int]] = PrivateAttr(
```
brian-dellabetta: Since we are making the value generic, can we update the class definition so that we can do something like this?
```suggestion
    _parent_args_cache: IntermediatesCache[tuple[Module, int], dict[str, torch.Tensor]] = PrivateAttr(
```

you can do this by adding a TValue like so:
```python
from typing import TypeVar, Generic

# Define type variables
TKey = TypeVar("TKey")
TValue = TypeVar("TValue")

# Inherit from Generic[TKey, TValue] to make the class generic
class MyPair(Generic[TKey, TValue]):
    def __init__(self, key: TKey, value: TValue):
        self.key = key
        self.value = value
```

src/llmcompressor/pipelines/sequential/pipeline.py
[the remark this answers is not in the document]
menogrey: Use 1, updated.

src/llmcompressor/pipelines/cache.py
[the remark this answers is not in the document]
menogrey: Updated, add self._prefix_counters and self._prefix_indices for cache.

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -68,7 +76,7 @@ def __call__(
             to the cpu between each batch in order to save memory
 
         This pipeline requires that the model be traceable with respect to data from the
-        data loader. This may be an issue for vision models with vision datasets, due
+        data loader. This may be an issue for vision models with vision datasets, due to
```
brian-dellabetta: possibly added by mistake?
```suggestion
        data loader. This may be an issue for vision models with vision datasets, due
```

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -208,6 +210,11 @@ def on_start(self, state: State, event: Event, **kwargs):
         # assume quantization has been initialized by this modifier or one before it
         QuantizationMixin.start_calibration(self, state.model)
 
+        # Configure hessians cache offload device
+        self._hessians_cache.offload_device = (
```
brian-dellabetta: rather than creating this with a default factory above and setting offload_device after the fact, better to set it as none above and create it here. We don't want to encourage users to update offload_device after creation, in the event something was already offloaded to the previous offload_device value

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -467,26 +466,27 @@ def cache_smooth_activations_hook(
...
-                self._smooth_activation_stats[smooth_name][0] += new_sum
-                self._smooth_activation_stats[smooth_name][1] += new_count
+                    self._smooth_activation_stats.update(
+                        smooth_name,
+                        [torch.zeros_like(new_sum), torch.zeros_like(new_count)],
+                    )
+                x_sum, count = self._smooth_activation_stats[smooth_name]
+                x_sum += new_sum
+                count += new_count
+                self._smooth_activation_stats.update(smooth_name, [x_sum, count])
```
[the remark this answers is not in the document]
brian-dellabetta: yeah since these are not big and need to be accumulated, it probably makes sense to not use IntermediatesCache for that

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -264,22 +271,19 @@ def calibrate_module(
...
 
-        # Accumulate hessian with input with optional offloading
-        with self._maybe_onload_hessian(module):
-            self._hessians[module], self._num_samples[module] = accumulate_hessian(
-                inp,
-                module,
-                self._hessians[module],
-                self._num_samples[module],
-            )
+        # Fetch hessian (onloads to original device automatically)
```
HDCharles, line 286: i think we need to resolve this problem

a<-a+b

onload: a
add: a = b+a
offload: a

is going to be way slower than

offload b
add a=b+a

I think this is going to require some kind of 'disable offloading' functionality

src/llmcompressor/modifiers/transform/awq/base.py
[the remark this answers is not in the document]
HDCharles: same issue here

tests/llmcompressor/pipelines/test_cache.py
```diff
@@ -1,166 +1,274 @@
...
+    # When onload == processing: ≈ (num_items + 1) * delay = 0.55s (vs 1.0s sequential)
+    # Expected speedup: ~1.8x
+    # Prefetch should be at least 70% faster than sequential (speedup >= 1.7x)
+    # With large delays, threading overhead becomes negligible
+    speedup = sequential_time / prefetch_time
+    assert speedup >= 1.7, (
+        f"Prefetch should provide significant speedup. "
+        f"Sequential: {sequential_time:.3f}s, Prefetch: {prefetch_time:.3f}s, "
+        f"Speedup: {speedup:.2f}x (expected >= 1.7x)"
+    )
```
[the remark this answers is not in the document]
brian-dellabetta: agree with this, @codreabbitai can you explain what you mean by "Prefer a deterministic overlap check with synchronization primitives"?

src/llmcompressor/modifiers/transform/awq/base.py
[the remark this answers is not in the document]
HDCharles: this is going to error for the same reason

src/llmcompressor/pipelines/cache.py
```diff
@@ -1,73 +1,58 @@
...
-    Currently supports nested offloading of dataclass instances and tuples
+    Uses pytree internally to handle nested containers (dicts, tuples, lists).
+    IntermediateValue wrappers are placed at tensor leaf positions only.
 
-    Construct using `empty` and `from_dataloader` class methods
+    Construct using `from_dataloader` class method or directly with offload_device
     """
 
-    batch_intermediates: list[IntermediateValues]
+    _store: dict[TKey, Any]
```
brian-dellabetta: can be this now right?
```suggestion
    _store: dict[TKey, TValue]
```

src/llmcompressor/pipelines/cache.py
```diff
@@ -250,122 +302,119 @@ def _fetch_and_record(batch_index):
...
+        self._prefix_indices.clear()
 
     def __len__(self) -> int:
-        return len(self.batch_intermediates)
+        """Return the number of entries in the cache."""
+        return len(self._store)
 
     @classmethod
-    def _onload_value(cls, intermediate: IntermediateValue) -> Any:
+    def _onload_value(cls, value: Any) -> Any:
```
brian-dellabetta: And I think this can be
```suggestion
    def _onload_value(cls, value: TValue) -> TValue:
```

src/llmcompressor/pipelines/cache.py
```diff
@@ -250,122 +302,119 @@ def _fetch_and_record(batch_index):
...
+                else:
+                    onloaded_leaves.append(tensor)
+            else:
+                onloaded_leaves.append(leaf)
+        return tree_unflatten(onloaded_leaves, spec)
 
     @classmethod
     def _offload_value(
         cls,
         value: Any,
```
brian-dellabetta:
```suggestion
        value: TValue,
```

src/llmcompressor/pipelines/cache.py
```diff
@@ -250,122 +302,119 @@ def _fetch_and_record(batch_index):
...
         offload_device: torch.device | None,
         onload_device: torch.device | None = None,
-    ) -> IntermediateValue:
+    ) -> Any:
```
brian-dellabetta:
```suggestion
    ) -> TValue:
```

brian-dellabetta (commented):
Thanks @menogrey for the updates, I have a bunch of small nits but otherwise i think this is really close. thanks for working through this with us, I really like the look of it

src/llmcompressor/pipelines/cache.py
```diff
@@ -76,144 +61,211 @@ def from_dataloader(
...
 
-    def update(self, batch_index: int, values: dict[str, Any]):
+        :param key: key to store value under
+        :param value: value to store (tensors will be offloaded)
         """
-        Update/put values belonging to a batch
+        self._store[key] = self._offload_value(value, self.offload_device)
+
+        # Update prefix counter and indices to ensure append doesn't overwrite
+        if isinstance(key, int):
```
HDCharles, line 121: huh? don't we want this to work for e.g. 

cache.update(module, value)

doesn't this break that?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -32,20 +32,28 @@ def _get_batches(
...
-    for batch_idx, inputs in tqdm(
-        enumerate(batch_source), total=num_batches, desc=desc
-    ):
-        yield batch_idx, inputs
+    all_keys = []
+    for batch_idx in range(num_batches):
+        for name in input_names:
+            full_key = (batch_idx, name)
+            if full_key in activations:
+                all_keys.append(full_key)
```
HDCharles: this is just itertools.product
```suggestion
all_keys = [(x,) for x in product(range(num_batches), input_names) if [x] in activations]
```

tests/llmcompressor/pipelines/test_cache.py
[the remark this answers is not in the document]
brian-dellabetta: @coderabbitai please explain what you mean by "Prefer a deterministic overlap check with synchronization primitives". can you suggest a code change to do this?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -32,20 +32,28 @@ def _get_batches(
...
+
+    iter_fn = activations.iter_prefetch if sequential_prefetch else activations.iter
+    value_iter = iter_fn(keys=all_keys)
+
+    for batch_idx in tqdm(range(num_batches), desc=desc):
```
HDCharles: this isn't doing prefetch correct, right? its not getting the whole dict

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -147,8 +155,17 @@ def __call__(
...
+                                if outputs is not None:
+                                    if isinstance(outputs, tuple) and len(outputs) == 1:
+                                        outputs = outputs[0]
+                                    for key, value in outputs.items():
+                                        activations.update((batch_idx, key), value)
+                                # Delete consumed names with granular keys
+                                for key in subgraph.consumed_names:
+                                    full_key = (batch_idx, key)
+                                    if full_key in activations:
+                                        activations.delete(full_key)
```
HDCharles: might be cleaner to make sure activations.delete(full_key) just noops if the key isn't found.

```
for key in subgraph.consumed_names:
        activations.delete((batch_idx, key))
```

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -147,8 +155,17 @@ def __call__(
 
                         if not dataset_args.propagate_error:
                             if subgraph_index < num_subgraphs - 1:
-                                activations.update(batch_idx, outputs)
-                                activations.delete(batch_idx, subgraph.consumed_names)
+                                # Update outputs with granular keys
+                                if outputs is not None:
+                                    if isinstance(outputs, tuple) and len(outputs) == 1:
+                                        outputs = outputs[0]
```
HDCharles: howd we get away without this stuff before?

HDCharles:
ok so there are some nits but there are a few structural issues still. I think we are going to need a way to interact with offloaded values without onloading them in a few instances.

1) prefetch is now broken this is my fault i didn't think about how the granular structure would break prefetch

because of the granular dict structure, we're prefetching 1 tensor at a time so when we do

for batch_idx
    for dict_key
         batch[key] = next(value_iter) # final next will prefetch next tensor but not whole batch
    yield batch # no real prefetching will happen here when there's actual io deadtime

this can be fixed in 2 ways

a) we go back to what you had before but add functionality for fetching without offloading

this change was initially made because we had to do things like

dict_to_update = cache[batch_idx] # onloads entire dict
updated_dict = do_updates(dict_to_update, update_dict)
catch[batch_idx] = updated_dict # offloads entire dict

if we can do

dict_to_update = cache.fetch_no_onload(batch_idx) # keeps dict offloaded
updated_dict = do_updates(dict_to_update, update_dict)
catch.update(batch_idx, updated_dict) # offloads everything that was updatedthen we won't have this issue. This may be the cleaner solution in all honesty.

b) instead of

iter_fn(keys=[(b1,k1),(,b1,k2),(b2,k1)(b2,k2)...etc]) you could do
iter_fn(keys=[[(b1,k1),(,b1,k2)],[(b2,k1),(b2,k2)],[...],..etc])

so we specify both what order and how many to prefetch

2) accumulation is going to be super slow

this will affect awq, gptq

we need to do

old_val = cache[module] # onloads old_val
updated_val = old_val + new_val
cache[module] = updated_val # offloads updated_val

as is we have no way of accessing the offloads without onloading them. This seems relatively easy to fix

old_val = cache.fetch_no_onload(module) # onloads old_val
updated_val = old_val + new_val
cache.update(module) = updated_val # offloads updated_val

So Proposal:

we add 2 new pieces of functionality:

1) fetch_no_onload
cache.fetch_no_onload ([@Brian Dellabetta](https://vllm-dev.slack.com/team/U08A4GJS7AR) better name?)  does the same thing as normal fetch but doesn't do the onload, so we just need to set it up to do the normal key handling and then leaf cleaning without onloading:

leaves, spec = tree_flatten(value)
cleaned_leaves = [x.value if isinstance(x, IntermediateValue) else x for x in leaves]
return tree_unflatten(cleaned_leaves, spec)

however this adds a new issue when we want to update all the leaves will be on the wrong device so we add

2) force onload_device update/append functionality

def update(key, value, onload_device=None):
      ...

def append(key_prefix, value, onload_device=None):
      ...
      self._store[full_key] = self._offload_value(value, self.offload_device, onload_device)

the onload device functionality already lives in _offload_value  so this should be a minor change.

3) sequential pipeline goes back to how it was before
4) fix accumulation in gptq/awq to stay offloaded

how does this sound?

src/llmcompressor/modifiers/gptq/base.py
menogrey: Previous comment from @HDCharles  mention that using default factory to initialize cache for type hint but I prefer using None , all updated.

src/llmcompressor/modifiers/transform/awq/base.py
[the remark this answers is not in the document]
menogrey: updated.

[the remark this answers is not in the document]
menogrey: updated. Now accumulate on offloading.

src/llmcompressor/modifiers/gptq/base.py
menogrey: Now if `offload_hessians` is True, the hessian will onload and offload, same as before. You mention to stay offloaded, should I change the `accumulate_hessian`?  the method seems still require execution on GPU. And also the sparsegpt using a different `accumulate_hessian`

src/llmcompressor/pipelines/cache.py
menogrey: This is a process for fast index search, add testcase for it now.

src/llmcompressor/pipelines/sequential/pipeline.py
menogrey: Updated

src/llmcompressor/pipelines/sequential/pipeline.py
menogrey: Updated in delete().

src/llmcompressor/pipelines/sequential/pipeline.py
menogrey: Maybe this is a mistake, delete it.

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -123,7 +123,9 @@ class GPTQModifier(Modifier, QuantizationMixin):
 
     # private variables
     _module_names: Dict[torch.nn.Module, str] = PrivateAttr(default_factory=dict)
-    _hessians: Dict[torch.nn.Module, torch.Tensor] = PrivateAttr(default_factory=dict)
+    _hessians_cache: IntermediatesCache[torch.nn.Module, torch.Tensor] = PrivateAttr(
+        default=None
```
HDCharles: typehint is bad again right? need to add | None

src/llmcompressor/modifiers/gptq/base.py
HDCharles: yeah i think it was bad before, we should also edit accumulate_hessian to not onload. Since we're changing how the offloading works. This is in gptq_quantize.py

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -445,23 +451,21 @@ def cache_smooth_activations_hook(
                 new_sum = masked_activations.float().sum(dim=0).cpu()
                 new_count = torch.tensor(masked_activations.size(0)).cpu()
```
HDCharles:
```suggestion
                new_sum = masked_activations.float().sum(dim=0).to(self.offload_device)
                new_count = torch.tensor(masked_activations.size(0)).to(self.offload_device)
```

src/llmcompressor/modifiers/gptq/base.py
HDCharles: also should probably do this in the same way as smooth_activation_stats

i.e. 

1) make this be [hessian, count]
2) change accumulate_hessian to calculate_hessian, calculate_hessian takes input of just inp (delete the conv2d/conv1d functionality) and return [hessian, num_samples]
3) then do the update in the same way as smooth_activation_stats

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -166,10 +168,14 @@ def __call__(
                             ):
                                 output = subgraph.forward(model, **inputs)
                                 if subgraph_index < num_subgraphs - 1:
-                                    activations.update(batch_idx, output)
-                                    activations.delete(
-                                        batch_idx, subgraph.consumed_names
-                                    )
+                                    # Get raw batch dict (no onload of existing tensors)
+                                    raw_batch = activations.fetch_no_onload(batch_idx)
```
HDCharles, line 179: nice

src/llmcompressor/pipelines/cache.py
HDCharles: oh i understande

src/llmcompressor/pipelines/cache.py
```diff
@@ -1,73 +1,58 @@
...
-    def __init__(
-        self,
-        batch_intermediates: list[IntermediateValues] | None = None,
-        offload_device: torch.device | None = "cpu",
-    ):
-        self.batch_intermediates = batch_intermediates or []
+    def __init__(self, offload_device: torch.device | None = None):
+        self._store = {}
+        self._prefix_counters: dict[tuple, int] = {}
+        self._prefix_indices: dict[tuple, list[int]] = {}
```
HDCharles: this should be a set

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -585,18 +589,22 @@ def _smooth(
...
+        cache = self._parent_args_cache
         use_prefetch = active_session().state.sequential_prefetch
-        batch_iter = cache.iter_prefetch() if use_prefetch else cache
+
+        if use_prefetch:
+            batch_iter = cache.iter_prefetch([(module)])
+        else:
+            batch_iter = cache.iter([(module)])
+
         outputs = [module(**batch_kwargs) for batch_kwargs in batch_iter]
```
HDCharles:
```suggestion
        iter_fn = cache.iter_prefetch if use_prefetch else cache.iter
        outputs = [module(**batch_kwargs) for batch_kwargs in iter_fn([module])]
```

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -637,7 +645,7 @@ def _compute_best_scale(
         x_sum, count = self._smooth_activation_stats[mapping.smooth_name]
         if is_distributed():
             x_sum, count = _allreduce_data_sum([x_sum, count])
-        x_mean = x_sum.to(device) / count.to(device)
+        x_mean = x_sum / count
```
HDCharles, line 654: wooooo we did it, no more .to(device)

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -445,23 +451,21 @@ def cache_smooth_activations_hook(
...
-                self._smooth_activation_stats[smooth_name][0] += new_sum
-                self._smooth_activation_stats[smooth_name][1] += new_count
+                    self._smooth_activation_stats.update(
+                        smooth_name,
+                        [torch.zeros_like(new_sum), torch.zeros_like(new_count)],
+                    )
+                x_sum, count = self._smooth_activation_stats.fetch_no_onload(smooth_name)
+                x_sum += new_sum
+                count += new_count
+                self._smooth_activation_stats.update(smooth_name, [x_sum, count], get_execution_device(_module))
```
HDCharles:
```suggestion
                device = get_execution_device(_module)
                if smooth_name not in self._smooth_activation_stats:
                    new = [torch.zeros_like(new_sum), torch.zeros_like(new_count)]
                    self._smooth_activation_stats.update(smooth_name, new)
                x_sum, count = self._smooth_activation_stats.fetch_no_onload(smooth_name)
                x_sum += new_sum
                count += new_count
                self._smooth_activation_stats.update(smooth_name, [x_sum, count], device)
```

src/llmcompressor/modifiers/pruning/sparsegpt/base.py
```diff
@@ -165,7 +157,8 @@ def on_finalize(self, state: State, **kwargs) -> bool:
         if len(self._num_samples) > 0:
             raise ValueError(f"Failed to compress {len(self._num_samples)} modules")
 
-        self._hessians = dict()
+        if self._hessians_cache is not None:
+            self._hessians_cache.clear()
```
HDCharles:
```suggestion
        self._hessians_cache.clear() if self._hessians_cache else None:

```

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -398,19 +408,8 @@ def on_finalize(self, state: State, **kwargs) -> bool:
         if len(self._num_samples) > 0:
             raise ValueError(f"Failed to compress {len(self._num_samples)} modules")
 
-        self._hessians = dict()
+        if self._hessians_cache is not None:
+            self._hessians_cache.clear()
```
HDCharles:
```suggestion
        self._hessians_cache.clear() if self._hessians_cache else None:
```

src/llmcompressor/pipelines/cache.py
```diff
@@ -250,122 +321,116 @@ def _fetch_and_record(batch_index):
...
+                        and torch.device(device).type
+                        == torch.accelerator.current_accelerator().type
+                    )
+                    onloaded_leaves.append(
+                        tensor.to(device=device, non_blocking=non_blocking)
+                    )
+                else:
+                    onloaded_leaves.append(tensor)
+            else:
+                onloaded_leaves.append(leaf)
```
HDCharles: i think we can further simplify/clean

```suggestion
            if not isinstance(leaf, IntermediateValue):
                onloaded_leaves.append(leaf)
                continue
            tensor = leaf.value
            device = leaf.device
            non_blocking = (
                tensor.is_pinned()
                and torch.accelerator.is_available()
                and torch.device(device).type
                == torch.accelerator.current_accelerator().type
            )
            tensor = tensor.to(device=device, non_blocking=non_blocking)
            onloaded_leaves.append(tensor)   
```

src/llmcompressor/pipelines/cache.py
```diff
@@ -250,122 +321,116 @@ def _fetch_and_record(batch_index):
...
+                    leaf,
                     (int, str, float, bool, torch.dtype, torch.device, type(None)),
                 ):
-                    warnings.warn(f"Offloading not implemented for type {type(value)}.")
-                return IntermediateValue(value=value, device=None)
+                    warnings.warn(
+                        f"Leaf type {type(leaf)} may not be handled correctly; "
+                        f"storing as-is"
+                    )
+                offloaded_leaves.append(leaf)
```
HDCharles, line 399: i don't think we need the warning anymore given the new handling method and we can clean this up a bunch. pinning is idempotent ...etc

```suggestion
        do_pin =  str(offload_device) == "cpu" and torch.accelerator.is_available()
        for leaf in leaves:
            if not isinstance(leaf, torch.Tensor):
                offloaded_leaves.append(leaf)
                continue
            
            with OverrideEqMode():
                # check for cache hit between shared tensors
                if leaf in cls.offload_values:
                   offloaded_tensor = cls.offload_values[leaf]
                else:
                   # move to offload if no hit
                   off = leaf.to(device=offload_device)
                   if offloaded is not leaf:
                       cls.offload_values[leaf] = off.pin_memory() if do_pin else off
                
            offloaded_leaves.append(
                IntermediateValue(
                    value=offloaded_tensor,
                    device=(onload_device if onload_device else leaf.device),
                )
            )
```

HDCharles (changes requested):
see comments but we're almost there!

src/llmcompressor/modifiers/gptq/base.py
menogrey: I try to run `mypy src/llmcompressor/modifiers/gptq/base.py`, if adding `| None`, there will be more error.
```
src/llmcompressor/modifiers/gptq/base.py:272: error: Item "None" of "IntermediatesCache[Module, list[Tensor]] | None" has no attribute "update"  [union-attr]
src/llmcompressor/modifiers/gptq/base.py:281: error: Value of type "IntermediatesCache[Module, list[Tensor]] | None" is not indexable  [index]
src/llmcompressor/modifiers/gptq/base.py:283: error: Item "None" of "IntermediatesCache[Module, list[Tensor]] | None" has no attribute "update"  [union-attr]
```

src/llmcompressor/modifiers/gptq/base.py
menogrey: Updated

src/llmcompressor/pipelines/cache.py
menogrey: Done.

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -262,43 +266,45 @@ def calibrate_module(
...
+                    torch.zeros(tuple(), device=device),
+                ],
             )
 
+        # Fetch hessian (onloads to original device automatically)
+        hessian, count = self._hessians_cache.fetch_no_onload(module)
+        new_hessian, new_count = calculate_hessian(inp, module)
+        hessian += new_hessian.to(hessian.device)
+        count += new_count
+        self._hessians_cache.update(module, [hessian, count], device)
```
HDCharles, line 291: this isn't going to work generally, if offload device is cpu it will be ok, if offload device is cuda
then count will get a device mismatch since the new one will be on cpu.

fix with something like the following (including fixing calculate_hessian to move things onto the right device (you probably want to do it before the matmul)

```suggestion
        new_hessian, new_count = calculate_hessian(inp, offload_device)
        
        hessian, count = self._hessians_cache.fetch_no_onload(module)
        hessian += new_hessian
        count += new_count
        self._hessians_cache.update(module, [hessian, count], device)
```

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -56,11 +53,11 @@ def accumulate_hessian(
             inp = inp.permute([1, 0, 2])
             inp = inp.flatten(1)
 
-    num_samples += num_added
+    num_samples = num_added
```
HDCharles: just name it correctly the first time

HDCharles:
see final fix

src/llmcompressor/modifiers/gptq/base.py
menogrey: Updated.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
menogrey: Updated.

HDCharles (approved):
looks good, just fix merge conflicts

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -28,41 +27,29 @@ def make_empty_hessian(
...
-            inp = unfold(inp)
-            inp = inp.permute([1, 0, 2])
-            inp = inp.flatten(1)
-
-    num_samples += num_added
+    if len(inp.shape) == 3:
+        inp = inp.reshape((-1, inp.shape[-1]))
+    inp = inp.t()
+
+    count = num_added
```
HDCharles, line 46:
```suggestion
    assert isinstance(module, torch.nn.Linear), "GPTQ only supports linear"
    inp = inp.reshape((-1, inp.shape[-1]))
    count = inp.shape[0]
    
    inp = inp.t()
```

oh this was wrong before i'm pretty sure. We look at the number of new datapoints before we reshape which is wrong.

HDCharles (changes requested):
see comment about gptq_quantize.py

this was wrong before and should be fixed.

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -149,13 +149,13 @@ class AWQModifier(Modifier):
...
-    )
-    # Dict[smooth layer name, [activation sums, activation counts]]
-    _smooth_activation_stats: dict[str, list[torch.Tensor]] = PrivateAttr(
-        default_factory=dict
+    # Cache of forward input args for each parent module, one dict for each batch
+    _parent_args_cache: IntermediatesCache[
+        tuple[Module, int], dict[str, torch.Tensor]
+    ] = PrivateAttr(default=None)
+    # Cache dict[smooth layer name, [activation sums, activation counts]]
+    _smooth_activation_stats: IntermediatesCache[str, list[torch.Tensor]] = PrivateAttr(
```
brian-dellabetta, line 158: I thought the overhead of offloading sums/counts was not worth the memory saved? These should be significantly smaller than the parent_args_cache

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -442,26 +448,33 @@ def cache_smooth_activations_hook(
...
 
         for mapping in self._resolved_mappings:
             # parent kwargs needed for future forward passes
             # same parent may appear multiple times in resolved mappings
-            if mapping.parent not in self._parent_args_cache:
-                self._parent_args_cache[mapping.parent] = IntermediatesCache(
-                    None,
-                    self.offload_device,
-                )
+            if len(self._parent_args_cache) == 0:
```
brian-dellabetta, line 478: This could lead to multiple hooks being registered, no?

src/llmcompressor/pipelines/cache.py
```diff
@@ -250,122 +326,100 @@ def _fetch_and_record(batch_index):
...
+                # check for cache hit between shared tensors
+                if leaf in cls.offload_values:
+                    offloaded_tensor = cls.offload_values[leaf]
+                else:
+                    # move to offload if no hit
+                    offloaded_tensor = leaf.to(device=offload_device)
+                    if offloaded_tensor is not leaf:
+                        cls.offload_values[leaf] = (
+                            offloaded_tensor.pin_memory()
+                            if do_pin
```
brian-dellabetta, line 412: do_pin is not including the `and not offloaded.is_pinned()` anymore, why is that?

src/llmcompressor/pipelines/cache.py
```diff
@@ -76,144 +62,234 @@ def from_dataloader(
...
+        :raises KeyError: if key is not found in cache
+        """
+        if key not in self._store:
+            raise KeyError(f"Key {key} not found in cache")
+        return self._onload_value(self._store[key])
 
-    def fetch(
-        self, batch_index: int, input_names: list[str] | None = None
-    ) -> dict[str, Any]:
+    def fetch_no_onload(self, key: TKey) -> TValue:
```
brian-dellabetta, line 93: why are we not onloading when fetching? Previously users didn't have to worry about this, I would prefer we not open up the API in a way users have to concern themselves about this

brian-dellabetta (commented):
Few comments on parts that have changed since my last review. This seems to have really expanded in scope, changing GPTQ, AWQ, and sparsegpt beyond just code that touches IntermediateCache instances. We should do extensive testing before merging this

src/llmcompressor/pipelines/cache.py
HDCharles: it makes no sense to since its idempotent

src/llmcompressor/pipelines/cache.py
HDCharles: we did have to worry about it, we just worried about it outside of intermediatescache, since intermediates cache couldn't do this

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -123,9 +123,9 @@ class GPTQModifier(Modifier, QuantizationMixin):
 
     # private variables
     _module_names: Dict[torch.nn.Module, str] = PrivateAttr(default_factory=dict)
-    _hessians: Dict[torch.nn.Module, torch.Tensor] = PrivateAttr(default_factory=dict)
-    _num_samples: Dict[torch.nn.Module, torch.Tensor] = PrivateAttr(
-        default_factory=dict
+    # cache for Module: [hessian, count]
+    _hessians_cache: IntermediatesCache[torch.nn.Module, list[torch.Tensor]] = (
```
brian-dellabetta, line 131:
```suggestion
    _hessians_cache: IntermediatesCache[torch.nn.Module, tuple[torch.Tensor, torch.Tensor]] = (
```

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -296,9 +299,9 @@ def compress_modules(self):
 
         # Assign modules to ranks
         module_list, rank_to_modules, module_to_rank = greedy_bin_packing(
-            list(self._hessians.keys()),
+            list(self._hessians_cache.iter_keys()),
             world_size,
-            item_weight_fn=lambda mod: self._hessians[mod].shape[0],
+            item_weight_fn=lambda mod: self._hessians_cache[mod].shape[0],
```
brian-dellabetta, line 310: This looks off, if hessians_cache[mod] returns a list (or tuple), it should be

```suggestion
            item_weight_fn=lambda mod: self._hessians_cache[mod][0].shape[0],
```

src/llmcompressor/pipelines/cache.py
brian-dellabetta: spoke about this, in agreement

src/llmcompressor/modifiers/transform/awq/base.py
menogrey: Seems like a issue caused by moving cache initialization，try to move it back.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
HDCharles: actually lets not make functional changes, i need to fix this incorrect counting thing and test it independently

suggestion should be 

```
    assert isinstance(module, torch.nn.Linear), "GPTQ only supports linear"
    # TODO fix count to be number of sequences rather than number of batches
    count = inp.shape[0] if inp.dim == 3 else 1
    inp = inp.reshape((-1, inp.shape[-1]))
    
    inp = inp.t()
```

src/llmcompressor/modifiers/gptq/base.py
menogrey: Updated.

PR #2668, opened 2026-04-29
feat: concurrent KLD evaluation without enforce_eager (closes #2667, refs #2646)

### LINKED ISSUE #2667: Fast KLD metric
followup to https://github.com/vllm-project/llm-compressor/issues/2646#issue-4323565620

looking for the fastest way to calculate an accurate kl divergence metric using vllm inference, i think this would involve 2 models being served on vllm concurrently to avoid disk writes and calculating the kld in an online manner.

jayakumarpujar:
## Summary

Implements the design proposed in #2667 - a fast, concurrent KLD metric that addresses the two core concerns raised on the closed PR #2659:

1. **No `enforce_eager=True`** - previous approach used Python forward hooks which are skipped during CUDA-graph replay, forcing eager execution. This PR replaces hooks with a **pre-allocated CUDA buffer** patched onto `LogitsProcessor.forward` before the first `generate()` call. Buffer writes (`Tensor.copy_`) are CUDA memory operations - they are recorded in CUDA graphs and replayed correctly.

2. **No disk I/O, both models concurrent** - both base and quantized models are loaded simultaneously (`gpu_memory_utilization=0.45` each, ~0.90 total). KLD is computed online after each prompt pair; no intermediate data is written to disk.

## Approach

```
LLM(base)  ──┐
              ├── apply_model(_worker_setup_buffer_capture)  [before first generate]
LLM(quant) ──┘

for prompt in prompts:
    base_llm.generate(prompt)  →  apply_model(collect_buffer)  →  h_base
    quant_llm.generate(prompt) →  apply_model(collect_buffer)  →  h_quant
    KLD computed immediately (online, in-memory)
```

The buffer capture patch:
- Pre-allocates `torch.zeros(max_tokens, hidden_dim, device='cuda')` on the model
- Replaces `lp.forward` instance method to `copy_` hidden states into the buffer before calling the original forward
- `copy_` is a CUDA kernel → captured in CUDA graph → replayed on every forward pass without Python overhead

## Validated results (Colab T4, `facebook/opt-125m` W4A16)

**Base vs W4A16 (8 prompts):**
```
KLDivergenceResult(mean_kld=0.216322, num_prompts=8, num_tokens=87, skipped=0)
```

**Sanity check (same model vs same model):**
```
KLD: 0.00000000  ✓
```

**WikiText-2 (64 samples):**
```
WikiText Mean KLD: 0.154980
Tokens evaluated:  6042
```

## Changes

| File | Change |
|------|--------|
| `src/llmcompressor/evaluation/kld.py` | Replace hook-based capture with CUDA buffer capture; concurrent dual-model loading; online KLD; default `gpu_memory_utilization=0.45` |
| `tests/unit/evaluation/test_kld.py` | 6 new tests for buffer setup/collect; default gpu_memory_utilization test; 33/33 passing |
| `examples/kld_evaluation/kld_metric.ipynb` | Updated to concurrent design; corrected `gpu_memory_utilization`; validated on Colab T4 |

## Usage

```python
from llmcompressor.evaluation import evaluate_kl_divergence

result = evaluate_kl_divergence(
    base_model_id="meta-llama/Meta-Llama-3-8B",
    quantized_model_id="./Meta-Llama-3-8B-W4A16",
    gpu_memory_utilization=0.45,  # per model; two models ~0.90 total
)
print(f"Mean KLD: {result.mean_kld:.6f}")
```


jayakumarpujar:
@HDCharles could you please have a look at this when you get a chance, and trigger the full CI suite? This addresses the two concerns from #2659 - no `enforce_eager`, both models run concurrently with no disk I/O. Validated on Colab T4 with OPT-125m W4A16, sanity check gives KLD ≈ 0. Would appreciate your feedback on the buffer capture approach.

HDCharles (changes requested):
looks good high level

1) lets get rid of this example notebook and use the format of the other examples
2) lets have the example be a more realistic usecase like e.g. wikitext
3) it seems like there's a lot of configuration that i'm not sure how it would work, e.g. changing temp, chat template, context length...etc we should probably set the API to be more aligned with the normal api
4) probably want something invocable from the command line like the existing 
```

        lm_eval \
            --model vllm \
            --model_args "pretrained=$save_dir,dtype=auto,max_model_len=$max_model_len,add_bos_token=True,tensor_parallel_size=$tp_size,gpu_memory_utilization=0.85" \
            --tasks "$task" \
            --num_fewshot "$num_fewshot" \
            --batch_size auto \
            $chat_args \
            --output_path "$eval_output_dir" 2>&1
```
maybe the api can accept a specific hf dataset?

jayakumarpujar:
@HDCharles, any update regarding this PR?

PR #2670, opened 2026-04-30
[GPTQ] Support modifier-level ActivationOrdering for all weight strategies

### LINKED ISSUE #2524: GPTQ ActivationOrdering support for per-channel quantization

elwhyjay:
Fixes #2524

## Summary
This PR closes the remaining modifier-level gap for `GPTQModifier(actorder=...)`. Before this PR, `resolve_quantization_config` only propagated modifier-level actorder to `strategy=GROUP` schemes, so the same setting was silently ignored for any other strategy. This PR extends the whitelist to **GROUP, TENSOR_GROUP, CHANNEL, TENSOR, and BLOCK**, with a `None` fallback (and warning) for combinations CT rejects on reload.

## Background
- vllm-project/compressed-tensors#682 relaxed the `QuantizationArgs` validator from `actorder is not None` to `actorder == GROUP` for non-group strategies, so yaml-level configs like `weights: {strategy: channel, actorder: weight}` already validate.
- #2616 generalized `quantize_weight`: `if actorder:` permutes for any strategy, `g_idx` is built for `{GROUP, TENSOR_GROUP, BLOCK}`, and is saved when `actorder == GROUP`.
- The remaining gap (this PR) is `GPTQModifier.resolve_quantization_config`: it still only resolved modifier-level actorder for `GROUP`, so the modifier-level path was inconsistent with the kernel-level path.

## `actorder=GROUP` fallback
`actorder=GROUP` is only accepted by compressed-tensors on GROUP / TENSOR_GROUP schemes (per CT #682). For CHANNEL / TENSOR / BLOCK, `resolve_quantization_config` warns and resets to `None` so the artifact stays loadable; `quantize_weight` mirrors the same guard for direct callers.

## Scope summary
| strategy       | modifier-level `actorder` (before) | (after this PR)                |
|----------------|------------------------------------|--------------------------------|
| GROUP          | propagated                         | propagated                     |
| TENSOR_GROUP   | silently ignored                   | propagated                     |
| CHANNEL        | silently ignored                   | propagated; `GROUP` → fallback |
| TENSOR         | silently ignored                   | propagated; `GROUP` → fallback |
| BLOCK          | silently ignored                   | propagated; `GROUP` → fallback |

### Automated
- `tests/llmcompressor/modifiers/gptq/test_base.py`
  - `test_config_resolution` parametrized over all five whitelisted strategies × actorder values.
  - `test_actorder_group_falls_back_to_none` parametrized over `{channel, tensor, block}`, asserts the warning is emitted.
- `tests/llmcompressor/modifiers/gptq/test_gptq_quantize.py`: `test_quantize_weight_channel_actorder_weight` (Linear sanity, asserts no `weight_g_idx`).
- `tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py`: `recipe_modifier_channel_actorder_weight` exercises the modifier-level CHANNEL path end-to-end.

```
pytest tests/llmcompressor/modifiers/gptq -v
# 57 passed (CPU)
```

## Notes
Rebased onto current `main` after #2616 landed; the `quantize_weight` actorder generalization that earlier revisions of this branch carried is now obsolete and was dropped in the rebase.


src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -111,23 +111,34 @@ def quantize_weight(
...
+    if is_grouped:
         # mapping from column index to group index
         g_idx = (
             torch.arange(num_columns, device=W.device, dtype=torch.int)
             // quant_args.group_size
         )
 
+    # column permutation is common to WEIGHT and GROUP actorder across all
+    # strategies; CHANNEL + GROUP is filtered out by GPTQModifier.resolve_quantization_config
+    if actorder in (ActivationOrdering.WEIGHT, ActivationOrdering.GROUP):
```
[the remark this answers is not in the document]
elwhyjay: This comment flags `STATIC` as missing, but the current code intentionally matches `main`'s behavior — `STATIC` isn't in the permutation branch on `main` either (see `gptq_quantize.py:114-130` on `main`, which only keys off `GROUP` / `WEIGHT`). Adding it here would silently change the quantization output for every existing `group + actorder=static` user, which is outside the scope of #2524.

The docstring vs. code mismatch for `STATIC` (the "no runtime cost" claim) looks like a pre-existing gap in `main` — I think it's better tracked as a separate issue.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -254,20 +265,18 @@ def quantize_weight(
...
-
-        elif actorder == ActivationOrdering.GROUP:
-            # restore original permutation
-            invperm = torch.argsort(perm)
-            W = W[:, invperm]
-            g_idx = g_idx[invperm]
-
-            # only save g_idx if mapping is not identity
-            has_gidx = True
+    if actorder == ActivationOrdering.WEIGHT:
```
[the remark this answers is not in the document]
elwhyjay: Same reasoning as the pre-process branch above — `main` doesn't invert a permutation for `STATIC` either (since `STATIC` never enters `_apply_activation_ordering`), so adding it here would be inconsistent with the pre-process side and would change behavior for existing users. Out of scope for this PR.

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -152,11 +152,28 @@ def resolve_actorder(existing):
...
         for scheme in config.config_groups.values():
             assert isinstance(scheme, QuantizationScheme)
-            if (
-                getattr_chain(scheme, "weights.strategy", None)
-                == QuantizationStrategy.GROUP
+            strategy = getattr_chain(scheme, "weights.strategy", None)
+            if strategy in (
+                QuantizationStrategy.GROUP,
+                QuantizationStrategy.CHANNEL,
             ):
```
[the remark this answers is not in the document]
elwhyjay: This comment correctly catches an inconsistency between the diff and the original PR description — I had written "unifies across GROUP / TENSOR_GROUP / CHANNEL", which oversold what the change actually does. The summary has been reworded to match the diff.

In practice this PR only adds `CHANNEL` to the `resolve_quantization_config` whitelist. `main` already only whitelists `GROUP`, so `GPTQModifier(actorder=...)` doesn't propagate to a `tensor_group` scheme today either (yaml-level config still works because `compressed-tensors`' validator allows it for `tensor_group`). The PR is scoped to the silent-ignore that #2524 is about, and `TENSOR_GROUP` modifier-level resolution has been added to the "Out of scope / follow-ups" list. Happy to do it as a quick follow-up PR right after this one merges.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -111,23 +111,34 @@ def quantize_weight(
...
             # use identity g_idx (invert permutation later)
-
-        elif actorder == ActivationOrdering.WEIGHT:
-            # permute weights and g_idx
-            W, H, perm = _apply_activation_ordering(W, H)
+        elif is_grouped:
+            # actorder == WEIGHT: permute g_idx so groups follow permuted columns
             g_idx = g_idx[perm]
+        # else: CHANNEL + WEIGHT — no g_idx, no scale recompute (per-channel
+        # scale is row-wise and invariant under column permutation)
```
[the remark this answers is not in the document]
elwhyjay: Applied in 2a1971c — added a guard at the top of quantize_weight that warns and falls back to actorder=None when a direct caller passes GROUP with a non-grouped strategy.

elwhyjay:
Pushed `2a1971c` addressing CodeRabbit's two suggestions and a small PR-description rewording prompted by Gemini's TENSOR_GROUP comment. Per-comment rationale is in the inline replies on each Gemini thread.

### Applied (from CodeRabbit)
- **Defensive guard in `quantize_weight`** — direct callers that bypass `GPTQModifier.resolve_quantization_config()` could pass `actorder=GROUP` with a non-grouped strategy and reach `g_idx[invperm]` while `g_idx` is `None`. Added a guard at the top of the function that warns and falls back to `actorder=None` for that combination.
- **Warning assertion in fallback test** — `test_channel_actorder_group_falls_back_to_none` now uses `unittest.mock.patch` to verify the warning is emitted in addition to checking `actorder is None`.

### Not applied (from Gemini), rationale inline
- `gptq_quantize.py:130` / `:268` — `STATIC` missing from permutation branches: `main` already excludes `STATIC` from these branches, and this PR preserves that behavior intentionally. Adding it would silently change quantization output for every existing `group + actorder=static` user. The docstring/code mismatch around `STATIC` looks like a pre-existing gap in `main`, better tracked as a separate issue.
- `base.py:159` — `TENSOR_GROUP` missing from the whitelist: `main` only whitelists `GROUP`, so `tensor_group` modifier-level actorder resolution is a pre-existing gap rather than something this PR introduces. Reworded the PR summary (the original "unifies across GROUP / TENSOR_GROUP / CHANNEL" was misleading) and added `TENSOR_GROUP` modifier-level resolution to the "Out of scope / follow-ups" list. Happy to ship that as a separate follow-up PR.

### Other
- Local pytest rerun: 49 passed on A100 (same as before, plus the new warning assertion).
- Small ruff line-length fix on the unified-permutation comment.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -111,23 +111,44 @@ def quantize_weight(
...
         if actorder == ActivationOrdering.GROUP:
-            W, H, perm = _apply_activation_ordering(W, H)
             # actually need scale/zp for permuted weight for this format
             scale, zero_point = observer(W)
             # use identity g_idx (invert permutation later)
-
-        elif actorder == ActivationOrdering.WEIGHT:
-            # permute weights and g_idx
-            W, H, perm = _apply_activation_ordering(W, H)
+        elif is_grouped:
```
HDCharles: This is not the correct logic

HDCharles: Act order group should have identity g_idx

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -111,23 +111,44 @@ def quantize_weight(
...
         # mapping from column index to group index
         g_idx = (
             torch.arange(num_columns, device=W.device, dtype=torch.int)
             // quant_args.group_size
         )
 
+    # GPTQModifier.resolve_quantization_config already filters CHANNEL+GROUP, but
+    # quantize_weight may be called directly; guard against it here to avoid
+    # indexing g_idx (which is None for non-grouped strategies).
+    if actorder == ActivationOrdering.GROUP and not is_grouped:
```
HDCharles: isn't this handled above?

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -111,23 +111,44 @@ def quantize_weight(
 
     scale, zero_point = observer(W)
     # handle g_idx and activation ordering
-    if strategy in (QuantizationStrategy.GROUP, QuantizationStrategy.TENSOR_GROUP):
```
HDCharles: the logic you've added is a bit mixed up, it should be noted that group actorder doesn't permute g_idx at start (since we WANT to use the scale/zp from the permuted W) while weight actorder does (since it needs the scale/zp from the unpermuted matrix).

i think in general, the logic i would go with is removing this if statement and letting everything have g_idx since its only used where its needed, then you can do:

```
s, zp = observer(W)
g_idx = torch.arange ....
perm = None
if actorder_weight:
    H,W,perm = apply_actorder
    g_idx = g_idx[perm]
if actorder_group:
    H,W,perm = apply_actorder

....main GPTQ code....

if actorder # either
    invperm = argsort(perm)
    W=W[invperm]
    g_idx = g_idx[invperm]

if actorder_group:
    qparams["g_idx"] = g_idx

```

this seems the easiest way to genearlize for per channel + actorder

src/llmcompressor/modifiers/gptq/gptq_quantize.py
elwhyjay: You're right, this is already handled in `resolve_quantization_config`. I will remove this guard.

HDCharles (changes requested):

src/llmcompressor/modifiers/gptq/gptq_quantize.py
elwhyjay: Thanks, updated to match your suggested flow. `g_idx` is now always initialized as identity, grouped strategies convert it to column-to-group indices, `WEIGHT` permutes it up front, and `GROUP` keeps it unpermuted during quantization while recomputing qparams on the permuted weight. Post-process restores both `W` and `g_idx`, and only saves `weight_g_idx` for grouped `actorder=GROUP`.

elwhyjay: Also verified locally: 49/49 pytest pass, and group + actorder=group output is bitwise-identical to `main`.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -253,23 +260,16 @@ def quantize_weight(
...
+    # alongside W so the saved mapping is back in original-column space
+    if perm is not None:
+        invperm = torch.argsort(perm)
+        W = W[:, invperm]
+        g_idx = g_idx[invperm]
 
-    if not has_gidx:
+    # only ship g_idx when it carries a non-identity mapping (actorder=GROUP on
+    # a grouped strategy); otherwise drop it so the consumer doesn't see one
+    if not (actorder == ActivationOrdering.GROUP and is_grouped):
```
HDCharles: just move this below, it makes no sense to separate this into 2 pieces, also don't need to worry about invalid configs, lets assume those are validated above.

```
if actorder==ActivationOrdering.GROUP:
   q_param_dict["weight_g_idx"] = g_idx
```

put this below where the save happens.

HDCharles (changes requested):
see comments, otherwise looks good

src/llmcompressor/modifiers/gptq/gptq_quantize.py
elwhyjay: Done. move the `weight_g_idx` save next to the rest of the qparams and dropped the separate `g_idx = None` step

elwhyjay:
@HDCharles thanks for the cleanup suggestion. it should be ready now — let me know if there's anything else.

HDCharles:
uh oh, you got beaten to the punch, the other GPTQ actorder refactor landed first, can you resolve merge conflicts and see if you can incorporate the previous feedback in still given the changes to the flow. 

I think high level it should all still work but may be a little more complicated

elwhyjay:
@HDCharles Quick scope check before pushing — looking at #2616, the round 2 (sketch flatten) and round 3 (save-position) feedback is essentially incorporated into the new `if actorder:` form already. The remaining gap is round 1's CHANNEL+GROUP guard.

So I think the rebased diff is small: `base.py` whitelist + fallback, channel-specific tests, and `quantize_weight` unchanged. Does that match your read, or want me to extend the scope?

Also re the "CT PR as well" comment on #2524 — I think the YAML-level piece already landed in vllm-project/compressed-tensors#682 (it relaxed the `QuantizationArgs` validator from `actorder is not None` to `actorder == GROUP` for non-group strategies, so yaml `channel + actorder=weight/static` already validates). Combined with #2616 (generalized `quantize_weight` actorder branches) and this PR (modifier-level resolve for CHANNEL + fallback for CHANNEL+GROUP), I think #2524 should be fully covered. The only optional remaining piece I see is the `TENSOR_GROUP` modifier-level whitelist (one line in `base.py` — happy to fold in or leave as a separate follow-up).

elwhyjay:
While digging into the rebase I noticed `TENSOR`, `TENSOR_GROUP`, and `BLOCK` strategies have the exact same modifier-level whitelist gap (yaml-level works after CT #682, but `GPTQModifier(actorder=...)` is silently ignored on those schemes too). Since #2616 already makes `quantize_weight` handle them all, it's a one-line-per-strategy fix in `base.py` plus one fallback for `+ actorder=GROUP`. Want me to fold those in here too, or keep #2670 channel-only and do them in a follow-up?

HDCharles:
I'd probably fold them in but you can keep separate if you find that easier

elwhyjay:
I'll fold them into this PR with a bounded scope

elwhyjay:
@HDCharles rebased on main and folded TENSOR / TENSOR_GROUP / BLOCK in per your suggestion. Description is also update. ready for a look

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -150,13 +150,40 @@ def resolve_actorder(existing):
...
+                QuantizationStrategy.TENSOR_GROUP,
+                QuantizationStrategy.CHANNEL,
+                QuantizationStrategy.TENSOR,
+                QuantizationStrategy.BLOCK,
             ):
+                # NOTE: setattr bypasses QuantizationArgs' model_validator,
+                # which only runs at construction time. After CT #682 the
+                # validator allows actorder on non-group strategies except
+                # for actorder=GROUP; we still go through setattr so that
+                # the modifier-level path is uniform across strategies.
```
HDCharles: What does this comment mean?

elwhyjay: Good call, that comment was muddy. I was trying to describe the modifier-level post-construction assignment, but the validator details made it more confusing than useful. I wil simpilfy it to say that we apply modifier-level actorder to already-constructed `QuantizationArgs`, and left the `actorder=GROUP` fallback below as the CT-compatibility guard.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -111,6 +111,22 @@ def quantize_weight(
 
     scale, zero_point = observer(W)
 
+    # Mirror GPTQModifier.resolve_quantization_config: actorder=GROUP requires
```
HDCharles: I feel like the code itself does a better job explaining than this comment tbh

tests/llmcompressor/modifiers/gptq/test_base.py
```diff
@@ -109,12 +110,36 @@ def test_actorder_resolution(
...
+
+
 @pytest.mark.parametrize(
     "strategies,actorder",
     [
         (["group"], None),
         (["group"], "static"),
         (["group"], "group"),
+        (["tensor_group"], None),
+        (["tensor_group"], "static"),
```
HDCharles: Static is weight, test should use one or the other

HDCharles (changes requested):
Minor comments will send to team this is pretty much ready

brian-dellabetta (approved):
Agree that some of the comments can be cleaned up as @HDCharles suggests. thanks for the contribution!

src/llmcompressor/modifiers/gptq/gptq_quantize.py
elwhyjay: got it. I will drop it

tests/llmcompressor/modifiers/gptq/test_base.py
elwhyjay: Will remain only  a`weight`

elwhyjay:
@HDCharles @brian-dellabetta thanks for the reviews — both cleanups pushed.

tests/llmcompressor/modifiers/gptq/test_base.py
```diff
@@ -109,12 +110,32 @@ def test_actorder_resolution(
...
+    return QuantizationArgs(strategy=strategy)
+
+
 @pytest.mark.parametrize(
     "strategies,actorder",
     [
         (["group"], None),
         (["group"], "static"),
         (["group"], "group"),
+        (["tensor_group"], None),
```
HDCharles, line 130: nit: keep this consistent, either change static to weight or weight to static across the entire test.

HDCharles (approved):

tests/llmcompressor/modifiers/gptq/test_base.py
HDCharles: Align rest of test?

tests/llmcompressor/modifiers/gptq/test_base.py
elwhyjay: Done — converted the remaining static cases to weight

PR #2672, opened 2026-04-30
[Transformers Tests] Don't hardcode pytest commands

dsikka:
- Rewrite to prevent hardcoded module pytest calls


brian-dellabetta (commented):
Just FYI there's a `llm-compressor/tests/llmcompressor/transformers/smoothquant` module that would get included with this test. Is that desired?

dsikka:
> Just FYI there's a `llm-compressor/tests/llmcompressor/transformers/smoothquant` module that would get included with this test. Is that desired?

Yeah that’s what this is trying to fix - skipping over modules with tests

brian-dellabetta (approved):

PR #2674, opened 2026-04-30
[bugfix] reduce memory requirements of `moe_calibration_context`

### LINKED ISSUE #2669: [Bug]: OOM during MoE module replacement when AWQ quantizing GLM-5.1

<details>

```text
Your output of `python collect_env.py` here
```

</details>

Hi, thanks for the great project.

I'm trying to quantize GLM-5.1 using AWQ with llm-compressor, but I'm consistently hitting an OOM issue during the MoE replacement (calibration) stage, even with extremely large RAM.

Environment
- Model: GLM-5.1 (MoE)
- Quantization: AWQ
- RAM: ~2TB allocated
- GPU: H100-80GB x 1
- llmcompressor==0.10.1 (built from main branch)
- torch==2.9.0
- transformers==5.5.4

### 🐛 Describe the bug

**Issue**
The process fails during the MoE module replacement step, specifically in the loop where modules are replaced:

```python
# Step 1: Collect all MoE modules that need replacement
modules_to_replace = []
    class_name = module.__class__.__name__
    if _is_registered(class_name, MoECalibrationModule):
        modules_to_replace.append((name, class_name))

# Step 2: Replace modules
for name, class_name in modules_to_replace:
    module = model.get_submodule(name)

    replacement = MoECalibrationModule.load_from_registry(
        class_name,
        original=module,
        config=model.config,
    )

    model.set_submodule(name, replacement)
```

The OOM happens during or right after load_from_registry() / set_submodule().

**Key Problem**
While replacing MoE modules layer by layer, I observe that:
- Memory usage keeps increasing after each replacement
- It does not stabilize or decrease, even after gc.collect()
- After replacing a certain number of layers, the process eventually runs out of memory and crashes

So this is not a one-time spike — it's a progressive memory growth during the replacement loop.

**Observations**
- Even with ~2TB RAM, the process eventually OOMs
- It appears that **previously replaced modules (or their tensors) are not fully released**
- I added weakref checks and GC logging, and some original module components still seem to remain alive
- This suggests possible:
   - lingering references
   - or duplication of large MoE expert weights during replacement

### 🛠️ Steps to reproduce

1. Load the GLM-5.1 MoE model with llm-compressor.
2. Configure AWQ quantization for the model.
3. Start the quantization / calibration process.
4. Monitor CPU RAM usage during replacement.
5. Observe that memory usage keeps increasing after each MoE layer replacement.
6. After several layers are replaced, the process is eventually killed by OOM, even with ~2TB RAM allocated.

brian-dellabetta:
SUMMARY:
Currently on main, `moe_calibration_context` retains a list of all modules to replace, even after the replacement module has been added to the module. The module was deleted, but because it was held as a reference elsewhere it wasn't being garbage collected, doubling the memory requirements inside the context.

Potentially resolves #2669 

TEST PLAN:
no net new logic


HDCharles (approved):
unsure how i didn't hit this when i was doing big model stuff because freeing this memory was something i also tried to do

brian-dellabetta:
> unsure how i didn't hit this when i was doing big model stuff because freeing this memory was something i also tried to do

@HDCharles good to know. There may be other places that also cause it not be garbage collected, but this should only help. i will merge and mark the issue as closed since i haven't heard back from the user

PR #2681, opened 2026-05-02
fix: add enable_thinking flag and reasoning dataset example for Qwen3-Next AWQ

### LINKED ISSUE #2680: AWQ calibration for Qwen3.5 thinking models: non-thinking leakage of <think> and failure to emit </think> after quantization
Hi,
I am quantizing Qwen/Qwen3.5-122B-A10B with AWQ using llmcompressor, and I am seeing a behavior regression that seems specific to thinking-mode calibration.

**Observed issue:**

After AWQ quantization:

1. In non-thinking inference with `enable_thinking=False`, the quantized model sometimes generates `<think>` tokens in the output.
2. In thinking inference with `enable_thinking=True`, the quantized model often produces very long reasoning traces, sometimes does not emit `</think>`, and may fail to transition to the final answer tokens.
3. The original base model does not show this behavior.
4. An open-source quantized variant I tested also does not show this behavior.

So the failure seems to be around preserving:
- suppression of thinking in non-thinking mode
- proper exit from thinking mode via `</think>`

**Model / setup:**
* Model: Qwen/Qwen3.5-122B-A10B
* Quantization: [AWQModifier(scheme="W4A16", targets=["Linear"])]
* Calibration dataset: Dataset name: HuggingFaceH4/ultrachat_200k

**Question:**
Do you have any guidance on the calibration datasets for Qwen3.5-122B model?

In particular, I would like to understand whether `ultrachat_200k` alone is insufficient for preserving the intended `<think>` / `</think>` behavior after quantization, which was not the case with Qwen3-4B or Qwen3-30B model.

Any suggestions would be very helpful. Thanks.

jayakumarpujar:
## Summary

Fixes #2680 - `<think>` leakage and missing `</think>` after AWQ quantization of Qwen3-Next thinking models.

**Root cause:** `apply_chat_template` in `qwen3_next_example.py` was called without `enable_thinking=False`. Qwen3-Next thinking tokenizers default to `enable_thinking=True`, so calibration samples were formatted with a thinking-mode prompt prefix - but `ultrachat_200k` has no `<think>...</think>` completions. AWQ fits weight scales to this mismatched activation distribution, corrupting the think-token generation pathway.

## Changes

- **`qwen3_next_example.py`** - pass `enable_thinking=False` to `apply_chat_template` for non-thinking deployment calibration
- **`qwen3_next_thinking_example.py`** - new example for thinking-mode deployment using `Magpie-Reasoning-V2` dataset with `loss_mask` to focus AWQ smoothing on assistant reasoning + answer tokens

## Ablation Results

Tested on `Qwen/Qwen3-14B` (H100 80GB), 128 calibration samples, seq_len 2048:

| Run | Dataset | `enable_thinking` | `loss_mask` | `<think>` leak (non-thinking) | `</think>` intact (thinking) |
|-----|---------|-------------------|-------------|-------------------------------|------------------------------|
| A | ultrachat_200k | not set (bug) | no | NO ✅ | YES ✅ |
| B | ultrachat_200k | `False` | no | NO ✅ | YES ✅ |
| C | Magpie-Reasoning-V2 | `True` | yes | NO ✅ | YES ✅ |

**Note:** The `<think>` leakage symptom from the issue was not reproduced at 14B scale - the bug appears scale-dependent and most pronounced at the reported 122B-A10B MoE model. The fix is architecturally correct regardless: calibration data distribution must match deployment mode. Run B confirms `enable_thinking=False` is safe and does not regress thinking-mode generation.

## Test Plan

- [ ] Maintainer validates Run A reproduces leakage on `Qwen/Qwen3.5-122B-A10B` internally
- [ ] Maintainer validates Run B (this fix) resolves leakage on the same model
- [ ] `qwen3_next_thinking_example.py` runs end-to-end on a thinking-mode deployment


jayakumarpujar:
Expanded scope to cover @Priya95715's exact model in [040ddb9](https://github.com/vllm-project/llm-compressor/pull/2681/commits/040ddb9):

The original issue is on 'Qwen/Qwen3.5-122B-A10B', which is exercised by 'examples/quantization_w4a4_fp4/qwen3_5_example.py' (and the sibling 'qwen3_6_example.py'). Both had the same 'apply_chat_template(...)' bug as 'qwen3_next_example.py' - calling it without 'enable_thinking=False' on a thinking-capable tokenizer while calibrating on 'ultrachat_200k'. Added the one-line fix to both.

The 'mxfp4'/'nvfp4' Qwen3.5 examples were checked but not modified - they use data-free quantization ('oneshot(model, recipe)' with no dataset), so the bug doesn't apply there.

examples/awq/qwen3_next_thinking_example.py
```diff
@@ -0,0 +1,118 @@
...
+
+# NOTE: Use this example when deploying Qwen3-Next in thinking mode.
+# For non-thinking deployment, use qwen3_next_example.py instead.
+# Using a reasoning dataset with loss_mask ensures AWQ weight scales
+# are fit to the thinking-mode activation distribution.
+
+MODEL_ID = "Qwen/Qwen3-Next-80B-A3B-Thinking"
+SAVE_DIR = MODEL_ID.split("/")[-1] + "-W4A16-awq-thinking"
+
+DATASET_ID = "Magpie-Align/Magpie-Reasoning-V2-250K-CoT-Deepseek-R1-Llama-70B"
```
brian-dellabetta, line 22: are there any ramifications to using a raesoning dataset made of responses from llama for a qwen model? I hadn't thought about this before.

brian-dellabetta (commented):
Thanks @jayakumarpujar for the nice PR. Things look good for the most part, but i am curious if you were able to run any basic ablations to see if the thinking_mode toggles or the loss mask were important for recovery on reasoning evals? Will also follow up on user issue this is intended to resolve

jayakumarpujar:
@brian-dellabetta ran token-level ablation (Qwen3-14B, H100), confirmed enable_thinking=False safe, but reasoning eval scores (gsm8k/MATH) not run needs 122B model.

examples/awq/qwen3_next_thinking_example.py
jayakumarpujar: Llama reasoning traces differ from Qwen's native thinking style in verbosity and formatting, which could slightly miscalibrate the AWQ scales toward Llama's activation patterns rather than Qwen's.

A cleaner alternative would be a Qwen-native reasoning dataset like Qwen/QwQ-LongCoT-130K-Filtered, which would better match the activation distribution of the model being quantized. I'll update the example to use that instead if you'd prefer?

brian-dellabetta: yeah i think that makes sense for Qwen. yet another thing we'd want to test lol

jayakumarpujar: Llama-generated traces differ from Qwen's native thinking style in verbosity and format, so ideally we'd use a Qwen-native dataset (QwQ-style). I checked HuggingFace and there's no Qwen-native reasoning dataset with a compatible schema available right now. Added a comment in the code flagging this so it's easy to swap when one becomes available.

brian-dellabetta: thanks for looking into it!

examples/awq/qwen3_next_thinking_example.py
```diff
@@ -0,0 +1,121 @@
...
+recipe = [
+    AWQModifier(
+        ignore=["lm_head", "re:.*mlp.gate$", "re:.*mlp.shared_expert_gate$"],
+        scheme="W4A16",
+        targets=["Linear"],
+    ),
+]
+
+
+def get_calib_dataset(tokenizer):
```
brian-dellabetta, line 38: can you clean this up to just use ds.map intead of looping through and creating a Dataset.from_list?

brian-dellabetta (commented):
Sorry for the delay on this, were you able to run the example and see if it performs better on reasoning evals? 

I am curious if the token mask is important here as well -- https://github.com/vllm-project/llm-compressor/blob/main/examples/awq/llama_example_with_masking.py

brian-dellabetta (commented):
I pushed some changes, but I am hitting OOM failures running the example that are resolved by

* #2813 

Once that is in I will retry.

jayakumarpujar:
@brian-dellabetta Ran the ablation you asked for. Compared 4 AWQ calibration strategies on **Qwen/Qwen3-8B** (hybrid thinking model with the `enable_thinking` toggle), evaluating both in thinking/chat mode.

**Setup:** W4A16 AWQ, 128 calibration samples. GSM8K run with `apply_chat_template=True` + `fewshot_as_multiturn=True`, 4-shot, `flexible-extract`, 250-sample subset, `max_gen_toks=2048` (long enough for `<think>` chains). Generation eval over 12 reasoning prompts, separating the real bug from this issue (opens `<think>`, never closes, stops on its own) from benign truncation.

| Variant | Dataset | `enable_thinking` | `use_loss_mask` | GSM8K | Proper `<think>` |
|---|---|---|---|---|---|
| Baseline (FP16) | — | — | — | 0.8760 | 12/12 |
| V1 | ultrachat_200k | False | False | 0.8440 | 12/12 |
| V2 | Magpie-Reasoning-V2 | True | False | 0.8680 | 12/12 |
| V3 | Magpie-Reasoning-V2 | True | True | 0.8600 | 12/12 |

(0 Broken, 0 Truncated, 0 Leaked across all variants.)

**Findings:**

1. **The pathology from this issue does not reproduce at 8B.** Every variant - including the ultrachat/no-think baseline — emits well-formed `<think>...</think>` with no leakage. The broken-`</think>` and non-thinking-leakage symptoms appear to be specific to the much larger MoE thinking models in the original report (Qwen3.5-122B, Qwen3-Next-80B). I don't have the hardware to reproduce those directly.

2. **Reasoning-dataset calibration recovers GSM8K better than ultrachat.** V2/V3 (reasoning + `enable_thinking=True`) land at 0.868/0.860 vs V1 (ultrachat + `enable_thinking=False`) at 0.844 - directionally consistent with the hypothesis.

3. **Honest caveat on significance + loss mask.** At n=250 the GSM8K stderr is approximately +/-2.2 pts, so the V1->V2 gap (~2.4 pts) is within noise - directional, not conclusive. And `use_loss_mask` shows **no measurable benefit** here (V2 no-mask >= V3 mask). The example keeps the mask since it's the right tool for the large target models, but I can't claim it helps at this scale from this data.

**Takeaway:** the new `qwen3_next_thinking_example.py` is a reasonable default for thinking-mode deployments (reasoning dataset + `enable_thinking=True`), but definitive validation of both the bug fix and the loss-mask benefit needs the actual large target models + a full GSM8K run. Happy to extend if someone with the hardware can pick it up.

brian-dellabetta (approved):
Thanks @jayakumarpujar for the example and for the updates, this should be good to merge in

kylesayrs (commented):
Needs `load_context`

examples/awq/qwen3_next_thinking_example.py
```diff
@@ -0,0 +1,124 @@
...
+    return ds.select(range(min(NUM_CALIBRATION_SAMPLES, len(ds))))
+
+
+def data_collator(batch):
+    assert len(batch) == 1
+    return {key: torch.tensor(val).unsqueeze(0) for key, val in batch[0].items()}
+
+
+if __name__ == "__main__":
+    with load_offloaded_model():
```
brian-dellabetta: please update to use `load_context` instead, as shown [here](https://github.com/vllm-project/llm-compressor/blob/main/examples/disk_offloading/qwen3_example.py#L14). this landed after you created the PR

brian-dellabetta:
> Needs `load_context`

@kylesayrs resolved, this should be good to go now

HDCharles:
if this is a bug with qwen3_next_example.py should that be corrected?

brian-dellabetta:
> if this is a bug with qwen3_next_example.py should that be corrected?

@HDCharles they're 2 different modalities. so if someone wants to use an awq-ed model in a thinking context, they should use this example. rather than changing it across the board, maintaining a separate example that we can direct users to. just like how we have an example specific for coding

HDCharles (approved):

PR #2686, opened 2026-05-05
[Tracing] Support tracing cache

kylesayrs:
## Purpose ##
* Support tracing `Cache` instances in fx graphs
  * This is needed for models like [DeepSeekV4](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v4/modeling_deepseek_v4.py#L1204-L1205) which use kv cache instances, regardless of whether the user sets the `use_cache` flag or not
* Prerequisite for #2647

## Background ##
```python3
class HFProxyableCacheMeta(type):
    """
    Metaclass that creates a class with its main methods wrapped to be proxyable.
    In the same way that objects can be replaced with `Proxy`s during trace-time,
    classes can also be replaced with `ProxyableClass`es during trace-time.
    This meta class acts as factory for creating `ProxyableClass`es of `Cache` classes.
    At trace-time, all references to `Cache` classes are monkeypatched with references
    to `ProxyableCache` classes. Whenever this class is used to construct a new instance
    of a `Cache`, an instance of `HFCacheProxy` is generated instead.
    """
```

## Changes ##
* Update how ProxyableClasses are monkeypatched at trace-time
  * Previously, `HFTracer` would look at the defining module of the model (usually `modeling_xxx.py` and monkeypatch the `Cache` imports there. However, this doesn't work for autowrapped forwards
  * Instead of monkeypatching the module dict, instead monkeypatch the `forward.__globals__` dict. This patch is more direct and relevant, and works for autowrapped forwards
* Rename `HFProxyableClassMeta` to `HFProxyableCacheMeta`, cause the class to emit instances of `HFCacheProxy` when constructed during tracing
  * `HFCacheProxy` is a proxy for an instance of a `Cache` and can be traced
* Fix minor typo in `HFProxyableClassMeta` implementation where fx graph nodes for `method` and `function` were swapped

## Testing ##
* Added `test_dynamic_cache_produces_hf_cache_proxy_node` which tests that cache construction and usage can be traced in the fx graph


src/llmcompressor/pipelines/sequential/transformers_helpers.py
```diff
@@ -1325,17 +1396,21 @@ def patch_for_tracing(self, root: torch.nn.Module | Callable[..., Any]):
...
+
+        for forward_fn in forward_fns:
             for orig_cls, patched_cls in self._CLASSES_TO_PATCH.items():
-                for attr_name, attr in mod.__dict__.items():
+                for attr_name, attr in forward_fn.__globals__.items():
                     if attr is orig_cls:
-                        patched.append((mod, attr_name, orig_cls))
-                        setattr(mod, attr_name, patched_cls)
+                        patched.append((forward_fn.__globals__, attr_name, orig_cls))
+                        forward_fn.__globals__[attr_name] = patched_cls
```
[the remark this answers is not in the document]
kylesayrs: This is still the same worst case asymptotic runtime. The runtime has been negligible for the very large models I'm using.

src/llmcompressor/pipelines/sequential/transformers_helpers.py
```diff
@@ -804,9 +805,17 @@ def check_proxy(a):
     return wrapper
 
 
-class HFProxyableClassMeta(type):
+class HFProxyableCacheMeta(type):
     """
```
HDCharles, line 809: Bad desc? Is this proxyable cache or class?

HDCharles: I'm confused about this abstraction this looks like it was a general class but now it's cache specific but all the descriptions seems to be a mix of class and cache, is this still being used generally?

kylesayrs: It used to be general, parameterized by the `proxy_factory_fn`. I've since changed it to be more specific.

brian-dellabetta (approved):

brian-dellabetta (approved):
We covered these changes in a team codereview

HDCharles (approved):

PR #2589, opened 2026-04-08
[Refactor] Refactor splits to only use the "calibration" split (#2551)

arpitkh101:
## Summary
Closes #2551
Simplifies the `splits` interface in [get_processed_dataset](cci:1://file:///data/arpit/projects/llm-compressor/src/llmcompressor/datasets/utils.py:28:0-94:32) by removing 
multi-split dict handling in favour of a plain string argument.

### Examples & Tests
- Updated all examples and tests to use `splits="train[:N]"` string format.
- Deleted `test_dataset_helpers.py` (helpers no longer exist).
- Added new unit tests to [test_dataset_loading.py](cci:7://file:///data/arpit/projects/llm-compressor/tests/llmcompressor/transformers/data/test_dataset_loading.py:0:0-0:0):
  - `{"calibration": ...}` dict backward compat
  - Deprecation warning is emitted for dict input
  - `splits=None` returns `None` (data-free flow)
  - Invalid type raises `ValueError`

## Before / After
```python
# Before (deprecated, still works with warning)
oneshot(model, dataset="ultrachat", splits={"calibration": "train_sft[:512]"})
# After (recommended)
oneshot(model, dataset="ultrachat", splits="train_sft[:512]")

PR #2691, opened 2026-05-07
[Tests][E2E][LM Eval] Configs Update and standardization

dsikka:
## Summary

This PR standardizes the lm-eval and e2e test infrastructure by introducing a shared
`BaseTestConfig` Pydantic model, flattening the e2e test folder structure, and cleaning
up both test configs and test utilities.

## `BaseTestConfig` — shared config class (`testing_utils.py`)

A single `BaseTestConfig(BaseModel)` now lives in `testing_utils.py` and is shared
across both test suites. It handles validated, typed YAML config loading with a
`scheme OR recipe` validator and sensible defaults for all common fields (`model_class`,
`num_calibration_samples`, `seed`, etc.).

- lm-eval `TestConfig` extends `BaseTestConfig` and only adds the `lmeval: LmEvalConfig` field
- e2e `TestvLLM` uses `BaseTestConfig` directly

This replaces the old pattern of manually parsing `eval_config.get(...)` with no
validation in the e2e tests.

## e2e folder restructure

Moved all contents of `tests/e2e/vLLM/` up to `tests/e2e/`:

```
tests/e2e/vLLM/test_vllm.py  →  tests/e2e/test_vllm.py
tests/e2e/vLLM/configs/      →  tests/e2e/configs/
tests/e2e/vLLM/recipes/      →  tests/e2e/recipes/
tests/e2e/vLLM/moe_configs/  →  tests/e2e/moe_configs/
```

## e2e config cleanup

All configs were moved and cleaned up during the restructure:

- `test_type` field removed from all configs
- `scheme` removed from recipe-based configs where it was only used as a label.
  Two FP8A16 configs retain `scheme` as it triggers the vLLM marlin kernel env var.
- `save_dir` removed where it can be derived from the config filename stem,
  consistent with how lm-eval already handles it.
- Recipe paths updated to reflect the new `tests/e2e/recipes/` location.
- `save_compressed=True` and `max_seq_length=2048` hardcoded in the test runner —
  neither needs to be configurable.

Two configs were renamed for clarity:

- `w4a16_grouped_quant_asym_awq` → `w4a16_asym_awq`
- `w4a16_grouped_quant_sym_awq` → `w4a16_sym_awq`

The following configs and their associated recipes were removed:

- `int8_tensor_weight_static_per_tensor_act`
- `kv_cache_phi3`
- `w4a16_actorder_none`

One new config was added:

- `nvfp4_moe_awq` — NVFP4 AWQ for MoE models (Qwen3-30B-A3B)

## lm-eval config changes

- Removed `w4a16_actorder_none` and `skipped_configs/awq_nvfp4a16`
- Added `vl_nvfp4_fp8` vision-language config
- Updated recipe paths for `awq_nvfp4`, `int8_w8a8_dynamic_per_token`,
  `w4a16_actorder_group`, `w4a16_actorder_weight`, `w4a16_awq_sym` to reflect the
  moved recipe location

## Removed timing instrumentation

`@log_time`, `get_singleton_manager`, `TIMINGS_DIR`, and `pandas` removed from both
`test_lmeval.py` and `test_vllm.py`.

## `testing_utils.py` cleanup

- Removed legacy `Cadence`, `TestType` enums and the old `TestConfig` dataclass
- Simplified `_validate_test_config` to check `cadence` directly instead of via enum membership
- Removed the module-level `TEST_DATA_FILE` variable (was only used inside `parse_params`)

# Test flows:
1. E2E: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/25692177817
2. LM Eval: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/25692935654


tests/e2e/test_vllm.py
```diff
@@ -105,16 +88,16 @@ def set_up(self, test_data_file: str):
...
-            scheme=self.scheme,
-            dataset_id=self.dataset_id,
-            dataset_config=self.dataset_config,
-            dataset_split=self.dataset_split,
-            recipe=self.recipe,
-            quant_type=self.quant_type,
+            model=self.config.model,
+            model_class=self.config.model_class,
+            num_calibration_samples=self.config.num_calibration_samples,
+            max_seq_length=2048,
```
HDCharles, line 94: Is this fine to hardcore?

tests/lmeval/vl_configs/vl_fp8_dynamic_per_token.yaml
```diff
@@ -11,7 +10,6 @@ lmeval:
   task: chartqa
   apply_chat_template: True
   num_fewshot: 0
-  batch_size: 100
```
HDCharles, line 14: Why get rid of batch size?

HDCharles: should we be using auto?

tests/lmeval/vl_configs/vl_nvfp4_fp8.yaml
```diff
@@ -0,0 +1,21 @@
+cadence: "weekly"
+model: Qwen/Qwen3-VL-8B-Instruct
+model_class: Qwen3VLForConditionalGeneration
+recipe: tests/e2e/recipes/non_uniform/recipe_nvfp4_fp8_mixed.yaml
+dataset_id: neuralmagic/calibration
+dataset_config: LLM
+dataset_split: "train[:20]"
```
HDCharles, line 7: Should we be changing these per https://github.com/vllm-project/llm-compressor/pull/2589

HDCharles (approved):
Seems fine, see comments

tests/lmeval/vl_configs/vl_fp8_dynamic_per_token.yaml
dsikka: The default batch size is already 100 in TestConfig args so this is redundant

tests/e2e/test_vllm.py
dsikka: It was already hardcoded / not set by any config

tests/lmeval/vl_configs/vl_nvfp4_fp8.yaml
dsikka: I think this is correct based on the PR?

PR #2703, opened 2026-05-12
Implement REAP pruning modifier for MoE expert pruning

eldarkurtic:
SUMMARY:
This PR implements REAP pruning from https://arxiv.org/abs/2510.13999 , following as much as possible the original codebase from https://github.com/CerebrasResearch/reap 

IMPLEMENTATION DETAILS:
REAP is implemented as `REAPPruningModifier`. This pruner is based on calibration data and therefore doesn't work with datafree pipeline. In short, it runs calibration, collects per-expert saliency scores via hooks, and then prunes experts. It also modifies corresponding field in `config.json` to indicate that the model has reduced number of experts.

The saliency score is: `S_j = mean(g_j * ||f_j||_2)`
where `g_j` is the router probability for expert `j`, and `f_j` is that expert’s output activation norm for routed tokens.

Example recipe:
```yaml
pruning_stage:
  pruning_modifiers:
    REAPPruningModifier:
      sparsity: 0.25
```

Modifier Lifecycle:
On initialize, it detects the MoE architecture. It has a class-name registry for known MoE blocks and a fallback heuristic looking for router/expert attributes. It then finds MoE layers and creates one saliency tracker per layer.
On calibration start, it registers three hook types:
  - MoE pre-hook: runs the router, computes softmax probabilities, and caches top-k expert indices.
  - Expert forward hook: just records `||expert_output||_2` for saliency score
  - MoE post-hook: combines router probabilities with expert output norms and updates `REAPSaliencyTracker`.

REAP deliberately prunes after the MoE calibration wrappers have been restored. That means saliency is collected on calibration modules, but pruning is applied to the real model modules.

TEST PLAN:
1) There are already some sanity check tests in `tests/llmcompressor/modifiers/pruning/reap/test_base.py`
2) I ran the example script `examples/reap_expert_pruning/reap_qwen3_30b_25pct.py` for 25% and 50% sparsity on Qwen3-30B-A3B-Instruct-2507 model with OpenPlatypus calibration data and obtained the following GSM8k scores with vLLM engine:
```bash
1) unpruned baseline
vllm ({'pretrained': '/home/eldarkurtic/hf_models/Qwen/Qwen3-30B-A3B-Instruct-2507', 'add_bos_token': True, 'tensor_parallel_size': 2, 'max_model_len': 8192}), gen_kwargs: ({}), limit: None, num_fewshot: 5, batch_size: auto
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8825|±  |0.0089|
|     |       |strict-match    |     5|exact_match|↑  |0.8734|±  |0.0092|

2) REAP sparsity = 0.25
vllm ({'pretrained': '/home/eldarkurtic/github/eldarkurtic/llm-compressor/output_dir/Qwen3-30B-A3B-Instruct-2507_REAP_sp25', 'add_bos_token': True, 'tensor_parallel_size': 2, 'max_model_len': 8192}), gen_kwargs: ({}), limit: None, num_fewshot: 5, batch_size: auto
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8870|±  |0.0087|
|     |       |strict-match    |     5|exact_match|↑  |0.8704|±  |0.0093|

3) REAP sparsity = 0.5
vllm ({'pretrained': '/home/eldarkurtic/github/eldarkurtic/llm-compressor/output_dir/Qwen3-30B-A3B-Instruct-2507_REAP_sp50', 'add_bos_token': True, 'tensor_parallel_size': 2, 'max_model_len': 8192}), gen_kwargs: ({}), limit: None, num_fewshot: 5, batch_size: auto
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8635|±  |0.0095|
|     |       |strict-match    |     5|exact_match|↑  |0.8582|±  |0.0096|
```
3) I've also checked the saved ckpts, and they are indeed pruned. Disk usage: `57GB --> 25%: 44GB --> 50%: 30GB`. Each model has updated `config.json` with the correct number of experts.

Note: before landing this modifier, it would be useful to run more testing. Specifically with other MoE architectures, possibly including the ones with fused experts (e.g. Llama), more eval benchmarks (e.g. calibrate on coding data, eval on coding and non-coding). Let's sync offline on the plan for these tests.

### DIFF
--- a/src/llmcompressor/modifiers/pruning/reap/base.py
+++ b/src/llmcompressor/modifiers/pruning/reap/base.py
@@ -1,0 +1,256 @@
+"""
+REAP (Router-weighted Expert Activation Pruning) modifier for MoE models.
+
+See: https://arxiv.org/abs/2510.13999
+"""
+
+from functools import partial
+from typing import Any
+
+import torch
+import torch.nn.functional as F
+from loguru import logger
+from pydantic import Field, PrivateAttr, model_validator
+
+from llmcompressor.core import Event, EventType, State
+from llmcompressor.modifiers import Modifier
+from llmcompressor.modifiers.pruning.reap.utils import (
+    MoEModelAttrs,
+    REAPSaliencyTracker,
+    detect_moe_attrs,
+    find_moe_layers,
+    get_num_experts,
+    prune_moe_layer,
+    update_model_config,
+)
+
+__all__ = ["REAPPruningModifier"]
+
+
+class REAPPruningModifier(Modifier):
+    """
+    Prunes experts from MoE layers using the REAP saliency metric:
+    S_j = mean(g_j * ||f_j||_2), averaged over tokens routed to expert j.
+
+    :param compression_ratio: fraction of experts to remove per layer (0, 1).
+    :param ignore: module name patterns to skip during MoE layer detection.
+
+    Example recipe::
+
+        REAPPruningModifier:
+          compression_ratio: 0.5
+    """
+
+    compression_ratio: float = 0.5
+    ignore: list[str] = Field(default_factory=list)
+
+    _attrs: MoEModelAttrs | None = PrivateAttr(default=None)
+    _moe_layer_names: list[str] = PrivateAttr(default_factory=list)
+    _saliency_trackers: dict[str, REAPSaliencyTracker] = PrivateAttr(
+        default_factory=dict
+    )
+    _top_k: int = PrivateAttr(default=2)
+    _n_experts_to_drop: int = PrivateAttr(default=0)
+    _norm_buffers: dict[str, dict[int, torch.Tensor]] = PrivateAttr(
+        default_factory=dict
+    )
+    _routing_cache: dict = PrivateAttr(default_factory=dict)
+
+    @model_validator(mode="after")
+    def _validate_compression_ratio(self) -> "REAPPruningModifier":
+        if not 0.0 < self.compression_ratio < 1.0:
+            raise ValueError(
+                f"compression_ratio must be in (0, 1), got {self.compression_ratio}"
+            )
+        return self
+
+    def on_initialize(self, state: State, **kwargs) -> bool:
+        model = state.model
+
+        self._attrs = detect_moe_attrs(model)
+        if self._attrs is None:
+            raise ValueError(
+                "Could not detect MoE architecture. Ensure the model is an "
+                "MoE model or specify 'targets' explicitly."
+            )
+
+        moe_layers = find_moe_layers(model, self._attrs, self.ignore)
+        if not moe_layers:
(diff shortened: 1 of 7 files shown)


HDCharles:
One thing I'd love to have a better idea about REAP is how it translates to runtime improvements. Is it still hitting tuned kernels at 25% sparsity or do we need to stick to powers of two or something like that for the tradeoff to he worth it?

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,248 @@
...
+        if event.type_ == EventType.CALIBRATION_EPOCH_END:
+            if not self.ended_:
+                self.on_end(state, None)
+
+    def on_end(self, state: State, event: Event, **kwargs):
+        self.ended_ = True
+        self.remove_hooks()
+
+    def on_finalize(self, state: State, **kwargs) -> bool:
+        # Prune in on_finalize (not on_end) so we act on actual modules,
```
HDCharles, line 150: I think we normally don't restore the actual modules in the moe calibration context so it seems like this approach won't work for a lot of models.

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,248 @@
...
+    update_model_config,
+)
+
+__all__ = ["REAPPruningModifier"]
+
+
+class REAPPruningModifier(Modifier):
+    """
+    Prunes experts from MoE layers using the REAP saliency metric:
+    S_j = mean(g_j * ||f_j||_2), averaged over tokens routed to expert j.
```
HDCharles, line 33: What are g and f?

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,248 @@
...
+                partial(self._moe_post_hook, layer_name),
+                "forward",
+            )
+
+    def on_event(self, state: State, event: Event, **kwargs):
+        if event.type_ == EventType.CALIBRATION_EPOCH_START:
+            if not self.started_:
+                self.on_start(state, None)
+
+        if event.type_ == EventType.CALIBRATION_EPOCH_END:
```
HDCharles, line 141: Not being able to prune in sequential epoch end is going to make us have to onload everything to do pruning and carry around the activation info the whole time

HDCharles: We should at least finalize the pruning decisions at sequential epoch end even if we want to do final pruning later, though we won't see the difference in propagate

Ryfernandes:
@kylesayrs Just rebased against transformers v5 on main, getting started on my planned changes

PR #2708, opened 2026-05-13
[Docs] update/add docs related to MX formats

dsikka:
SUMMARY:
- Add  and update documentation related to MXFp4 and MXFp8


dsikka:
@coderabbitai you should only apply labels when the content has significant relevance to a label. For example, we dont really mention wna16 here at all.

brian-dellabetta (approved):

PR #2710, opened 2026-05-13
[LM Eval] Update lm eval tests to spin-up vLLM

dsikka:
SUMMARY:
- Add run_lmeval.py which runs lm-eval using vLLM, in a separate env, similar to our e2e tests. As a result, we no longer need to keep lm-eval as a dependency 
- Significantly clean-up and simplify the vLLM test 
- Clean-up configs 

Test Flow: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/26009646892

## Summary by CodeRabbit

* **Tests**
  * Refactored evaluation pipeline to use subprocess-based execution instead of in-process evaluation.
  * Updated vision-language model evaluation configurations to support vLLM backend.
  * Adjusted evaluation metric thresholds across quantization configurations.
  * Changed default GPU memory utilization setting to 0.70.

* **Chores**
  * Removed version constraint on evaluation dependency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vllm-project/llm-compressor/pull/2710?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tests/e2e/test_vllm.py
```diff
@@ -189,8 +189,7 @@ def _run_vllm(self, logger):
         else:
             os.environ.pop("VLLM_TEST_FORCE_FP8_MARLIN", None)
 
-        if self.config.gpu_memory_utilization is not None:
-            llm_kwargs["gpu_memory_utilization"] = self.config.gpu_memory_utilization
```
kylesayrs, line 193: The config default is still `None`, right?
https://github.com/vllm-project/llm-compressor/blob/main/tests/testing_utils.py#L178-L181

tests/lmeval/configs/w4a4_nvfp4.yaml
```diff
@@ -6,11 +6,7 @@ dataset_split: train_sft
...
 lmeval:
-  # NVFP4 (4-bit weights + 4-bit activations) has lower recovery than FP8/INT8
-  # Observed: strict-match ~92.81%, flexible-extract ~89.59%
-  recovery_threshold:
-    exact_match,strict-match: 0.92
-    exact_match,flexible-extract: 0.89
   metrics:
-    exact_match,flexible-extract: 0.70
-    exact_match,strict-match: 0.65
+    # this is higher as tests are run on h100, not b200
```
kylesayrs: This comment won't make much sense outside of the context of this PR

tests/lmeval/run_lmeval.py
```diff
@@ -0,0 +1,60 @@
...
+    model, config = parse_args()
+    lmeval_config = config["lmeval"]
+    seed = config["seed"]
+
+    model_args = (
+        f"pretrained={model},"
+        f"dtype={lmeval_config['dtype']},"
+        f"add_bos_token={lmeval_config['add_bos_token']},"
+        f"trust_remote_code={lmeval_config['trust_remote_code']},"
+        f"max_model_len=2048,"
```
kylesayrs: future make this configurable.

I'm almost certain that this is too small for chartqa for some models/tokenizers

tests/lmeval/vl_configs/vl_int8_w8a8_dynamic_per_token.yaml
```diff
@@ -6,18 +6,13 @@ dataset_id: neuralmagic/calibration
 dataset_config: LLM
 dataset_split: "train[:512]"
 lmeval:
-  model: "hf-multimodal"
-  model_args:
-    dtype: bfloat16
-    convert_img_format: True
+  model: vllm-vlm
+  dtype: bfloat16
```
kylesayrs: Delete this arg from configs, already defaults to bfloat16

tests/lmeval/test_lmeval.py
```diff
@@ -123,22 +115,34 @@ def set_up(self, test_data_file: str):
...
-        if self.config.lmeval.metrics:
-            logger.info("Absolute metrics provided - will show warnings if outside ±5%")
 
     def test_lm_eval(self, test_data_file: str):
         # Run vLLM with saved model
         self.set_up(test_data_file)
 
         # Always evaluate base model for recovery testing
         logger.info("================= Evaluating BASE model ======================")
+        # TODO: need to free memory between vLLM / OneShot Calls
```
kylesayrs: Remove todo

tests/lmeval/test_lmeval.py
```diff
@@ -123,22 +115,34 @@ def set_up(self, test_data_file: str):
...
         base_results = self._eval_base_model()
 
+        gc.collect()
+        torch.cuda.empty_cache()
+        torch.cuda.synchronize()
```
kylesayrs, line 149: I love python /s

tests/lmeval/test_lmeval.py
```diff
@@ -316,44 +281,19 @@ def _check_absolute_warnings(self, results):
...
-
-            actual_val = metrics.get(metric_key)
-            if actual_val is None:
-                logger.warning(
-                    f"Metric {metric_key} in config not found in results, "
-                    f"skipping warning check"
-                )
-                continue
-
-            higher_is_better = (
```
kylesayrs, line 333: Is this no longer available from lm_eval?

kylesayrs (commented):
My only callouts are
1. `max_model_len` is likely too low (which may explain the lower VLM scores)
2. Why was `higher_is_better` removed?

tests/lmeval/configs/w4a4_nvfp4.yaml
dsikka: I'll make it more clear

tests/lmeval/run_lmeval.py
dsikka: I can make it configurable and then rerun

tests/lmeval/test_lmeval.py
dsikka: I removed it because none of our tests use tasks where this field was set. I think we can add it back if we decide to add a perplexity test but the way it was currently written wasn't great / hard to follow. Our standard evals used by MLR use the default / have higher_is_better as True

tests/e2e/test_vllm.py
dsikka: The test config we're now using sets a default to 0.7 through this PR

kylesayrs: I missed that, thanks

tests/lmeval/test_lmeval.py
kylesayrs: Please add an assert so we don't footgun in the future

dsikka: Assert for what?

dsikka: Done

kylesayrs (approved):

PR #2713, opened 2026-05-15
enhance AutoRoundModifier performance by skipping useless model forward in calibration stage.

xin3he:
SUMMARY:
This pull request improves the calibration and tuning workflow in the `AutoRoundModifier` class by introducing a no-op forward method for decoder layers during calibration, ensuring efficient input capture without unnecessary computation. It also adds a new configuration option and ensures robust restoration of original module state after calibration.

Key changes:

**Calibration and Forward Pass Handling:**
* Introduced a `_noop_forward` function that replaces the decoder layer's `forward` method during calibration, allowing input capture hooks to run without executing the full forward computation. The original `forward` is restored before tuning and again at the end to ensure correct module behavior. 

**Configuration and Argument Passing:**
* Added a `disable_opt_rtn` boolean option to `AutoRoundModifier` and ensured it is passed to the `AutoRound` initialization arguments. [[1]](diffhunk://#diff-7d09d56952a42ef7887f966d67c46ef24b542ea6b7689e16db5cd47fc9dffba2R167) [[2]](diffhunk://#diff-7d09d56952a42ef7887f966d67c46ef24b542ea6b7689e16db5cd47fc9dffba2R302)

**Robustness and Efficiency:**
* Updated `suspend_offloading` to skip non-weight layers (like SiLU), preventing errors and unnecessary processing.

**Code Cleanliness:**
* Removed a redundant call to `_update_device_map_for_dp`, streamlining the `apply_autoround` logic.

TEST PLAN:
"Previous unit tests could cover it"


brian-dellabetta (commented):
Hi @xin3he , thanks for preparing this. Rather than handle this at the modifier level, check out [this chunk of code](https://github.com/vllm-project/llm-compressor/blob/59d8809570bfc89bf614d1b925dff7d005484cbd/src/llmcompressor/pipelines/sequential/pipeline.py#L129-L172) in SequentialPipeline. If you don't want to propagate quantization error to the next layer, you can set `propagate_error=False` to only do the calibration stage. Does this resolve your use case? If we know we don't want to do this for AutoRoundModifier, we could add some logic during pre-processing stage to set propagate_error instead of always defaulting to True

xin3he:
Thanks, @brian-dellabetta 
It's updated with `dataset_args.propagate_error = False`

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -86,6 +86,13 @@ def __call__(
         offload_device = torch.device(dataset_args.sequential_offload_device)
         set_onload_device(model, onload_device)
 
+        # AutoRoundModifier optimizes each layer independently using its own
+        # forward passes, so quantization error should not be propagated between
+        # layers during the calibration stage
+        modifiers = session.lifecycle.recipe.modifiers
+        if any(type(m).__name__ == "AutoRoundModifier" for m in modifiers):
+            dataset_args.propagate_error = False
```
HDCharles, line 94: isn't this going to break e.g. AWQ -> autoround flows?

brian-dellabetta: If AWQ+AutoRound is run in a single sequential pipeline, yes AWQ would always have propagate_error=False. This won't break per se, error just won't be propagated. The default "independent" pipeline though will split them up into separate pipelines so this will only run during AutoRound.

brian-dellabetta (approved):
Thanks, this looks good to me. Can you resolve merge conflicts and then i'll kick off tests?

HDCharles (approved):

xin3he:
> Thanks, this looks good to me. Can you resolve merge conflicts and then i'll kick off tests?

Resolved, thank you.

xin3he:
@brian-dellabetta Can you help merge this PR? Thank you.

brian-dellabetta (approved):

Issue #2698, opened 2026-05-11
Remove iMatrixGatherer


HDCharles:
In https://github.com/vllm-project/llm-compressor/pull/2473 we went back and forth about the implementation and finalized a temporary design. The plan was to implement it then with an iMatrixGatherer modifier and later remove it when required llmc functionality was added. That day has come.

The iMatrix technique is fundamentally about how you quantize values, making it an observer, but to support it we needed to add an iMatrixGatherer modifier because the normal quantization modifier did weight quantization at the start of the lifecycle, before calibration which would break the iMatrixObserver. This is no longer the case as of https://github.com/vllm-project/llm-compressor/pull/2585 which consolidated the lifecycle to always do quantization after calibration. This allows us to do the following:

1) remove iMatrixGatherer
2) store imatrix statistics always on the observer, never on the modifier
3) we need to make sure to do calibration if doing weight only quantization with the imatrix observer.

dshane1903:
I’d like to take this one.

My plan:
1. Remove `IMatrixGatherer` exports, registry entries, docs/examples, and tests that depend on the temporary modifier.
2. Keep `imatrix_mse` support in `IMatrixMSEObserver`, with importance statistics stored on the observer/module path rather than on a separate modifier.
3. Update the quantization/GPTQ flow so weight-only `imatrix_mse` still performs activation calibration now that quantization happens after calibration.
4. Add/update CPU tests proving recipes using `weights.observer: imatrix_mse` work without `IMatrixGatherer`, including an end-to-end comparison that iMatrix affects scales versus the fallback path.

Please let me know if you want backward compatibility for old recipes that still include `IMatrixGatherer`, or if removing it outright is preferred.

HDCharles:
Outright removal is preferred, go ahead I'll assign you

vasuag09:
Hey @HDCharles, Is this issue still open to work upon?

dshane1903:
Hey @vasuag09, I’m currently working on this in vllm-project/llm-compressor#2716. That PR implements the outright `IMatrixGatherer` removal requested above. It’s currently waiting on the companion compressed-tensors PR vllm-project/compressed-tensors#710, which moves the calibration-data check into `QuantizationConfig.requires_calibration_data()` so the llm-compressor PR can be simplified cleanly. Happy to coordinate if you were looking at a related piece.

PR #2716, opened 2026-05-18
Remove IMatrixGatherer

dshane1903:
SUMMARY:
- remove the temporary IMatrixGatherer modifier and its dedicated tests
- keep imatrix_mse importance statistics on the observer instead of modules
- route QuantizationModifier recipes using imatrix_mse through calibration automatically
- update iMatrix docs and examples to configure imatrix_mse directly as the weight observer

TEST PLAN:
- `.venv/bin/python -m pytest tests/llmcompressor/observers/test_imatrix.py tests/llmcompressor/pipelines/test_registry.py`
- `.venv/bin/ruff check src/llmcompressor/observers/imatrix.py src/llmcompressor/modifiers/quantization/quantization/mixin.py src/llmcompressor/pipelines/registry.py tests/llmcompressor/observers/test_imatrix.py tests/llmcompressor/pipelines/test_registry.py tests/llmcompressor/modifiers/transform/imatrix/test_e2e_integration.py examples/imatrix/llama3_imatrix_example.py`


src/llmcompressor/pipelines/registry.py
```diff
@@ -63,12 +63,10 @@ def _modifier_requires_calibration(modifier: Modifier):
                 "GPTQModifier",
                 "AWQModifier",
                 "AutoRoundModifier",
-                "IMatrixGatherer",
             ):
                 return True
             elif isinstance(modifier, QuantizationModifier):
-                config = modifier.resolve_quantization_config()
-                return config.requires_calibration_data()
```
HDCharles, line 72: What's going on here? Looks like you removed a bunch of existing functionality

HDCharles:
@Yatimai also should review

src/llmcompressor/pipelines/registry.py
dshane1903: Good catch, the previous diff made this look like it removed the existing config-based calibration check...I pushed a follow-up commit that keeps `config.requires_calibration_data()` explicit in `registry.py`

dshane1903: In the original version of this PR, I moved the `config.requires_calibration_data()` call into a helper on `QuantizationModifier`, then added the `imatrix_mse` weight-observer case there

HDCharles: It makes more sense to move this change here:

https://github.com/vllm-project/compressed-tensors/blob/c763417afecf5fef123cbc1dbca16dc449ab8e9a/src/compressed_tensors/quantization/quant_config.py#L257

instead of duplicating this calibration check, can you update that?

tests/llmcompressor/modifiers/transform/imatrix/test_e2e_integration.py
```diff
@@ -74,13 +72,12 @@ def test_pipeline_produces_quantized_model(self):
...
-        # Verify importance was cleaned up after finalization
         targeted_names = _get_linear_layer_names(model, ignore=["lm_head"])
         for name in targeted_names:
             mod = _get_module_by_name(model, name)
-            assert not hasattr(
-                mod, "_imatrix_importance"
-            ), f"{name} should not have _imatrix_importance after finalization"
+            assert hasattr(mod, "weight_scale"), f"{name} should be quantized"
+            assert not hasattr(mod, "_imatrix_sum")
+            assert not hasattr(mod, "_imatrix_count")
```
Yatimai: These assertions are tautological under this PR : `_imatrix_sum` / `_imatrix_count` only live on the observer instance, never on the module. They pass regardless of whether cleanup works. The meaningful check would be on `module.weight_observer` (or that its accumulators don't outlive calibration). 
Same in `test_pipeline_with_regex_targets`.

Yatimai (commented):
This is a clean execution of the plan from #2698. `IMatrixGatherer` was always meant as a temporary scaffold in #2473 :  the plan was to remove it once the observer lifecycle could run quantization after calibration, which #2585 made possible.

One small observation below on the tests. Non-blocking.

src/llmcompressor/pipelines/registry.py
dshane1903: I opened a companion compressed-tensors PR to move this into `QuantizationConfig.requires_calibration_data()` where the existing calibration inference lives: https://github.com/vllm-project/compressed-tensors/pull/710/changes

Once that lands, I can simplify this PR back to only calling `config.requires_calibration_data()`

tests/llmcompressor/modifiers/transform/imatrix/test_e2e_integration.py
dshane1903: Cleaned up integration tests

src/llmcompressor/observers/imatrix.py
```diff
@@ -66,25 +69,15 @@ def __init__(self, *args, **kwargs):
...
         if not hasattr(module, "in_features"):
             return
+        if self._imatrix_hook is not None:
+            self._imatrix_hook.remove()
 
         in_features = module.in_features
-        module._imatrix_sum = torch.zeros(in_features, dtype=IMATRIX_PRECISION)
-        module._imatrix_count = torch.tensor(0, dtype=torch.int64)
+        self._imatrix_sum = torch.zeros(in_features, dtype=IMATRIX_PRECISION)
+        self._imatrix_count = torch.tensor(0, dtype=torch.int64)
```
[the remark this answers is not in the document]
dshane1903: Initialized ` _imatrix_sum` and `_imatrix_count` on the observed module’s parameter device.

src/llmcompressor/observers/imatrix.py
```diff
@@ -66,25 +69,15 @@ def __init__(self, *args, **kwargs):
...
-            self._imatrix_count = module._imatrix_count
-            del module._imatrix_sum
-            del module._imatrix_count
-            return
-
+        """Attach a forward-pre hook to accumulate E[x²] per input channel."""
         if not hasattr(module, "in_features"):
             return
+        if self._imatrix_hook is not None:
+            self._imatrix_hook.remove()
```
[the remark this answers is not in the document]
dshane1903: Removed existing `_imatrix_hook` before returning early for unsupported modules.

kylesayrs:
Implemented by https://github.com/vllm-project/llm-compressor/pull/2920/changes

PR #2722, opened 2026-05-18
DDP docs updates

HDCharles:
SUMMARY:
Updating DDP functionality in docs to reflect actual state in LLMC


docs/guides/observers.md
```diff
@@ -142,6 +142,35 @@ Observers can be configured with optional keyword arguments via `QuantizationArg
 | `norm`        | `3.0`   | Exponent used when computing the importance-weighted error. |
 | `strict`      | `False` | If `True`, raise an error instead of falling back to uniform MSE when importance data is unavailable. |
 
+## DDP (Distributed) Support
+
+Observers support distributed data-parallel calibration. When activation quantization runs across multiple ranks (each rank processing a disjoint partition of the calibration dataset), observer statistics must be synchronized before quantization parameters are computed so all ranks produce identical results.
+
+### How it works
+
+Each observer class declares a `_act_sync_dict` class attribute mapping statistic attribute names to their `torch.distributed.ReduceOp`. The base class `sync_activation_stats()` method iterates this dict and issues async `dist.all_reduce` calls:
```
dsikka, line 207: `class attribute mapping statistic attribute names` - attribute twice?

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,51 @@ ds = load_dataset(
...
+
+After `oneshot` returns, call `dispatch_model` to move the model to a single GPU for inference or validation:
+
+```python
+from compressed_tensors.offload import dispatch_model
+
+dispatch_model(model)
+```
+
+To save the model, `save_pretrained` handles distributed internals automatically — only rank 0 writes to disk. You should still call it from all ranks (or guard on rank 0 for the save path), and call `destroy_process_group()` at the end:
```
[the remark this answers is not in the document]
HDCharles: not a docs concern

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,51 @@ ds = load_dataset(
...
+
+```python
+import torch.distributed as dist
+
+model.save_pretrained(SAVE_DIR, save_compressed=True)
+tokenizer.save_pretrained(SAVE_DIR)
```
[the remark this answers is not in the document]
HDCharles: no

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -68,4 +110,55 @@ Now, your script is ready to run using distributed processes. To start, simply r
 | Meta-Llama-3-8B-Instruct | 4 | 264.07 | 5.82 | 52.50 | 0.7180 | 96.74 |
 | Qwen3-30B-A3B | 1 | 14207.53 | 6.56 | 748.23 | 0.8704 | 209.93 |
 | Qwen3-30B-A3B | 2 | 7018.25 | 6.36 | 696.65 | 0.8810 | 205.89 |
-| Qwen3-30B-A3B | 4 | 3694.46 | 6.36 | 723.05 | 0.8832 | 217.62 |
\ No newline at end of file
+| Qwen3-30B-A3B | 4 | 3694.46 | 6.36 | 723.05 | 0.8832 | 217.62 |
```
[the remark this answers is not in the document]
HDCharles: no

docs/guides/observers.md
```diff
@@ -142,6 +142,39 @@ Observers can be configured with optional keyword arguments via `QuantizationArg
...
+Only **activation** statistics are synchronized. Weight statistics are never synced because weights are identical across all ranks (broadcast during model load).
+
+### Per-observer sync behavior
+
+| Observer | Synced statistics | Reduce op | Notes |
+|----------|-------------------|-----------|-------|
+| `static_minmax` | `min_vals`, `max_vals` | MIN, MAX | Global min/max across all ranks |
+| `minmax` (EMA) | `min_vals`, `max_vals` | MIN, MAX | Global min/max across all ranks |
+| `memoryless_minmax` | *(none)* | — | Stateless; each rank's data is independent |
+| `mse` (EMA) | `min_vals`, `max_vals` | AVG | Averages the per-rank MSE-optimal ranges |
```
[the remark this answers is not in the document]
HDCharles: no

docs/guides/observers.md
```diff
@@ -142,6 +142,39 @@ Observers can be configured with optional keyword arguments via `QuantizationArg
...
+| `memoryless_mse` | *(none)* | — | Stateless; each rank's data is independent |
+| `imatrix_mse` | `_imatrix_sum`, `_imatrix_count` | SUM | Accumulates importance scores across ranks before normalization |
```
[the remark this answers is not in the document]
HDCharles: ok

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,51 @@ ds = load_dataset(
...
+To save the model, `save_pretrained` handles distributed internals automatically — only rank 0 writes to disk. You should still call it from all ranks (or guard on rank 0 for the save path), and call `destroy_process_group()` at the end:
+
+```python
+import torch.distributed as dist
+
+model.save_pretrained(SAVE_DIR, save_compressed=True)
+tokenizer.save_pretrained(SAVE_DIR)
+
+dist.destroy_process_group()
+```
```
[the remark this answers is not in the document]
HDCharles: not a docs concern

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -1,5 +1,5 @@
...
+LLM Compressor supports distributed oneshot to greatly speed up model calibration and compression. Each rank processes a disjoint partition of the calibration dataset (data-parallel calibration); modifiers then synchronize statistics across ranks so all ranks produce identical quantization parameters. For the GPTQ modifier, compression itself is also distributed: modules are assigned to ranks by a greedy bin-packing algorithm and the compressed weights are broadcast back. For more information on the design, see [[RFC] [Performance Refactor][Distributed] Sequential Onloading with Data-Parallel Calibration and Weight-Parallel Optimization](https://github.com/vllm-project/llm-compressor/issues/2180) as well as [[GPTQ][ddp] enabling DDP for GPTQ](https://github.com/vllm-project/llm-compressor/pull/2333).
```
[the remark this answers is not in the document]
HDCharles: its not broen

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,51 @@ ds = load_dataset(
...
+Alternatively, if you pass a `Dataset` object directly to `oneshot`, the sampler automatically detects when all ranks have the same dataset (via fingerprint comparison) and partitions it across ranks for you — no manual partitioning needed in that case.
+
 ### 4. Call your script with `torchrun` ###
 
-Now, your script is ready to run using distributed processes. To start, simply run your script using `torchrun --nproc_per_node=2 YOUR_EXAMPLE.py` to run with two GPU devices. For a complete example script, see [llama_ddp_example.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/llama3_ddp_example.py). The below table shows results and speedups as of LLM Compressor v0.10.0, future changes will bring these numbers closer to linear speedups.
+Now, your script is ready to run using distributed processes. To start, simply run your script using `torchrun --nproc_per_node=2 YOUR_EXAMPLE.py` to run with two GPU devices.
+
+### 5. Post-quantization: dispatch and save ###
+
+After `oneshot` returns, call `dispatch_model` to move the model to a single GPU for inference or validation:
```
dsikka: I think this is misleading. We offload if the model doesn't fit on a single GPU, right? Whereas this sounds like it will just fail.

dsikka (dismissed):

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -68,4 +110,55 @@ Now, your script is ready to run using distributed processes. To start, simply r
 | Meta-Llama-3-8B-Instruct | 4 | 264.07 | 5.82 | 52.50 | 0.7180 | 96.74 |
 | Qwen3-30B-A3B | 1 | 14207.53 | 6.56 | 748.23 | 0.8704 | 209.93 |
 | Qwen3-30B-A3B | 2 | 7018.25 | 6.36 | 696.65 | 0.8810 | 205.89 |
-| Qwen3-30B-A3B | 4 | 3694.46 | 6.36 | 723.05 | 0.8832 | 217.62 |
\ No newline at end of file
+| Qwen3-30B-A3B | 4 | 3694.46 | 6.36 | 723.05 | 0.8832 | 217.62 |
+
+### QuantizationModifier (activation quantization / W8A8) ###
```
dsikka: Remove W8A8 and activation quantization - this sounds like these are the only two things supported by this which is untrue

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -1,5 +1,5 @@
 ## Distributed Oneshot ##
```
dsikka, line 1: I think it may also be useful to mention how the sequential pipeline interacts with DDP

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,51 @@ ds = load_dataset(
...
+
+### GPTQModifier ###
+
+GPTQ uses **weight-parallel compression**: after each rank accumulates Hessians from its data partition, the modifier:
+
+1. Uses a greedy bin-packing algorithm to assign modules to ranks (balanced by Hessian size).
+2. Each rank reduces (`dist.reduce`) the Hessians for modules assigned to it, accumulating contributions from all ranks.
+3. Each rank compresses only its assigned modules independently.
+4. Compressed parameters (`weight`, `weight_scale`, `weight_zero_point`) are broadcast back to all ranks via `dist.broadcast`.
+
```
dsikka, line 107: I found this a little bit confusing to differentiate between the calibration vs compression steps - could we make it more explicit, maybe something like this?

GPTQ uses weight-parallel compression: each rank runs the full model forward pass on its data partition, accumulating partial Hessians for every module. After calibration, every rank holds
  partial Hessians for all modules, but each partial Hessian only reflects that rank's data slice. The modifier then:

  1. Uses a greedy bin-packing algorithm to assign each module to exactly one "owner" rank (balanced by Hessian size).
  2. For each module, all ranks send their partial Hessian to the owner rank via dist.reduce (SUM), giving the owner the full dataset's Hessian.
  3. Each owner rank compresses only its assigned modules independently (avoiding redundant computation across ranks).
  4. Compressed parameters (weight, weight_scale, weight_zero_point) are broadcast back to all ranks via dist.broadcast.

docs/guides/big_models_and_distributed/distributed_oneshot.md
kylesayrs: Agree. We should also be careful to make it clear that sample generation is not required for saving.

Start by introducing sample generation. For large models, consider skipping this step or using just a few tokens, as onloading offloaded weights can take a long time.

```python3
print("\n\n")
sample = tokenizer("Hello my name is", return_tensors="pt")
sample = {key: value.to(model.device) for key, value in sample.items()}
output = model.generate(**sample, max_new_tokens=100)
```

If the model is small enough, you can speed up sample generation by dispatching to GPU before running generation.
```python3
dispatch_model(model)
```

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -68,4 +110,55 @@ Now, your script is ready to run using distributed processes. To start, simply r
...
+See [smoothquant_ddp_example.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w8a8_int8/smoothquant_ddp_example.py) for a complete W8A8 + SmoothQuant example.
+
+### AWQModifier ###
+
+AWQ uses **all-reduce of activation statistics** across ranks before computing smoothing scales and clipping parameters. Similar to SmoothQuant, each rank observes its data partition and the statistics are globally reduced. The scale computation is then replicated identically on each rank.
+
+See [llama_example_ddp.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/awq/llama_example_ddp.py) for a LLaMA W4A16 example, and [qwen3_moe_example_ddp.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/awq/qwen3_moe_example_ddp.py) for a Qwen3-MoE example.
+
+## Observer DDP Support ##
```
kylesayrs, line 164: Can you provide a little more information about what the observer lifecycle looks like now? What exactly are the code steps to calibrating and updating weights? What exactly are the code steps to calibrating and updating activations?

kylesayrs (commented):
Strongly consider adding diagrams from the RFC to the DDP algorithms section. Not only are they good visuals, but they really assist the text explanations.

docs/guides/big_models_and_distributed/distributed_oneshot.md
HDCharles: it's entirely DDP-unaware, i can add a section about this but it feels like its more confusing than helpful. Like it we explicate every place where DDP functionality exists/doesn't exist we'll list the whole repo and if we just limit it to places where DDP functionality does exist, should we include things like the model save code, data pre-processing..etc that are kind of there to avoid foot guns?

see `DDP Support - General Framework`

for my attempt to do this

docs/guides/big_models_and_distributed/distributed_oneshot.md
HDCharles: shouldn't that be left to the observers.md?

the only DDP relevent part is the sync

docs/guides/observers.md
HDCharles: fixed

docs/guides/big_models_and_distributed/distributed_oneshot.md
HDCharles: I rewrote this, let me know your thoughts.

docs/guides/big_models_and_distributed/distributed_oneshot.md
HDCharles: fixed and added more details

docs/guides/big_models_and_distributed/distributed_oneshot.md
HDCharles: I rewrote based on your feedback

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,61 @@ ds = load_dataset(
...
+
+dist.destroy_process_group()
+```
+
+## Quantization Modifier DDP Support ##
+
+The following quantization modifiers are DDP-aware. Each section describes its synchronization strategy.
+
+### QuantizationModifier ###
+Since weight information is the same across ranks we only have to syncrhonize activation information and activation information is only needed in 2 cases for the QuantizationModifier. Case 1: when doing non-dynamic activation quantization like FP8 (static activation quantization) and NVFP4 (local activation quantization) quantization schemes. Case 2: when using a weight observer that makes use of activation information like the iMatrixObserver.
```
brian-dellabetta:
```suggestion
Since weight information is the same across ranks we only have to synchronize activation information and activation information is only needed in 2 cases for the QuantizationModifier. Case 1: when doing non-dynamic activation quantization like FP8 (static activation quantization) and NVFP4 (local activation quantization) quantization schemes. Case 2: when using a weight observer that makes use of activation information like the iMatrixObserver.
```

docs/guides/big_models_and_distributed/distributed_oneshot.md
kylesayrs: I think it's very clear now

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,61 @@ ds = load_dataset(
...
+GPTQ, being a more expensive technique, can make further use of parallelization by doing weight-parallel compression. Once rank has accumulated a partial Hessian for every module during the forward pass on its data partition, the modifier then:
+
+1. Uses a greedy bin-packing algorithm to assign each module to exactly one "owner" rank (balanced by Hessian size).
+2. For each module, all ranks send their partial Hessian to the owner rank via dist.reduce (SUM), giving the owner the full dataset's Hessian.
+3. Each owner rank compresses its assigned modules independently (avoiding redundant computation across ranks).
+4. Compressed parameters (`weight`, `weight_scale`, `weight_zero_point`) are broadcast back to all ranks via `dist.broadcast`.
+
+![Diagram showing DDP GPTQ](DDP_GPTQ.png)
+
+This means compression itself is parallelized across ranks, not just calibration.
```
kylesayrs, line 110: In the future, I'd like to start referring to this step as "optimization", and the actual packing as compression

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,61 @@ ds = load_dataset(
...
+Since weight information is the same across ranks we only have to syncrhonize activation information and activation information is only needed in 2 cases for the QuantizationModifier. Case 1: when doing non-dynamic activation quantization like FP8 (static activation quantization) and NVFP4 (local activation quantization) quantization schemes. Case 2: when using a weight observer that makes use of activation information like the iMatrixObserver.
+
+In both cases, activation information is **all-reduced across ranks at sequential layer boundaries** — so all ranks share identical quantization parameters before moving to the next layer. The reduce operation varies by observer type (see [Observer DDP support](#observer-ddp-support) below). Thus while the calibration is parallelized, the actual compression happens identically across all ranks after any activation statistics have been synchrnized.
+
+### GPTQModifier ###
+
+GPTQ, being a more expensive technique, can make further use of parallelization by doing weight-parallel compression. Once rank has accumulated a partial Hessian for every module during the forward pass on its data partition, the modifier then:
```
brian-dellabetta: not sure what this should be? a rank? each rank?
```suggestion
GPTQ, being a more expensive technique, can make further use of parallelization by doing weight-parallel compression. Once each rank has accumulated a partial Hessian for every module during the forward pass on its data partition, the modifier then:
```

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -68,4 +120,74 @@ Now, your script is ready to run using distributed processes. To start, simply r
...
+| `memoryless_mse` | *(none — data-free)* | — |
+| `imatrix_mse` | `_imatrix_sum`, `_imatrix_count` | SUM |
+
+Weight statistics are never synced: weights are identical across ranks (broadcast at load time), so only activation-derived statistics need synchronization.
+
+The `QuantizationMixin.sync_obs_act_stats()` method (called at each sequential layer boundary) iterates all observers on the modules in the current subgraph and calls `observer.sync_activation_stats()`, collecting async communication handles and waiting for completion before proceeding to the next layer.
+
+## DDP Support - General Framework
+
+In general we've design our DDP implementation to affect as few abstractions as possible to reduce complication and ease maintence. While modifiers have to be generally DDP-aware and observers provide an interface that modifiers can use to synchronize activation statistics, the rest of llm-compressor is largely DDP agnostic. Things like the sequential pipeline are entirely blind to whether or not DDP is being used. While a few areas like saving and data preparation include some DDP-awareness to avoid foot-guns, these instances are rare and the system will work without them.
```
kylesayrs:
```suggestion
In general, the DDP implementation is designed to affect as few abstractions as possible to reduce complication and ease maintenance. While modifiers have to be generally DDP-aware and observers provide an interface that modifiers can use to synchronize activation statistics, the rest of llm-compressor is largely DDP agnostic. Things like the sequential pipeline are entirely blind to whether or not DDP is being used. While a few areas like saving and data preparation include some DDP-awareness to avoid foot-guns, these instances are rare and the system will work without them.
```

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,61 @@ ds = load_dataset(
...
+
+In both cases, activation information is **all-reduced across ranks at sequential layer boundaries** — so all ranks share identical quantization parameters before moving to the next layer. The reduce operation varies by observer type (see [Observer DDP support](#observer-ddp-support) below). Thus while the calibration is parallelized, the actual compression happens identically across all ranks after any activation statistics have been synchrnized.
+
+### GPTQModifier ###
+
+GPTQ, being a more expensive technique, can make further use of parallelization by doing weight-parallel compression. Once rank has accumulated a partial Hessian for every module during the forward pass on its data partition, the modifier then:
+
+1. Uses a greedy bin-packing algorithm to assign each module to exactly one "owner" rank (balanced by Hessian size).
```
brian-dellabetta: is this the same as "source" rank? if so, would help with consistency
```suggestion
1. Uses a greedy bin-packing algorithm to assign each module to exactly one "source" rank (balanced by Hessian size).
```
Would need to update below as well if we do this

kylesayrs (approved):

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -57,9 +57,61 @@ ds = load_dataset(
...
+2. For each module, all ranks send their partial Hessian to the owner rank via dist.reduce (SUM), giving the owner the full dataset's Hessian.
+3. Each owner rank compresses its assigned modules independently (avoiding redundant computation across ranks).
+4. Compressed parameters (`weight`, `weight_scale`, `weight_zero_point`) are broadcast back to all ranks via `dist.broadcast`.
+
+![Diagram showing DDP GPTQ](DDP_GPTQ.png)
+
+This means compression itself is parallelized across ranks, not just calibration.
+
+See [llama3_ddp_example.py](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/llama3_ddp_example.py) for a complete W4A16 example.
```
brian-dellabetta: to point to the living example, regardless of branch/commit user is viewing on
```suggestion
See [llama3_ddp_example.py](../../../examples/quantization_w4a16/llama3_ddp_example.py) for a complete W4A16 example.
```

docs/guides/big_models_and_distributed/distributed_oneshot.md
```diff
@@ -68,4 +120,74 @@ Now, your script is ready to run using distributed processes. To start, simply r
...
\ No newline at end of file
+| Qwen3-30B-A3B | 4 | 3694.46 | 6.36 | 723.05 | 0.8832 | 217.62 |
+
+### AutoRoundModifier ###
+
+AutoRound defines a 3rd archetype of modifier as far as DDP support. Because the compression step repeatedly uses the actual activations which are sharded across ranks, Autoround uses a kind of data-parallel synchronized compression in contrast to GPTQ's weight-parallel or QuantizationModifier's identical-across-ranks compression. During calibration the inputs to each module are recorded, from there:
+
+1. For each module, for some number of iterations, quantize the weight and then do a forward and a (modified) backward pass (using the recorded module inputs) to obtain gradients (SignSGD)
+2. synchronize gradients across ranks using `dist.all_reduce` (SUM)
+3. Each rank indepdently updates the weights and then repeats
```
brian-dellabetta:
```suggestion
3. Each rank independently updates the weights and then repeats
```

brian-dellabetta (approved):
LGTM, just typo/link nits

docs/guides/big_models_and_distributed/distributed_oneshot.md
HDCharles: yeah do you want to do that in this PR though?

kylesayrs: Later

docs/guides/big_models_and_distributed/distributed_oneshot.md
kylesayrs: Yep that's fine. It should dedicated observer docs

kylesayrs (approved):

PR #2732, opened 2026-05-20
fix bug fused obs + sequential targets

HDCharles:
Summary:

observers need stats of their fused observers to calculate global_scale properly.

we run all observers in a chunk at the same time so normally this isn't an issue.

if sequential targets is set to linear or something else small, the fused observers will not be in the current chunk and so when get_qparams is called, it will error. see new unit test which fails on main

fix: observers have a record of other observers in their fusion list, now it includes those observers' parent modules so they can run them when needed. Mostly this won't matter but in the situation where the sequential targets is really small, it will solve the problem.

Test Plan:
see unit test

### DIFF
--- a/src/llmcompressor/observers/helpers.py
+++ b/src/llmcompressor/observers/helpers.py
@@ -185,21 +185,32 @@ def fuse_weight_observers(model: Module):
     from llmcompressor.observers import Observer
 
     for submodule in model.modules():
-        for layers_to_fuse in FUSED_LAYER_NAMES:
-            if not all(hasattr(submodule, name) for name in layers_to_fuse):
+        for fusion_name_group in FUSED_LAYER_NAMES:
+            if not all(hasattr(submodule, name) for name in fusion_name_group):
                 continue
-
-            layers = [getattr(submodule, name) for name in layers_to_fuse]
-            observers = []
-            for layer in layers:
+            layers_to_fuse = [getattr(submodule, name) for name in fusion_name_group]
+            
+            only_obs = True
+            only_tensor_group = True
+            observers_and_modules = []
+            for layer in layers_to_fuse:
                 obs = getattr(layer, "weight_observer", None)
                 if obs is None:
-                    break
+                    only_obs = False
+                    continue
                 if obs.args.strategy != QuantizationStrategy.TENSOR_GROUP:
-                    break
-                observers.append(obs)
-            else:
-                Observer.fuse(observers)
+                    only_tensor_group = False
+                    continue
+                observers_and_modules.append((obs, layer))
+            if len(observers_and_modules) == 0:
+                continue
+            
+            start = f"Some layers in fused group: {fusion_name_group} have"
+            end = ", need all fused layers to have same quantization"
+            assert only_obs, f"{start} no weight observer{end}"
+            assert only_tensor_group, f"{start} non-TENSOR_GROUP quantization{end}"
+
+            Observer.fuse(observers_and_modules)
 
 
 def lerp(start: torch.Tensor, end: torch.Tensor, weight: float) -> torch.Tensor:
--- a/src/llmcompressor/observers/helpers.py
+++ b/src/llmcompressor/observers/helpers.py
@@ -189,7 +189,7 @@ def fuse_weight_observers(model: Module):
             if not all(hasattr(submodule, name) for name in fusion_name_group):
                 continue
             layers_to_fuse = [getattr(submodule, name) for name in fusion_name_group]
-            
+
             only_obs = True
             only_tensor_group = True
             observers_and_modules = []
@@ -204,7 +204,7 @@ def fuse_weight_observers(model: Module):
                 observers_and_modules.append((obs, layer))
             if len(observers_and_modules) == 0:
                 continue
-            
+
             start = f"Some layers in fused group: {fusion_name_group} have"
             end = ", need all fused layers to have same quantization"
             assert only_obs, f"{start} no weight observer{end}"
--- a/tests/llmcompressor/observers/test_fused_observers.py
+++ b/tests/llmcompressor/observers/test_fused_observers.py
@@ -1,0 +1,73 @@
+"""
+Test that fused observers work correctly with sequential calibration.
+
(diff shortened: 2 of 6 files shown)


tests/llmcompressor/observers/test_fused_observers.py
```diff
@@ -0,0 +1,70 @@
+"""
+Test that fused observers work correctly with sequential calibration.
+
+This tests the fix for fused observer handling where sequential calibration
+with TENSOR_GROUP quantization would fail because only some fused observers
+would be observed (e.g., q_proj observed but k_proj and v_proj skipped).
+"""
+
+from llmcompressor import oneshot
```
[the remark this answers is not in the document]
HDCharles: already fixed

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -30,6 +31,7 @@
     "calibrate_query_hook",
     "calibrate_key_hook",
     "calibrate_value_hook",
+    "get_modules_with_fused_observers",
```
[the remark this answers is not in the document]
HDCharles: late

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -83,20 +85,70 @@ def initialize_observer(
...
+        if getattr(mod, obs_name, None) in fused_no_stats:
+            all_modules.add(mod)
+
+    return all_modules
+
+
 def observe(
     module: Module | Iterable[Module],
     base_name: str,
+    also_do_fused: bool = True,
```
brian-dellabetta: we call observe in a handful of places. do we want to also_do_fused for those? wondering if defaulting to False for original behavior is preferable

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -83,20 +85,70 @@ def initialize_observer(
...
+    fused_no_stats = set()
+
+    # Collect all fused observers without statistics from the initial modules
+    for mod in modules:
+        obs = getattr(mod, obs_name, None)
+        if obs is not None:
+            cur = {o for o in obs.get_fused_observers() if not o.has_statistics}
+            fused_no_stats.update(cur)
+
+    model = active_session().state.model
```
brian-dellabetta: this is needed because the fused observers data model is unidirectional?

HDCharles: its needed because we need the modules, not the observers. and observers don't have a reference to modules anymore

do you think i should make the fused observer thing hold references to modules instead?

then i could get rid of most of this.

i could even hide it in [get_qparams](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/observers/base.py#L87-L96)
```

global_absmax = torch.max(-self.min_vals.min(), self.max_vals.max())
for mod in self.fused_obs_modules
   obs = getattr(mod, self.base_name)
   if not obs.has_statistics
      observe(mod, self.base_name)
   global_absmax = torch.max(global_absmax, -obs.min_vals.min())
   ....etc
```

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: its only an issue in a minority of cases (model free stuff) so i just went through and edited it there. By default this is what we should be doing.

HDCharles: actually its not an issue anywhere because if there's no model it just doesn't get any fused observers, so it works fine in all cases but i removed it from the model_free just to be clear about how it was supposed to function.

brian-dellabetta: does the set of fused observers only get set for quant schemes that require it? model free will never have NVFP4, so maybe it's a no-op.

src/llmcompressor/modifiers/quantization/calibration.py
brian-dellabetta: i think that would preferable to pulling in active_session, yeah. unless there was a good reason we don't want observers referencing modules?

HDCharles: well there's other ways around this than using active session, the modifiers can pass the model directly

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: no fused observers are always set though that could probably be changed

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -83,20 +84,86 @@ def initialize_observer(
...
     :param module: module or iterable of modules with observer attributes
     :param base_name: substring used to fetch the observer and value to observe
+    :param model: the model to search for modules with fused observers (required if
+        also_do_fused is True)
+    :param also_do_fused: whether to also observe fused observers of the
+        observer(s)
     """
+    assert not (
+        also_do_fused and model is None
+    ), "Model must be provided if also_do_fused is True"
```
brian-dellabetta: to avoid invalid configuration, what if the logic wrapped in `also_do_fused` is done only if model is passed in, otherwise not?

src/llmcompressor/observers/helpers.py
```diff
@@ -185,21 +185,32 @@ def fuse_weight_observers(model: Module):
...
-            else:
-                Observer.fuse(observers)
+                    only_tensor_group = False
+                    continue
+                observers_and_modules.append((obs, layer))
+            if len(observers_and_modules) == 0:
+                continue
+
+            start = f"Some layers in fused group: {fusion_name_group} have"
+            end = ", need all fused layers to have same quantization"
```
brian-dellabetta, line 209: consider moving up in scope to avoid redundant re-setting of vars

brian-dellabetta (approved):
👍

src/llmcompressor/modifiers/quantization/calibration.py
HDCharles: outdated

kylesayrs (approved):

Issue #2735, opened 2026-05-20
DSv4 canonical example drops MTP layer; load_quantizable_moe regex anchored at ^layers excludes mtp.* block


pasta-paul:
## Summary

The canonical `examples/quantizing_moe/deepseek_v4_example.py` (`kylesayrs/transformers-v5` branch, commit `8c533c21f`, 2026-05-20) calibrates the main 43 routed-expert layers of DeepSeek-V4-Flash but **silently drops the MTP (multi-token-prediction) draft block** (`mtp.0.ffn.experts.0..255`). This makes the resulting artifact unusable with vLLM's `--speculative-config '{"method":"mtp","num_speculative_tokens":N}'`.

Two independent root causes compound:

### Root cause 1 — `transformers` 5.8.1 silently drops `mtp.*` on load

`transformers.models.deepseek_v4.modeling_deepseek_v4.DeepseekV4PreTrainedModel` carries:

```python
_keys_to_ignore_on_load_unexpected = [r"(^|\.)mtp\..*"]
```

`from_pretrained` filters these keys before they reach the state dict. Even after this filter is neutralized, `transformers/models/deepseek_v4/` has **no MTP module class** — only `num_nextn_predict_layers: int = 1` in the config. So the MTP block has no submodule for the keys to attach to.

### Root cause 2 — `load_quantizable_moe` / `linearize_moe` is MTP-unaware

In `src/llmcompressor/modeling/moe/conversion_mappings.py`:

```python
ARCH_TO_2D_MAPPINGS = {
    "deepseek_v4": (
        ["mlp.experts.gate_up_proj", "mlp.experts.down_proj"],
        [
            WeightRenaming(
                source_patterns=r"^layers\.(\d+)\.mlp\.experts\.(\d+)\.w1\.",
                target_patterns=r"layers.\1.mlp.experts.\2.gate_proj.",
            ),
            # ... w2, w3 ...
        ],
    )
}
```

The `^layers\.` anchor excludes `mtp.0.mlp.experts.*` paths. Even if root cause 1 were fixed (MTP module class added), `linearize_moe` would skip MTP's MoE block.

Grep across `src/llmcompressor/modeling/` confirms: **zero references to `mtp`/`MTP`/`nextn`.**

## Repro

```python
from llmcompressor.modeling.moe.linearize import load_quantizable_moe
from transformers import AutoModelForCausalLM
with load_quantizable_moe():
[...]

PR #2739, opened 2026-05-20
[deepseek_v4] Extend ARCH_TO_2D_MAPPINGS for MTP block

pasta-paul:
## Summary

DeepSeek-V4-Flash has a Multi-Token-Prediction (MTP) draft block (`num_nextn_predict_layers = 1`) with the same MoE structure as a main DecoderLayer but under an `mtp.<M>.` prefix instead of `layers.<L>.`.

The existing `ARCH_TO_2D_MAPPINGS["deepseek_v4"]` regex is anchored at `^layers\.`, so `linearize_moe`'s `named_modules()` walk for `FusedExpertsProtocol` instances picks up the 43 main decoder layers but skips the MTP block — making the MTP MoE ship uncalibrated.

This PR adds three additional `WeightRenaming` entries mirroring the existing main-layer ones but anchored at `^mtp\.(\d+)\.` to cover the MTP block's MoE.

## Dependency

Depends on [huggingface/transformers#46127](https://github.com/huggingface/transformers/pull/46127) — adds `DeepseekV4NextNPredictor` so `mtp.*` keys load into real submodules instead of being filtered out by the `_keys_to_ignore_on_load_unexpected` regex. Without that PR landing, the MTP keys never reach `named_modules()` for `linearize_moe` to walk.

## Acceptance

Single-rank dry calibration via `load_quantizable_moe()` on a DSv4-Flash BF16 checkpoint with MTP preserved → after `save_pretrained`, the resulting `model.safetensors.index.json` should contain `mtp.0.mlp.experts.*.{weight_packed,weight_scale,*_shape}` keys for all 256 routed experts (matching the main-layer pattern).

## Base branch

This PR is based on the `kylesayrs/transformers-v5` branch (PR #2647) since `conversion_mappings.py` doesn't exist on `main` yet. Once #2647 merges into main, this can be rebased trivially.

Closes the MTP gap discussed at #2735.

cc @kylesayrs — companion to your active DSv4 iteration. Paired with [huggingface/transformers#46127](https://github.com/huggingface/transformers/pull/46127).


pasta-paul:
## Design note — extension enables both MTP quantization AND deliberate MTP exclusion

Worth noting alongside this PR's "make MTP visible to the conversion mapping" goal: with the regex extension applied, users now have a **choice** about whether to quantize the MTP block.

Two valid recipes for `linearize_moe_model()` + DSv4 calibration:

```python
# Option X: quantize MTP alongside main MoE
ignore=["lm_head"]                       # default — MTP picked up by targets=

# Option Y: preserve MTP at higher precision  (DeepSeek's own MXFP4 release pattern)
ignore=["lm_head", r"re:.*mtp\..*"]      # MTP block excluded; stays at base dtype
```

Our shipping artifact uses Option Y — MTP stays BF16 while the main 43 layers' routed experts go to W4A16. Rationale: MTP is the speculative-decoding draft head; aggressive quantization degrades token-acceptance-rate (proportional to spec-decode throughput). 7% extra disk size is trivial vs 30%+ potential throughput hit.

This PR enables the mapping that makes MTP keys visible to the pipeline. The `ignore=` choice is recipe-level and orthogonal to whether the mapping should exist. **Both recipes need this PR to land.**

The `examples/quantizing_moe/deepseek_v4_example.py` could be extended with an Option Y variant that demonstrates the BF16-MTP recipe as a documented mixed-precision pattern.

pasta-paul:
> **Disclosure:** this comment was generated with AI assistance.

**Tracking note — MTP-class cluster.** This filing is one of four artifacts for what is fundamentally a single bug — "DSv4 MTP weights have no home in the canonical loader path":

| Filing | Layer | Role |
|---|---|---|
| huggingface/transformers#46127 (PR) | transformers | Add `DeepseekV4NextNPredictor` class so the MTP block has a module to load into |
| huggingface/transformers#46129 (issue) | transformers | `conversion_mapping` doesn't cover `mtp.*` paths — companion to #46127 |
| vllm-project/llm-compressor#2735 (issue) | llm-compressor | Canonical DSv4 calibration example drops MTP; regex `^layers\.` anchored excludes `mtp.*` |
| vllm-project/llm-compressor#2739 (PR) | llm-compressor | Extend `ARCH_TO_2D_MAPPINGS` for the MTP block — fixes #2735 |

Posting this header on all four so maintainers see one effort rather than four independent threads. Happy to consolidate further if reviewers prefer a single tracking issue.

PR #2748, opened 2026-05-21
feat: add layerwise quantization pipeline for memory-efficient large …

ManikandanBalasubramanian:
## Summary

Adds a new `layerwise` pipeline that enables AWQ quantization of models too large to fit in GPU memory by processing one subgraph (layer) at a time. This allows quantizing models of any size on a single GPU including GLM-5 (744B), which requires ~1.5TB just to load weights in fp16, yet can be quantized on a single H100 with this pipeline.

## Problem

Models like GLM-5 (744B, 160 experts) require ~1500GB VRAM just to load weights far beyond any single GPU. Even "smaller" large models like Qwen3.5-35B-A3B (256 experts, ~70GB fp16) or GLM-4.7-Flash (64 experts, ~59GB) exceed single-GPU capacity for standard quantization workflows.

The existing pipelines require the full model to reside in GPU memory during calibration and quantization, forcing users toward expensive multi-node setups or giving up on quantizing these models entirely.

This layerwise pipeline removes that constraint: since layer-local quantization techniques (AWQ, GPTQ) only need one layer's weights and activations at a time, we never need the full model in memory. The pipeline loads, calibrates, quantizes, and saves each layer independently  bounded only by the size of the largest single layer.

**Practical benefits:**
- Quantize a 744B model on a single H100 (80GB)
- Run multiple quantization jobs in parallel on different GPUs (different models, or same model with different calibration data)
- Well-suited for any layer-local quantization technique

### New Pipeline (`src/llmcompressor/pipelines/layerwise/`)
- Sequential onloading of model weights per subgraph
- Compress-as-you-go: quantize and save each subgraph immediately, then free memory
- On-demand shard downloads for HuggingFace Hub models
- Sequential prefetch for calibration batches
- Support for fused MoE experts, VL models, and tied weights

### AWQ Modifier Optimizations (`modifiers/transform/awq/base.py`)
- VRAM-aware fp16 output placement (CPU vs GPU based on available memory)
- Pre-move fp16 outputs to GPU before grid search to reduce CPU→GPU transfers
- Streaming MSE loss computation to avoid materializing all quantized outputs
- GPU memory cleanup after grid search

### Entrypoint & Supporting Changes
- `oneshot()` accepts `layerwise=True` and `sequential_prefetch=True`
- Offset-norm weight handling for layerwise pipeline
- GPU memory leak fix: offload compressed params to meta device
- Skip tied weight aliases in compress-as-you-go save
- AWQ mappings for additional model architectures

## Usage

```python
from llmcompressor.modifiers.transform.awq import AWQModifier

oneshot(
    model="Qwen/Qwen3.5-35B-A3B",
    dataset=calib_dataset,
    recipe=[AWQModifier(duo_scaling="True"), QuantizationModifier(targets="Linear", scheme="W4A16")],
    max_seq_length=2048,
    num_calibration_samples=512,
    output_dir="./Qwen3.5-35B-AWQ-INT4",
    layerwise=True,
    sequential_prefetch=True,
)
```

## Validated On

| Model | Params | Layers | Experts | GPU | Status |
|-------|--------|--------|---------|-----|--------|
| GLM-5 | 744B | 92 | 160 (8 active) | 1× H100 80GB | ✅ |
| Qwen3.5-35B-A3B | 35B | 40 | 256 (8 active) | 1× L40S 46GB | ✅ |
| GLM-4.7-Flash | ~30B | 47 | 64 (4 active) | 1× L40S 46GB | ✅ |
| Qwen3.5-4B | 4B | 36 | dense | 1× L40S 46GB | ✅ |
| Qwen3-4B | 4B | 36 | dense | 1× L40S 46GB | ✅ |

## Tests

- Unit tests for layerwise helpers (weight map building, key remapping, subgraph partitioning, offload/move)
- Integration tests: full AWQ+W4A16 and W4A16-only e2e on Qwen3-0.6B
- Pipeline registry test for LayerwisePipeline discoverability

## Limitations

- Designed for **layer-local** quantization techniques (AWQ, GPTQ) where each layer can be quantized independently. Not applicable to methods requiring global model access.

### DIFF
--- a/src/llmcompressor/args/dataset_arguments.py
+++ b/src/llmcompressor/args/dataset_arguments.py
@@ -290,6 +290,14 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
+    layerwise_resume_from: int = field(
+        default=0,
+        metadata={
+            "help": "Subgraph index to resume layerwise quantization from. "
+            "Subgraphs before this index are skipped (forward-only replay to "
+            "rebuild intermediates). Useful for resuming after a crash. Default 0."
+        },
+    )
 
     def is_dataset_provided(self) -> bool:
         return self.dataset is not None or self.dataset_path is not None
--- a/src/llmcompressor/args/model_arguments.py
+++ b/src/llmcompressor/args/model_arguments.py
@@ -86,3 +86,13 @@ class ModelArguments:
             "(can be a branch name, tag name or commit id)"
         },
     )
+    layerwise: bool = field(
+        default=False,
+        metadata={
+            "help": "Enable layerwise quantization mode. When True, the model is "
+            "loaded on meta device (no memory) and weights are loaded from "
+            "safetensors files per-subgraph during calibration. This allows "
+            "quantizing models that are too large to fit in memory. "
+            "Requires the model to be saved in safetensors format."
+        },
+    )
--- a/src/llmcompressor/core/state.py
+++ b/src/llmcompressor/core/state.py
@@ -105,6 +105,7 @@ class State:
     loss_masks: list[torch.Tensor] | None = None
     current_batch_idx: int = -1
     sequential_prefetch: bool = False
+    pipeline_type: str = "basic"
 
     @property
     def compression_ready(self) -> bool:
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -9,11 +9,14 @@ with various pipeline configurations for efficient model optimization.
 
 from __future__ import annotations
 
+import json
 import os
+import shutil
 from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Callable
 
+from compressed_tensors import ModelCompressor
 from loguru import logger
 from torch.utils.data import DataLoader
 from transformers import PreTrainedModel, PreTrainedTokenizerBase, ProcessorMixin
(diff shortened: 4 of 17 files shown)


brian-dellabetta:
Hi @ManikandanBalasubramanian , reading through your PR summary it sounds like you are reproducing the [sequential pipeline](https://github.com/vllm-project/llm-compressor/tree/main/src/llmcompressor/pipelines/sequential) to process the model one layer at a time. If you run your example snippet above on main, it already does this on a layer-by-layer basis unless you specify it use a different pipeline.

ManikandanBalasubramanian:
Hi @brian-dellabetta , thanks for looking at this! 

Yes, both pipelines process the model layer by layer, but the key difference is in how the model is loaded into memory.

The sequential pipeline requires [from_pretrained](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/entrypoints/utils.py) which places all weights in CPU RAM, then onloads layers to GPU one at a time. This works great for models that fit in CPU memory (like Llama-3.3-70B on a 150GB RAM machine).

But for models like GLM-5 (744B, ~1.5TB in fp16), there's no practical way to hold all weights in CPU RAM. The layerwise pipeline solves this by:

- Loading only the model structure on meta device (no real weights allocated)
- Loading one layer's weights directly from safetensors on disk → GPU
- Calibrating and quantizing that layer
- Saving the compressed layer to disk, then freeing all memory
- Repeating for the next layer

Peak memory is bounded by the largest single layer (~5-10GB), not the full model. This is what makes quantizing a 744B model on a single H100 possible, you'd need ~1.5TB RAM to even attempt it with the sequential pipeline.

Additionally, the pipeline supports:
- On-demand shard downloads (no need to download the full model upfront)
- Resume from any layer if interrupted
- Background shard prefetching to overlap I/O with compute

I'd love to hear your thoughts on whether this fits well as a separate pipeline or is this better suited as a standalone script outside the project?

brian-dellabetta:
Hi @ManikandanBalasubramanian , please review our examples for using disk offloading. We allow users to quantize models without loading the entire model up into memory
https://github.com/vllm-project/llm-compressor/tree/main/examples/disk_offloading

ManikandanBalasubramanian:
Thanks for pointing this out, @brian-dellabetta  I wasn't aware of the disk offloading support with "auto_offload", that combined with the sequential pipeline already covers the use case I was targeting. Closing this PR. Appreciate the review!

Issue #2006, opened 2025-11-09
[RFC]: MR-GPTQ (GPTQ+NVFP4)


mratsim:
A new paper came out that tailors GPTQ to MXFP4 and NVFP4 storage format:
- https://arxiv.org/pdf/2509.23202

> The recent hardware-accelerated microscaling 4-bit floating-point formats such as MXFP4 and NVFP4, supported on NVIDIA and AMD GPUs, promise to revolutionize large language model (LLM) inference. Yet, their practical benefits remain unproven. We present the first comprehensive study of MXFP4 and NVFP4 for post-training quantization, revealing gaps between their promise and real-world performance. Our analysis shows that state-of-the-art methods struggle with FP4, due to two key issues: (1) NVFP4's small group size provably neutralizes traditional outlier mitigation techniques; (2) MXFP4's power-of-two scale quantization severely degrades accuracy due to high induced error. To bridge this gap, we introduce Micro-Rotated-GPTQ (MR-GPTQ), a variant of the classic GPTQ quantization algorithm that tailors the quantization process to FP4's unique properties, by using block-wise Hadamard transforms and format-specific optimizations. We support our proposal with a set of high-performance GPU kernels that enable the MR-GPTQ format with negligible overhead, by rotation fusion into the weights, and fast online computation of the activations. This leads to speedups vs. FP16 of up to 3.6x layer-wise, and 2.2x end-to-end on NVIDIA B200, and of 6x layer-wise and 4x end-to-end on RTX5090. Our extensive empirical evaluation demonstrates that MR-GPTQ matches or outperforms state-of-the-art accuracy, significantly boosting MXFP4, to the point where it can near the accuracy that of NVFP4. We conclude that, while FP4 is not an automatic upgrade over INT4, format-specialized methods like MR-GPTQ can unlock a new frontier of accuracy-performance trade-offs. 

The models are supported in vLLM since https://github.com/vllm-project/vllm/pull/24440 and reference quantization framework is https://github.com/IST-DASLab/FP-Quant

kylesayrs:
Hi @mratsim!

This technique has been implemented in LLM Compressor through transforms + GPTQ. More work needs to be done to verify these results, but you can try out this configuration yourself using the below example and changing QuantizationModifier to GPTQModifier.

https://github.com/vllm-project/llm-compressor/blob/main/examples/transform/quip_example.py

mratsim:
[...]

PR #2751, opened 2026-05-23
[Examples] Add MR-GPTQ (QuIP + GPTQ + NVFP4A16) example

Yatimai:
## Purpose

Add `examples/transform/nvfp4_quip_gptq_example.py`, demonstrating the composition of `QuIPModifier` block-wise Hadamard rotations with `GPTQModifier` calibration for NVFP4A16 weight quantization. This composition is the MR-GPTQ recipe described in arXiv:2509.23202.

Follows @kylesayrs's suggestion in #2006 to combine transforms with GPTQ for NVFP4. The example mirrors the existing `examples/transform/quip_example.py` structure, adapted to use `GPTQModifier` with WikiText-2 calibration for the data-dependent path.

## Context

The composition (`QuIPModifier` + `GPTQModifier` for NVFP4A16) requires a runtime fix in vLLM's `hadacore_transform` kernel on Blackwell consumer GPUs (RTX 5090 / SM120), addressed in vllm-project/vllm#43462. Without that fix, the kernel produces garbage outputs under `torch.compile` + CUDA graph capture; with it applied, the composition runs end-to-end with measurable accuracy.

A `NOTE:` header in the example file points users to the vLLM PR so they know which runtime version is required on consumer Blackwell.

## Changes

* Add `examples/transform/nvfp4_quip_gptq_example.py` mirroring `examples/transform/quip_example.py` structure, with `GPTQModifier` + WikiText-2 calibration

## Testing

End-to-end on `meta-llama/Llama-3.1-8B` (base, not Instruct, to match the standard PPL benchmark protocol):

- Quantization completes via QuIP rotations + GPTQ calibration on WikiText-2 (256 chunks × 2048 tokens)
- Sample generation post-quantization is coherent
- WikiText-2 test PPL = **6.4491** (vs FP16 baseline 6.1886), 141 chunks of 2048 tokens, measured via vLLM built from #43462

## Related

- vllm-project/vllm#43462: runtime kernel fix (required on RTX 5090 / SM120)
- vllm-project/vllm#22486: foundational online Hadamard rotations support
- #2006: original symptom report (@mratsim) and recipe suggestion (@kylesayrs)
- `examples/transform/quip_example.py`: reference W4A16 datafree variant this file builds on


examples/transform/nvfp4_quip_gptq_example.py
```diff
@@ -0,0 +1,109 @@
...
+    weights=QuantizationArgs(
+        num_bits=4,
+        actorder="static",
+        type=QuantizationType.FLOAT,
+        strategy=QuantizationStrategy.TENSOR_GROUP,
+        symmetric=True,
+        dynamic=False,
+        group_size=16,
+        scale_dtype=FP8_E4M3_DATA.dtype,
+        zp_dtype=FP8_E4M3_DATA.dtype,
```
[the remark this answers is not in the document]
Yatimai: This is a false positive. `FP8_E4M3_DATA.dtype` exists and resolves to `torch.float8_e4m3fn`. Verified locally:

```python
>>> from compressed_tensors.quantization import FP8_E4M3_DATA
>>> FP8_E4M3_DATA.dtype
torch.float8_e4m3fn
```

The reference `examples/quantization_w4a4_fp4/llama3_gptq_example.py` uses the exact same `scale_dtype=FP8_E4M3_DATA.dtype` / `zp_dtype=FP8_E4M3_DATA.dtype` pattern, currently merged and working.

lantudou:
I tested this recipe on Qwen3.6 35B. it will cause nan value during the calibration, but nvfp4a16+gptq is fine. I guess quip is not suitable for MOE model or it is just a bug?

Yatimai:
Thanks for testing it, and good instinct: your isolation is right, it's the QuIP step, not the quantization. The fact that nvfp4a16+gptq works fine for you confirms the transform is the source of the NaNs.

QuIP applies block-wise Hadamard rotations, which are defined over standard dense linears. MoE layers are a different shape (grouped expert tensors plus a router), so the rotations don't map onto them as-is. This is a known boundary for rotation methods in general, not specific to this recipe, which runs end-to-end on dense models (6.45 PPL on Llama-3.1-8B base here). Extending QuIP to MoE means handling the expert structure explicitly, which is its own piece of work.

For your case it wouldn't change the outcome anyway: on NVFP4A16 the activations stay in high precision, so the rotations are neutral by construction (in my ablation GPTQ alone matches QuIP+GPTQ on this format). nvfp4a16+gptq is the right call for your Qwen3 MoE. The composition is built for the activation-quant case (MXFP4 W4A4), where the input rotation actually does work.

examples/transform/nvfp4_quip_gptq_example.py
```diff
@@ -0,0 +1,109 @@
...
+from llmcompressor.modifiers.gptq import GPTQModifier
+from llmcompressor.modifiers.transform import QuIPModifier
+
+# Select model and load it.
+model_id = "meta-llama/Llama-3.1-8B"
+model = AutoModelForCausalLM.from_pretrained(model_id, dtype="auto")
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+# Select calibration dataset.
+# WikiText-2 is the standard calibration corpus in quantization research
```
brian-dellabetta, line 27: hmm, would this not work as well using one of the datasets we more commonly use in examples? The additional overhead of chunking below kinda gnarls up the example

brian-dellabetta (commented):
Thanks @Yatimai for preparing this! One request

Sorry for the delay, we are through some work now and reviewing the backlog.

examples/transform/nvfp4_quip_gptq_example.py
brian-dellabetta, line 75: and to make this a little cleaner, can you just use the `NVFP4A16` or `NVFP4` preset scheme directly?

Yatimai:
@brian-dellabetta Thanks for the review.

**Dataset**: agreed, I'll switch to `ultrachat_200k` (matching `llama3_gptq_example.py`).

**Preset / direction**: here's where I'd like your call before refactoring. Since #43229 is building the NVFP4 W4A4 Qutlass runtime (your smoke tests use `quipv16-nvfp4`), it feels like this example's natural role is the producer side of that W4A4 story, i.e. `NVFP4` (W4A4) preset, coupled to #43229. To match your test models exactly, I'd also move to v-only block-16 (dropping the output `u` rotation, which my ablation showed is neutral on NVFP4 anyway), that's exactly the `quipv16` setup your runtime serves. The catch is W4A4 + QuIP needs `QutlassNvFP4LinearMethod.apply()` from #43229, so it's not runnable on main until that lands.

Two ways I see it:

- **Couple to #43229**: refactor to `NVFP4` (W4A4) + v-only block-16 and land this alongside #43229 when the runtime is ready, one coherent end-to-end story.
- **Standalone now**: ship `NVFP4A16` (runnable on main today via hadacore), follow up with W4A4 once #43229 merges.

W4A4 couples to your runtime but waits on #43229; A16 ships now but stays orphaned from where the format pays off. Let me know what you'd prefer.

Also FYI: the PyTorch 2.12 aliasing warning you `.clone()`-suppressed in #43229 is what vllm-project/vllm#43462 fixes at root, once it merges, `ops.hadacore_transform(value)` works without the clone, as Kyle noted.

kylesayrs (approved):
Looks good! I think non-standard dataset is fine

examples/transform/nvfp4_quip_gptq_example.py
```diff
@@ -0,0 +1,109 @@
...
+    ]
+)
+
+# Configure the quantization algorithm to run.
+#   * apply quip transforms to model in order to make quantization easier
+#   * quantize the weights to nvfp4 (fp4 weights, fp16 activations) with a
+#     hardware-imposed group size of 16
+#   * use GPTQ with mse observer for activation-ordering-aware calibration
+#     (combination follows the MR-GPTQ recipe, arXiv:2509.23202)
+NVFP4A16 = dict(
```
brian-dellabetta: can't we just use the `NVFP4A16` preset here?

examples/transform/nvfp4_quip_gptq_example.py
```diff
@@ -0,0 +1,109 @@
...
+        group_size=16,
+        scale_dtype=FP8_E4M3_DATA.dtype,
+        zp_dtype=FP8_E4M3_DATA.dtype,
+        observer="mse",
+    ),
+    targets=["Linear"],
+)
+recipe = [
+    QuIPModifier(
+        rotations=["v", "u"], transform_block_size=128, transform_type="hadamard"
```
brian-dellabetta: suggest not adding u here
```suggestion
        rotations=["v"], transform_block_size=128, transform_type="hadamard"
```

brian-dellabetta (commented):
Hi @Yatimai , we can probably land this, but i have a couple nits

brian-dellabetta (approved):
Thanks @Yatimai !

examples/transform/nvfp4_quip_gptq_example.py
```diff
@@ -0,0 +1,109 @@
...
+ds = Dataset.from_list(
+    [
+        {
+            "input_ids": input_ids_all[
+                i * MAX_SEQUENCE_LENGTH : (i + 1) * MAX_SEQUENCE_LENGTH
+            ].tolist()
+        }
+        for i in range(n_chunks)
+    ]
+)
```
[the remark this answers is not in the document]
Yatimai: With the pinned dataset and defaults this always yields 256 chunks; keeping the example minimal.

examples/transform/nvfp4_quip_gptq_example.py
Yatimai: Addressed in a35d485 (NVFP4A16 preset), resolving.

examples/transform/nvfp4_quip_gptq_example.py
Yatimai: Done, dropped u. Matches my ablation (u is neutral on NVFP4A16) and the v-only default from #2815

examples/transform/nvfp4_quip_gptq_example.py
Yatimai: Done in a35d485. Switched to the NVFP4A16 preset, keeping the mse observer via weight_observer since it is part of the MR-GPTQ recipe and what the reported PPL was measured with.

PR #2762, opened 2026-05-27
minor docs: remove non existent training args

JINO-ROHIT:
remove training args since it doesnt exist


brian-dellabetta (approved):
thanks!

orestis-z (approved):
Thanks!

PR #2782, opened 2026-06-02
[Misc] Remove oneshot `output_dir` warning

kylesayrs:
## Purpose ##
* Remove warning for gold-standard and expected behavior
  * For context, this warning was added when passing `output_dir` to `oneshot` or `train` was the default flow
  * Now, we discourage users from passing `output_dir`
  * It's been almost two years since we made this change, it's reasonable for users to have an expectation that their model will not be automatically saved if `output_dir` is not passed.

## Changes ##
* Remove warning


brian-dellabetta (approved):
yeah this is noisy and worth removing imo

kylesayrs:

PR #2783, opened 2026-06-02
[Lifecycle] Make calibration events first-class citizens

kylesayrs:
## Purpose ##
* Remove confusing and boiler-plate modifier code for handling calibration events
* Rename `calibration_epoch_start/end` to `calibration_start/end` for clarity, as requested by @brian-dellabetta

## Changes ##
* Add calibration lifecycle event hooks as optional methods that modifiers can implement
  * `on_calibration_start`
  * `on_sequential_epoch_end`
  * `on_calibration_end`
* All logic related to `start_`ing and `end_`ing is now handled by the `Modifier` base class
* Updated documentation 

## Testing ##
* All tests in `tests/llmcompressor/modifiers/` folder pass


brian-dellabetta (commented):
I agree with the goal, but I think the name of calibration epoch vs sequential epoch is confusing. I never understood that terminology beforehand, but I think it makes more sense to change the event names than the `on_start`/ `on_end` parts of the Modifier API. what about `on_calibration_(start|end)`?

kylesayrs:
@brian-dellabetta I agree the names are non-ideal (they were chosen based on the functionality that was available at the time that they were created), but renaming the events is out of scope of what I'm trying to accomplish. Let's do a rename in a separate PR where we also change docs.

HDCharles:
i actually like this better, @brian-dellabetta i was always confused by on init and on start which both seemed different than the calibration/sequential epoch and i wasn't sure where they would go if you had multiple modifiers but this seems clearer

```
on init
for each modifier
    on calib start (clearly goes in the for loop rather than before like on init)
    for each sequential target
        on seq start 
        on seq end
    on calib end (clearly goes in the for loop rather than after like finalize)
on finalize

```

HDCharles (approved):

brian-dellabetta (approved):
Thanks Kyle, the updates look good to me

PR #2785, opened 2026-06-02
[Distributed] Module parallel calibration for `QuantizationModifier`

kylesayrs:
## Purpose ##
* Accelerate distributed calibration for `QuantizationModifier` by performing weight calibration in parallel
  * The speedups from this change are minor: runtime is typically dominated by activation calibration, not weight calibration

## Changes ##
* Implement weight-parallel calibration of weights for `QuantizationModifier`

## Testing ##
The following times are taken from quantizing `Qwen/Qwen3.5-122B-A10B` to RTN `NVFP4` with full disk offloading and one calibration sample.

### Main ###
| world_size | oneshot time | e2e time |
| - | - | - |
| 1 | 8.67 | 15m |
| 2 | 10.81 | 19m |
| 3 | 8.77 | 17m |

### This Branch ###
| world_size | oneshot time | e2e time |
| - | - | - |
| 1 | 8.69m | 14m |
| 2 | 7.72m | 15m |
| 3 | 8.09m | 16m |

Non-linear speedups are due to fixed costs associated with calibration

### DIFF
--- a/src/llmcompressor/modifiers/quantization/quantization/base.py
+++ b/src/llmcompressor/modifiers/quantization/quantization/base.py
@@ -1,5 +1,11 @@
 import torch
-from compressed_tensors.offload.dist_utils import is_source_process as is_src
+import torch.distributed as dist
+from compressed_tensors.distributed import (
+    greedy_bin_packing,
+    is_distributed,
+    wait_for_comms,
+)
+from compressed_tensors.offload import get_execution_device
 from compressed_tensors.quantization.utils import is_module_quantized
 
 from llmcompressor.core import Event, State
@@ -12,6 +18,8 @@ from llmcompressor.modifiers.quantization.quantization.mixin import Quantization
 from llmcompressor.observers import ACTIVATION_OBS
 
 __all__ = ["QuantizationModifier"]
+
+_WEIGHT_Q_PARAMS = ["weight_scale", "weight_zero_point", "weight_global_scale"]
 
 
 class QuantizationModifier(Modifier, QuantizationMixin):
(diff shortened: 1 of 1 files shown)


src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -13,6 +18,8 @@
 
 __all__ = ["QuantizationModifier"]
 
+_WEIGHT_Q_PARAMS = ["weight", "weight_scale", "weight_zero_point"]
```
brian-dellabetta: weight_packed? weight_shape? We should use the CT [QuantizationMetadata](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_metadata.py#L25-L40) as source of truth here

brian-dellabetta (commented):
Looks good, but one change request

src/llmcompressor/modifiers/quantization/quantization/base.py
kylesayrs: Weight packing occurs during/after compression. You will never see a packed weight during calibration/optimization.

If we used QuantizationMetadata, we would be including broadcasting activation qparams, which would be a waste of work; We already assume that activation qparams are identical across ranks.

Note that this logic will be replaced once [[Offload] [2/2] Better abstractions for updating parameters](https://github.com/vllm-project/compressed-tensors/pull/703) lands.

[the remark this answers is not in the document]
kylesayrs: Yep this makes sense

brian-dellabetta: ok, i missed that this was not a PR into main but your other branch. so it is the list of all pre-compression weight q params, excepting "weight" itself?

src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -78,11 +86,49 @@ def on_sequential_epoch_end(
...
+            if get_execution_device(module) != torch.device("cpu"):
+                for qparam_name in _WEIGHT_Q_PARAMS:
+                    if (qparam := getattr(module, qparam_name, None)) is not None:
+                        pending_comms.append(
+                            dist.broadcast(
+                                qparam, src=module_to_rank[module], async_op=True
+                            )
+                        )
+
+        wait_for_comms(pending_comms)
```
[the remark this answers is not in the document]
kylesayrs: Good catch.

brian-dellabetta (approved):

HDCharles (approved):

PR #2791, opened 2026-06-03
Add HYV3 MoE calibration adapter

selalipop:
SUMMARY: Adding support for calibrated quantization for HY3

TEST PLAN:
- Converted finetuned hy3-preview checkpoint and confirmed it loads in vllm as expected
- Measured perplexity vs BF16 and model free ptq checkpoints on held-out portion of my calibration samples: BF16 hadslowest perplexity as expected (+0%), followed by calibrated (+3.68%) then model free (+13.98%)


brian-dellabetta (commented):
Hi @selalipop , the moe calibration context will be removed as part of the work to support transformers v5:
* #2647 

Perhaps we can revisit after the upgrade to see if additional logic is needed to support HYV3MoE?

kylesayrs:
This has landed as part of https://github.com/vllm-project/llm-compressor/pull/2647.

Please create an issue if this does not enable the model!

PR #2794, opened 2026-06-04
fix: correct mypy files path from guidellm to llmcompressor

soyr-redhat:
SUMMARY:
The `[tool.mypy]` config in `pyproject.toml` had `files = "src/guidellm"`, which appears to be an artifact from guidellm. Updated to `files = "src/llmcompressor"` so mypy checks the correct source tree.

TEST PLAN:
Verified that `src/llmcompressor` exists and `src/guidellm` does not. This is a config-only fix with no runtime impact.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools", "wheel", "setuptools_scm==8.2.0"]
 build-backend = "setuptools.build_meta"
 
 [tool.mypy]
-files = "src/guidellm"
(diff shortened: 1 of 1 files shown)


soyr-redhat:
Brief update, I noticed that mypy was commented out of the Makefile. Is this something we want removed?

brian-dellabetta (approved):
thanks!

kylesayrs (approved):
Thanks for the catch

PR #2798, opened 2026-06-04
fix: re-tie word embeddings before save to avoid duplicate lm_head

EdalatiAli:
## Summary
When a model with tied word embeddings is saved while offloaded, the offload conversion (`to_accelerate`) splits `embed_tokens` and `lm_head` into separate parameters. This defeats transformers' save-time de-duplication, so a redundant, bit-identical `lm_head.weight` gets written.

This re-ties the weights right after `to_accelerate` and before saving, so the tied head is dropped as intended. It's a cheap attribute reassignment (no onloading) and a no-op for models with untied embeddings.

## Test plan
- Exported an AWQ (sequential-pipeline) checkpoint and confirmed `lm_head.weight`  no longer appears in `model.safetensors.index.json`.
- Verified datafree/untied exports are unaffected.
- passed the added test: `pytest tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py -k no_duplicate_tied_lm_head -v`


brian-dellabetta (approved):
thanks! Will confirm with @kylesayrs

kylesayrs (changes requested):
Will write up more information in a bit, but temporarily blocking this for now for the following changes:
1. This logic needs a guard on `if model.config.tie_embeddings:`. Sometimes `oneshot` will untie embeddings, as is the case for spinquant, so checking for the presence of `tie_weights` is not alone.
2. Some models implement `tie_weights`, but the implementation raises an error. We should try to catch this case and warn if possible

kylesayrs:
Outside of the required changes mentioned above, I'll also mention that this usage creates a data configuration that `accelerate` isn't expecting and might break in the future, depending on the implementation of `save_pretrained`

`tie_weights` will simply reassign the `weight` parameter of `lm_head` to the value of `embed_tokens`. This works because transformers, when saving, only checks the id of the meta tensor ([source](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L474-L475), I actually wrote some of this logic).

However, typically accelerate expects all tied tensors to not only share the same meta tensor value, but also that you specify them in an explicit `tied_params_map` ([source](https://github.com/huggingface/accelerate/blob/8d0796721e9b5d233233e17186810c0f11bf9c98/src/accelerate/big_modeling.py#L440)). By doing the above, we break this assumption.

This is nbd, but it might break in the future if accelerate changes how it implements `save_pretrained`.

I think a more stable pattern might be to perform the tying before converting to accelerate

```python3
if model.config.tie_embeddings:  # might need a more robust check
  _retie_offloaded_weights(model)  # add warning/logger if weights are retied

to_accelerate(model)
```

EdalatiAli:
Thanks for the detailed review @kylesayrs @brian-dellabetta   ! Addressed the three points:

1. **Guard on tie config** — `_retie_offloaded_weights` now early-returns unless `config.tie_word_embeddings` is set (checked via `get_text_config(decoder=True)` so multimodal configs work too). Models that `oneshot` explicitly unties (e.g. SpinQuant via `untie_word_embeddings`, which sets `tie_word_embeddings = False`) are left untouched, so checking for `tie_weights` alone is no longer the only condition.
2. **Catch + warn on `tie_weights` failures** — wrapped the call in a `try/except` that logs a warning instead of failing the save, for models whose `tie_weights` implementation raises.
3. **Re-tie before `to_accelerate`** — moved the re-tie to run before the offload conversion, per your note about not creating a tied layout that accelerate isn't expecting (shared meta tensor + explicit `tied_params_map`).


I also added a regression test (`test_no_duplicate_tied_lm_head_on_save` in `test_compress_tensor_utils.py`) covering offloaded/non-offloaded × tied/untied, asserting `lm_head.weight` is dropped on disk for tied models and kept for untied ones.

kylesayrs (approved):
Great job!

brian-dellabetta (approved):
Thanks @EdalatiAli !

brian-dellabetta:
@EdalatiAli can you run `make style` / `make quality`?

brian-dellabetta:
`pytest -s tests/llmcompressor/transformers/gptq/test_gptq_oneshot.py` is failing with this warning in the logs:
```
[transformers] The tied weights mapping and config for this model specifies to tie model.embed_tokens.weight to lm_head.weight, but both are present in the checkpoints with different values, so we will NOT tie them. You should update the config with `tie_word_embeddings=False` to silence this warning.
```
Apparently our smoke model is incorrectly configured, its [config](https://huggingface.co/nm-testing/tinysmokellama-3.2/blob/main/config.json#L25) has `"tie_word_embeddings": true` but both `model.embed_tokens.weight` and `lm_head.weight` appear in model.safetensors. The test succeeds with `meta-llama/Llama-3.2-1B-Instruct`. 

I will confirm with original creator of smoke model that changing won't break anything or get overwritten by some other process

brian-dellabetta:
Updating the model config to `tie_word_embeddings:False` causes [new errors](https://github.com/vllm-project/llm-compressor/actions/runs/27441892662/job/81126664867?pr=2798#step:11:147) with `tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py::test_no_duplicate_tied_lm_head_on_save`. i've reverted the model config.json for now, will need to debug this next week and then we can get it in after release

EdalatiAli:
Thank you @brian-dellabetta for fixing the issue with the failed test! Would be great to merge the PR if there are no other concerns

brian-dellabetta:
> Thank you @brian-dellabetta for fixing the issue with the failed test! Would be great to merge the PR if there are no other concerns

Thanks @EdalatiAli , I'll cover this with Kyle when we meet this afternoon, hope to get it in today

kylesayrs (approved):

kylesayrs:
I think this solution may be incorrect, as it messes with cases where only the input embedding is quantized

brian-dellabetta:
> I think this solution may be incorrect, as it messes with cases where only the input embedding is quantized

Yeah, this PR and the PR that introduced embedding quantization ( #2830 and https://github.com/vllm-project/compressed-tensors/pull/718) were considered independently, so we're hitting an issue at the intersection of them. But we can resolve in #2835

PR #2770, opened 2026-05-28
Add AWQ model mappings for Step3p5

wanadzhar913:
- Closes: https://github.com/vllm-project/llm-compressor/issues/2386

## SUMMARY:
- Add mappings for https://huggingface.co/stepfun-ai/Step-3.5-Flash/blob/main/modeling_step3p5.py
- Add example script (untested) in `step3p5_example.py`

Thanks so much!

## Architecture Notes
Step-3.5-Flash:
- **GQA attention** with standard `q_proj`, `k_proj`, `v_proj`, plus optional head-wise attention gating via `g_proj` (which we included)
- **Mixed dense/MoE FFN**:
  - Dense layers use standard SwiGLU MLP (`gate_proj`, `up_proj`, `down_proj`)
  - MoE layers use routed experts implemented with packed `MoELinear`
  - MoE blocks also include a dense `share_expert` branch with standard MLP structure

## TEST PLAN:
None at the moment as I don't have enough resources/GPU VRAM. However, the following files/tests should be ran:

- `python3 ./examples/awq/step3p5_example.py`
- `pytest ./tests/llmcompressor/modifiers/transform/awq/test_base.py ./tests/llmcompressor/modifiers/transform/awq/test_dynamic_mappings.py`

PR #2768, opened 2026-05-28
Add AWQ Mappings for Glm4MoeLiteForCausalLM

Priya95715:
Fixes #2696

## Summary:
Adds AWQ smoothing mappings for `Glm4MoeLiteForCausalLM` (GLM-4.7-Flash model).

## Architecture Notes
GLM-4.7-Flash:
- **MLA attention** (`q_a_proj`, `kv_a_proj_with_mqa`, `q_b_proj`, `kv_b_proj`) instead of standard `q/k/v_proj`
- **Mixed dense/MoE FFN**: `first_k_dense_replace=1` means layer 0 is a standard dense FFN; layers 1–46 are MoE with 64 routed experts + 1 shared expert

## Key Design Decision
`mlp.gate` (the MoE router) is intentionally excluded from balance layers. Including it causes `match_modules_set` to never complete a group for the dense layer 0 (which has no router), accumulating all `post_attention_layernorm` entries across all 47 layers into a single group.

## Testing
Quantized `zai-org/GLM-4.7-Flash` with AWQ + W4A16 successfully. Verified inference via vLLM produces coherent outputs.

PR #2804, opened 2026-06-05
[Skill] Automated PR Validator

brian-dellabetta:
SUMMARY:
Basic Proof-of-concept for validating PRs following structure of [this skill](https://github.com/opendatahub-io/skills-registry/blob/main/skills/generate-site/SKILL.md?plain=1) in the opendatahub skills registry. 

This skill allows us to tell claude "Validate PR #2768", it will inspect the diff for new mappings, and run awq for a model matching the added architecture in a fresh venv, report its findings onto the PR, and tear down.

Currently runs locally, in future we could extend this to run in RHOAI.

- [x] Validate AWQ mappings
- [ ] Validate SmoothQuant mappings
- [ ] Validate new examples

TEST PLAN:
- [x] Validation works for correct AWQ mappings -- https://github.com/vllm-project/llm-compressor/pull/2768#issuecomment-4635617614
- [x] Validation works for incorrect AWQ mappings  -- https://github.com/vllm-project/llm-compressor/pull/2770#issuecomment-4635251905


kylesayrs:
I'm a little bit confused why we need a full agentic workflow for this? I think it's good to run tests against mappings prs, but I feel like this could be better handled by just adding a unit test and pointing users towards adding a parameterization of the test.

```python3
@pytest.mark.parameterize("model_type,layer_cls", [... "PR committers add a parameterization here" ...])
def test_mappings(model_type, layer_cls):
  layer = layer_cls(...)
  inputs = ...
  awq = AWQ(...)

  awq._on_start(layer)
  true_outputs = layer(inputs)
  assert awq._parent_args_cache ...

  awq._apply_smoothing(layer)
  outputs = layer(inputs)
  assert torch.functional.mse(outputs, true_outputs) <= 1e-10
```

.claude/skills/validate-pr/SKILL.md
```diff
@@ -0,0 +1,121 @@
...
+
+Then diff against `main` to understand what changed:
+```bash
+git diff main...HEAD --stat
+git diff main...HEAD
+```
+
+### Step 3: Identify new AWQ mappings
+
+Inspect the diff from step 2. Look for new or modified AWQ mapping files (typically under `src/llmcompressor/modifiers/transform/awq/`). If the PR contains **no new AWQ mappings**, skip to step 8 and post a comment noting that no AWQ validation was needed.
```
orestis-z: Maybe better to not post anything if there are no  AWQ mapping changes? I can imagine we could have more validators later and if each of them posts a comment noting that there was no action needed from their end it could get spammy.

.claude/skills/validate-pr/SKILL.md
```diff
@@ -0,0 +1,121 @@
+---
+name: validate-pr
```
orestis-z, line 2: I believe `validate-pr` is a bit too generic since the validation is specifically about AWQ-mappings. Maybe sth like `validate-awq-mappings-pr` would be more honest?

orestis-z:
> I'm a little bit confused why we need a full agentic workflow for this? I think it's good to run tests against mappings prs, but I feel like this could be better handled by just adding a unit test and pointing users towards adding a parameterization of the test.
> 
> ```python
> @pytest.mark.parameterize("model_type,layer_cls", [... "PR committers add a parameterization here" ...])
> def test_mappings(model_type, layer_cls):
>   layer = layer_cls(...)
>   inputs = ...
>   awq = AWQ(...)
> 
>   awq._on_start(layer)
>   true_outputs = layer(inputs)
>   assert awq._parent_args_cache ...
> 
>   awq._apply_smoothing(layer)
>   outputs = layer(inputs)
>   assert torch.functional.mse(outputs, true_outputs) <= 1e-10
> ```

I agree as long as the output we're checking is (semi-) deterministic I don't see why we should add an LLM to validate.

orestis-z (commented):
Thanks for this Brian! My implementation comments are gated by the architectural question from @kylesayrs https://github.com/vllm-project/llm-compressor/pull/2804#issuecomment-4642970505

PR #2806, opened 2026-06-07
fix(gptq): drop non-GROUP actorder for non-grouped strategies (#2805)

### LINKED ISSUE #2805: [Bug]: GPTQModifier writes actorder: "static" for non-group strategies, breaking vLLM serving

<details>

```text
Operating System: `Linux-5.14.0-570.107.1.el9_6.x86_64-x86_64-with-glibc2.35`
Python Version: `3.12.13 (main, Mar  4 2026, 09:23:07) [GCC 11.4.0]`
llm-compressor Version: 0.11.0
compressed-tensors Version: 0.16.0
transformers Version: 4.57.6
torch Version: 2.11.0+cu130
CUDA Devices: ['NVIDIA H200']
AMD Devices: None
NPU Devices: None

```

### 🐛 Describe the bug

</details>
I ran into this while quantizing Llama-3.2-1B-Instruct and Qwen1.5-MoE-A2.7B across four GPTQ schemes (W4A16, W8A8, FP8, NVFP4). The W4A16 and NVFP4 models serve fine on vLLM 0.22.1, but W8A8 and FP8 fail at load time with:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for VllmConfig
Value error, Must use group or tensor_group quantization strategy in order to apply
activation ordering [type=value_error, ...]
```

The issue seems to be that `GPTQModifier` defaults `actorder` to `Sentinel("static")` and writes it into `config.json` for all schemes unconditionally. That's correct for `group` and `tensor_group` strategies, but meaningless for `channel` (W8A8) and `tensor` (FP8) .  There are no groups to reorder across and so vLLM rightfully rejects the combination.

There's already a guard in `resolve_quantization_config()` ([base.py:177-186](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/base.py#L177-L186)) that resets `actorder` to `None` for non-grouped strategies, but it only checks for `actorder == ActivationOrdering.GROUP`. 

The `STATIC` and `WEIGHT` cases slip through. The same partial guard exists in `quantize_weight()` ([gptq_quantize.py:98-106](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L98-L106)).

I would be willing to take a stab at correcting this. My plan is to investigate the condition logic and possibly widen  from `actorder == GROUP` to `actorder is not None` so any actorder gets reset to `None` on non-grouped strategies. This would include me going ahead and extending the existing `test_actorder_group_falls_back_to_none` test to cover `"static"` and `"weight"`. 

Let me know if y'all would like for me to get started, if more discussion is required!

### 🛠️ Steps to reproduce

Minimal repro:

```python

    "meta-llama/Llama-3.2-1B-Instruct", dtype="auto"
)

oneshot(
    model=model,
    dataset="ultrachat_200k",
    recipe=GPTQModifier(targets="Linear", scheme="W8A8", ignore=["lm_head"]),
    max_seq_length=2048,
    num_calibration_samples=512,
)

model.save_pretrained("/tmp/test-w8a8")
# config.json now has weights.actorder="static" + weights.strategy="channel"
# vLLM rejects this at load time
```

Anai-Guo:
## Summary

Fixes #2805.

`GPTQModifier.actorder` defaults to `Sentinel("static")`, which `resolve_actorder` resolves to `ActivationOrdering.STATIC` (an alias of `WEIGHT`) for every scheme whose `actorder` is unset. The fall-back guard in `resolve_quantization_config()` then only reset `actorder` to `None` when it equalled `ActivationOrdering.GROUP`:

```python
if (
    scheme.weights.actorder == ActivationOrdering.GROUP
    and strategy not in grouped_strategies
):
    ...
    scheme.weights.actorder = None
```

So the default `STATIC` (and an explicit `WEIGHT`) **slipped through** for `channel` / `tensor` / `block` strategies and got written into `config.json`. This contradicts the comment right above the block, which already states that non-grouped strategies should *"fall back to None"*.

## Impact

Activation ordering is meaningless without groups to reorder across. compressed-tensors rejects it on non-grouped strategies on reload, so quantizing e.g. `W8A8` (channel) or `FP8` (tensor) produces models that fail to load in vLLM:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for VllmConfig
Value error, Must use group or tensor_group quantization strategy in order to
apply activation ordering
```

(reproduced by the reporter on vLLM 0.22.1 + compressed-tensors 0.16.0). `W4A16`/`NVFP4` (grouped) are unaffected and keep serving fine.

## Fix

Reset `actorder` to `None` for non-grouped strategies regardless of the ordering value, matching the comment's stated intent. The `GROUP`-incompatibility warning is preserved for the case where a user *explicitly* requested group ordering on an unsupported strategy; the harmless default `STATIC`/`WEIGHT` is dropped silently.

```python
if (
    scheme.weights.actorder is not None
    and strategy not in grouped_strategies
):
    if scheme.weights.actorder == ActivationOrdering.GROUP:
        logger.warning(...)
    scheme.weights.actorder = None
```

## Scope note

This only touches config **serialization** (`resolve_quantization_config`). The runtime guard in `gptq_quantize.py` is intentionally left unchanged: applying `WEIGHT`/`STATIC` ordering *during calibration* is valid for any strategy (it improves accuracy and restores the original column order without a format change), so it should still run — only the reloaded config must not advertise an ordering the format doesn't carry.

> Note: recent `compressed-tensors` `main` has relaxed its validator to reject only `GROUP` actorder on non-grouped strategies, so the crash above no longer reproduces on bleeding-edge installs. This fix is still warranted: it restores the documented intent, unbreaks the released vLLM + compressed-tensors stack, and stops emitting a meaningless `actorder` field.

### DIFF
--- a/src/llmcompressor/modifiers/gptq/base.py
+++ b/src/llmcompressor/modifiers/gptq/base.py
@@ -174,15 +174,22 @@ class GPTQModifier(Modifier, QuantizationMixin):
                 # Apply modifier-level actorder to already-constructed QuantizationArgs.
                 scheme.weights.actorder = resolve_actorder(scheme.weights.actorder)
 
+                # Activation ordering is only meaningful for grouped strategies;
+                # non-grouped strategies must fall back to actorder=None so the
+                # serialized config reloads cleanly (e.g. in vLLM). This applies
+                # to every ordering value, not just GROUP -- the default STATIC
+                # (alias of WEIGHT) would otherwise be written for channel/tensor
+                # schemes and rejected on reload.
                 if (
-                    scheme.weights.actorder == ActivationOrdering.GROUP
+                    scheme.weights.actorder is not None
                     and strategy not in grouped_strategies
                 ):
-                    logger.warning(
-                        f"ActivationOrdering.GROUP is not compatible with "
-                        f"strategy={strategy}; falling back to actorder=None "
-                        f"for this scheme."
-                    )
+                    if scheme.weights.actorder == ActivationOrdering.GROUP:
+                        logger.warning(
+                            f"ActivationOrdering.GROUP is not compatible with "
+                            f"strategy={strategy}; falling back to actorder=None "
+                            f"for this scheme."
+                        )
                     scheme.weights.actorder = None
         return config
 


kylesayrs (changes requested):
Weight activation ordering is still meaningful for non-group strategies: weight activation ordering has higher accuracy recovery due to better error propagation.

The issue is that these configs are not backwards compatible with vLLM <=0.22

kylesayrs:
https://github.com/vllm-project/llm-compressor/issues/2805#issuecomment-4642929520

Anai-Guo:
Thanks @kylesayrs for the explanation, and apologies @soyr-redhat for the overlap — you're right that I should have read the full issue thread before opening this. Closing in favor of the compressed-tensors 0.16.0 fix. Appreciate the context on weight activation ordering for non-grouped strategies.

PR #2812, opened 2026-06-09
Add Mistral 4 MoE calibration support

robertforbes68:
SUMMARY:

Adds `CalibrationMistral4MoE` so llm-compressor can calibrate **Mistral 4** (`transformers>=5.4`) MoE models.

**Problem.** `Mistral4MoE` (`transformers.models.mistral4.modeling_mistral4`) stores routed experts as fused 3D tensors and routes through a data-dependent loop in `Mistral4NaiveMoe.forward`:

```python
expert_hit = torch.greater(expert_mask.sum(dim=(-1, -2)), 0).nonzero()
for expert_idx in expert_hit:
    expert_idx = expert_idx[0]
    if expert_idx == self.num_experts:
        continue
    ...
```

The data-dependent `expert_hit` loop + `continue` are not `torch.fx`-traceable, so `SequentialPipeline` builds a degenerate subgraph and calibration fails with `TypeError: unsupported operand type(s) for +: 'Tensor' and 'NoneType'`.

**Fix.** `CalibrationMistral4MoE` (registered for `"Mistral4MoE"`) combines the two existing patterns in `modeling/`, because Mistral 4's MoE is a hybrid:
- like `llama4.py` / `qwen3_5_moe.py` — un-fuses the fused 3D experts into per-expert `Mistral4MLP` (`nn.Linear`) modules so `targets="Linear"` can quantize each expert;
- like `deepseek_v3.py` — a static, FX-traceable expert loop with `calibrate_all_experts` semantics + a shared expert (Mistral 4 also uses DeepSeek-V2-style group routing).

Follows the recent-arch conventions: `from __future__ import annotations` + `TYPE_CHECKING` type imports, with the real `transformers.models.mistral4` import done lazily inside `SequentialMistral4Experts.__init__`, so importing `llmcompressor.modeling` does not require a Mistral-4-capable transformers. Registered in `modeling/__init__.py`.

Notes:
- Quantizing Mistral 4 also needs `transformers>=5` to load the model (see #2608).
- Attention carries a *separate* FX-trace blocker (`apply_rotary_pos_emb` is `@use_kernel_func_from_hub`-decorated); that's handled by adding `re:.*self_attn` to the recipe `ignore` list, not by this wrapper.

TEST PLAN:

- **End-to-end validated** producing a working **NVFP4 (W4A4)** quant of a 119B Mistral 4 MoE (`darkc0de/Mistral-Small-4-119B-2603-heretic`), `transformers==5.5.3`, this wrapper, `pipeline="basic"`:
  - calibration log: `Found 36 MoE modules to replace` → `Replaced 36`;
  - `oneshot()` completed and saved a `compressed-tensors nvfp4-pack-quantized` checkpoint;
  - the resulting model **serves coherently on vLLM** — fluent output (no `!!!!`), and exact reproduction of figures/IDs (the failure mode a broken MoE calibration would cause).
- `ruff format --check` and `ruff check` pass on the new/changed files.
- No unit test added (consistent with the existing per-arch wrappers, which require the model arch + a checkpoint to exercise) — happy to add a lightweight registration test if you'd like one.

### DIFF
--- a/src/llmcompressor/modeling/__init__.py
+++ b/src/llmcompressor/modeling/__init__.py
@@ -16,6 +16,7 @@ from .glm4_moe import CalibrationGlm4MoeMoE  # noqa: F401
 from .glm4_moe_lite import CalibrationGlm4MoeLiteMoE  # noqa: F401
 from .glm_moe_dsa import CalibrationGlmMoeDsaMoE  # noqa: F401
 from .llama4 import SequentialLlama4TextMoe  # noqa: F401
+from .mistral4 import CalibrationMistral4MoE  # noqa: F401
 from .qwen3_moe import CalibrationQwen3MoeSparseMoeBlock  # noqa: F401
 from .qwen3_5_moe import CalibrationQwen3_5MoeSparseMoeBlock
 from .qwen3_vl_moe import CalibrateQwen3VLMoeTextSparseMoeBlock  # noqa: F401
--- a/src/llmcompressor/modeling/mistral4.py
+++ b/src/llmcompressor/modeling/mistral4.py
@@ -1,0 +1,212 @@
+"""
+Calibration wrapper for Mistral 4 MoE (``transformers>=5.4``).
+
+``Mistral4MoE`` (in ``transformers.models.mistral4.modeling_mistral4``) stores
+its routed experts as fused 3D tensors and routes through a data-dependent loop
+in ``Mistral4NaiveMoe.forward``::
+
+    expert_hit = torch.greater(expert_mask.sum(dim=(-1, -2)), 0).nonzero()
+    for expert_idx in expert_hit:
+        expert_idx = expert_idx[0]
+        if expert_idx == self.num_experts:
+            continue
+        ...
+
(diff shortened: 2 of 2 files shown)


brian-dellabetta:
Hi @robertforbes68 , we just landed transformers v5 support on main, removing the MoE calibration context entirely. Can you retry with the new changes and see if further work is needed to support mistral 4?

* #2647

robertforbes68:
Converting to draft. #2647 (Transformers v5) merged after this was opened and reworked the MoE-calibration subsystem (`MoECalibrationModule` → the new `modeling/moe/` `LinearExperts2D` + conversion-mapping architecture). I'll re-target this to add Mistral 4 in the new style (`moe/mistral4.py` modeled on `moe/llama4.py`; its `w1/w2/w3 → gate/down/up` expert convention matches the existing `deepseek_v4` mapping) and validate against transformers 5.x before marking it ready again. Mistral 4 still isn't supported upstream, so the contribution stands.

brian-dellabetta:
Sounds good, thanks @robertforbes68

kylesayrs:
Supported via https://github.com/vllm-project/llm-compressor/pull/2647

PR #2814, opened 2026-06-09
fix imatrix + awq

HDCharles:
Summary:

imatrix creates its hook differently from hooks_mixin so it wasn'te respecting this

kind of hacky but we're going to rewrite this anyway when we get rid of imatrixgatherer

TEST PLAN:
pytest /home/HDCharles/repos/llm-compressor/tests/llmcompressor/observers -k test_hook


kylesayrs (approved):
`¯\_(ツ)_/¯ `

brian-dellabetta (approved):
HooksMixin would probably clean this up a bit, but we don't have any usage of that inside Observer, so probably better not to subclass it

PR #2816, opened 2026-06-09
gemma4 example

HDCharles:
Summary:

previously this was fixing a bug, the bugfix was moved https://github.com/vllm-project/compressed-tensors/pull/739 but this example is still useful


src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -20,6 +20,51 @@
...
+    result = []
+    for name in ignore_list:
+        for rev in inverted:
+            src, tgt = rev.source_patterns[0], rev.target_patterns[0]
+            idx = name.find(src)
+            if idx >= 0:
+                name = name[:idx] + tgt + name[idx + len(src) :]
+        result.append(name)
+
+    return result
```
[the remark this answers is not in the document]
HDCharles: fixed

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -20,6 +20,50 @@
...
+    uses HF names, but safetensors keys use checkpoint names. This applies the
+    same reverse mapping that save_pretrained uses for weights.
+    """
+    try:
+        from transformers.conversion_mapping import (
+            PrefixChange,
+            get_model_conversion_mapping,
+        )
+    except ImportError:
+        return ignore_list
```
kylesayrs: I think this should be safe to rely on now that we're in v5 land

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -20,6 +20,50 @@
...
+
+    weight_conversions = getattr(model, "_weight_conversions", None)
+    if weight_conversions is None:
+        weight_conversions = get_model_conversion_mapping(model, add_legacy=False)
+        weight_conversions = [
+            x for x in weight_conversions if not isinstance(x, PrefixChange)
+        ]
```
kylesayrs: I know that this comes from [here](https://github.com/huggingface/transformers/blob/main/src/transformers/core_model_loading.py#L1524), but I think in this case, we still want prefix changes. We want to add "unwanted" prefixes so that it matches checkpoint format as closely as possible.

Not that I really think the `if weight_conversions is None:` will ever be hit

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -20,6 +20,50 @@
...
+    # Renamings chain — every matching one fires sequentially.
+    inverted = [conv.reverse_transform() for conv in reversed(weight_conversions)]
+
+    result = []
+    for name in ignore_list:
+        for rev in inverted:
+            renamed, matched = rev.rename_source_key(name)
+            if matched is not None:
+                name = renamed
+        result.append(name)
```
orestis-z: maybe worth hiding this logic under a separate method

orestis-z (approved):
Nice fix! I believe it would be good to have a lightweight integration test that exercises the actual bug path. Something like:

- Load a small model that has weight renamings (or mock one)
- Run oneshot with an ignore list using the HF runtime names
- Call save_pretrained
- Read the saved config.json and assert the ignore list contains checkpoint names, not HF names

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
orestis-z: could add a comment that this is expected to fail with transformers v4

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
```diff
@@ -0,0 +1,133 @@
+# Quantize gemma-4-12B-it to FP8 using RTN (round-to-nearest)
+#
+# Requires transformers >= 5.0 for correct weight name mapping in config.json.
+#
```
brian-dellabetta: can remove these now that transformers v5 PR is in

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
```diff
@@ -0,0 +1,133 @@
...
+        "re:.*embed_vision.*",
+        "re:.*embed_audio.*",
+        "re:.*vision_embedder.*",
+    ],
+)
+
+oneshot(
+    model=model,
+    recipe=recipe,
+    dataset=ds,
```
brian-dellabetta: we shouldn't need a dataset for FP8_DYNAMIC?

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
```diff
@@ -0,0 +1,133 @@
...
+    max_seq_length=MAX_SEQUENCE_LENGTH,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+    data_collator=data_collator,
+)
+
+print("\n\n")
+print("========== SAMPLE GENERATION ==============")
+dispatch_model(model)
+input_ids = torch.tensor(
```
brian-dellabetta: is this a specific prompt?

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -20,6 +20,50 @@
...
+        return ignore_list
+
+    # Match revert_weight_conversion: reverse order, then reverse each transform.
+    # Renamings chain — every matching one fires sequentially.
+    inverted = [conv.reverse_transform() for conv in reversed(weight_conversions)]
+
+    result = []
+    for name in ignore_list:
+        for rev in inverted:
+            renamed, matched = rev.rename_source_key(name)
```
brian-dellabetta: does this work with regexes?

brian-dellabetta (commented):
Hey Charles, a few comments. The ignore conversion logic is good, should we also do targets for posterity?

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
kylesayrs: We've dropped support for v4, so no need

kylesayrs (changes requested):
Agree with Brian and Orestis' reviews

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
HDCharles: yeah its the start of
```
<bos><|turn>user
The reason the sky is blue is because<turn|>
<|turn>model
<|channel>thought
<channel|>The reason the sky is blue is because of a phenomenon called **Rayleigh scattering**.

Here is the step-by-step breakdown of how it works:

### 1. Sunlight is a mix of all colors
Although sunlight looks white, it is actually made up of all the colors of the rainbow (red, orange, yellow, green, blue, indigo, and violet). Light travels as waves, and each color has a different wavelength. Red light has long,
```

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
HDCharles: yes the ignore list is resolved at this point so there's no issue, we can just match module names

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
HDCharles: thats all the logic though, the rest is just corner case checking

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
brian-dellabetta: ok, looks like you have the tokenizer loaded, might be worth switching to that or leaving a comment on why, if we want this to persist as an example

brian-dellabetta (approved):
👍

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
```diff
@@ -0,0 +1,84 @@
...
+
+oneshot(
+    model=model,
+    recipe=recipe,
+)
+
+print("\n\n")
+print("========== SAMPLE GENERATION ==============")
+dispatch_model(model)
+input_ids = torch.tensor(
```
kylesayrs: Does normal tokenization not work?

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -68,6 +109,10 @@ def save_pretrained_wrapper(
             compressor = ModelCompressor.from_pretrained_model(
                 model, quantization_format=quantization_format
             )
+            _map_ignore_to_checkpoint_names(
+                model, getattr(compressor, "quantization_config", None)
+            )
```
kylesayrs: Injecting here seems a little brittle, as it introspects `ModelCompressor`. What do you think about implementing this logic in `QuantizationConfig.from_pretrained`? That way we'll have this (model -> checkpoint) logic at the source of where it's needed

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
```diff
@@ -0,0 +1,84 @@
...
+#     --apply_chat_template --batch_size auto
+
+import torch
+import transformers
+from compressed_tensors.offload import dispatch_model
+
+if int(transformers.__version__.split(".")[0]) < 5:
+    raise RuntimeError(
+        f"transformers >= 5.0 required, found {transformers.__version__}"
+    )
```
kylesayrs: Remove

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
```diff
@@ -0,0 +1,84 @@
...
+    ]
+).to(model.device)
+output = model.generate(
+    input_ids,
+    max_new_tokens=100,
+)
+print(processor.tokenizer.decode(output[0]))
+print("==========================================\n\n")
+
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-FP8-RTN"
```
kylesayrs:
```suggestion
SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-FP8"
```

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
```diff
@@ -0,0 +1,84 @@
...
+            105,
+            4368,
+            107,
+        ]
+    ]
+).to(model.device)
+output = model.generate(
+    input_ids,
+    max_new_tokens=100,
+)
```
kylesayrs:
```suggestion
output = model.generate(input_ids, max_new_tokens=20)
```

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -20,6 +24,43 @@
...
+
+    result = []
+    for name in ignore_list:
+        for rev in inverted:
+            renamed, matched = rev.rename_source_key(name)
+            if matched is not None:
+                name = renamed
+        result.append(name)
+
+    quantization_config.ignore = result
```
kylesayrs: Noting that this is scoped to just ignores for now, since `QuantizationConfig.from_pretrained` doesn't support non-class-name scheme targets

kylesayrs (commented):
It seems like your test coverage is good, you can ignore my offline suggestion if you like (but it would still be nice to have).

My biggest suggestion is the one to move this logic into `QuantizationConfig.from_pretrained` as it will ensure that this logic exists at the source of where it's needed

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
HDCharles: I can do that

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
HDCharles: thats not enough tokens imo

examples/quantization_w8a8_fp8/gemma4_12b_fp8_rtn.py
HDCharles: this was copied from @dsikka

brian-dellabetta (approved):

kylesayrs (approved):

Issue #2808, opened 2026-06-08
AWQ GQA support


HDCharles:
# Background

for v -> o layers in GQA models like e.g. [llama3 repeats the kv heads by repeating each one](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L187).

AWQ currently just checks that [layer sizes match](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/transform/awq/base.py#L975) and ignores it if not but in theory it doesn't seem hard to extend this functionality to handle GQA.

the only issues are that when duo_scaling is enabled at either True/"both" your scales are going to differ. This could be handled by averaging or something.

so you could

1) generate scale
2) if GQA, you need to do something along the lines of:

```
scale = scale.view(num_repeats, -1).mean(dim=0).repeat_interleave(num_repeats)
```

from there i think the rest is as normal, you just need to verify the smoothing, though i think this is [already handled:](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/transform/awq/base.py#L975)

# Task

add support for GQA, probably need some way to specify GQA layers in the mapping, add above outlined functionality, test it on a few models with GQA to show that this works.

soyr-redhat:
Hey! @HDCharles, I'd be willing to take a stab at this.

My understanding of the problem:
AWQ skips v -> o layer mappings in GQA models because ```_check_layers_are_compatible()``` rejects them when ```smooth_layer.out_features != balance_layer.in_features```. For GQA, this mismatch is expected since KV heads are
repeated to match the number of attention heads.

My proposed approach:
1. Update the compatibility check to detect GQA. The size mismatch should be a clean multiple (num_attention_heads / num_key_value_heads)

2. After generating scales, average across the repeated groups and expand back like you said: 
```scale = scale.view(num_repeats, -1).mean(dim=0).repeat_interleave(num_repeats)```
3. Validate on a smaller GQA model and compare quantized model quality
  
Happy to hear any guidance or work through plan revisions before I start. Thanks!

Edwardssss:
Hi all — I ran into this while tracing why `v_proj → o_proj` mappings get dropped on GQA models, and I'm partly commenting because the issue was marked stale on 2026-09-07. Rather than let the direction lapse silently, I'd like to check whether it still has a home.

Two things worth recording, in case the thread is picked up again.

[...]

PR #2818, opened 2026-06-10
add: GQA support to AWQ

soyr-redhat:
SUMMARY:
Addresses #2808

AWQ currently skips v_proj -> o_proj smoothing for GQA models because _check_layers_are_compatible() rejects mappings where ```v_proj.out_features != o_proj.in_features```. In GQA models (Llama 3, Mistral, Qwen2, etc.), this mismatch is expected, KV heads are repeated via repeat_kv before feeding into o_proj.

This PR adds GQA support by:
- Detecting valid GQA dimension ratios in the compatibility check instead of rejecting them 
- Compressing scales from o_proj's input dimension to v_proj's output dimension at application time, by averaging across KV head repeat groups to match the repeat_kv layout
- Leaving the grid search and activation hooks unchanged, scales are computed at (hidden_dim,) for o_proj and only compressed when applying to v_proj

Fused qkv_proj with GQA is still lacking support. I'm unsure as to if this can be addressed by just knowing model dimensions.

Currently with runs, the logger outputs:

```
GQA detected for model.layers.0.self_attn.v_proj: num_repeats=n_0, head_dim=n_1
```
For every GQA layer.

TEST PLAN:
- [x] All 22 existing AWQ unit tests pass
- [x] 4 new GQA-specific unit tests pass ----- Attribution to Claude Code
- [x] End-to-end AWQ on nm-testing/tinysmokellama-3.2 (GQA, num_repeats=2, head_dim=4)
- [x] End-to-end AWQ on meta-llama/Llama-3.2-1B (GQA, num_repeats=4, head_dim=64):  RTX 3090, ~42s


soyr-redhat:
Ran a quick perplexity check on Llama 3.2 1B (wikitext-2-raw-v1, 2048 token window, 32 calibration samples):

| Config                                    | Perplexity |
|-------------------------------------------|------------|
| FP16 baseline                             | 7.25       |
| AWQ W4A16 without v_proj->o_proj smoothing | 9.07       |
| AWQ W4A16 with GQA smoothing              | 9.21       |

The results seem to be comparable, but with a 1B model, its not saying too much, will get a deployment running on h200s to test a larger model, but a 1.5% variance seems convincing.

soyr-redhat:
### Results (`acc_norm` where available, `acc` for winogrande)
  
| Model | GQA Ratio | Config | arc_easy | hellaswag | piqa | winogrande |
|-------|-----------|--------|----------|-----------|------|------------|
| **Llama 3.1 8B** | 4:1 | FP16 | 0.825 | 0.793 | 0.812 | 0.745 |
| | | AWQ (upstream) | 0.789 | 0.783 | 0.801 | 0.726 |
| | | AWQ + GQA smooth | 0.790 | 0.787 | 0.796 | 0.710 |
| **Mistral 7B v0.3** | 4:1 | FP16 | 0.802 | 0.807 | 0.821 | 0.741 |
| | | AWQ (upstream) | 0.789 | 0.799 | 0.813 | 0.729 |
| | | AWQ + GQA smooth | 0.789 | 0.802 | 0.819 | 0.731 |
| **Gemma 2 9B** | 2:1 | FP16 | 0.878 | 0.800 | 0.829 | 0.741 |
| | | AWQ (upstream) | 0.871 | 0.789 | 0.826 | 0.747 |
| | | AWQ + GQA smooth | 0.873 | 0.793 | 0.826 | 0.740 |

### Summary
  
The implemented GQA smoothing strategy seems appears to be quality-neutral across all three architectures and both GQA ratios (4:1 and 2:1). 

All differences between upstream and GQA smooth are within a good range of eachother. The new `v_proj->o_proj` smoothing path does not degrade model quality.

soyr-redhat (commented):
Reviewed Gemini Code Review

src/llmcompressor/modifiers/transform/awq/mappings.py
```diff
@@ -316,3 +316,4 @@ class ResolvedMapping:
     parent: Module
     parent_name: str
     activation_hook_target: Module | None = None
+    gqa_head_dim: int | None = None
```
brian-dellabetta: can you add this to doc string above?

brian-dellabetta (commented):
Hi @soyr-redhat , thanks for setting this up. The changes make sense, lemme get @HDCharles 's thoughts on the complexity vs improvement tradeoff here.

You expect this to be generally applicable to any GQA model, yes?

src/llmcompressor/modifiers/transform/awq/mappings.py
soyr-redhat: sure thing, on it now!

soyr-redhat:
Hey @brian-dellabetta! Thanks for taking a look. 

I personally do expect this to be generally applicable, as I ran some tests across a couple different architectures/ratios. At least as far as I'm aware, this should be agnostic since we're directly deriving the ratio from the model config!

src/llmcompressor/modifiers/transform/awq/mappings.py
soyr-redhat: resolving for now

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -977,31 +1015,52 @@ def _check_layers_are_compatible(
     smooth_name: str,
     balance_layers: list[Module],
     balance_names: list[str],
-):
+    head_dim: int | None = None,
```
HDCharles: is the head_dim needed?

you're checking

1) v.out mod o.in == 0
2) v.out mod head_dim == 0 # <-----\
3) o.in mod head_dim == 0 # <-------- aren't these both always true

also don't 1 and 3 imply 2?

if head_dim is a factor of o.in and o.in is a factor of v.out then all factors of o.in must be factors of v.out, right? 


when is just checking 1 going to be not sufficient?

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -558,11 +586,19 @@ def _smooth(
                             * scales.view(1, -1),
                         )
                     elif module == smooth_layer:
+                        smooth_scales = scales
+                        if mapping.gqa_head_dim is not None:
+                            smooth_scales = _compress_scales_for_gqa(
```
HDCharles: this can't be happening here, this has to be happening in compute_best_scales otherwise the error estimates we took will be meaningless

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles, line 660: i think you do the GQA adjustment right here, check for smooth_layer.out vs balance_layer.in get GQA factor, then take x_mean and/or w_mean and do the average + expand thing. Then all your scales will be defined correctly going forward

HDCharles:
1) the scales need to be set during the compute_best_scales step, there's not really a point to doing this unless we can actually calculate the correct output error. 

 so e.g. when you go to compute the scales for the grid search, you have to check for GQA and then do the average+expand thing when necessary.

2) I'm trying to understand if all the new logic is necessary. It seems like we can just always check for compatible dimensions A.in mod B.out == 0 and treat it like GQA and if a model can't handle that, they can edit the mappings to not include that mapping. So then can we just wait until the point where we generate the scales in compute_best_scales? I don't see why not. But i could be missing something?

HDCharles (changes requested):

src/llmcompressor/modifiers/transform/awq/base.py
soyr-redhat: Good point, theyre now moved out. The GQA averaging now happens in _compute_best_scale on x_mean/w_mean before the grid search. 

_smooth no longer calls _compress_scales_for_gqa. Since best_scales now comes out at kv_dim, we have to do an extra expand back to hidden_dim when applying to balance layers (o_proj),  the dimension mismatch is unavoidable.

src/llmcompressor/modifiers/transform/awq/base.py
soyr-redhat: I'll remove the redundant divisibility checks. The new check will just be:
balance_layer.in_features % smooth_layer.out_features == 0

head_dim would still used in the caller to moderate acceptance (if GQA is detected but no head_dim is available from model config, we skip the mapping).

Is that along the lines of what you were thinking?

src/llmcompressor/modifiers/transform/awq/base.py
soyr-redhat: Now compressing x_mean and w_mean to kv_dim right here. The grid search runs at kv_dim granularity, expanding to hidden_dim inside the loop for W*s on o_proj.

src/llmcompressor/modifiers/transform/awq/base.py
HDCharles: yeah that makes sense

src/llmcompressor/modifiers/transform/awq/mappings.py
```diff
@@ -316,3 +318,4 @@ class ResolvedMapping:
...
```
HDCharles: lets get rid of this

soyr-redhat: Alright! In doing so, _compute_best_scale and _smooth still need head_dim. Where would you prefer it gets passed in? ------ Or did you have a different idea? Happy to entertain either!

soyr-redhat: Addressed in latest commit, Removed gqa_head_dim from ResolvedMapping. GQA is now detected inline in _compute_best_scale by checking hidden_dim > kv_dim and hidden_dim % kv_dim == 0 like noted previously, and head_dim is passed down from the model config at point of use.

HDCharles: yeah i think this is fine, if its bad to do, then it shouldn't be in the mapping. We don't need any specific information ahead of time.

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -321,12 +337,18 @@ def _set_resolved_mappings(self, model: Module) -> None:
                     ]
                 )
 
-                all_compatible = _check_layers_are_compatible(
-                    smooth_layer, smooth_name, balance_layers, balance_names
+                compatible, is_gqa = _check_layers_are_compatible(
```
HDCharles: no longer a need for this right?

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -321,12 +337,18 @@ def _set_resolved_mappings(self, model: Module) -> None:
...
+                    smooth_layer,
+                    smooth_name,
+                    balance_layers,
+                    balance_names,
                 )
 
+                if is_gqa and head_dim is None:
```
HDCharles: or this

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -489,6 +511,7 @@ def _apply_smoothing(self, model: Module) -> None:
             for mapping in self._resolved_mappings
             if mapping.smooth_name in self._smooth_activation_stats
         ]
+        head_dim = self._get_head_dim(model)
```
HDCharles, line 510: or this

soyr-redhat: For this one, head_dim is still needed in _compute_best_scale for the reshape

Would it make sense to then move  _get_head_dim call into _compute_best_scale directly, or keep it in _apply_smoothing and pass it down?

src/llmcompressor/modifiers/transform/awq/base.py
soyr-redhat: I suppose you're right, getting on those changes now

src/llmcompressor/modifiers/transform/awq/base.py
soyr-redhat: +1

soyr-redhat:
@HDCharles Latest round has been addressed! Would appreciate another look if you had any availability. Thank you!

PR #2820, opened 2026-06-10
[Docs] Add MoE + Transformers V5 docs

kylesayrs:
## Purpose ##
* Explain how the new MoE loading system works
* Explain how to add support for MoE models which do not follow the `use_experts_implementation` convention

### DIFF
--- a/docs/developer-tutorials/add-moe-support.md
+++ b/docs/developer-tutorials/add-moe-support.md
@@ -1,241 +1,167 @@
-# Adding MoE Calibration Support for a New Model
+# Background
 
-Mixture of Experts (MoE) models route each token to only a subset of expert layers. This creates a calibration problem: experts that are not activated for a given token never see calibration data, which can result in poorly calibrated quantization parameters, numerical instability, or NaNs.
+## How are Mixture of Experts (MoE) models calibrated?
 
-LLM Compressor solves this by replacing MoE modules with calibration-friendly versions that route all tokens through all experts during calibration, while keeping only the routed expert outputs for the final result.
+Mixture of Experts (MoE) models route each token to only a subset of expert layers. This creates a calibration problem: Given that calibration datasets are relatively small, some experts will not be activated, or activated very infrequently. This can result in poorly calibrated quantization parameters, numerical instability, or NaNs.
 
-For background, see [Quantizing MoEs with a custom definition](../../examples/quantizing_moe/README.md#quantizing-moes-with-a-custom-definition).
+LLM Compressor and other quantization frameworks resolve this by replacing MoE modules with calibration-friendly versions that route all tokens through all experts during calibration, while keeping only the routed expert outputs for the final result.
 
-## When Do You Need This?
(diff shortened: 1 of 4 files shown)


docs/developer-tutorials/add-moe-support.md
```diff
@@ -1,241 +1,167 @@
...
-                expert_out = self.experts[i](hidden_states[token_idx])
-
-            if len(token_idx) > 0:
-                weighted_output = expert_out * router_scores[:, i][token_idx].reshape(-1, 1)
-                out[token_idx] += weighted_output
+                expert_out = self[i](inputs.split(expert_size, dim=0)[i])
+            output_list.append(expert_out)
 
-        return out, router_logits
+        return torch.cat(output_list, dim=0)
```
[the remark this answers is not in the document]
kylesayrs: You don't have enough context

docs/developer-tutorials/add-moe-support.md
```diff
@@ -1,241 +1,167 @@
...
-from transformers.models.your_model.modeling_your_model import YourModelMoE
-print(YourModelMoE.__name__)  # e.g. "YourModelMoE"
-```
+Adding a conversion mapping is the most efficient way to load your model. For models which do not have conversion mappings, they will fallback to performing conversion after loading.
 
-### Step 2: Determine whether experts are packed
+## MoE Conversion After Loading
 
-Inspect the original MoE module to see how experts are stored:
+The vast majority of models can be converted after loading. Conversion after loading is often slow, since it may require converting from 2D -> 3D and then back to 2D. However, it is guaranteed to work for nearly all MoE model definitions.
```
brian-dellabetta, line 44: > Conversion after loading is often slow

Is this is a significant overhead w.r.t. GPTQ or AWQ? i.e. does conversion cause load time to triple, but that ultimately only adds a few minutes to a ~1hr run?

docs/developer-tutorials/add-moe-support.md
```diff
@@ -1,241 +1,167 @@
...
-from transformers.models.your_model.modeling_your_model import YourModelMoE as OriginalYourModelMoE
+class GraniteMoeLinearExperts(LinearExperts2D):
+    is_concatenated = False
+    is_transposed = False
+    has_bias = False
+    has_gate = False
+
+    @classmethod
+    @torch.no_grad()
+    def from_experts_module(
```
brian-dellabetta, line 63: consider moving this below init, or changing the order you have above after `This definition should specify an __init__ method ...`

brian-dellabetta (approved):
Docs look great! 🤩

docs/developer-tutorials/add-moe-support.md
kylesayrs: From my experience implementing these, it's better to start with the init function, then from_experts_module. However, when following code conventions, class methods should be stated first.

docs/developer-tutorials/add-moe-support.md
kylesayrs: I measured the load time of DSV4-Flash + disk offloading to be ~14 minutes. DSV4-Pro is about 4x bigger, so I would estimate the load time to be ~1hr.

kylesayrs: Smaller models (~100B) which fit in CPU memory typically take ~2 minutes on our machines.

HDCharles (approved):

PR #2823, opened 2026-06-10
[Bugfix] Fix KV cache tests with transformers v5

kylesayrs:
## Purpose ##
* Fix KV Cache tests when using `init_empty_weights` with `accelerate==1.6.0` and `transformers>=5.9.0`
  * Newer versions of transformers attach an attribute `_is_hf_initialized` to parameters. However, `accelerate::init_empty_weights` treats this as a class constructor param. When it tries to reconstruct the parameter with this argument, the constructor fails.
  * This is [fixed by upstream accelerate](https://github.com/huggingface/accelerate/pull/3955), but we can avoid bumping versions by instead just not using `init_empty_weights` and instead loading on the meta device instead.

## Changes ##
* Remove all uses of `init_empty_weights`, instead load on meta device

### DIFF
--- a/tests/llmcompressor/transformers/kv_cache/test_kv_cache.py
+++ b/tests/llmcompressor/transformers/kv_cache/test_kv_cache.py
@@ -2,7 +2,6 @@ import os
 from pathlib import Path
 
 import pytest
-from accelerate import init_empty_weights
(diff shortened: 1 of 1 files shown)


brian-dellabetta (approved):
👍

PR #2825, opened 2026-06-11
[MoE] Standardize MoE identification

kylesayrs:
## Purpose ##
* Standardize on a single source of truth for detecting MoE models

### DIFF
--- a/src/llmcompressor/modeling/moe/helpers.py
+++ b/src/llmcompressor/modeling/moe/helpers.py
@@ -15,6 +15,21 @@ from llmcompressor.sentinel import Sentinel
 from llmcompressor.typing import TorchModuleProtocol
 
 
+def is_moe_model(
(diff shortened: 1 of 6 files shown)


src/llmcompressor/modeling/moe/helpers.py
```diff
@@ -15,6 +15,21 @@
 from llmcompressor.typing import TorchModuleProtocol
 
 
+def is_moe_model(
+    model: torch.nn.Module, check_fused: bool = True, check_unfused: bool = True
+) -> bool:
+    from .linear_experts import LinearExperts2D
```
HDCharles, line 21:
```suggestion
    from .linear_experts import LinearExperts2D
    if not (check_fused or check_unfused):
       raiseValueError("is_moe_model was invoked with both check_fused and check_unfused as False, one must be true")
```

HDCharles (approved):
beyond that comment, looks good.

PR #2826, opened 2026-06-12
[Calibration] Ensure all experts are calibrated

dsikka:
## Summary

**Problem:** `moe_calibrate_all_experts` (default `True`) was accepted as a parameter in `DatasetArguments` and `oneshot()` but was never wired up — `get_calibrate_all_experts_flag()` always returned `False`. As a result, MoE experts only received their routed tokens during calibration. Experts that were never activated by the calibration dataset produced NaN `input_global_scale` values (empty activation statistics → undefined max → NaN scale), causing corrupted quantization for those experts.

**Fix:** Replaced the unused global flag mechanism with a `calibrate_all_experts` attribute stored directly on each `LinearExperts2D` module. The flag is now threaded from `DatasetArguments.moe_calibrate_all_experts` → `linearize_moe()` → `from_experts_module()` → `self.calibrate_all_experts`, and read in each module's `forward()`. This ensures the flag is correctly set at linearization time and requires no global mutable state.

**Changes:**
- `linear_experts.py`: `LinearExperts2D.__init__` and `from_experts_module` accept `calibrate_all_experts`; `forward` uses `self.calibrate_all_experts` instead of the global flag
- `granitemoe.py` / `llama4.py`: Same treatment for their `from_experts_module` overrides and `forward` methods; removed `get_calibrate_all_experts_flag` imports
- `linearize.py`: `linearize_moe` and `load_quantizable_moe` accept and thread `calibrate_all_experts` through to `from_experts_module`
- `oneshot.py`: Passes `self.dataset_args.moe_calibrate_all_experts` to `linearize_moe`

**Validation:** Re-ran W4A4 FP4 quantization on Qwen3.6-35B-A3B. Before this fix, 255 of 30,880 `input_global_scale` tensors were NaN (83 unactivated MoE experts × 3 projections). After this fix, all 30,880 `input_global_scale` tensors are valid (range: 5.16–4544, no NaNs, zeros, or Infs).

### DIFF
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -223,7 +223,10 @@ class Oneshot:
                 "model `with llmcompressor.modeling.moe.linearize.load_quantizable_moe`"
                 " before passing to `oneshot`. Falling back to post-load linearization."
             )
-            linearize_moe(self.model)
+            linearize_moe(
+                self.model,
+                calibrate_all_experts=self.dataset_args.moe_calibrate_all_experts,
+            )
 
         # (Helen INFERENG-661): validate recipe modifiers before initialization
         # Apply calibration contexts for the entire calibration process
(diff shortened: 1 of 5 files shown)


src/llmcompressor/entrypoints/oneshot.py
kylesayrs, line 233:
```suggestion
        with norm_calibration_context(self.model), moe_calibration_context():
```

kylesayrs (changes requested):
Ugh. I somehow missed to add this. The correct solution is to add the above suggestion. Adding the flag to the modules themselves makes it impossible to run sample generation at the end without sending to all tokens.

kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/2827

PR #2827, opened 2026-06-12
[MoE] [Bugfix] Add missing `moe_calibration_context`

kylesayrs:
## Purpose ##
* Fix calibration of all experts
  * Need to enter `moe_calibration_context` during oneshot. In the future, I'd like to add this to `calibration_forward_context` and deprecate the `moe_calibrate_all_experts` argument
  * I'm not sure why this error was never seen by any of the other NVFP4/ GPTQ tests

## Changes ##
* Added `moe_calibration_context` to oneshot call

## Testing ##
* Added `tests/llmcompressor/modeling/test_linear_experts.py` which tests that the `LinearExperts2D` class properly sends all tokens to all experts in context
* Added `tests/llmcompressor/modeling/test_moe_context.py` which asserts that all expert qparams get calibrated after `oneshot` call (finishes in 3.6s)

* Ran `examples/quantization_w4a4_fp4/qwen_30b_a3b.py`
```
[transformers] The attention mask is not set and cannot be inferred from input because pad token is same as eos token. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
Hello my name is John and I'm 30 years old. I'm from the USA. I have a B.S. in computer science. I'm looking for a job in the software industry. I have experience in Java, Python, and C++. I have a few projects that I've worked on. I'm looking for a job in the software industry. I'm looking for a job in the software industry. I'm looking for a job in the software industry. I'm looking for a job in the
```

### DIFF
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -10,6 +10,7 @@ with various pipeline configurations for efficient model optimization.
 from __future__ import annotations
 
 import os
+from contextlib import ExitStack
 from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Callable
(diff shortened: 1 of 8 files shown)


dsikka (changes requested):
We need a test to make sure that it is True by default. We’re calling the context in the tests which is why they passed but that wasn’t reflective of our examples

brian-dellabetta (approved):
LGTM

HDCharles (approved):

dsikka (approved):

PR #2828, opened 2026-06-12
Add REAP pruning modifier

eldarkurtic:
This PR addresses @HDCharles comments from https://github.com/vllm-project/llm-compressor/pull/2703 but I am opening a new branch because of forced push made a lot of mess in the original PR: https://github.com/vllm-project/llm-compressor/pull/2703

@Ryfernandes I will close that branch, and please feel free to take over upgrade to transformers v5 from this branch.

New changes (relative to the old PR):
- Lifecycle: collect saliency during sequential calibration, finalize the
  drop decision per layer at SEQUENTIAL_EPOCH_END (and free that layer's
  activations then), apply the structural prune at on_finalize. Works for both
  "permanent" and "restored" MoE calibration wrappers and survives the pipeline's
  error-propagation second pass.
- Correct routing per architecture: read the router's actual output instead of
  assuming softmax. Modes: softmax (Qwen3 family), tuple indices/weights
  (DeepSeek-V3, GLM4), group-limited sigmoid (GLM-MoE-DSA / GLM4-MoE-Lite),
  scores+logits (Llama4), scores+experts (AfMoE).
- Group-limited routers (DeepSeek/GLM): prune equally per group so routing stays
  valid, resize e_score_correction_bias, update n_routed_experts on router+config.
- Fail fast if the requested sparsity leaves fewer experts than top_k.
- Offload-safe pruning; vectorized, on-device saliency (no per-step CPU syncs).
- Reject unsupported archs with a clear error: Mixtral (no calibration wrapper),
  Gemma4 (router lives outside the experts block).

Tested: GSM8k, 5-shot, vLLM; calib = 512 OpenPlatypus samples, seq 2048.
```
  Model                         Config        flexible   strict
  ----------------------------  ------------  --------   ------
  Qwen3-30B-A3B (softmax,128e)  baseline       0.8802    0.8741
                                REAP 25% ->96  0.8999    0.8923
                                REAP 50% ->64  0.8711    0.8643
  Moonlight-16B-A3B (DeepSeek   baseline       0.7627    0.7566
  -V3 arch, sigmoid, 64e)       REAP 25% ->48  0.6869    0.6831
                                REAP 50% ->32  0.2449    0.2426
```

- Qwen3-30B matches the originally reported numbers at both sparsities -> no
  regression in the softmax path; full lifecycle works end-to-end on a real 30B.
- Moonlight validates the new sigmoid/top-k path (incl. correction-bias resize).
  Smooth monotonic degradation (0.76 -> 0.69 @25% -> 0.24 @50%) = correctly
  working pruner; the steep 50% drop is genuine sparsity-sensitivity of a small
  MoE (few experts, little redundancy), not a bug.
- Practical range: ~25% sparsity is comfortable across families; 50% is fine on
  large expert-rich MoEs (Qwen3) but too aggressive for small ones (Moonlight).

### DIFF
--- a/examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
+++ b/examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
@@ -1,0 +1,76 @@
+"""
+REAP Expert Pruning: Qwen3-30B-A3B at 25% compression.
+
+Usage:
+    CUDA_VISIBLE_DEVICES=0,1,2,3 uv run python examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
+"""
+
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+
+# ── Configuration ──────────────────────────────────────────────────────────
+
+MODEL_ID = "/home/eldarkurtic/hf_models/Qwen/Qwen3-30B-A3B-Instruct-2507"
+OUTPUT_DIR = "/home/eldarkurtic/github/eldarkurtic/llm-compressor/output_dir/Qwen3-30B-A3B-Instruct-2507_REAP_sp25"
+COMPRESSION_RATIO = 0.25
+NUM_CALIBRATION_SAMPLES = 512
+MAX_SEQ_LENGTH = 2048
+DATASET = "open_platypus"
+
+# ── Recipe ─────────────────────────────────────────────────────────────────
+
+RECIPE = """
+pruning_stage:
+  pruning_modifiers:
+    REAPPruningModifier:
+      compression_ratio: 0.25
+"""
+
+# ── Run pruning ────────────────────────────────────────────────────────────
+
+print(f"Model:             {MODEL_ID}")
+print(f"Compression ratio: {COMPRESSION_RATIO}")
+print(f"Output:            {OUTPUT_DIR}")
+print(f"Calibration:       {NUM_CALIBRATION_SAMPLES} samples, {MAX_SEQ_LENGTH} max seq len")
+print()
+
+oneshot(
+    model=MODEL_ID,
+    recipe=RECIPE,
+    dataset=DATASET,
+    output_dir=OUTPUT_DIR,
+    num_calibration_samples=NUM_CALIBRATION_SAMPLES,
+    max_seq_length=MAX_SEQ_LENGTH,
+)
+
+print(f"\nDone! Pruned model saved to: {OUTPUT_DIR}")
+
+# ── Quick smoke test ───────────────────────────────────────────────────────
+
+print("\nRunning generation smoke test...")
+
+tokenizer = AutoTokenizer.from_pretrained(OUTPUT_DIR)
+model = AutoModelForCausalLM.from_pretrained(
+    OUTPUT_DIR,
+    torch_dtype=torch.bfloat16,
+    device_map="auto",
+)
+model.eval()
+
+prompt = "Write a Python function that checks if a number is prime."
+messages = [{"role": "user", "content": prompt}]
+text = tokenizer.apply_chat_template(
+    messages, tokenize=False, add_generation_prompt=True, enable_thinking=False
+)
+inputs = tokenizer(text, return_tensors="pt").to(model.device)
+
+with torch.no_grad():
+    output_ids = model.generate(**inputs, max_new_tokens=1000, do_sample=False)
+
+new_tokens = output_ids[0][inputs["input_ids"].shape[1] :]
+response = tokenizer.decode(new_tokens, skip_special_tokens=True)
+
+print(f"\nPrompt: {prompt}")
+print(f"Response: {response}")
--- a/examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
+++ b/examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
@@ -1,8 +1,8 @@
 """
-REAP Expert Pruning: Qwen3-30B-A3B at 25% compression.
+REAP Expert Pruning: Qwen3-30B-A3B at 25% sparsity.
 
 Usage:
-    CUDA_VISIBLE_DEVICES=0,1,2,3 uv run python examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
+CUDA_VISIBLE_DEVICES=0,1 python examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
 """
 
 import torch
@@ -10,31 +10,22 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
 
-# ── Configuration ──────────────────────────────────────────────────────────
-
-MODEL_ID = "/home/eldarkurtic/hf_models/Qwen/Qwen3-30B-A3B-Instruct-2507"
-OUTPUT_DIR = "/home/eldarkurtic/github/eldarkurtic/llm-compressor/output_dir/Qwen3-30B-A3B-Instruct-2507_REAP_sp25"
-COMPRESSION_RATIO = 0.25
+MDL = "Qwen/Qwen3-30B-A3B-Instruct-2507"
+SPARSITY = 0.25
+BASE_PATH = "/home/eldarkurtic/"
+OUT_PATH = "github/eldarkurtic/llm-compressor/output_dir/"
+MODEL_ID = BASE_PATH + "hf_models/" + MDL
+OUTPUT_DIR = BASE_PATH + OUT_PATH + MDL + "_REAP_sp" + str(int(SPARSITY * 100))
 NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQ_LENGTH = 2048
 DATASET = "open_platypus"
-
-# ── Recipe ─────────────────────────────────────────────────────────────────
 
 RECIPE = """
 pruning_stage:
   pruning_modifiers:
     REAPPruningModifier:
-      compression_ratio: 0.25
-"""
-
-# ── Run pruning ────────────────────────────────────────────────────────────
-
-print(f"Model:             {MODEL_ID}")
-print(f"Compression ratio: {COMPRESSION_RATIO}")
-print(f"Output:            {OUTPUT_DIR}")
-print(f"Calibration:       {NUM_CALIBRATION_SAMPLES} samples, {MAX_SEQ_LENGTH} max seq len")
-print()
+      sparsity: {SPARSITY}
+""".format(SPARSITY=SPARSITY)
 
 oneshot(
     model=MODEL_ID,
@@ -46,9 +37,6 @@ oneshot(
 )
 
 print(f"\nDone! Pruned model saved to: {OUTPUT_DIR}")
-
-# ── Quick smoke test ───────────────────────────────────────────────────────
-
 print("\nRunning generation smoke test...")
 
 tokenizer = AutoTokenizer.from_pretrained(OUTPUT_DIR)
--- a/examples/reap_expert_pruning/reap_qwen3_30b_50pct.py
+++ b/examples/reap_expert_pruning/reap_qwen3_30b_50pct.py
@@ -3,35 +3,29 @@ REAP Expert Pruning: Qwen3-30B-A3B at 25% sparsity.
 
 Usage:
 CUDA_VISIBLE_DEVICES=0,1 python examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
-
-The model is loaded from the Hugging Face Hub by default. Override the source
-model or output location with the MODEL_ID / OUTPUT_DIR environment variables.
 """
-
-import inspect
-import os
 
 import torch
 from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from llmcompressor import oneshot
 
-MODEL_ID = os.environ.get("MODEL_ID", "Qwen/Qwen3-30B-A3B-Instruct-2507")
-SPARSITY = 0.25
-OUTPUT_DIR = os.environ.get(
-    "OUTPUT_DIR",
-    f"{MODEL_ID.split('/')[-1]}_REAP_sp{int(SPARSITY * 100)}",
-)
+MDL = "Qwen/Qwen3-30B-A3B-Instruct-2507"
+SPARSITY = 0.5
+BASE_PATH = "/home/eldarkurtic/"
+OUT_PATH = "github/eldarkurtic/llm-compressor/output_dir/"
+MODEL_ID = BASE_PATH + "hf_models/" + MDL
+OUTPUT_DIR = BASE_PATH + OUT_PATH + MDL + "_REAP_sp" + str(int(SPARSITY * 100))
 NUM_CALIBRATION_SAMPLES = 512
 MAX_SEQ_LENGTH = 2048
 DATASET = "open_platypus"
 
-RECIPE = f"""
+RECIPE = """
 pruning_stage:
   pruning_modifiers:
     REAPPruningModifier:
       sparsity: {SPARSITY}
-"""
+""".format(SPARSITY=SPARSITY)
 
 oneshot(
     model=MODEL_ID,
@@ -55,11 +49,9 @@ model.eval()
 
 prompt = "Write a Python function that checks if a number is prime."
 messages = [{"role": "user", "content": prompt}]
-chat_kwargs = {"tokenize": False, "add_generation_prompt": True}
-# enable_thinking is only accepted by some Qwen tokenizers
-if "enable_thinking" in inspect.signature(tokenizer.apply_chat_template).parameters:
-    chat_kwargs["enable_thinking"] = False
-text = tokenizer.apply_chat_template(messages, **chat_kwargs)
+text = tokenizer.apply_chat_template(
+    messages, tokenize=False, add_generation_prompt=True, enable_thinking=False
+)
 inputs = tokenizer(text, return_tensors="pt").to(model.device)
 
 with torch.no_grad():
--- a/src/llmcompressor/modifiers/pruning/reap/base.py
+++ b/src/llmcompressor/modifiers/pruning/reap/base.py
@@ -8,7 +8,6 @@ from functools import partial
 from typing import Any
 
 import torch
-import torch.nn.functional as F
 from loguru import logger
 from pydantic import Field, PrivateAttr, model_validator
 
@@ -17,9 +16,14 @@ from llmcompressor.modifiers import Modifier
 from llmcompressor.modifiers.pruning.reap.utils import (
     MoEModelAttrs,
     REAPSaliencyTracker,
+    assert_routing_feasible,
+    compute_retained_experts,
     detect_moe_attrs,
+    extract_routing,
     find_moe_layers,
     get_num_experts,
+    get_router_num_groups,
+    get_router_topk_group,
     prune_moe_layer,
     update_model_config,
 )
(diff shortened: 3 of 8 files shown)


src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+            )
+            return attrs
+
+    for _, module in model.named_modules():
+        router = _find_router_attr(module)
+        if router is not None and hasattr(module, "experts"):
+            attrs = MoEModelAttrs(
+                router_attr=router,
+                experts_attr="experts",
+                num_experts_config_key="num_experts",
```
Ryfernandes, line 160: Just pointing out that we don't validate the existence of this field in the same manner as "experts," but it is only used to update the number of experts in the config at the end, so leaving out a potentially blocking validation seems correct

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+    if hasattr(experts, "num_experts"):
+        return experts.num_experts
+
+    for _, param in experts.named_parameters(recurse=False):
+        if param.dim() >= 2:
+            return param.shape[0]
+
+    raise ValueError(
+        f"Cannot determine number of experts from {module.__class__.__name__}"
+    )
```
Ryfernandes, line 217: In the transformers v5 update to main, all models should be loaded in a 2D-experts format. Thus, these extra checks should be removed, and the value error should point out that the model loaded in an unexpected format

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+            attrs = MoEModelAttrs(
+                router_attr=router,
+                experts_attr="experts",
+                num_experts_config_key="num_experts",
+                routing_mode=SOFTMAX,
+            )
+            logger.info(
+                f"Auto-detected MoE block: {module.__class__.__name__} "
+                f"(router={router}, routing={SOFTMAX})"
+            )
```
Ryfernandes, line 166: Should also add the expert attribute to this log, like above

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+    Number of expert groups used by group-limited routers (DeepSeek-V3, GLM4,
+    GLM-MoE-DSA). Returns 1 for routers without grouping. ``n_group`` may live on
+    the router module or on the (calibration) block.
+    """
+    router = getattr(module, attrs.router_attr)
+    for holder in (router, module):
+        n_group = getattr(holder, "n_group", None)
+        if isinstance(n_group, int) and n_group > 0:
+            return n_group
+    return 1
```
Ryfernandes, line 231: Should add a logger.info() for the n_group value used, to help users debug if incorrectly detected

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+            )
+        else:
+            # fail fast (before calibration) if the requested sparsity would
+            # leave the router unable to select top_k experts per token
+            assert_routing_feasible(
+                num_experts,
+                self._n_experts_to_drop,
+                self._n_group,
+                get_router_topk_group(sample_module, self._attrs),
+                self._top_k,
```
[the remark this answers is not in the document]
Ryfernandes: All models that we have tried to calibrate have uniform topology, but it would make sense to add a uniformity check that raises a ValueError if there is inhomogeneity between layers. Compared to per-layer support, this seems like small overhead to protect users from silent failures

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+            module = model.get_submodule(layer_name)
+
+            if not hasattr(module, "calibrate_all_experts"):
+                raise RuntimeError(
+                    f"REAP requires the MoE block '{layer_name}' "
+                    f"({module.__class__.__name__}) to support all-expert "
+                    "calibration, but it has no 'calibrate_all_experts' attribute. "
+                    "Ensure the model runs inside moe_calibration_context."
+                )
+            module.calibrate_all_experts = True
```
Ryfernandes, line 156: Need to confirm with @kylesayrs how it is determined whether MoE calibration context is used in oneshot(). I believe that setting this within the modifier is outdated, but maybe a check with get_calibrate_all_experts_flag would be appropriate for a RuntimeError

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+
+            num_experts = get_num_experts(module, self._attrs)
```
Ryfernandes, line 158: R.e. the CodeRabbit comment above, we should decide if we want to enforce MoE layer homogeneity or support non-homogenous layers for REAP. If we go with the former, we can just have an attribute of the class that stores num_experts for all layers

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+            return
+
+        with torch.no_grad():
+            topk_indices, topk_weights = extract_routing(
+                router_output, module, self._attrs, self._top_k
+            )
+            expert_norms = torch.stack(
+                [norm_buffer[i] for i in range(tracker.num_experts)], dim=1
+            )
+            tracker.update(topk_indices, topk_weights, expert_norms)
```
Ryfernandes, line 328: Based on the original [REAP paper](https://arxiv.org/pdf/2510.13999), the only L2 norm expert outputs that are used are the outputs of the expert when it is selected by the gate/router. Because of this, calibrate_all_experts behavior shouldn't be necessary. Here, it is convenient for vectorized accumulation operations (stacking all expert output norms into a tensor of [num_tokens, num_experts]). However, given the tradeoff between extra forward passes and this clean, vectorized operation, it might be more efficient to forego calibrate_all_experts, and deal with the expert outputs in another way.

In a non-calibrate_all_experts implementation, the norm buffer would be treated as a queue, where as tokens are processed by an expert, their L2 norms are added to the queue. Then, instead of doing expert_norms.gather (reap/utils.py, line 390), the accumulator would need to iterate through the topk_indices in token order, dequeuing from the right expert queue based on the index. This would still ensure that the correct norms get multiplied by the correct router weights. I'm not sure if a torch implementation of this tensor operation exists as a replacement for .gather(), but if not, a custom implementation could be used that I believe would still be cheaper than the extra expert forward passes.

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+    if n_experts_to_drop >= num_experts:
+        raise ValueError(
+            f"Cannot drop {n_experts_to_drop} experts from a layer with only "
+            f"{num_experts} experts"
+        )
+
+    if n_group <= 1:
+        _, drop_idx = torch.topk(saliency, n_experts_to_drop, largest=False)
+        drop_set = set(drop_idx.tolist())
+        return sorted(i for i in range(num_experts) if i not in drop_set)
```
Ryfernandes, line 441: Is sorted necessary here? Would list comprehension be enough since range() should always be sorted

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+        )
+
+    retained: list[int] = []
+    for g in range(n_group):
+        lo = g * group_size
+        grp = saliency[lo : lo + group_size]
+        _, drop_local = torch.topk(grp, drop_per_group, largest=False)
+        drop_set = {lo + int(i) for i in drop_local.tolist()}
+        retained.extend(i for i in range(lo, lo + group_size) if i not in drop_set)
+    return sorted(retained)
```
Ryfernandes, line 467: Again, I think that this sorted() is redundant

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+
+            retained = compute_retained_experts(
+                tracker.mean_saliency, self._n_experts_to_drop, self._n_group
+            )
+            self._prune_decisions[layer_name] = retained
+
+            # free this layer's accumulators / buffers now
+            del self._saliency_trackers[layer_name]
+            self._norm_buffers.pop(layer_name, None)
+            self._routing_cache.pop(layer_name, None)
```
Ryfernandes, line 267: I see this comes from @HDCharles original comments on the PR, but can we also execute the pruning here? If the layer will no longer be used again and there is no restore() behavior with the new moe_calibration_context in v5, I believe that we should be able to prune here, unless it would affect other modifiers

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+
+    # -- decision finalization ----------------------------------------------
+
+    def _finalize_decisions(self):
+        """Finalize drop decisions for any tracked layer whose saliency is
+        complete, then release its activation buffers."""
+        if self._n_experts_to_drop == 0:
+            return
+
+        for layer_name, tracker in list(self._saliency_trackers.items()):
```
Ryfernandes, line 253: @kylesayrs Can forward passes for multiple layers be computed in each SEQUENTIAL_EPOCH, or is it always just one layer at a time? If it is just one, then we don't need multiple saliency trackers nor norm buffers. We can just have a single one that is cleared and reused between layers. If multiple layers can be processed in each SEQUENTIAL_EPOCH, I agree that this is the best approach.

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+    attr_name = attrs.num_experts_config_key
+
+    updated = False
+    for cfg in (config, getattr(config, "text_config", None)):
+        if cfg is not None and hasattr(cfg, attr_name):
+            old_val = getattr(cfg, attr_name)
+            setattr(cfg, attr_name, new_num_experts)
+            logger.info(f"Updated config.{attr_name}: {old_val} -> {new_num_experts}")
+            updated = True
+            break
```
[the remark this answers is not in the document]
Ryfernandes: I agree here. The break should be removed, just in case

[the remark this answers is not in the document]
Ryfernandes: ^^ See above

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+
+    if not updated:
+        logger.warning(
+            f"Config attribute '{attr_name}' not found, skipping config update"
+        )
```
Ryfernandes, line 618: How big of an issue is it if this is skipped. Would other codebases rely on num_experts in the config? If so, this should be checked as I pointed out in the start of the compression process, to ensure that the new value can be written at the end

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+    """
+    moe_block = model.get_submodule(layer_name)
+    router = getattr(moe_block, attrs.router_attr)
+    experts = getattr(moe_block, attrs.experts_attr)
+
+    if isinstance(experts, nn.ModuleList):
+        old_experts = experts
+        new_experts = nn.ModuleList([experts[i] for i in retained])
+        setattr(moe_block, attrs.experts_attr, new_experts)
+        del old_experts
```
Ryfernandes, line 532: Here, would it be better to just del all of the experts that need to be removed from the original ModuleList, rather than creating a new one?

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+    else:
+        _prune_fused_experts(experts, retained)
```
Ryfernandes, line 534: No longer needed, transformers v5. Should just raise a RunTime error for a non-linearized model

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: From a conversation with @HDCharles , it seems that as long as other modifier will not hit errors due to the pruning, it makes most sense to do the pruning in SEQUENTIAL_EPOCH, so that later quantization modifiers run on the pruned model. However, this will require additional insight on if this will cause errors and/or testing with some multi-step recipes involving REAP

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: From a conversation with @HDCharles , it is best to leave the multiple saliency tracker behavior as-is, since multiple layers can be processed during a single SEQUENTIAL_EPOCH

examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
```diff
@@ -0,0 +1,72 @@
...
+inputs = tokenizer(text, return_tensors="pt").to(model.device)
+
+with torch.no_grad():
+    output_ids = model.generate(**inputs, max_new_tokens=1000, do_sample=False)
+
+new_tokens = output_ids[0][inputs["input_ids"].shape[1] :]
+response = tokenizer.decode(new_tokens, skip_special_tokens=True)
+
+print(f"\nPrompt: {prompt}")
+print(f"Response: {response}")
```
Ryfernandes, line 72: @HDCharles pointed out that this is somewhat non-standard among our examples for a smoke test. Should we change this to just use dispatch_model() from compressed tensors and do a single prompt genration?

examples/reap_expert_pruning/reap_qwen3_30b_25pct.py
```diff
@@ -0,0 +1,72 @@
...
+  pruning_modifiers:
+    REAPPruningModifier:
+      sparsity: {SPARSITY}
+"""
+
+oneshot(
+    model=MODEL_ID,
+    recipe=RECIPE,
+    dataset=DATASET,
+    output_dir=OUTPUT_DIR,
```
Ryfernandes, line 40: It seems that other examples save the model manually with .save_pretrained(). Is that preferred over this?

examples/reap_expert_pruning/reap_qwen3_30b_50pct.py
```diff
@@ -0,0 +1,64 @@
+"""
+REAP Expert Pruning: Qwen3-30B-A3B at 25% sparsity.
+
+Usage:
```
Ryfernandes, line 4: Is it worth even having this as a separate example? It seems to be the exact same code as 25pct, just with the SPARSITY parameter changed. The handling of output directories and model ID's is better in the 25pct example. In my opinion, this file should be removed.

Ryfernandes (commented):
Overall, this seems close to ready in my opinion, although there are two main areas where I would like further opinions.

First of all, rather than deciding pruning in SEQUENTIAL_EPOCH and executing it in on_finalize(), would it be better to also execute pruning in SEQUENTIAL_EPOCH, since restore() behavior of MoE wrappers is no longer a concern in transformers v5?

Secondly, it seems much more efficient to me to not use the moe_calibration_context, since the saliency scores only use the output norms from the top_k experts. It would require some reconfiguration and de-vectorization of part of the accumulation operation, but it seems clear to me that this would still be much faster, considering the overhead of unused expert forward passes. @kylesayrs , with the new moe_calibration_context implementation, is it supported for individual modifiers to choose whether or not to use it? If not, is it safe/supported to override the _CALIBRATE_ALL_EXPERTS global flag within the modifier, and/or is it worth making changes to support the configurability of the context within modifiers?

Would appreciate thoughts on other comments that I made, whether from @eldarkurtic or others before going ahead with changes, but I feel like the rest are fairly standard.

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+                    f"REAP could not register per-expert hooks for '{layer_name}': "
+                    f"the experts module ({experts.__class__.__name__}) has no "
+                    "child modules. Fused experts are not supported."
+                )
+
+            self.register_hook(
+                module, partial(self._moe_post_hook, layer_name), "forward"
+            )
+
+    def on_event(self, state: State, event: Event, **kwargs):
```
brian-dellabetta, line 188: @Ryfernandes just FYI, we recently merged in #2783 which changes the modifier API to no longer use `on_event`. We'll have to propagate those changes to this new modifier

src/llmcompressor/modifiers/pruning/reap/base.py
brian-dellabetta: yeah, check out the `moe_calibrate_all_experts` flag in oneshot.

src/llmcompressor/modifiers/pruning/reap/base.py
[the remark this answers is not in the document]
brian-dellabetta: a validation error seems appropriate to me for now, if we don't want to worry about adding support for heterogeneous MoEs now. that way we still catch it right away and avoid undesired behavior or downstream errors.

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Got it, is it supported at all to set this within a recipe, or should it just be set from oneshot? Should modifiers support behavior for both True and False, or is it valid to have a check that fast fails if moe_calibrate_all_experts is not set as desired?

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Got it. Looks like it should be simple enough to just use on_calibration_start, on_calibration_end, and on_sequential_epoch_end as replacements for these three conditionals

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,331 @@
...
+            f"{self._n_experts_to_drop} ({self.sparsity:.0%})"
+            + (f", n_group={self._n_group}" if self._n_group > 1 else "")
+        )
+
+        return True
+
+    def on_start(self, state: State, event: Event, **kwargs):
+        self.started_ = True
+        model = state.model
+
```
kylesayrs, line 145:
```suggestion
current_session().dataset_args.calibrate_all_experts = False
# + add warning
```

kylesayrs: Restore on end

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+            continue
+        if any(re.search(pattern, name) for pattern in ignore):
+            continue
+        moe_layers[name] = module
+
+    logger.info(f"Found {len(moe_layers)} MoE layers")
+    return moe_layers
+
+
+def get_num_experts(module: nn.Module, attrs: MoEModelAttrs) -> int:
```
kylesayrs, line 203: consider using https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/moe/helpers.py#L118

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,618 @@
...
+
+def _find_router_attr(module: nn.Module) -> str | None:
+    for attr_name in ("gate", "router"):
+        val = getattr(module, attr_name, None)
+        if val is not None and isinstance(val, nn.Linear):
+            return attr_name
+    return None
+
+
+def find_moe_layers(
```
kylesayrs, line 180: This is guaranteed to be standardized to an instance of https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/moe/linear_experts.py#L139 by this point.

kylesayrs (commented):
1. Set dataset_args.moe_calibrate_all_experts on calibration start, restore flag on calibration end
2. Do model pruning on sequential end (allow for propagation), do config update on calibration end. Optionally, according to empirical data, set dataset_args.propagate_error = False or whatever is best
3. Use utils from https://github.com/vllm-project/llm-compressor/tree/main/src/llmcompressor/modeling/moe

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: Decided to add the validation

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: Also added for topk group with the fallback to n_group

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Going to implement this, but in a separate commit, so that it can be reversed

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: Without restore() for the MoE copy of a model, doing del on individual experts no longer presents much benefit, so the explicitness and safety of this approach is preferred

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Will also implement in a separate commit after conversation with @kylesayrs . Pruning in SEQUENTIAL_EPOCH and config update in calibration_end

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Adding a check for the get_calibrate_all_experts_flag() that will raise an error in the interim, but going to set this to False explicitly in a separate commit

kylesayrs:
https://github.com/vllm-project/llm-compressor/pull/2864

PR #2830, opened 2026-06-13
Add input-embedding quantization example

KKothuri:
## Purpose

Adds `examples/quantization_embedding/` — an example showing how to quantize a model's **input embedding table** to weight-only `intN` (WNA16) with a data-free `QuantizationModifier`. vLLM loads these checkpoints (embedding-quant support added in vllm-project/vllm#44340) and runs a fused gather + dequant over the looked-up rows, so the packed table is never densified.

The recipe targets the `Embedding` module **by class name** (`["Embedding"]`), which is portable across architectures and independent of a model's module prefix. This matters because name-based targets (e.g. `re:.*embed_tokens$`) require the model to forward `prefix` to its `VocabParallelEmbedding`, which not all vLLM models do (see vllm-project/vllm#45535).

Embedding quantization is weight-only and **data-free** (no calibration set), near-lossless, and most useful for large-vocabulary models where the embedding table is a meaningful fraction of memory.

## Changes

- `examples/quantization_embedding/llama3_example.py` — data-free embedding quant (int4, group size 64), sample generation, compressed save.
- `examples/quantization_embedding/README.md` — walkthrough, channel / 8-bit variants, how to compose with linear-weight quantization, and an accuracy table.

## Testing

Ran the example flow end-to-end (load → `oneshot` → `dispatch_model` → generate → save); exits clean. Accuracy (`lm-eval`) on `pythia-1.4b` shows embedding quantization is near-lossless:

| scheme | wikitext ppl | arc_easy acc |
| --- | --- | --- |
| baseline (fp16) | 14.733 | 0.6048 |
| embedding W8 channel | 14.732 | 0.6052 |
| embedding W4 group-64 | 14.752 | 0.6061 |

The example references `meta-llama/Meta-Llama-3-8B-Instruct` per the repo convention (each scheme folder has a llama3 example); the identical recipe was validated on `pythia-1.4b` and `Mistral-7B-v0.1`, and the resulting checkpoints load and generate in vLLM.

---

This change was developed with AI assistance (Claude Code). All changed lines were reviewed by the submitter.


kylesayrs (approved):
Looks great, thanks for the contribution!

examples/quantization_embedding/llama3_example.py
```diff
@@ -0,0 +1,54 @@
...
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+
+# Select model and load it.
+model_id = "meta-llama/Meta-Llama-3-8B-Instruct"
+model = AutoModelForCausalLM.from_pretrained(
+    model_id, device_map="auto", torch_dtype="auto"
+)
```
kylesayrs:
```suggestion
model = AutoModelForCausalLM.from_pretrained(model_id)
```

examples/quantization_embedding/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+### 1) Load the model
+
+```python
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+model_id = "meta-llama/Meta-Llama-3-8B-Instruct"
+model = AutoModelForCausalLM.from_pretrained(
+    model_id, device_map="auto", torch_dtype="auto"
+)
```
kylesayrs:
```suggestion
model = AutoModelForCausalLM.from_pretrained(model_id)
```

brian-dellabetta (approved):
Thanks for the contribution @KKothuri !

PR #2833, opened 2026-06-16
[model_free] fix: skipped quantization on split fused moe experts

LeonEricsson:
**SUMMARY**:

#2677 removed the .weight suffix from split tensor names based on the assumption that safetensors tensor names always include `.weight`. That was a valid fix for the case introduced in #2464: when the input tensor already ended in .weight, the #2464 logic could produce invalid names like `...gate_proj.weight.weight`.

For example, this input:

`model.layers.0.mlp.experts.gate_up_proj.weight`

must split to:

`model.layers.0.mlp.experts.0.gate_proj.weight`

not:

`model.layers.0.mlp.experts.0.gate_proj.weight.weight`

However, the assumption does not hold generally, some fused expert tensors, such as the ones in Gemma 4 appear as:

```
model.layers.0.mlp.experts.gate_up_proj
model.layers.0.mlp.experts.down_proj
```

without `.weight` suffix entirely. After #2677, those are now split into names like:

```
model.layers.0.mlp.experts.0.gate_proj
model.layers.0.mlp.experts.0.up_proj
model.layers.0.mlp.experts.0.down_proj
```

and are subsequently skipped by match_quantizable_tensors, because compressed-tensors matches the final parameter segment and defaults to param_targets=("weight",).

This PR appends `.weight` to split fused expert tensor names only when the split key does not already end in `.weight`. ~It also applies fused expert splitting during validation so validation and quantization operate on the same tensor names.~

**TEST PLAN**:
minor regression


src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -60,10 +60,11 @@ def validate_file(
     :param converter: optional converter to apply to the checkpoint,
         e.g. conversion of some layers from some format to compressed-tensors
     """
-    # device is ignored: all validation is done on meta device
-    device = torch.device("meta")
+
     tensors = load_tensors_from_inverse_weight_map(inverse_weight_map, device)
 
+    tensors = split_fused_moe_experts(tensors)
```
brian-dellabetta: validate should only raise warnings/errors as a pre-flight check. splitting of tensors should really be handled during process, and `split_fused_moe_experts` is already called in process_file and process_file_microscale_scheme. Does your update to `split_fused_moe_experts` resolve the issue without requiring calling it during validate?

brian-dellabetta (commented):
Hi @LeonEricsson , thanks for raising this and creating a PR. One request, otherwise this LGTM!

src/llmcompressor/entrypoints/model_free/process.py
LeonEricsson: Got it. The fix does not depend on the call in `validate_file`, I've removed it.                               
 
My motivation was to strengthen the pre-flight behavior of `validate_file`: validating the same tensor view we later quantize, including fused MoE tensors after splitting. The split didn't persist or save any transformed tensors; it only created a temp view for validation.

brian-dellabetta (approved):
Thanks for raising this and providing a resolution

kylesayrs (approved):
Thanks for the catch and fix!

HDCharles:
nice

src/llmcompressor/entrypoints/model_free/process.py
brian-dellabetta: ok, makes sense. i see your reasoning now. we can get this in for now and add in future if needed.

PR #2834, opened 2026-06-16
refactor: modernize core/ module with Python 3.10+ type hints

AsadShahid04:
## Summary
- Replace `Optional[X]` with `X | None` in `events/event.py` and `session_functions.py`
- Remove now-unused `Optional` imports from `typing`
- Fix import block ordering to satisfy ruff's isort check

## Problem
The `core/` module still has legacy `Optional[...]` type hint syntax from Python 3.9. Since Python 3.9 support was dropped in #1910, these can be modernized to the cleaner PEP 604 `X | None` form.

## Solution
Updated the two remaining files in `core/` that still used `Optional`:
- `src/llmcompressor/core/events/event.py`: 3 dataclass fields + 3 `should_update()` params
- `src/llmcompressor/core/session_functions.py`: 2 `LifecycleCallbacks` method params

The other files in `core/` (`state.py`, `session.py`, `lifecycle.py`, `model_layer.py`) were already using the modern `|` syntax.

## Testing
- ruff check: passed
- mypy: 179 pre-existing errors in 44 unrelated files; none in the 2 changed files
- Unit tests: torch not installed in this environment; changes are syntax-only with no functional impact

Part of #1927


AsadShahid04:
Pushed 6a124871 addressing the gemini-code-assist review:

**`Any | None` → `Any`** in both `batch_start` and `loss_calculated` signatures — `Any` already encompasses `None`, so the union was redundant. The `Optional` import was already removed in the original commit; these two call sites were the only remaining cases where the redundancy appeared. Ruff passes clean.

AsadShahid04:
Pushed c351014b: fixed the remaining `:type invocations_per_step: Optional[int]` docstring in the `Event` dataclass to plain `int` — matching the actual field annotation (`invocations_per_step: int = 1`). All `Optional` references are now removed from both changed files.

brian-dellabetta (approved):
thanks!

PR #2835, opened 2026-06-17
Re-tie identically-quantized input/output embeddings to avoid a duplicate lm_head

KKothuri:
## Summary

Quantizing a **tied** model's embedding is currently counterproductive. When embeddings are targeted, `QuantizationMixin.start_calibration` unties them (existing upstream behavior) so the input and output embeddings can be quantized independently — but for a tied model the one shared matrix then becomes a packed table **plus** a new full-precision `lm_head`, so the checkpoint grows. This matters most for small SOTA models (`Qwen3-0.6B`, `LFM2.5-350M`) where that shared matrix dominates file size.

Per review (thanks @kylesayrs), this adds a **save-time** step that re-ties the input and output embeddings when they were quantized to identical values — the same matrix quantized twice — so they serialize as a single shared table:

- **Re-tie identically-quantized embeddings before saving** (`compressed_tensors_utils.py`). If the (compressed) input and output embeddings have identical tensors, the output embedding's storage is aliased to the input's so transformers' save-time de-duplication writes a **single shared table**, and `tie_word_embeddings` is restored so the tie is reconstructed at load. If they differ — quantized differently, or only one was quantized — they are left untied and both are kept, preserving the model's integrity. The decision is driven purely by tensor equality, so nothing needs to track the original tie state.

The re-tie uses the existing compressed-tensors offload API — `setattr`/`getattr` under `OffloadCache.disable_onloading()` — so the two modules share one tensor with no copy and no new compressed-tensors API. (The `disable_onloading` context can be removed once compressed-tensors#709 makes `__setitem__` a non-copying replacement.)

### Relationship to #2798

Builds on (does not duplicate) the merged #2798. #2798 re-ties **uncompressed** tied weights that offloading splits apart, via `model.tie_weights()`. That path can't handle a **compressed** embedding, which has no plain `.weight` to tie to (`tie_weights()` raises and logs a warning), and it only runs while `tie_word_embeddings` is still set. This PR covers exactly that gap. The two run back-to-back at save and are complementary: `_retie_offloaded_weights` handles the dense case, `_retie_quantized_embeddings` the compressed one.

### Why this is useful

The single packed table is reused for **both** the input lookup and the logits matmul by runtimes that support it — vLLM consumer support: vllm-project/vllm#45535. Accuracy (lm-eval): tied **W8** embedding is near-lossless (Qwen3-0.6B wikitext bits-per-byte −0.0%, LFM2.5-350M +0.45%); **W4** is heavier on these compression-sensitive models (+1.2% / +4.4%).

## Test Plan

```
pytest tests/llmcompressor/transformers/compression/test_compress_tensor_utils.py -k tied
```

`test_tied_quantized_embedding_no_duplicate_head` quantizes `Qwen3-0.6B`'s embedding **and** `lm_head` (data-free) and asserts the saved checkpoint stays tied (`tie_word_embeddings=True`) with the packed embedding present and **no** `lm_head` weight tensors.

## Test Result

- New test passes.
- Existing `test_no_duplicate_tied_lm_head_on_save` (tied/untied × offload) still passes — an explicitly-untied model keeps its separate head (the re-tie is scoped to the compressed case).
- End-to-end: the produced `Qwen3-0.6B` checkpoint has a single shared packed table (no fp16 head), config declares both embeddings quantized, `tie_word_embeddings=True`, 930 MB vs 1.5 GB fp16 — loads and generates coherently in vLLM (with #45535).
- `ruff check` / `ruff format --check` clean.

---

This change was developed with AI assistance (Claude). All changed lines were reviewed by the submitter.


brian-dellabetta (commented):
Hi @KKothuri , thanks for raising the issue and providing a resolution. This involves 2 PRs (#2798 and your #2830 ) that were considered independently, so we'll have to think about the best way to handle this:
1. User wants to quantize just embed_tokens and leave lm_head (https://github.com/vllm-project/vllm/pull/44340)
2. User wants to quantize both (your open PR https://github.com/vllm-project/vllm/pull/45535)

It looks like you are defaulting to re-tying, but what about the case where user wants lm_head dense? It would lead to a slightly larger model but possibly faster runtime.

KKothuri:
> Hi @KKothuri , thanks for raising the issue and providing a resolution. This involves 2 PRs (#2798 and your #2830 ) that were considered independently, so we'll have to think about the best way to handle this:
> 
> 1. User wants to quantize just embed_tokens and leave lm_head ([vllm-project/vllm#44340](https://github.com/vllm-project/vllm/pull/44340))
> 2. User wants to quantize both (your open PR [vllm-project/vllm#45535](https://github.com/vllm-project/vllm/pull/45535))
> 
> It looks like you are defaulting to re-tying, but what about the case where user wants lm_head dense? It would lead to a slightly larger model but possibly faster runtime.

Thanks @brian-dellabetta ! The below 3 cases are supported and controlled by tie_word_embeddings:

1. Quantize embed_tokens, leave lm_head dense (tie_word_embeddings=False, target ["Embedding"])

2. Quantize one shared tied matrix (default, tie_word_embeddings=True)

3. Quantize both as separate tensors (tie_word_embeddings=False, target ["Embedding"] + lm_head)

The only behavioral change here is the default for tied models: previously the embedding-quant path force-untied (giving case1 by default, contradicting the documented tie_word_embeddings "leave tied if possible"); now the default keeps it tied (case2). Cases 1 and 3 need tie_word_embeddings=False.

Let me know if I should change the default.

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -245,7 +245,21 @@ def start_calibration(self, model: torch.nn.Module):
...
+        )
+        if targets_embeddings(model, targets) and not getattr(
+            text_config, "tie_word_embeddings", False
+        ):
+            # Respect `tie_word_embeddings`: only untie when the user opted out of
+            # tying (the entrypoint sets the config flag accordingly) or the model
+            # is natively untied. Keeping a tied model tied lets the shared
+            # embedding/lm_head matrix be quantized once; consumers use the packed
+            # table for both the input lookup and the logits matmul.
             untie_word_embeddings(model)
```
kylesayrs, line 249: If `tie_word_embeddings` is already `False`, then untying word embeddings is a no-op, right?

kylesayrs (commented):
I'll give a deeper review soon. I think this implementation is slightly problematic, as it does not attach qparams to the lm_head (so sample generations will be incorrect) and does not utilize the system for declaring tied parameters during saving which is already provided by transformers.

What do you think about a flow like this? I think a flow like this should cover what is needed while working within the systems set up by `transformers`.
1. always untie (if embeddings are targeted). Leave model config unchanged
2. explicitly quantize + compress input and output embeddings separately
3. retie embeddings (and embedding qparams) before saving (if their values are identical, and tie_word_embeddings=True in the config). Do not drop, which can destroy the model's integrity

This way, if a user accidentally specifies a recipe in which input and output embeddings and quantized differently, we can explicitly warn that this occurred and still save both weights in the checkpoint. **It also creates a config which specifies that both input and output embeddings are both quantized**, which allows us to reload the model with the [transformers integration](https://huggingface.co/docs/transformers/en/quantization/compressed_tensors).

KKothuri:
> What do you think about a flow like this? I think a flow like this should cover what is needed while working within the systems set up by `transformers`.
> 
> 1. always untie (if embeddings are targeted). Leave model config unchanged
> 
> 2. explicitly quantize + compress input and output embeddings separately
> 
> 3. retie embeddings (and embedding qparams) before saving (if their values are identical, and tie_word_embeddings=True in the config). Do not drop, which can destroy the model's integrity
> 
> 
> 
> This way, if a user accidentally specifies a recipe in which input and output embeddings and quantized differently, we can explicitly warn that this occurred and still save both weights in the checkpoint. **It also creates a config which specifies that both input and output embeddings are both quantized**, which allows us to reload the model with the [transformers integration](https://huggingface.co/docs/transformers/en/quantization/compressed_tensors).

I agree with this. This sounds like a cleaner flow. In this setting if the user wants to quantize the input embedding for a tied model do we also automatically quantize the output embedding? Or should they explicitly target both in the config? I think the second one is more correct even though there is only one matrix in a tied model. I will rework this PR to match your suggested flow.

KKothuri:
@kylesayrs I started on this and hit a snag in step 3 (re-tie before save): re-tieing the packed params doesn't dedup like the dense .weight tie does. compress_model offloads weight_packed/scale/shape into per-module offload dicts, so pointing the output module's packed params at the input's doesn't actually share storage, and transformers' save dedup (which keys off shared tensor storage + _tied_weights_keys, only covering .weight) never collapses them.  So two identical packed tables get written even though the model is tied (repro'd on CPU and GPU). I think the fix needs a compressed-tensors side primitive to make two modules share a single offloaded packed tensor that survives to_accelerate/state_dict. Does that sound fine, or is there an existing primitive I should use?

src/llmcompressor/modifiers/quantization/quantization/mixin.py
KKothuri: Yes it is a no-op if the embeddings are already untied. The existing flow had this call unguarded

KKothuri:
@kylesayrs I reworked this PR to your proposed flow: always untie when embeddings are targeted, quantize the input and output embeddings separately, then re-tie at save when their packed values are identical. Verified on Qwen3-0.6B.

The re-tie needs a small compressed-tensors primitive, tie_offload_parameter ([vllm-project/compressed-tensors#749](https://github.com/vllm-project/compressed-tensors/issues/749)): the packed params live in per-module offload caches whose assignment copies, so two modules can't share one tensor and without that, state_dict keeps separate storage and the dedup never fires. This PR imports it with a fallback for CT versions that don't have it yet.

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -20,6 +25,111 @@
...
+        return
+
+    try:
+        input_embed = model.get_input_embeddings()
+        output_embed = model.get_output_embeddings()
+    except (AttributeError, NotImplementedError):
+        return
+
+    if input_embed is None or output_embed is None or input_embed is output_embed:
+        return
```
kylesayrs: This snippet duplicates [`get_embeddings`](https://github.com/vllm-project/llm-compressor/blob/2e7456ae9362b6a7d14b3e9889dcdb5c2930759d/src/llmcompressor/utils/transformers.py#L68C5-L68C19)

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -246,7 +246,19 @@ def start_calibration(self, model: torch.nn.Module):
...
+            # results into a single shared tensor when they end up identical.
+            config = getattr(model, "config", None)
+            get_text_config = getattr(config, "get_text_config", None)
+            text_config = (
+                get_text_config(decoder=True) if callable(get_text_config) else config
+            )
+            was_tied = getattr(text_config, "tie_word_embeddings", False)
             untie_word_embeddings(model)
+            if was_tied and text_config is not None:
+                text_config.tie_word_embeddings = True
```
kylesayrs: It seems more appropriate to wait until the tensors are tied again before readjusting the config.

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -78,6 +188,11 @@ def save_pretrained_wrapper(
             # tied-parameter bookkeeping consistent.
             _retie_offloaded_weights(model)
 
+            # Embeddings are untied during calibration so input and output can be
+            # quantized independently; if they ended up identical, re-tie the
+            # packed weights so save de-duplicates them into one shared table.
+            _retie_quantized_embeddings(model)
```
kylesayrs:
```suggestion
input_embed, output_embed = get_embeddings()
input_params = get_tensors(input_embed)
output_params = get_tensors(output_embed)
if len(input_params) == len(output_params) and all(torch.equal(ip, op), for zip(input_params, output_params)):
    with disable_onloading():
        for name in input_params.keys():
            setattr(output_embed, name, getattr(input_embed, name))
    model.config.tie_word_embeddings = True
```

kylesayrs (commented):
@KKothuri Thanks so much for your patience here, it's been a busy few weeks with glm5.2 🙃. I think there a couple places which can make the code changes much smaller, let me know what you think

KKothuri:
Hi @kylesayrs. Thanks for the review. I have addressed your comments and reduced the changes. I am still keeping the retie_quantized_embeddings function but I am not duplicating the get_embeddings code anymore. Please take a look.

kylesayrs (approved):
Looks great, thanks for bringing it to attention!

brian-dellabetta (approved):
Thanks for the contribution!

kylesayrs:
@KKothuri This new implement should cover `_retie_offloaded_weights`, so we should remove `_retie_offloaded_weights`, right?

KKothuri:
> @KKothuri This new implement should cover `_retie_offloaded_weights`, so we should remove `_retie_offloaded_weights`, right?

Should I roll them into one function?

KKothuri:
> @KKothuri This new implement should cover `_retie_offloaded_weights`, so we should remove `_retie_offloaded_weights`, right?

I merged both functions into `_retie_embeddings`.

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -18,10 +18,80 @@
...
+    config = getattr(model, "config", None)
+    get_text_config = getattr(config, "get_text_config", None)
+    text_config = get_text_config(decoder=True) if callable(get_text_config) else config
+
+    # Quantized embeddings carry packed weights and are always re-tied when their
+    # values match. A dense model is only re-tied when its config still declares
+    # it tied, so a deliberately-untied model is left untouched.
+    compressed = "weight_packed" in input_tensors
+    if not compressed and not getattr(text_config, "tie_word_embeddings", False):
+        return
```
kylesayrs: Can we delete this part? Not all compressed modules have "weight_packed", for example fp8 still calls it "weight".

We can just make it our policy to always retie if the weights are exactly the same

kylesayrs (changes requested):
Last change, looks good! Thanks for combining them!

kylesayrs (approved):
Looks great, thanks!

PR #2839, opened 2026-06-18
fix: widen oneshot() recipe type hint to accept Modifier and Recipe objects

### LINKED ISSUE #2837: [Bug]: oneshot() recipe parameter type hint rejects Modifier objects

soyr-redhat:
## Summary
- Widen the `recipe` parameter type on `oneshot()` from `str | list[str] | None` to `RecipeInput | None`
- Update docstring to reflect accepted input types

`RecipeArguments.recipe` is left as `str | None` since `HfArgumentParser` needs primitive types.

Fixes #2837

### DIFF
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -33,6 +33,8 @@ __all__ = ["Oneshot", "oneshot"]
 if TYPE_CHECKING:
     from datasets import Dataset, DatasetDict
 
+    from llmcompressor.recipe import RecipeInput
+
 
 TOKENIZERS_PARALLELISM_ENV = "TOKENIZERS_PARALLELISM"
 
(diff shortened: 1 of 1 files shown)


soyr-redhat:
This isn't a hard a fast requirement for operation, so low-prio for sure

brian-dellabetta (approved):

kylesayrs (approved):

PR #2841, opened 2026-06-18
fix: skip None layers in fuse_weight_observers

HDCharles:
Summary

when any of the fused layers are missing (shared k,v in qkv) this will error without this change, see the unit test

### DIFF
--- a/src/llmcompressor/observers/helpers.py
+++ b/src/llmcompressor/observers/helpers.py
@@ -194,6 +194,8 @@ def fuse_weight_observers(model: Module):
             only_tensor_group = True
             observers_and_modules = []
             for layer in layers_to_fuse:
+                if layer is None:
+                    continue
                 obs = getattr(layer, "weight_observer", None)
                 if obs is None:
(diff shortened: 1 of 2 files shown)


src/llmcompressor/observers/helpers.py
```diff
@@ -194,6 +194,8 @@ def fuse_weight_observers(model: Module):
             only_tensor_group = True
             observers_and_modules = []
             for layer in layers_to_fuse:
+                if layer is None:
```
brian-dellabetta, line 197: wouldn't this just continue for `layer is None` anyway? if layer is None, then
```
obs = getattr(layer, "weight_observer", None)
```
would be None, and that continues in the next conditional?

brian-dellabetta: nm only_obs setting must matter

brian-dellabetta (approved):
makes sense to me

kylesayrs (approved):

PR #2844, opened 2026-06-21
[Autoround][DDP] Add Qwen MoE Example

yiliu30:
SUMMARY:

- ~~Depends on https://github.com/vllm-project/compressed-tensors/pull/742~~
- Depends on https://github.com/vllm-project/llm-compressor/pull/2896
```bash
  
# vllm ({'pretrained': 'Yi30/Qwen3-235B-A22B-Instruct-2507-W4A16-AutoRound-iters100-nsamples256-DDP2', 'tensor_parallel_size': 4, 'max_model_len': 8192, 'max_num_batched_tokens': 32768, 'max_num_seqs': 128, 'add_bos_token': True, 'gpu_memory_utilization': 0.8, 'dtype': 'bfloat16', 'max_gen_toks': 2048, 'enable_prefix_caching': False}), gen_kwargs: ({}), limit: 1000.0, num_fewshot: None, batch_size: 128
# |Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
# |-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
# |gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.956|±  |0.0065|
# |     |       |strict-match    |     5|exact_match|↑  |0.957|±  |0.0064|
```

TEST PLAN:


src/llmcompressor/utils/dev.py
```diff
@@ -160,7 +160,7 @@ def get_main_device() -> torch.device:
 
     elif torch.accelerator.is_available():
         accel_type = torch.accelerator.current_accelerator().type
-        return torch.device(accel_type, rank)
+        return torch.device(accel_type, torch.accelerator.current_device_index())
```
brian-dellabetta: this indeed looks like a bugfix, perhaps we just never hit this previously

examples/autoround/ddp/ddp_qwen3_moe_example.py
```diff
@@ -0,0 +1,117 @@
+"""
+DDP AutoRound quantization example for large MoE models.
+
+Uses the standard compressed-tensors DDP path: load_offloaded_model()
+broadcasts weights from rank 0 to rank 1.  GPUS_PER_GROUP controls how
+many GPUs each rank uses for per-block model parallelism.
```
brian-dellabetta: Can you provide a little context here? I am unfamiliar with this sort of grouping beyond `--nproc_per_node`. 

If we want to keep this, we should also prefix with `AUTOROUND_GPUS_PER_NODE` or `LLMCOMRESSOR_...` to make it clear this env var is entirely owned by us and not used by torch distributed

examples/autoround/ddp/ddp_qwen3_moe_example.py
```diff
@@ -0,0 +1,117 @@
...
+ITERS = 100
+NSAMPLES = 256
+
+###### DDP INIT #####
+gpus_per_group = int(os.environ.get("GPUS_PER_GROUP", "1"))
+local_rank = int(os.environ["LOCAL_RANK"])
+main_gpu = local_rank * gpus_per_group
+torch.cuda.set_device(main_gpu)
+dist.init_process_group(
+    backend="nccl",
```
brian-dellabetta: seems like this example is hard-coded specifically for CUDA, is DDP not compatible with XPU?

brian-dellabetta (commented):
Hi @yiliu30 , a few questions

examples/autoround/ddp/ddp_qwen3_moe_example.py
yiliu30: Hi @brian-dellabetta, good catch. It should work on XPU. Our upcoming GPU, CRI, has more than 100GB of VRAM, so it should be able to tune large models. I updated it.

examples/autoround/ddp/ddp_qwen3_moe_example.py
yiliu30: Sure, updated to `LLMCOMRESSOR_GPUS_PER_GROUP` and added more context.

src/llmcompressor/utils/dev.py
yiliu30: Yeah, `torch.distributed.get_rank()` is the global worker id, while `torch.accelerator.current_device_index()` is the actual device selected in the current process.
In this AutoRound case is our grouped-GPU setup on a single 8-GPU node:
```bash
LLMCOMRESSOR_GPUS_PER_GROUP=2 torchrun --nproc_per_node=2 ...
  In that launch:

  - rank 0 uses GPUs 0-1
  - rank 1 uses GPUs 2-3
```
So for rank 1, `get_rank()` is 1, but the process’s main device is GPU 2, meaning `current_device_index()` is 2 after device setup.

yiliu30:
Hi @kylesayrs @HDCharles please take a look when you have time, thanks!

examples/autoround/ddp/ddp_qwen3_moe_example.py
```diff
@@ -0,0 +1,131 @@
...
+    backend = "gloo"
+gpus_per_group = int(os.environ.get("LLMCOMPRESSOR_GPUS_PER_GROUP", "1"))
+local_rank = int(os.environ["LOCAL_RANK"])
+main_gpu = local_rank * gpus_per_group
+torch.accelerator.set_device_index(main_gpu)
+dist.init_process_group(
+    backend=backend,
+    init_method="env://",
+    device_id=torch.device(f"{accel_type}:{main_gpu}"),
+)
```
brian-dellabetta: there's a lot of overlap here with our `init_dist` helper method, can we move any new logic into the helper instead to make the example more aligned with our other DDP examples?

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -30,6 +33,15 @@
 __all__ = ["AutoRoundModifier", "fix_batch_if_needed"]
 
 
+def _get_local_gpu_group_size() -> int:
+    return int(
+        os.environ.get(
+            "LLMCOMPRESSOR_GPUS_PER_GROUP",
+            os.environ.get("GPUS_PER_RANK", "1"),
```
brian-dellabetta: We should just pick one here env var here, my thinking for suggesting the LLMCOMPRESSOR prefix was to help indicate to users it was scoped purely to our repo and not something torch would use. But if there is precedent for GPUS_PER_RANK elsewhere, we can stick to it

brian-dellabetta (commented):
Thanks @yiliu30 , a couple more comments, and I'll ping for reviews from the maintainers on DDP

examples/autoround/ddp/ddp_qwen3_moe_example.py
HDCharles: yeah I agree, it seems like all this is mostly just altering the device assignment to be 

`gpus_per_group * local_rank ` rather than just local_rank, that can be added into init_dist pretty easily if this is a feature we want to support

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Hi @brian-dellabetta, `LLMCOMPRESSOR_GPUS_PER_GROUP` sounds good. I’ve updated the code to keep this one.

examples/autoround/ddp/ddp_qwen3_moe_example.py
yiliu30: Hi @brian-dellabetta @HDCharles  Thanks for the suggestion. I think this behavior is fairly AutoRound-specific, since, AutoRound is currently the only flow that may need multiple GPUs to tune a single decoder layer.

Would it be okay to keep this scoped to AutoRound rather than adding it to the shared `init_dist` helper? We could implement it either:

- as an AutoRound-specific helper used by the example; or
- directly in the example, if we prefer not to introduce additional shared API surface.

This would keep the common DDP initialization path simpler while still supporting AutoRound’s multi-GPU tuning use case.

brian-dellabetta: Yeah, since DDP generally means 1 GPU per rank from what I've read, maybe it makes sense to leave it out of base `init_dist` for now, though I do like the idea of moving it to a shared helper in autoround src code

yiliu30: Thanks @brian-dellabetta , I moved it into the `autoround/utils.py`.

brian-dellabetta (approved):
Thanks @yiliu30 , LGTM!

HDCharles (approved):
looks good, just fix quality

yiliu30:
Hi @brian-dellabetta @HDCharles  It looks like the failing unit test was caused by the FP4 packing Triton kernel. I’ve fixed it here. Could you please take a look? Thanks!
 https://github.com/vllm-project/compressed-tensors/pull/781

PR #2852, opened 2026-06-23
Enable torch.compile for GPTQ block quantization

Bias92:
## summary

this is the gptq follow-up to #2384. that pr made the mse observer compile path work; this one applies the same opt-in `enable_compile` flow to the main gptq quantization loop.

the compiled path now handles `tensor`, `channel, group`, `tensor_group`, and `block` weights.

## details

the main change is pulling the inner block loop into `_quantize_block`, then calling the compiled version from `quantize_weight` when `active_session().state.enable_compile=True`.

`quantize_weight` reads the session state directly, matching the observer compile path and avoiding extra modifier-local state. the compile path normalizes each strategy's qparams into per-row column qparams, then temporarily patches `quant_args.strategy` to `channel` for the compiled helper.

`_quantize_block` returns `Q1`, `Err1`, and `losses1`. `W1` is only mutated inside the helper for the normal gptq error-feedback recurrence, and is not read after the compiled call. `Hinv1` is made contiguous before the compiled call to avoid specializing on the parent hessian stride.

## benchmarks

fresh process on a100 sxm 80gb, torch 2.12.1+cu130, qwen3-8b, w4a16 group, 32 samples, seq 2048:

| mode | time | peak memory |
| --- | ---: | ---: |
| eager | 776.7s | 19175.9 mb |
| enable_compile | 468.8s | 19174.3 mb |

this is about 1.66x cold e2e with effectively unchanged memory. `TORCH_LOGS=graph_breaks,recompiles` showed 0 graph breaks and 0 recompiles after making `Hinv1` contiguous.

first compile is still the main overhead: the first q_proj compile is about 280s, while the rest of the quantization run after compile is about 190s.

## validation

- `python -m ruff check src/llmcompressor/modifiers/gptq/base.py src/llmcompressor/modifiers/gptq/gptq_quantize.py tests/llmcompressor/modifiers/gptq/test_base.py tests/llmcompressor/modifiers/gptq/test_gptq_quantize.py`
- `python -m compileall src/llmcompressor/modifiers/gptq tests/llmcompressor/modifiers/gptq`
- `git diff --check`

### DIFF
--- a/src/llmcompressor/modifiers/gptq/base.py
+++ b/src/llmcompressor/modifiers/gptq/base.py
@@ -131,6 +131,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
     _num_samples: dict[torch.nn.Module, torch.Tensor] = PrivateAttr(
         default_factory=dict
     )
+    _enable_compile: bool = PrivateAttr(default=False)
 
     def resolve_quantization_config(self) -> QuantizationConfig:
         config = super().resolve_quantization_config()
@@ -206,6 +207,8 @@ class GPTQModifier(Modifier, QuantizationMixin):
         return True
 
     def on_calibration_start(self, state: State, event: Event, **kwargs):
+        self._enable_compile = state.enable_compile
+
         # register quantization calibration hooks
         # assume quantization has been initialized by this modifier or one before it
         QuantizationMixin.start_calibration(self, state.model)
@@ -242,6 +245,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
         """
         Finish calibrating by removing observers and calibration hooks
         """
+        self._enable_compile = False
         QuantizationMixin.end_calibration(self, state.model)
         self.remove_hooks()  # remove gptq hooks
 
@@ -328,6 +332,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
                     hessian=self._hessians.pop(module) / self._num_samples.pop(module),
                     blocksize=self.block_size,
                     percdamp=self.dampening_frac,
+                    enable_compile=self._enable_compile,
                 )
                 comp_logger.set_results(name="GPTQ", loss=loss)
 
--- a/src/llmcompressor/modifiers/gptq/base.py
+++ b/src/llmcompressor/modifiers/gptq/base.py
@@ -131,7 +131,6 @@ class GPTQModifier(Modifier, QuantizationMixin):
     _num_samples: dict[torch.nn.Module, torch.Tensor] = PrivateAttr(
         default_factory=dict
     )
-    _enable_compile: bool = PrivateAttr(default=False)
 
     def resolve_quantization_config(self) -> QuantizationConfig:
         config = super().resolve_quantization_config()
@@ -207,8 +206,6 @@ class GPTQModifier(Modifier, QuantizationMixin):
         return True
 
     def on_calibration_start(self, state: State, event: Event, **kwargs):
-        self._enable_compile = state.enable_compile
-
         # register quantization calibration hooks
         # assume quantization has been initialized by this modifier or one before it
         QuantizationMixin.start_calibration(self, state.model)
@@ -245,7 +242,6 @@ class GPTQModifier(Modifier, QuantizationMixin):
         """
         Finish calibrating by removing observers and calibration hooks
         """
-        self._enable_compile = False
         QuantizationMixin.end_calibration(self, state.model)
         self.remove_hooks()  # remove gptq hooks
 
@@ -332,7 +328,6 @@ class GPTQModifier(Modifier, QuantizationMixin):
                     hessian=self._hessians.pop(module) / self._num_samples.pop(module),
                     blocksize=self.block_size,
                     percdamp=self.dampening_frac,
-                    enable_compile=self._enable_compile,
                 )
                 comp_logger.set_results(name="GPTQ", loss=loss)
 
--- a/src/llmcompressor/modifiers/gptq/gptq_quantize.py
+++ b/src/llmcompressor/modifiers/gptq/gptq_quantize.py
@@ -2,6 +2,7 @@ import math
 from copy import copy
 
 import torch
+import torch._dynamo.config
 import transformers
 from compressed_tensors.quantization import (
     ActivationOrdering,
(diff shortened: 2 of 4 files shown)


src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -173,16 +213,49 @@ def quantize_weight(
         )
         Hinv = H = torch.eye(num_columns, dtype=H.dtype, device=H.device)
 
+    if strategy in (QuantizationStrategy.GROUP, QuantizationStrategy.TENSOR_GROUP):
+        column_quant_args = copy(quant_args)
+        column_quant_args.strategy = QuantizationStrategy.CHANNEL
+    else:
+        column_quant_args = quant_args
```
HDCharles: get rid of this and just use

```with patch_attr(quant_args, "strategy", QuantizationStrategy.CHANNEL):```

in the compile path

File: src/llmcompressor/modifiers/gptq/base.py [hunk 2]
```diff
@@ -242,6 +245,7 @@ def on_calibration_end(self, state: State, event: Event, **kwargs):
         """
         Finish calibrating by removing observers and calibration hooks
         """
+        self._enable_compile = False
```
HDCharles: why?

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -173,16 +213,49 @@ def quantize_weight(
...
+
+    use_compiled_block = (
+        enable_compile
```
HDCharles:
```suggestion
        active_session().state.enable_compile
```

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -131,6 +131,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
     _num_samples: dict[torch.nn.Module, torch.Tensor] = PrivateAttr(
         default_factory=dict
     )
+    _enable_compile: bool = PrivateAttr(default=False)
```
HDCharles: we can get rid of all this scaffolding and just use the state within gptq_quantize

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -173,16 +213,49 @@ def quantize_weight(
...
+        and strategy in (QuantizationStrategy.GROUP, QuantizationStrategy.TENSOR_GROUP)
```
HDCharles: shouldn't we be able to support QuantizationStrategy.CHANNEL as well?

also why can't we support per tensor and per block?

HDCharles:
a few questions

1) why not compile the whole outer loop as well? What ablations did you do and what was tested but didn't work?
2) why is this limited by quantization strategy at all? shouldn't per channel/tensor/block be trivial?
3) how do these changes affect the runtime of normal GPTQ before/after these changes?

HDCharles (changes requested):

src/llmcompressor/modifiers/gptq/gptq_quantize.py
Bias92: done. moved this to `patch_attr` around the compiled call only.

src/llmcompressor/modifiers/gptq/base.py
Bias92: removed.  compile now reads from `active_session().state.enable_compile` directly, so this reset isn't needed.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
Bias92: done.

src/llmcompressor/modifiers/gptq/base.py
Bias92: removed the modifier-level compile flag and read from session state inside gptq_quantize now.

src/llmcompressor/modifiers/gptq/gptq_quantize.py
Bias92: oh yeah, agreed. i widened this to tensor/channel/group/tensor_group/block for non-sparse weights. preserve_zeros still stays eager.

Bias92:
for ablations:
- first try was compiling only the per-column fake_quantize helper. that was only ~1.1x, so not enough.
- this PR is the first step from that thread: i pulled out the block/column loop rather than just fake_quantize.
- i haven't tried the wider 177-264 region yet; i can check that next if you want.

for strategies:
- agreed, it shouldn't be limited to group/tensor_group. i widened the compiled path to tensor/channel/group/tensor_group/block for non-sparse weights.
- preserve_zeros still stays eager since this helper doesn't handle the sparse mask path.

for normal gptq:
- this is still opt-in through the existing enable_compile flag from #2384.
- enable_compile=false stays on the original eager path. i kept the eager group path unchanged so it doesn't add an inner-loop context manager.
- i also checked a direct quantize_weight microbench with compile off: upstream/main median 2.267s, this branch median 2.290s on the same a100 env, so no meaningful regression there.
- targeted tests pass for flag on/off, including the real cuda compiled path. with TORCH_LOGS=graph_breaks,recompiles on that test, i didn't see graph breaks or recompiles.

> a few questions
> 
> 1. why not compile the whole outer loop as well? What ablations did you do and what was tested but didn't work?
> 2. why is this limited by quantization strategy at all? shouldn't per channel/tensor/block be trivial?
> 3. how do these changes affect the runtime of normal GPTQ before/after these changes?

HDCharles:
yeah i'd like to see if its possible to do the 177-264 region, i'm also wondering if we can do something a bit cleaner like

```
def inner_loop(inputs):
    stuff
    return things

def big_fn(stuff)
    setup_vars
    
     if state.use_compile and not isinstance(inner_loop, torch._dynamo.OptimizedModule):
         global inner_loop
         inner_loop = torch.compile(inner_loop)
    outs = inner_loop(setup_vars)
    
    return do_final_stuff(outs)

```

this way we don't have 2 implementations running around that are basically the same

Bias92:
closing in favor of #2948 , this one is faster and has no compile overhead.

PR #2784, opened 2026-06-02
[Lifecycle] Add `modules` argument to `on_sequential_epoch_end`

kylesayrs:
## Purpose ##
* Solidify the interface for `on_sequential_epoch_end` by passing modules directly
* Reduce boilerplate for extracting `modules` from method kwargs

PR #2855, opened 2026-06-23
fix: remove is_module_quantized filter that made AutoRound a no-op

HDCharles:
## Summary
when we [updated the lifecycle](https://github.com/vllm-project/llm-compressor/pull/2784) autoround was broken because apply autoround expects a decoder_layer specifically and we added a filter to only pass the quantized linear layers instead (like all the other modifiers) removing this filter fixes the issue.

we can now also re-enable the DDP smoke tests for autoround.

also added an autoround unit test to catch this in the future

## Test plan
`pytest -vs -rs tests/llmcompressor/transformers/compression/test_compression_ddp.py`

`pytest -vs -rs tests/llmcompressor/modifiers/autoround/test_base.py`

PR made with claude

### DIFF
--- a/src/llmcompressor/modifiers/autoround/base.py
+++ b/src/llmcompressor/modifiers/autoround/base.py
@@ -15,7 +15,6 @@ from compressed_tensors.quantization import (
     QuantizationStrategy,
     enable_quantization,
 )
-from compressed_tensors.quantization.utils import is_module_quantized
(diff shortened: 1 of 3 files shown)


brian-dellabetta (approved):
nice find

kylesayrs (approved):
Thanks

Issue #1936, opened 2025-10-15
[Bug]: v0.8.0 Causes OOM and Performance Degradation for Deepseek R1


linyu09-oss:
### ⚙️ Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
### Environment Information ###
Operating System: `Linux-5.4.250-4-velinux1u1-amd64-x86_64-with-glibc2.35`
Python Version: `3.11.13 (main, Jun  5 2025, 13:12:00) [GCC 11.2.0]`
llm-compressor Version: `0.7.1`
compressed-tensors Version: `0.11.0`
transformers Version: `4.55.2`
torch Version: `2.8.0`
CUDA Devices: `['NVIDIA H20', 'NVIDIA H20', 'NVIDIA H20', 'NVIDIA H20', 'NVIDIA H20', 'NVIDIA H20', 'NVIDIA H20', 'NVIDIA H20']`
AMD Devices: `None`
```

</details>

### 🐛 Describe the bug

When executing `examples/quantizing_moe/deepseek_r1_example.py`, an OOM (Out of Memory) error occurs at step 235 (model.layers.3.mlp.experts.226.down_proj), with the process taking approximately 1 hour. However, when using version v0.7.1, the memory usage is normal, and it only takes 25 minutes to reach step 235.

<img width="1032" height="588" alt="Image" src="https://github.com/user-attachments/assets/5fb8f7dc-8e7b-4439-a2d8-5d7c425d254c" />

kylesayrs:
Hi @linyu09-oss!

I assume the OOM you're referring to is a CPU OOM? Note that in both versions, deepseek 685B requires ~1.3TB of CPU memory.

Could you please post the memory graph/ maximum memory used by your LLM-Compressor 0.7.1 run?

linyu09-oss:
> Hi [@linyu09-oss](https://github.com/linyu09-oss)!
> 
> I assume the OOM you're referring to is a CPU OOM? Note that in both versions, deepseek 685B requires ~1.3TB of CPU memory.
> 
> Could you please post the memory graph/ maximum memory used by your LLM-Compressor 0.7.1 run?

Thanks for your reply.  I test the script on a machine with 1.8TB RAM. I am currently using v0.7.1 to process DeepSeek R1. I have completed 600 steps and used 170GB, and everything is going well.

PR #2861, opened 2026-06-24
[Bugfix] Fix observer memory leak

kylesayrs:
## Purpose ##
* Reduce memory usage when quantizing large models

## Changes ##
* Delete statistics after qparams are computed

## Testing ##
* TODO: Show before and after of memory usage

### DIFF
--- a/src/llmcompressor/observers/base.py
+++ b/src/llmcompressor/observers/base.py
@@ -97,6 +97,8 @@ class Observer(InternalModule, RegistryMixin):
                     fused_obs(fused_mod.weight)
                 global_absmax = torch.max(global_absmax, -fused_obs.min_vals.min())
                 global_absmax = torch.max(global_absmax, fused_obs.max_vals.max())
+                del fused_obs.min_vals
+                del fused_obs.max_vals
             global_scale = generate_gparam(
                 -global_absmax.reshape(1), global_absmax.reshape(1)
             )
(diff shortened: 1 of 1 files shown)


HDCharles:
doesn't this get freed once the observer is deleted?

HDCharles:
also this destroys fusion, right?

HDCharles (changes requested):
why do you say this is a memory leak, i think this is expected, this info is expected to hang around as long as observers do

this change as is will break nvfp4 since those statistics are needed to calculate global_scale by other observers.

kylesayrs:
@HDCharles That's a fair assumption to make, but leaked observer memory is a blocker for large model compression, see #1936.

The issue I see with deleting observers after sequential epoch is that recipes like AWQ + GPTQ will fail, since we've broken the lifecycle assumption that observers will be attached during calibration.

src/llmcompressor/observers/base.py
```diff
@@ -108,6 +108,9 @@ def get_qparams(self) -> QParamsDict:
             global_scale=global_scale,
         )
 
+        del self.min_vals
+        del self.max_vals
```
HDCharles, line 114:
```suggestion
        if not self._fusions:
            del self.min_vals
            del self.max_vals

HDCharles:
yeah fair enough, its less gptq or awq, its more the whole flow. if we could ensure fused_sequential targets stay together this would be a lot easier but as is you have to do annoying things to handle fusion, or just delete when there's no fusion.

PR #2864, opened 2026-06-24
REAP Pruning Modifier

Ryfernandes:
# REAP Expert Pruning for MoE Models

This PR implements REAP (Router-weighted Expert Activation Pruning) for structurally compressing Mixture of Experts (MoE) models.

## Overview

REAP reduces the VRAM footprint of MoE models by structurally removing individual experts in their MoE layers based on a saliency score computed during calibration. The algorithm is introduced in [REAP the Experts: Why Pruning Prevails for One-Shot MoE Compression](https://arxiv.org/pdf/2510.13999).

### Algorithm

REAP uses a calibration dataset to track router and expert outputs that construct a saliency score:

```
S_j = (1/N_j) ∑ (g_j(t) * ||f_j(t)||_2)
```

where the sum is over the N_j tokens routed to expert j, and:

- `g_j(t)` is the router gate weight assigned to expert `j` for token `t` (the coefficient that multiplies the expert's output when combining experts)
- `f_j(t)` is expert `j`'s output activation for token `t`
- `||f_j(t)||_2` is the L2 norm of that activation
- `N_j` is the total number of tokens routed to expert `j` during calibration

After calibration, experts with the lowest saliency in each layer are removed, up to a specified sparsity level.

## Implementation Details

This implementation now relies on `LinearExperts2D` for router weight extraction. The router weights are captured from the forward pass inputs to the experts block. Compared to the old method of detecting the MoE block class and applying logic to reconcile the router output format, this stays within the LLM Compressor framework and will likely be easier to maintain.

### Major Changes to Eldar's Original PR

- `moe_calibrate_all_experts` explicitly disabled, since it is not necessary in REAP. Saliency tracking logic updated accordingly.
- Pruning now executed in `sequential_epoch_end`, rather than `on_finalize`
- Implementation now relies on the `LinearExperts2D` standardization in LLM Compressor transformers v5 changes. Removed a lot of MoE scanning and extraction logic that can be abstracted away to `LinearExperts2D`

### Current Limitations

- **Granite and Llama MoE blocks are not supported** in this PR, since they do not expose router weights as inputs of the experts block. Their experts are implemented in LLM Compressor as classes that extend `LinearExperts2D` with a different forward pass. Support can be added in a follow-up PR.
- **Group-limited routers are supported**, enabling REAP to work with DeepSeek-V3 architecture models.

## Usage

REAP is useful for reducing the VRAM footprint of MoE models without sacrificing model quality. When applied, it can allow models to fit on fewer GPUs, drastically reducing inference costs.

However, excess pruning can lead to degradation in model quality. Additionally, the dataset choice for calibration is important. If calibration samples leave out certain features or topics, experts normally relevant for these samples may be marked as irrelevant overall and removed. Exercise care and know your application before pruning.

## Results

The end-to-end examples in `examples/reap_expert_pruning` were used to compress two different models.

The first, `Qwen/Qwen3-30B-A3B-Instruct-2507`, is a 30.5B parameter model, with only 3.3B parameters active during inference. It has 48 layers with 128 experts per layer. Only 8 experts are activated per token.

The second, `moonshotai/Moonlight-16B-A3B-Instruct`, is a relatively smaller 16B parameter model, with 3B parameters active during inference. It has 27 layers with 64 experts per layer. 6 experts are activated per token. The model uses a DeepSeek-V3 architecture with a group-limited router. This is supported by the `llm-compressor` REAP implementation.

To test, both models were pruned to 25% and 50% sparsity. Pruning for each was calibrated with 512 samples (max token length 2048) from the `train_sft` split of HuggingFace's `ultrachat_200k` dataset.

Then, the compressed models were evaluated on the `gsm8k_platinum_cot_llama` task from [LM Eval](https://github.com/EleutherAI/lm-evaluation-harness). Evaluation was done with the `local_chat_completions` backend of LM Eval, sending requests to a vLLM server. `MAX_LENGTH` was set to 8192 tokens and `MAX_GEN_TOKS` to 4096. For the Qwen models, the generation config parameters of `temperature=0.7, top_p=0.8, top_k=20` were used. For the Moonlight models, reasonable defaults of `temperature=0.7, top_p=0.9, top_k=0` were used. The task was executed with an `N_SHOT` of 5, and the results were averaged across seeds 1234-1236.

| Model | Config | Flexible | Strict | Recovery (Strict) | Size (GiB) |
|-------|--------|----------|--------|-------------------|------------|
| Qwen3-30B-A3B (128e) | baseline | 0.9672 | 0.9672 | - | 56.93 |
|  | REAP 25% -> 96 | 0.9559 | 0.9562 | 98.86% | 43.43 |
|  | REAP 50% -> 64 | 0.9653 | 0.9653 | 99.80% | 29.92 |
|       |        |          |        |                   |            |
| Moonlight-16B-A3B (DeepSeek-V3 arch, 64e) | baseline | 0.8348 | 0.8326 | - | 29.88 |
|  | REAP 25% -> 48 | 0.6647 | 0.6642 | 79.77% | 23.17 |
|  | REAP 50% -> 32 | 0.1368 | 0.1414 | 16.99% | 16.47 |

As can be seen, the Qwen3 model has very high recovery, both at 25% and 50% sparsity. The smaller Moonlight model has less redundancy across its experts, so it has a modest recovery at 25% sparsity and an extremely poor recovery at 50% sparsity. This demonstrates that REAP is only effective to the extent to which the original model's experts are redundant.

## Files Changed

- Added `REAPModifier` in `src/llmcompressor/modifiers/pruning/reap/`
- Added example scripts and documentation in `examples/reap_expert_pruning/`
- Added tests in `tests/llmcompressor/pruning/reap`
- Modified `oneshot.py`, `state.py`, and `helpers.py` to allow `moe_calibrate_all_experts` behavior to be configurable at a modifier level through the session state

## Summary by CodeRabbit

* **New Features**
  * Added REAP expert pruning support for Mixture-of-Experts models, including calibration-based expert reduction and model saving after compression.
  * Added example workflows for pruning two supported models and running a quick sanity-check generation.
* **Documentation**
  * Added guidance on how the pruning method works, key usage considerations, and sample evaluation results.
* **Bug Fixes**
  * Improved pipeline handling so pruning jobs now use the correct calibration flow when REAP pruning is selected.


src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+    
+    for _, module in model.named_modules():
+        for attr in EXPERTS_ATTRS:
+            if hasattr(module, attr):
+                experts_attr = attr
+                break
+        for attr in ROUTER_ATTRS:
+            if hasattr(module, attr):
+                router_attr = attr
+                break
```
kylesayrs: It seems like you detect moes by their attributes, but then use these attributes to find the layers in `find_moe_layers`. Not only does this make for two redundant functions, but a module might pass the first check while failing the second.

Consider condensing these into a single source of truth: `get_moe_mappings` which returns ([expert and router modules], config_path)

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+    attrs: MoEModelAttrs,
+    new_num_experts: int,
+):
+    config = model.config
+    attr_name = attrs.num_experts_config_key
+
+    updated = False
+    for cfg, cfg_name in [
+        (config, "config"),
+        (getattr(config, "text_config", None), "text_config"),
```
kylesayrs: Consider SSOT, config_path returned by `get_moe_mappings`. The path could include `text_config` if necessary

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+            )
+
+        moe_layers = find_moe_layers(model, self._attrs, self.ignore)
+        if not moe_layers:
+            raise ValueError("No valid MoE layers found in model.")
+        self._moe_layer_names = list(moe_layers.keys())
+
+        config = model.config
+        text_config = getattr(config, "text_config", config)
+        self._top_k = getattr(text_config, "num_experts_per_tok", 2)
```
kylesayrs: Consider SSOT, config_path returned by `get_moe_mappings`

kylesayrs: `num_experts_per_tok` should be included in `get_moe_mappings`

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # fast fail if all layers do not have same num_experts
+        assert_homogeneous_moe(moe_layers, self._attrs)
```
kylesayrs: You can probably remove this assertion; we currently don't support non-homogenous moes in `LinearExperts2D`

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+        sample_module = next(iter(moe_layers.values()))
+        self._num_experts = get_num_experts(sample_module, self._attrs)
```
kylesayrs: This should be gotten from the config

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+        self._n_experts_to_drop = int(self._num_experts * self.sparsity)
+
+        if self._n_experts_to_drop == 0:
+            logger.warning(
```
kylesayrs:
```suggestion
            raise ValueError(
```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+                f"sparsity={self.sparsity} results in 0 "
+                f"experts to drop (out of {self._num_experts}). No pruning will "
+                "be performed."
+            )
+        else:
+            # fail fast (before calibration) if the requested sparsity would
+            # leave the router unable to select top_k experts per token
+            assert_routing_feasible(
```
kylesayrs: Inline this function, it's a very thin wrapper

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+        logger.info(
+            f"REAP initialized: {len(moe_layers)} MoE layers, "
+            f"{self._num_experts} experts/layer, will drop "
+            f"{self._n_experts_to_drop} ({self.sparsity:.0%})"
+        )
+
+        return True
+
+    def on_calibration_start(self, state: State, event: Event, **kwargs):
+        self.started_ = True
```
kylesayrs:
```suggestion

```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+        for layer_name in self._moe_layer_names:
+            module = model.get_submodule(layer_name)
+
+            self._saliency_trackers[layer_name] = REAPSaliencyTracker(self._num_experts)
+            self._norm_buffers[layer_name] = {}
+
+            # one hook per expert to record its per-token output norm
+            experts = getattr(module, self._attrs.experts_attr)
+            expert_list = [expert for expert in experts.children() if isinstance(expert, ExpertMLP)] # Filter out the activation function
```
kylesayrs:
```suggestion
            expert_list = [expert for expert in experts.children() if isinstance(expert, ExpertMLP)] # Filter out the activation function submodule
```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+            for idx, expert in enumerate(expert_list): 
+                self.register_hook(
+                    expert, partial(self._expert_hook, layer_name, idx), "forward"
```
kylesayrs, line 193:
```suggestion
                    expert, self._make_experts_hook(layer_name, idx), "forward"
```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+                )
+            
+            # hook for the experts block to capture the router's top-k routing decisions and weights
+            self.register_hook(
+                experts, partial(self._experts_block_hook, layer_name), "forward"
```
kylesayrs, line 202: I think you want to capture the inputs to experts, right?

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+            )
+
+    def on_calibration_end(self, state: State, event: Event, **kwargs):
+        self.ended_ = True
```
kylesayrs:
```suggestion

```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+        # fallback for pipelines that never fire SEQUENTIAL_EPOCH_END
+        self.on_sequential_epoch_end(state, event, **kwargs)
+        self.remove_hooks()
```
kylesayrs: All pipelines now fire `on_sequential_epoch_end` before firing `on_calibration_end`

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+                "REAP: restored original moe_calibrate_all_experts="
+                f"{self._original_moe_calibrate_all_experts} after calibration."
+            )
+
+    def on_finalize(self, state: State, **kwargs) -> bool:
+        # Apply structural pruning here (after the calibration context has
+        # exited): the decisions were finalized at SEQUENTIAL_EPOCH_END, so this
+        # only mutates module structure -- it does not need calibration data.
+        if not self.ended_:
+            self.on_calibration_end(state, None)
```
kylesayrs:
```suggestion
```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+        if self._n_experts_to_drop == 0:
+            logger.info("REAP: nothing to prune (n_experts_to_drop=0)")
+            return True
```
kylesayrs: Consider erroring earlier.

```suggestion
```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+        model = state.model
+
+        sample_module = model.get_submodule(self._moe_layer_names[0])
+        new_num_experts = get_num_experts(sample_module, self._attrs)
```
kylesayrs:
```suggestion
        new_num_experts = self._n_routed_experts - self._n_experts_to_drop
```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+        update_model_config(model, self._attrs, new_num_experts)
+
+        self._saliency_trackers.clear()
+        self._norm_buffers.clear()
```
kylesayrs, line 217:
```suggestion
        self._saliency_trackers.clear()
        self._norm_buffers.clear()
        self._pruned.clear()
```

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+            if layer_name in self._pruned:
+                continue
+            if tracker.total_count <= 0:
+                continue
+
+            retained = compute_retained_experts(
+                tracker.mean_saliency, self._n_experts_to_drop
+            )
+            prune_moe_layer(model, layer_name, retained, self._attrs)
+            self._pruned.add(layer_name)
```
kylesayrs: Why do you need `_pruned`?

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+            # one hook per expert to record its per-token output norm
+            experts = getattr(module, self._attrs.experts_attr)
+            expert_list = [expert for expert in experts.children() if isinstance(expert, ExpertMLP)] # Filter out the activation function
+            for idx, expert in enumerate(expert_list): 
+                self.register_hook(
+                    expert, partial(self._expert_hook, layer_name, idx), "forward"
+                )
+            
+            # hook for the experts block to capture the router's top-k routing decisions and weights
```
kylesayrs: Consider adding a comment on why this is a `forward` hook and not a `pre_forward` hook, ie explain how norm buffers works

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+
+        for layer_name, tracker in list(self._saliency_trackers.items()):
+            if layer_name in self._pruned:
+                continue
+            if tracker.total_count <= 0:
+                continue
+
+            retained = compute_retained_experts(
+                tracker.mean_saliency, self._n_experts_to_drop
+            )
```
kylesayrs:
```suggestion
            retained = compute_retained_experts(
                tracker.mean_saliency, self._n_experts_to_drop
            )
            assert len(retained) == self._n_routed_experts - self._n_experts_to_drop
```

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+        # Preserve non-expert modules (e.g., act_fn in LinearExperts2D)
+        # These are modules that are not instances of ExpertMLP subclasses
+        non_expert_modules = {}
+        for key, module in experts._modules.items():
+            if not isinstance(module, ExpertMLP):
+                non_expert_modules[key] = module
+
+        # Rebuild with retained experts
+        str_indices = [str(i) for i in range(len(retained))]
+        new_modules = OrderedDict(zip(str_indices, [experts[i] for i in retained]))
```
kylesayrs:
```suggestion
        new_modules = OrderedDict(((str(i), experts[i]) for i in retained))
```

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # Re-add non-expert modules
+        new_modules.update(non_expert_modules)
+
+        experts._modules = new_modules
+        if hasattr(experts, "num_experts"):
+            experts.num_experts = len(retained)
```
kylesayrs:
```suggestion
        experts.num_experts = len(retained)
```

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+            " experts are not in 2D format as expected"
+        )
+
+    _prune_router(router, retained)
+
+    for holder in (moe_block, router):
+        if isinstance(getattr(holder, "num_experts", None), int):
+            holder.num_experts = len(retained)
+        if isinstance(getattr(holder, "n_routed_experts", None), int):
+            holder.n_routed_experts = len(retained)
```
kylesayrs: Does this ever occur?

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+        raise ValueError(
+            f"Cannot prune experts from layer {layer_name}:"
+            " experts are not in 2D format as expected"
+        )
+
+    _prune_router(router, retained)
+
+    for holder in (moe_block, router):
+        if isinstance(getattr(holder, "num_experts", None), int):
+            holder.num_experts = len(retained)
```
kylesayrs: Isn't this already covered by
```
if hasattr(experts, "num_experts"):
    experts.num_experts = len(retained)
```

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+
+    # Direct attribute assignment replaces a parameter/buffer with a different
+    # shape and is correct for both offloaded modules (routed through the
+    # OffloadCache, which re-offloads the new shape) and ordinary modules.
+    router.weight = nn.Parameter(new_weight, requires_grad=False)
+    if new_bias is not None:
+        router.bias = nn.Parameter(new_bias, requires_grad=False)
+
+    if isinstance(getattr(router, "out_features", None), int):
+        router.out_features = len(retained)
```
kylesayrs, line 414:
```suggestion
    if isinstance(router, torch.nn.Linear):
        router.out_features = len(retained)
```

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+
+    return retained
+
+
+def _prune_router(router: nn.Module, retained: list[int]):
+    retained_t = torch.tensor(retained, dtype=torch.long)
+
+    with align_module_device(router):
+        retained_t = retained_t.to(router.weight.device)
+        new_weight = router.weight.detach()[retained_t].contiguous()
```
kylesayrs, line 391: Weights don't have gradients, so there should be no reason to detach, right?

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+                "NUM_EXPERTS_ATTRS in utils.py to add more support."
+            )
+
+        moe_layers = find_moe_layers(model, self._attrs, self.ignore)
+        if not moe_layers:
+            raise ValueError("No valid MoE layers found in model.")
+        self._moe_layer_names = list(moe_layers.keys())
+
+        config = model.config
+        text_config = getattr(config, "text_config", config)
```
kylesayrs: What if a vision LLM is passed through here? It would still try to operate on the vision experts, right? Do you think we can support updating the vision config as well? (ie iterate through all subconfigs)

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+            setattr(cfg, attr_name, new_num_experts)
+            logger.info(
+                f"Updated {cfg_name}.{attr_name}: {old_val} -> {new_num_experts}"
+            )
+            updated = True
+
+    if not updated:
+        logger.warning(
+            f"Config attribute '{attr_name}' not found, skipping config update"
+        )
```
kylesayrs: It should be a requirement that we're able to find these earlier

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+        count = self.count.detach().to("cpu", torch.float64)
+        return sum_saliency / count.clamp(min=1.0)
+
+
+# ---------------------------------------------------------------------------
+# Pruning
+# ---------------------------------------------------------------------------
+
+
+def compute_retained_experts(
```
kylesayrs: Consider making this a method of SaliencyTracker

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+        if not expert_norms_dict:
+            return
+
+        self._ensure(next(iter(expert_norms_dict.values())).device)
+        
+        # For each expert, find which tokens/slots routed to it and compute contributions
+        for expert_idx in range(self.num_experts):
+            if expert_idx not in expert_norms_dict:
+                continue
```
kylesayrs: Should this be a value error?, ie remove this code?

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,275 @@
...
+        self,
+        layer_name: str,
+        expert_idx: int,
+        module: torch.nn.Module,
+        args: tuple,
+        output: Any,
+    ):
+        """Record expert ``f_j`` output norms for every token this batch."""
+        if layer_name not in self._norm_buffers:
+            return
```
kylesayrs, line 262: Should this be an assert?

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+
+            # Get corresponding norms and weights
+            expert_norms = expert_norms_dict[expert_idx]
+            if len(expert_norms) != len(token_indices):
+                raise RuntimeError(
+                    f"REAP saliency tracker: expert {expert_idx} has "
+                    f"{len(expert_norms)} norms but router sent "
+                    f"{len(token_indices)} tokens to it. This indicates a bug in "
+                    f"the expert hook or routing extraction logic."
+                )
```
kylesayrs, line 281: Nice check

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+    def total_count(self) -> float:
+        if self.count is None:
+            return 0.0
+        return float(self.count.sum().item())
+
+    @property
+    def mean_saliency(self) -> torch.Tensor:
+        if self.sum_saliency is None:
+            return torch.zeros(self.num_experts, dtype=torch.float64)
+        sum_saliency = self.sum_saliency.detach().to("cpu", torch.float64)
```
kylesayrs: Why detach?

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+            f"Detected heterogeneous num_experts across MoE layers: {num_experts}"
+        )
+
+
+# ---------------------------------------------------------------------------
+# Saliency tracking
+# ---------------------------------------------------------------------------
+
+
+@dataclass
```
kylesayrs: Consider not making this a data class, instead just a normal class

src/llmcompressor/utils/helpers.py
```diff
@@ -134,6 +136,8 @@ def calibration_forward_context(model: torch.nn.Module):
         stack.enter_context(eval_context(model))
         stack.enter_context(disable_hf_kernels(model))
         stack.enter_context(disable_lm_head(model))
+        if active_session().state.moe_calibrate_all_experts:
+            stack.enter_context(moe_calibration_context())
```
kylesayrs: Thank youuu

kylesayrs (commented):
Lots of nits around early erroring/ asserts, otherwise looks great! Awesome job

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Great, removed!

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Got it, removed this call

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: @kylesayrs What are your thoughts on when to use assert vs. an if / raise error pattern? Should I try to be consistent across this modifier?

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Good point, the deletion of the saliency tracker is enough to avoid double-pruning. I'll remove this

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: @kylesayrs If you mean each of the individual experts, I don't think so. The experts would just receive the hidden states, while this `experts` expert block receives the router output that I'm looking for.

If you're asking why I used a "forward" hook instead of "pre_forward," it's because I also update the saliency in this hook, so I want it to run after the individual expert forward passes have completed, while still being able to access the inputs. This way I don't need to cache them from pre_forward and update saliency in a different hook

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: @kylesayrs I don't believe so because now that we are not enforcing `calibrate_all_experts`, some experts may not receive any tokens for a particular batch, and they will not have a `norm_buffers` entry in `expert_norms_dict`. If you would rather not have this check, I can iterate through `expert_norms_dict.items()` rather than all expert indices.

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+                )
+
+            expert_weights = topk_weights[token_indices, slot_indices]
+
+            # Compute contributions
+            contrib = expert_weights.to(torch.float64) * expert_norms.to(torch.float64)
+
+            # Sum contributions for this expert
+            self.sum_saliency[expert_idx] += contrib.sum()
+            self.count[expert_idx] += len(contrib)
```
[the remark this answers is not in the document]
Ryfernandes: We explicitly set moe_calibrate_all_experts to false in on_calibration_start. Not an issue

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: I agree, and I will just add n_experts_to_drop as another property of the tracker

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: Agree since I will be adding computer_retained_experts to the class

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: This updates for the layer and router if they have num_experts-type variables that they rely on, but I will check to see if this is ever the case. If so, this is still necessary in my opinion

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: Checking

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: @kylesayrs Yeah for example, the [Qwen3 router has a num_experts attribute](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_moe/modeling_qwen3_moe.py#L258). Doesn't seem to be used in the forward pass, but still makes sense to update it.

[Ernie4.5 is an example of a MoE block with the attribute](https://github.com/huggingface/transformers/blob/main/src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py#L385)

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: @kylesayrs Yep, [GLM4 MoE](https://github.com/huggingface/transformers/blob/main/src/transformers/models/glm4_moe/modeling_glm4_moe.py#L292) has this naming convention. I'm going to change this to have a list of possible keys, similar to NUM_EXPERTS_CONFIG_KEYS, so that it's easy to update if we need to add support for something else in the future

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: Didn't end up adding n_experts_to_drop as a property of the tracker, but did add the method

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: @kylesayrs Are you proposing to have _make_experts_hook follow the same behavior as partial()? Creating a callable modification of _experts_hook with layer_name and idx fixed? What's the reason not to use partial, and would you recommend that this also be done on line 170?

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: @kylesayrs from scanning all vision LLMs in transformers with MoE, the MoE layers are only in the text decoder, and the vision encoder is always dense. As far as the config goes, models will only have a text_config if they are a vision LLM (one text_config, one vision_config in the main config). The MoE attributes always live in text_config, since the vision part is dense.

Because of this, we should not need to update the vision config. We should only need to update one config, which is either "text_config" or "config". I will have this embodied in get_moe_mappings

src/llmcompressor/modifiers/pruning/reap/base.py
kylesayrs: > Are you proposing to have _make_experts_hook follow the same behavior as partial()?

Yes, it's just a style suggestion. I should have specified this as a nit, feel free to keep as is.

src/llmcompressor/modifiers/pruning/reap/base.py
kylesayrs: My bad, I meant to delete this comment. Forward makes sense!

src/llmcompressor/modifiers/pruning/reap/base.py
kylesayrs: Anything which could be the result of a user error should be a `raise`, anything that is the result of a programmer error should be an `assert`.

src/llmcompressor/modifiers/pruning/reap/utils.py
kylesayrs: Sounds good, feel free to resolve this comment

src/llmcompressor/modifiers/pruning/reap/utils.py
kylesayrs: Yeah got it, I missed that this was applying to routers as well.

src/llmcompressor/modifiers/pruning/reap/base.py
kylesayrs: > From scanning all vision LLMs in transformers with MoE, the MoE layers are only in the text decoder, and the vision encoder is always dense

I believe that you are correct. Starting with text for now is good, it should be easy to include vision later if needed

src/llmcompressor/modifiers/pruning/reap/utils.py
kylesayrs: Makes sense! Consider adding a comment to clarify

src/llmcompressor/modifiers/pruning/reap/base.py
kylesayrs: Closing, see other comment https://github.com/vllm-project/llm-compressor/pull/2864#discussion_r3475385836

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Thanks!

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: Now a requirement in get_moe_attrs

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Added ✅

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: @kylesayrs I'm not sure. I'm less familiar with PyTorch's autograd, but from what I read online, it seemed that weights can have requires_grad = True. So the slice that happens here would be tracked if we didn't detach. If you think we should remove it though (similarly for the detach on the bias), then I will defer to your judgement

src/llmcompressor/modifiers/pruning/reap/utils.py
```diff
@@ -0,0 +1,362 @@
...
+        new_bias = None
+        if getattr(router, "bias", None) is not None:
+            new_bias = router.bias.detach()[retained_t].contiguous()
+
+    # Direct attribute assignment replaces a parameter/buffer with a different
+    # shape and is correct for both offloaded modules (routed through the
+    # OffloadCache, which re-offloads the new shape) and ordinary modules.
+    router.weight = nn.Parameter(new_weight, requires_grad=False)
+    if new_bias is not None:
+        router.bias = nn.Parameter(new_bias, requires_grad=False)
```
Ryfernandes: @kylesayrs Similarly here, I'm not sure why we're freezing the new weights and biases. It seems to me like we should remove requires_grad=False in case the pruned model has continued training/fine-tuning. Again, I'm newer to this, so please let me know what you think is the best move.

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: @kylesayrs I think this is for autograd tracking that potentially comes from expert_weights and expert_norms. Maybe it would be better to just wrap update() in @torch.no_grad()?

src/llmcompressor/modifiers/pruning/reap/utils.py
Ryfernandes: This was slightly off actually, as it didn't relabel the expert indices. Changed to:

`new_modules = OrderedDict(((str(i), experts[pos]) for i, pos in enumerate(retained)))`

src/llmcompressor/modifiers/pruning/reap/utils.py
kylesayrs: Yeah 100% let's not calculate gradients if we're not going to use them

src/llmcompressor/modifiers/pruning/reap/utils.py
kylesayrs: Leaving the detach here is fine. You're right that, in theory, we could have some gradient calculations, but in practice none of our algorithms expose gradients to each other, and we assume that no weights have gradients for the rest of this codebase.

src/llmcompressor/modifiers/pruning/reap/utils.py
kylesayrs: Again, in the strictest sense, it is possible to have a model which has gradients. However, these weights won't have gradients in practice, and nobody will be using gradients afterwards.

But I think your suggestion is valid. Let's copy the `requires_grad` attribute from the original weight.

```python3
router.weight = nn.Parameter(new_weight, requires_grad=router.weight.requires_grad)
```

examples/reap_expert_pruning/reap_qwen3_30b_50pct.py
```diff
@@ -0,0 +1,64 @@
+"""
```
kylesayrs: These examples should be standardized to follow the format of our other examples, see https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/llama3_example.py

kylesayrs (commented):
Please clean up the examples, otherwise looks great!

kylesayrs (approved):
Looks great! Nice job handling all the weird/ inconsistent moe definitions provided by transformers

src/llmcompressor/core/state.py
```diff
@@ -85,6 +85,8 @@ class State:
     :type optim_wrapped: bool
     :param loss: The loss function being used for training
     :type loss: Any
+    :param moe_calibrate_all_experts: Flag to calibrate all experts in a MoE model
+    :type moe_calibrate_all_experts: bool
```
HDCharles: ?

Ryfernandes: @HDCharles Before, moe_calibrate_all_experts was pulled from the dataset_args and then would be used to conditionally apply the moe_calibration_context in oneshot. However, this meant that moe_calibration_context behavior could not be selectively turned on/off between modifiers. 

The REAP modifier makes most sense to use with moe_calibration_context off, so I consulted with @kylesayrs and moved the conditional application of moe_calibration_context to the calibration_forward_context, so that it can be different for each modifier in an independent pipeline. The moe_calibrate_all_experts flag that controls this was moved to the session state from the dataset_args, since dataset_args is just an input to oneshot; it is not accessible during compression

brian-dellabetta (commented):
I need to wrap up some things before PTO next week, so i didn't have a chance to give this a thorough look-through, but high level structure, examples, and benchmarks look awesome! nice work @Ryfernandes

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def on_calibration_start(self, state: State, event: Event, **kwargs):
+        model = state.model
+
+        # Prevent unnecessary calibration of all experts
+        session = active_session()
+        self._original_moe_calibrate_all_experts = (
+            session.state.moe_calibrate_all_experts
+        )
+        session.state.moe_calibrate_all_experts = False
```
dsikka: I think this is the wrong approach. By doing this, we've essentially proven that moe_calibrate_all_experts is driven now at the modifier level, given REAP will need it off while quantization will need it enabled. Why not make this a modifier arg then instead of using a general dataset arg, and then enable or disable the context based on the modifier?

That's what you're doing here anyway, by capturing the value before (which the context would do) and after with the state

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,292 @@
...
+    _saliency_trackers: dict[str, REAPSaliencyTracker] = PrivateAttr(
+        default_factory=dict
+    )
+    _n_experts_to_drop: int = PrivateAttr(default=0)
+    _n_experts_to_drop_per_group: int | None = PrivateAttr(default=None)
+    _norm_buffers: dict[str, dict[int, torch.Tensor]] = PrivateAttr(
+        default_factory=dict
+    )
+    _original_moe_calibrate_all_experts: bool | None = PrivateAttr(default=None)
+
```
dsikka, line 71: docstrings

dsikka (changes requested):

src/llmcompressor/modifiers/pruning/reap/base.py
kylesayrs: > Why not make this a modifier arg then instead of using a general dataset arg, and then enable or disable the context based on the modifier?

If I understand what you're proposing, then the diff would look something like this?

```python
if any(modifier.moe_calibrate_all_experts for modifier in session.modifiers):
  stack.enter_context(calibrate_experts())
```

I think this approach has a couple downsides:
1. We have to modify all modifiers to expose an interface/ attribute to be read by the context
2. This potentially revokes the user's ability to set `moe_calibrate_all_experts` from `oneshot`

I think that driving pipeline-level arguments from the modifier is inevitable and fine so long as we warn the user that this is happening and the potential consequences, which we do.

src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -0,0 +1,288 @@
...
+                "the saliency tracking and pruning. Please only use one modifier "
+                "or use the independent pipeline."
+            )
+
+        # Warn if calibrate_all_experts is enabled, which is not necessary for REAP
+        if get_calibrate_all_experts_flag():
+            logger.warning(
+                "REAP: calibrate_all_experts is enabled, which is not necessary"
+                " for REAP. If no other modifiers depend on this being set,"
+                " please disable it explicity with the dataset_args flag "
```
dsikka: We should just list the exact arg name that should be updated

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -244,6 +244,7 @@ def apply_recipe_modifiers(
                 recipe_args=self.recipe_args.recipe_args,
                 calib_data=calibration_dataloader,
                 sequential_targets=self.dataset_args.sequential_targets,
+                moe_calibrate_all_experts=self.dataset_args.moe_calibrate_all_experts,
```
dsikka: is this needed?

examples/reap_expert_pruning/reap_qwen3_30b.py
```diff
@@ -0,0 +1,79 @@
...
+        truncation=True,
+        add_special_tokens=False,
+    )
+
+
+ds = ds.map(tokenize, remove_columns=ds.column_names)
+
+# Prune 25% of the experts in each MoE layer, based on saliency.
+# You can adjust this value to prune more or less aggressively.
+recipe = REAPPruningModifier(sparsity=0.25)
```
dsikka, line 54: Once we land this (and verify it) we should add an example with FP8_Dynamic

dsikka (commented):
LGTM. Some small comments

examples/reap_expert_pruning/reap_qwen3_30b.py
Ryfernandes: @dsikka Do you mean doing two modifiers, first one to get to FP8_Dynamic, and then REAP afterwards? Using an independent pipeline?

src/llmcompressor/entrypoints/oneshot.py
Ryfernandes: No longer needed, thank you

src/llmcompressor/modifiers/pruning/reap/base.py
Ryfernandes: Done 👍

Ryfernandes:
Should be ready to merge from my perspective. Just added the change mentioned in standup to modify the moe_calibrate_all_experts warning message. Kept moe_calibrate_all_experts=False as the default in examples

examples/reap_expert_pruning/reap_qwen3_30b.py
dsikka: REAP followed by applying the QuantizationModifier with FP8 so that we have an FP8 checkpoint for vLLM

dsikka (approved):

kylesayrs (approved):
Nice job

PR #2866, opened 2026-06-25
Fix untie_word_embeddings not untying offloaded models

ZewenShen-Cohere:
## Purpose

`untie_word_embeddings` does not actually untie the input/output embeddings when the model is offloaded with `compressed_tensors` (`OffloadCache`). The two embeddings keep sharing the same storage, so at save time `save_pretrained` de-duplicates them and silently drops `lm_head.weight`. 

This surfaced after #2846 ("Remove all device movement from pipelines"), which switched the `datafree`/`basic` pipelines from `dispatch_model` (materialized weights) to `set_onload_device` (offloaded weights via `OffloadCache`). Any flow that unties a tied model while it is offloaded is affected (e.g. SpinQuant, which must export an untied, rotated `lm_head`).

## Root cause
https://github.com/vllm-project/llm-compressor/blob/3707778241acf08eeb2f6aee7f4f50b02b2b45ca/src/llmcompressor/utils/transformers.py#L27-L33

For an offloaded module, `module._parameters` is an `OffloadCache`. `register_parameter` triggers `OffloadCache.__setitem__`, which for an existing key of the same size calls `update_offload(...)`—i.e. it writes into the **existing** storage in place instead of allocating new storage. When the embeddings are tied, both modules' cache entries point at the same offloaded storage, so both clones are written back into that one shared storage and the weights remain tied.

With the old `dispatch_model` path, `_parameters` is a plain dict and `register_parameter` simply replaces the entry, so the clone is genuinely independent—which is why this worked before #2846.

## Fix

Delete the existing parameter before re-registering, forcing the offload cache to allocate fresh storage:

```python
delattr(module, "weight")
module.register_parameter("weight", param)
```

This is a no-op for non-offloaded modules and correctly separates storage for offloaded ones.

## Reproduction

Run the below script:
```python
import torch
from transformers import LlamaConfig, LlamaForCausalLM

from compressed_tensors.offload import set_onload_device
from llmcompressor.utils import untie_word_embeddings


def storage_id(tensor: torch.Tensor):
    return (tensor.device, tensor.untyped_storage().data_ptr())


def main():
    config = LlamaConfig(
        hidden_size=32,
        intermediate_size=64,
        num_hidden_layers=2,
        num_attention_heads=4,
        num_key_value_heads=4,
        vocab_size=128,
        tie_word_embeddings=True,
    )
    model = LlamaForCausalLM(config)

    in_embed, out_embed = model.get_input_embeddings(), model.get_output_embeddings()
    assert storage_id(in_embed.weight) == storage_id(out_embed.weight), "should start tied"

    # offload the model the way the data-free / basic pipelines now do (CPU keeps it GPU-free)
    set_onload_device(model, torch.device("cpu"))

    untie_word_embeddings(model)

    in_embed, out_embed = model.get_input_embeddings(), model.get_output_embeddings()
    in_store, out_store = storage_id(in_embed.weight), storage_id(out_embed.weight)
    print(f"input  embedding storage: {in_store}")
    print(f"output embedding storage: {out_store}")

    assert in_store != out_store, (
        "BUG: input and output embeddings still share storage after "
        "untie_word_embeddings; save_pretrained will drop lm_head.weight"
    )
    print("OK: embeddings are untied (separate storage)")


if __name__ == "__main__":
    main()
```

### Output — before the fix (current `main`)

```
input  embedding storage: (device(type='cpu'), 764511680)
output embedding storage: (device(type='cpu'), 764511680)
  ...
AssertionError: BUG: input and output embeddings still share storage after untie_word_embeddings; save_pretrained will drop lm_head.weight
```

### Output — after the fix

```
input  embedding storage: (device(type='cpu'), 853154880)
output embedding storage: (device(type='cpu'), 853171328)
OK: embeddings are untied (separate storage)
```


kylesayrs (approved):
Thanks. This can be removed after
https://github.com/vllm-project/compressed-tensors/pull/709

brian-dellabetta (approved):

PR #2868, opened 2026-06-26
[Distributed] Add `suspend_distributed_timeout` to enable large model saving

kylesayrs:
## Purpose ##
* Support saving models which take more than 10 minutes to save
  * Longer save times can be exacerbated by very large models and disk/cpu offloading

## Changes ##
* Add `suspend_distributed_timeout`

## Testing ##
* Add `test_suspend_distributed_timeout`
* Now able to save models which take +20min to save

### DIFF
--- a/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
+++ b/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
@@ -208,8 +208,8 @@ def update_and_save_recipe(model_stub: str, save_directory: str):
 
 @contextmanager
 def suspend_distributed_timeout(
-    current_group: dist.ProcessGroup | None = None,
     timeout: datetime.timedelta = datetime.timedelta(hours=3),
+    current_group: dist.ProcessGroup | None = None,
 ):
     """
     Context manager that extends the timeout for distributed operations.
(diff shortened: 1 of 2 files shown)


HDCharles:
makes sense

HDCharles (approved):

HDCharles (approved):

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -202,3 +204,38 @@ def update_and_save_recipe(model_stub: str, save_directory: str):
...
+    :param current_group: The current process group to synchronize. If None,
+        defaults to dist.group.WORLD
+    """
+    if not dist.is_initialized():
+        yield
+        return
+
+    if current_group is None:
+        current_group = dist.group.WORLD
+    suspend_group = dist.new_group(backend="gloo", timeout=timeout)
```
brian-dellabetta, line 233: we can just hard-code "gloo" here?

kylesayrs: Yes, and it is safe as gloo backends are install as part of torch by default

Issue #2872, opened 2026-06-29
[RFC] Agent skill for hardware-aware quantization + example generation


2imi9:
**Is your feature request related to a problem? Please describe.**

Two recurring friction points when adopting LLM Compressor:

1. **"Which scheme should I use on my GPU?"** Picking between FP8 / NVFP4 / W4A16
   / INT8 / MXFP4 means knowing what your hardware can actually accelerate
   (FP4 → Blackwell, FP8 → Ada/Hopper+, INT8 → Turing+) and whether the model
   even fits VRAM for calibration *and* serving.
2. **Writing the first example.** Each new model/scheme means copying an
   `examples/*.py`, swapping model/recipe/scheme/ignore, and remembering current
   API details (e.g. AWQ now lives in `modifiers.transform.awq` paired with a
   `QuantizationModifier`).

**Describe the solution you'd like**

An optional **agent skill** (`SKILL.md` + small helper scripts) that:

- detects the local GPU (or a `--target` GPU) and recommends a scheme;
- estimates VRAM fit from HF metadata (no download);
- **generates a flat `examples/`-style script** for the chosen model/scheme —
  templates copied from the canonical examples, current APIs only — and can
  optionally run it;
- ships `verify_apis.py` to confirm every import/scheme it uses exists in the
  installed `llmcompressor` / `compressed-tensors`.

It **mirrors verified examples and delegates to real APIs** rather than
reimplementing library logic, so its output is itself a mergeable example. It is
also safe by construction: no telemetry, network limited to the HF reads any
example does, read-only hardware/verify steps, and it only writes the file you
name (no shell, no remote exec).

**Questions before a PR:**

1. Do you want an agent skill of this kind in the repo at all?
2. Placement — a new top-level `skills/`, the existing `.claude/skills/`, or
   `docs/`?
3. Is an optional NVIDIA Model Optimizer comparison welcome, or better left out
   to avoid an out-of-tree dependency?

**Describe alternatives you've considered**

- Keep the skill local and only contribute the **generated `examples/*.py`** it
  produces.
- Place it under `.claude/skills/` next to the existing `style.md` / `test.md`.

**Additional context**

A working version is up as a draft PR for reference (linked below). It is
API-correct against `main` and `ruff` / `ruff format`-clean; on an RTX 5090 the
generated FP8/NVFP4/AWQ scripts run end-to-end and produce vLLM-servable
compressed-tensors checkpoints. Happy to adjust placement and scope to your
preference.

[...]

PR #2851, opened 2026-06-22
[Agents] Add skills that can be used for automatic quantized checkpoint generation

dsikka:
# Summary
- Skills which write out examples for models, given a compression scheme
- Based on the LLM Compressor Compression Guide and existing examples 

# Testing
- Requires more testing

PR #2873, opened 2026-06-29
[Agents] Add quantization skill with hardware detection and example generation

2imi9:
SUMMARY:

Adds an agent skill under `.claude/skills/` (the location established by #2445)
that turns "make this model smaller/faster" into a correct, hardware-matched
quantization, and can generate a canonical `examples/`-style script.

It is a directory-form skill (`SKILL.md` + helper scripts) so it keeps the
executable tools rather than being prose-only:

- `detect_hardware.py` — GPU (local or `--target`) → recommended scheme
- `estimate_fit.py` — quantized footprint + VRAM fit from HF metadata (no download)
- `generate_example.py` — emit a canonical `examples/`-style script (optional `--run`)
- `verify_apis.py` — confirm every import/scheme is real in the install
- `benchmark_compare.py` — OPTIONAL dev tool, vs NVIDIA Model Optimizer
- `references/` — scheme-selection logic + ModelOpt comparison

### Try it

```bash
python .claude/skills/quantize-with-llmcompressor/scripts/detect_hardware.py
python .claude/skills/quantize-with-llmcompressor/scripts/generate_example.py \
    --model meta-llama/Llama-3.1-8B-Instruct --scheme FP8_DYNAMIC -o fp8_example.py
python .claude/skills/quantize-with-llmcompressor/scripts/verify_apis.py
```

### Low-risk to review and use

- **Mirrors verified repo code** — generator templates are copied from current
  `examples/` (FP8 / GPTQ / AWQ / NVFP4 / INT8 / vision); no library internals
  reimplemented.
- **Self-verifying** — `verify_apis.py` confirms every import and scheme string
  exists in the installed `llmcompressor` / `compressed-tensors`.
- **Safe by construction** — no telemetry; network limited to the HF reads any
  example performs; hardware/fit/verify steps are read-only; the generator only
  writes the file you name and runs it locally (no shell, no remote exec).
- `ruff` / `ruff format`-clean; lives under `.claude/` (outside `CHECKDIRS`).

Reshaped from an earlier top-level `skills/` proposal into `.claude/skills/` to
match the convention from #2445. Discussion: #2872.

TEST PLAN:

- Audited against `main`: import paths (`modifiers.transform.awq` /
  `modifiers.transform.smoothquant`), automatic MoE expert calibration, and
  scheme names vs `PRESET_SCHEMES`.
- `ruff check` / `ruff format --check` clean; `verify_apis.py` passes against the
  installed packages.
- On an RTX 5090 (Blackwell): generated FP8_DYNAMIC and AWQ (W4A16_ASYM) scripts
  run end-to-end and save vLLM-servable `compressed-tensors` checkpoints.


2imi9:
Thanks for the review. Pushed updates (and signed off all commits for DCO):

**Applied**
- **Sliding-window perplexity** (`benchmark_compare.py`): now tracks `prev_end`
  and scores only the new tokens per window, so overlapping context is
  conditioned on but never double-counted. I also re-ran the comparison and
  refreshed the numbers in `references/modelopt_comparison.md` (baseline 12.23;
  FP8 12.40 vs ModelOpt 12.41 — still a dead heat; NVFP4 14.84).
- **Gated-model UX** (`estimate_fit.py`): `model_info()` is wrapped in
  `try/except` with an actionable message (log in / set `HF_TOKEN`, or pass
  `--params N`).
- Also fixed a relative-link depth bug in `SKILL.md` (`../../../` → `../../`).

**Respectfully kept as-is: `dtype="auto"` and no `device_map`**
- This repo pins `transformers>=5.9` (`setup.py`), where the arg was renamed
  `torch_dtype` → `dtype` (`torch_dtype` now emits a deprecation warning). The
  README quick-tour uses `from_pretrained(MODEL_ID, dtype="auto")`, so the
  generated scripts match current repo style.
- The canonical text examples (`quantization_w8a8_fp8/llama3_example.py`,
  `quantization_w4a16/llama3_example.py`, `awq/llama_example.py`) call
  `from_pretrained(MODEL_ID)` **without** `device_map` and let `oneshot`
  handle device placement / sequential onloading. Adding `device_map="auto"`
  would diverge from those examples and can conflict with llmcompressor's own
  dispatch. The generated FP8 and AWQ scripts run end-to-end on GPU as-is.

Happy to adjust if you'd prefer otherwise. The core ask in #2872 still stands:
guidance on whether you want this skill in-repo and where it should live.

2imi9:
Reshaped to match the `.claude/skills/` convention from #2445: the PR is now a single `.claude/skills/quantization.md` (down from the earlier multi-file proposal). Rebased onto current `main` and signed off (DCO green). The earlier review points were folded in before slimming. Open question from #2872 remains: is a quantization skill welcome here? Happy to adjust the content.

2imi9:
Update: kept the executable tools (hardware detection, VRAM-fit estimate, example generator, API verifier) but moved the whole skill into `.claude/skills/quantize-with-llmcompressor/` as a directory-form `SKILL.md` + scripts, matching the `.claude/skills/` convention from #2445. Earlier I'd over-trimmed it to a single markdown file; this keeps the actual functionality while respecting the established location. DCO green, `ruff`/`ruff format` clean, `verify_apis.py` passes.

dsikka:
Left a comment on your PR but see: https://github.com/vllm-project/llm-compressor/pull/2851

2imi9:
Closing in favor of #2851, which is the maintainer's canonical source of truth for quantization example generation. Offered a complementary hardware-detection/fit helper + testing in #2872 if useful.

PR #2875, opened 2026-06-29
[Agents] Update fp8 skill with MoE/VLM architecture guidance and clarified defaults

dsikka:
## What this PR does

Adds a Claude Code skill (`/fp8`) that generates a working FP8 quantization example script for any HuggingFace model and saves a compressed-tensors checkpoint. When a user asks to quantize a model to FP8, the skill gathers the necessary inputs, selects the right template, and writes a ready-to-run `*_example.py` file into the appropriate `examples/` subdirectory.

The skill handles three model types and three FP8 schemes:

**Model types:**
- Dense — uses `AutoModelForCausalLM` with `oneshot`
- MoE — fetches `config.json` to get the exact architecture class, requires gate/router layer names before writing (quantizing gates degrades routing)
- Multimodal (vision/audio) — fetches `config.json` for the specific class, ignores vision/audio towers

**Schemes:**
- `FP8_BLOCK` — 128×128 block-scaled weights, dynamic activations; preferred for Hopper/Blackwell
- `FP8_DYNAMIC` — per-channel weights, dynamic per-token activations; fallback with broadest hardware support (Ampere+)
- `MXFP8` — MX fp8 format targeting AMD MI300X

**Two quantization paths:**
- `oneshot` — standard path for models with a transformers class definition
- `model_free_ptq` — for custom architectures not in transformers, or ~1TB+ models that should be quantized directly from safetensors

Neither scheme requires a calibration dataset.


examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+5. **Architecture class (MoE and VLM models only, REQUIRED)** — for any MoE or multimodal model, fetch the model's `config.json` from HuggingFace to get the exact architecture class before writing the file:
+   - Fetch `https://huggingface.co/{MODEL_ID}/raw/main/config.json` and read the `architectures` field (e.g. `["LagunaForCausalLM"]`, `["Llama4ForConditionalGeneration"]`)
+   - Use that class directly in the import and `from_pretrained` call instead of `AutoModelForCausalLM`
+   - To check if the class is in standard `transformers`, run: `python -c "from transformers import <ClassName>"`. If it imports successfully, no `trust_remote_code` needed. If it fails with `ImportError`, add `trust_remote_code=True` to both `from_pretrained` calls.
+   - Do **not** use `AutoModelForCausalLM` for MoE or VLM models — always use the specific class
+
+6. **Gate/router layer names (MoE models only, REQUIRED)** — for any MoE model you **must** determine the exact names of the gate/router layers before writing the file. Do one of the following:
```
HDCharles, line 31: we have a bunch of mappings that give layer names, should we direct the LLM to check that?

HDCharles (approved):
see nit, otherwise lgtm

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+        MODEL_ID, trust_remote_code=True, low_cpu_mem_usage=True
+    )
+```
+Common MoE gating patterns (always verify against the model's actual layer names before using):
+- Qwen MoE: `"re:.*mlp.gate$"`, `"re:.*shared_expert_gate.*"`
+- Llama4 / Gemma4 MoE: `"re:.*router"`
+- Laguna-M.1: `"re:.*mlp\.gate$"` (targets `mlp.gate` routing layers; avoids false match on `gate_proj`)
+
+If the gate/router layer is an `nn.Parameter` (not `nn.Linear`) it will not be targeted by `targets="Linear"`, but it should still be explicitly listed in `ignore` to document intent and guard against future changes.
+
```
kylesayrs:
```suggestion
Some deepseek-like architectures use an `attn.indexer` and `attn.indexer.compressor`. These weights are sensitive to quantization and should be ignored.
```

- DeepSeekV3.2: `"re:.*indexer.*"`

kylesayrs: Technically in DSV32, `attn.indexer.wq_b` also known as `attn.indexer.q_b_proj` is quantized to fp8, but we can skip it.

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+### Multimodal (vision / audio)
+- Use `AutoProcessor` instead of `AutoTokenizer`
+- Always fetch the model's `config.json` to get the specific class from the `architectures` field (e.g. `Gemma4ForConditionalGeneration`, `Llama4ForConditionalGeneration`) — **never use `AutoModelForCausalLM` for VLMs**
+- Import and use that class directly; wrap in `load_context`:
+  ```python
+  from transformers import Llama4ForConditionalGeneration  # example
+  from llmcompressor.utils import load_context
+
+  with load_context(Llama4ForConditionalGeneration):
+      model = Llama4ForConditionalGeneration.from_pretrained(MODEL_ID, low_cpu_mem_usage=True)
```
kylesayrs: `low_cpu_mem_usage` is removed in v5.

```suggestion
      model = Llama4ForConditionalGeneration.from_pretrained(MODEL_ID)
```

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+  ```
+- Add `trust_remote_code=True` if the class is not in standard transformers
+- Always ignore vision tower layers using a single regex that covers the model's actual layer names (verify against the architecture — one pattern is sufficient since regex search matches substrings):
+  - Vision: `"re:visual.*"` (matches `visual.*` and `model.visual.*`)
```
kylesayrs: Transformers is mostly standardized on `visual` at this point, but some models like llava still have `vision_tower`.

```suggestion
  - Vision: `["re:.*visual.*, ".*vision_tower.*"]`
```

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+print("==========================================")
+
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-<SCHEME>"
+model.save_pretrained(SAVE_DIR)
+tokenizer.save_pretrained(SAVE_DIR)
+```
+
+### `model_free_ptq` (no transformers model definition, or very large models ~1TB+)
+
+```python
```
kylesayrs: Consider breaking out templates into separate files under skills/fp8

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+Always wrap the load in `load_context` using the specific architecture class:
+```python
+from transformers import SpecificClass  # e.g. Llama4ForConditionalGeneration
+from llmcompressor.utils import load_context
+
+with load_context(SpecificClass):
+    model = SpecificClass.from_pretrained(MODEL_ID, low_cpu_mem_usage=True)
+```
+
+For custom architectures not in standard transformers (requires `trust_remote_code=True`), pass the class name string or use `AutoModelForCausalLM` as the proxy — but always add `trust_remote_code=True`:
```
kylesayrs:
```suggestion
For custom architectures not in standard transformers (requires `trust_remote_code=True`), pass the class name string or use `AutoModelForCausalLM` as the proxy — but always add `trust_remote_code=True` if required:
```

kylesayrs: I'm actually not sure what you mean here. Is Llama4ForConditionalGeneration not "in standard transformers"?

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+print("==========================================")
+
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-<SCHEME>"
+model.save_pretrained(SAVE_DIR)
+tokenizer.save_pretrained(SAVE_DIR)
+```
+
+**MoE and VLM models** — always use the specific architecture class from `config.json`. **Never use `AutoModelForCausalLM` for MoE or VLM models.**
+
+```python
```
kylesayrs: Consider breaking out templates into separate files under skills/fp8

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+
+MODEL_ID = "<MODEL_ID>"
+
+model = AutoModelForCausalLM.from_pretrained(MODEL_ID, low_cpu_mem_usage=True)
+tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
+
+recipe = QuantizationModifier(
+    targets="Linear",
+    scheme="<SCHEME>",  # FP8_DYNAMIC | FP8_BLOCK | MXFP8
+    ignore=["lm_head"],
```
kylesayrs:
```suggestion
    ignore=["lm_head"],  # add model-specific vision tower and gating/routing layers — see Step 3
```

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+output = model.generate(input_ids, max_new_tokens=20)
+print(tokenizer.decode(output[0]))
+print("==========================================")
+
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-<SCHEME>"
+model.save_pretrained(SAVE_DIR)
+tokenizer.save_pretrained(SAVE_DIR)
+```
+
+**MoE and VLM models** — always use the specific architecture class from `config.json`. **Never use `AutoModelForCausalLM` for MoE or VLM models.**
```
kylesayrs: `AutoModelForCausalLM` is fine/ required for most MoE models, right???

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+   - Once you have the layer name, construct the regex as follows:
+     - Escape any literal dots: `mlp.gate` → `mlp\.gate`
+     - Anchor with `$` to prevent partial matches on layers sharing the same prefix (e.g. `mlp\.gate$` matches `mlp.gate` but not `mlp.gate_proj`)
+     - Prepend `.*` to match at any depth: `"re:.*mlp\.gate$"`
+
+## Step 2 — Choose the template
+
+### `oneshot` with `QuantizationModifier` (standard path)
+
+**Dense models** — use `AutoModelForCausalLM`:
```
kylesayrs: Why break these into separate parts? Why not make it simple and give both the same template?

examples/.claude/skills/fp8/SKILL.md
kylesayrs: For `model_free_ptq`, you might want to ask the model to inspect the keys using [`inspect_tensors.py`](https://github.com/vllm-project/llm-compressor/pull/2753/changes#diff-3dadf498e923c99bb16fb32c70644e1a09f25a82661ab65f734daf4d3cf74aa7). Very often the key names in the checkpoint aren't going to match the names used by transformers

dsikka: Like model_free_ptq vs oneshot?

examples/.claude/skills/fp8/SKILL.md
dsikka: Updated - would you want it simpler than this?

examples/.claude/skills/fp8/SKILL.md
dsikka: Thanks - let me see about adding it in a follow-up

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+```
+Common MoE gating patterns (always verify against the model's actual layer names before using):
+- Qwen MoE: `"re:.*mlp.gate$"`, `"re:.*shared_expert_gate.*"`
+- Llama4 / Gemma4 MoE: `"re:.*router"`
+- Laguna-M.1: `"re:.*mlp\.gate$"` (targets `mlp.gate` routing layers; avoids false match on `gate_proj`)
+
+If the gate/router layer is an `nn.Parameter` (not `nn.Linear`) it will not be targeted by `targets="Linear"`, but it should still be explicitly listed in `ignore` to document intent and guard against future changes.
+
+### Multimodal (vision / audio)
+- Use `AutoProcessor` instead of `AutoTokenizer`
```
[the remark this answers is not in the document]
kylesayrs: That's not true, all the processors I've seen also accept text-only inputs

examples/.claude/skills/fp8/SKILL.md
```diff
@@ -0,0 +1,201 @@
...
+dispatch_model(model)
+input_ids = processor("Hello my name is", return_tensors="pt").input_ids.to(model.device)
+output = model.generate(input_ids, max_new_tokens=20)
+print(processor.decode(output[0]))
+print("==========================================")
+
+SAVE_DIR = MODEL_ID.rstrip("/").split("/")[-1] + "-<SCHEME>"
+model.save_pretrained(SAVE_DIR)
+processor.save_pretrained(SAVE_DIR)
+```
```
kylesayrs: Sample generation will cause an OOM for very large models. Consider asking the agent to remove the sample generation step if the model is too large, or coding it into the template

```python3
with disable_onloading():
  num_parameters = model.num_parameters()
if num_parameters <= 60e9:
  # do sample generation
```

examples/.claude/skills/fp8/SKILL.md
kylesayrs: Yes. It seems like right now you have two oneshot templates We should use the same template in both cases, and that one template should use `load_context` and `AutoProcessor`.

examples/.claude/skills/fp8/SKILL.md
kylesayrs: Yeah, one file for model_free_ptq template and one file for oneshot. See https://github.com/vllm-project/llm-compressor/pull/2753

kylesayrs: The benefit is that the model only has to read the template corresponding to which file it wants to use.

examples/.claude/skills/fp8/SKILL.md
dsikka: makes sense

examples/.claude/skills/fp8/SKILL.md
dsikka: fair but will do it in a follow-up once your skill lands

PR #2695, opened 2026-05-11
Fix AutoRound ignore-layer metadata handling and add Qwen3-30B to mxfp8 example.

changwangss:
**Summary**
This PR mainly includes two targeted fixes:

1. Ensure AutoRound ignored FP layers are not re-marked as quantized by clearing/restoring quantization metadata correctly in post-processing.
2. Update the Qwen3 MXFP8 example for `Qwen/Qwen3-30B-A3B-Instruct-2507`.

**bug fix**

Although mlp.gate is listed in AutoRound predefined ignore layers and reported as ignored in logs, it could still appear quantized after LLMC post-processing.
Root cause was that LLMC restored legacy qparams/metadata even for layers that AutoRound had already finalized as FP.
This fix skips qparam restoration for FP-finalized layers and clears stale quantization metadata, so ignored MoE gate layers remain truly unquantized end-to-end.

**Model Test**
test mxfp8 example，the mxfp8 Qwen3-30B-A3B-Instruct-2507 model has uploaded to https://huggingface.co/changwangss/Qwen3-30B-A3B-Instruct-2507-W8A8-MXFP8-AutoRound.
```
root@12bc54b086b6:/data3/changwa1/qwen_example/llmc/examples/autoround/quantization_w8a8_mxfp8# CUDA_VISIBLE_DEVICES=5 p
ython qwen3_example.py
Failed to load /usr/local/lib/python3.12/dist-packages/torchao/_C_cutlass_100a.abi3.so: Could not load this library: /us
r/local/lib/python3.12/dist-packages/torchao/_C_cutlass_100a.abi3.so
Failed to load /usr/local/lib/python3.12/dist-packages/torchao/_C.abi3.so: Could not load this library: /usr/local/lib/p
ython3.12/dist-packages/torchao/_C.abi3.so
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████| 16/16 [00:01<00:00,  8.16it/s]
2026-05-12 02:44:43 INFO calib_dataset.py L912: Preprocessing calibration dataset in a subprocess to avoid memory leaks.
..
2026-05-12T02:44:57.2141 | __init__ | WARNING - Disabling tokenizer parallelism due to threading conflict between FastTo
kenizer and Datasets. Set TOKENIZERS_PARALLELISM=false to suppress this warning.
2026-05-12T02:44:57.6223 | reset | INFO - Compression lifecycle reset
[...]

PR #2877, opened 2026-06-30
disable moe_calibrate_all_experts for Autoround example.

changwangss:
SUMMARY:

1. `llmcompressor` now applies `linearize_moe` in oneshot, so MoE experts may already be linearized before `AutoRound` runs.
Because of this, we added a skip in `AutoRound`  (https://github.com/intel/auto-round/pull/1966) to avoid redundant fused_moe replacement when linearized experts are detected.
For Llama4, the default linearized flow calibrates all experts at once, which can OOM on an 80GB GPU.
Setting `moe_calibrate_all_experts=False` avoids this, and with 128 samples + 200 iterations we still get sufficient expert coverage during calibration.

2. `moe_calibrate_all_experts=False` for `Qwen-30B-A3B mxfp8` was already introduced in an earlier PR ([vllm-project/llm-compressor#2695]. It was likely lost after multiple example updates, so this change restores it.

TEST PLAN:


yiliu30 (approved):
LGTM, thanks!

brian-dellabetta (approved):

kylesayrs:
> For Llama4, the default linearized flow calibrates all experts at once, which can OOM on an 80GB GPU.

In my understanding, AutoRound's memory usage is purely a function of (number of gradients which is proportional to number of activated experts), NOT the number of samples

The samples are reduced into gradients, so as long as each expert gets at least one sample, then adding more samples via calibrate_all_experts shouldn't increase memory usage, right?

As long as each expert gets at least one sample, I don't see a reason why `moe_calibrate_all_experts=False` would decrease memory usage.

kylesayrs:
I think the things I need to understand are
1. Why does `moe_calibrate_all_experts=False` use less memory than `moe_calibrate_all_experts=True`?
2. Is there ever a risk that a module goes uncalibrated or badly calibrated? What if someone modifies the example to use (a different model/ a different dataset/ a different seed)?
3. What if someone want to use activation quantization with AutoRound? Will experts with very few activations fail to calibrate correctly?

changwangss:
sorry for the late response. Let me answer the questions.

> 1. Why does moe_calibrate_all_experts=False use less memory than moe_calibrate_all_experts=True?

The main memory usage is not from the final gradient tensors. With moe_calibrate_all_experts=True, every expert processes all tokens, producing much larger QDQ inputs and intermediate activations. With False, each token is processed only by its top-8 routed experts, significantly reducing peak memory.

> 2. Is there ever a risk that a module goes uncalibrated or badly calibrated? What if someone modifies the example to use (a different model/ a different dataset/ a different seed)?

With 128 calibration samples, a sequence length of 2048, and top-8 routing, each MoE layer receives approximately 2.1 million expert-token assignments. In our DeepSeek-R1 test with the same calibration configuration, every expert was activated. The distribution was not perfectly uniform, as expected for MoE routing, but no expert was extremely underrepresented. Therefore, we expect this configuration to provide sufficient coverage in practice, although the exact distribution remains model- and dataset-dependent.

> 3. What if someone want to use activation quantization with AutoRound? Will experts with very few activations fail to calibrate correctly?

The current MXFP8 example uses dynamic activation quantization, where activation scales are computed at runtime, See the [MXFP8 configuration](https://github.com/intel/auto-round/blob/v0.14.0/auto_round/schemes.py#L416-L426)
For static activation quantization, fewer activations may make the collected range less representative. If an expert is not activated at all, AutoRound fills its missing amax using the maximum value from calibrated experts, see the [set_amax_for_uncalibrated_experts()](https://github.com/intel/auto-round/blob/v0.14.0/auto_round/utils/model.py#L1709-L1782)

brian-dellabetta:
@yiliu30 are you satisfied with the changes and explanation for us to merge this in?

kylesayrs (approved):
@changwangss Good to merge from my side, just adding my comments here

> 1. With moe_calibrate_all_experts=True, every expert processes all tokens, producing much larger QDQ inputs and intermediate activations

Note that LLM Compressor [linearizes experts](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/moe/linear_experts.py). This means that experts are executed sequentially, and therefore processing more experts does not increase peak memory usage, only increases runtime.

Even when `sequential_targets="ExpertWithGate"` and you care about offload memory, the sequential pipeline's tracing will guarantee that the expert outputs are added to the fixed size summing tensor, not accumulated before adding.

> 2. In our DeepSeek-R1 test with the same calibration configuration, every expert was activated.

That's interesting to hear. We initially added this feature for DSV3 and found that many experts received insufficient samples. This could be reflective of either models having better expert distribution, or autoround being more resilient to low calibration data than GPTQ

> 3. If an expert is not activated at all, AutoRound fills its missing amax using the maximum value from calibrated experts

Sounds good

changwangss:
All required checks and merge protections are passing now. Could a maintainer please merge this PR? @brian-dellabetta

yiliu30 (approved):
Looks good from my side. Just two points to clarify.

For expert imbalance, disabling `moe_calibrate_all_experts` should be less of a concern for AutoRound. Most experts are expected to receive routed tokens, and AutoRound has a fallback for missing amax. We have not observed a quality issue so far.

For memory, I agree with Kyle that the linearized experts run sequentially, so expert memory should not accumulate into the peak.

Given this, I think we can keep it disabled for now and revisit it if enabling it shows a measurable accuracy benefit.

changwangss:
> Note that LLM Compressor [linearizes experts](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/moe/linear_experts.py). This means that experts are executed sequentially, and therefore processing more experts does not increase peak memory usage, only increases runtime.

> Even when sequential_targets="ExpertWithGate" and you care about offload memory, the sequential pipeline's tracing will guarantee that the expert outputs are added to the fixed size summing tensor, not accumulated before adding.

@kylesayrs  Thanks for the clarification. I agree that the linearized experts run sequentially. Two details still cause higher peak memory:

1.  With all-expert calibration, each expert processes the full token tensor before selecting its routed tokens:
```
if get_calibrate_all_experts_flag():
    expert_output = expert(hidden_states)[token_indices]
else:
    expert_output = expert(hidden_states[token_indices])
```
Therefore, each expert creates larger QDQ and MLP intermediate tensors.

2. AutoRound tuning runs with gradients enabled, and `backward()` is called only after the full block forward. Although outputs are immediately accumulated into a fixed-size tensor, the tensors saved for backward across expert calls cannot be released. The retained expert-token computation grows from roughly `N × top_k` to `N × num_experts`.

This explains the observed OOM on the 80GB A100.

References: [LLM Compressor expert forward](https://github.com/vllm-project/llm-compressor/blob/870c529f49cb4a53d408cf7d51cfe286a3a78af1/src/llmcompressor/modeling/moe/linear_experts.py#L248-L265) and [AutoRound forward/backward](https://github.com/intel/auto-round/blob/v0.14.0/auto_round/algorithms/quantization/sign_round/quantizer.py#L225-L239).

PR #2878, opened 2026-06-30
[Agents] Add quantization skills: kv-cache, attention, w4a8, autoround, embedding, model-free-ptq

2imi9:
SUMMARY:

Adds agent skills under `examples/.claude/skills/`, following the same template as the `fp8` / `int` / `mxfp4` / `nvfp4` skills in #2851, covering quantization capabilities that set doesn't yet:

- **kv-cache** — FP8 KV cache (optionally with FP8 weights/activations), incl. per-head note
- **attention** — FP8 attention-activation quantization (targets the attention module class)
- **w4a8** — `W4AFP8` (GPTQ, calibrated) and `W4A8` (RTN, data-free) as two distinct templates
- **autoround** — AutoRound learned weight rounding (W4A16 / NVFP4 / MXFP4)
- **embedding** — weight-only int4/int8 of the `Embedding` table (data-free)
- **model-free-ptq** — data-free PTQ for very large models directly from safetensors (FP8 / NVFP4A16 / MXFP4A16), incl. the `reindex_fused_weights` step

Each mirrors the corresponding `examples/` script and uses current APIs. The skill directories are independent of #2851's, so this complements that PR.

TEST PLAN:

Smoke-tested every template end-to-end on an RTX 5090 (Blackwell, sm_120; llmcompressor 0.12.0) with `Qwen/Qwen2.5-0.5B-Instruct` and reduced calibration samples. Each ran and saved a valid compressed-tensors checkpoint:

| Skill | Result | Format |
|---|---|---|
| kv-cache | ✅ | `float-quantized` + `kv_cache_scheme` |
| attention | ✅ | attention `input_activations` quantized |
| w4a8 (W4AFP8, GPTQ) | ✅ | `pack-quantized` |
| w4a8 (W4A8, data-free RTN) | ✅ | `pack-quantized` |
| autoround (W4A16) | ✅ | `pack-quantized` |
| embedding (int4 group64) | ✅ | `pack-quantized` |
| model-free-ptq (FP8_DYNAMIC) | ✅ | `float-quantized` |

### DIFF
--- a/examples/.claude/skills/attention/SKILL.md
+++ b/examples/.claude/skills/attention/SKILL.md
@@ -1,0 +1,136 @@
+---
+name: attention
+description: >
+  Generate a working attention quantization example script (FP8 attention activations)
+  and save a compressed-tensors checkpoint.
+  Triggers on: "attention quantization", "quantize attention", "fp8 attention", "attention fp8", "qkv quantization".
+allowed-tools: [Read, Write, Glob, Bash(make style), Bash(ls *), Bash(find *)]
+---
+
+# Write Attention Example
+
+Generate a working Python example script that quantizes the attention activations
+of a model to FP8 and saves a compressed-tensors checkpoint.
+
+Attention quantization targets the model's attention module directly (not the
+`Linear` layers), so the target is the **attention class name** for the model
+architecture. It requires calibration data.
+
+## Step 1 — Gather information
+
+Ask the user (or infer from context) for:
+
+1. **MODEL_ID** — HuggingFace model ID (e.g. `meta-llama/Meta-Llama-3-8B-Instruct`)
+2. **Attention module class** — the model-specific attention class to target, e.g.
+   - Llama: `LlamaAttention`
+   - Qwen2 / Qwen2.5: `Qwen2Attention`
+   - Mistral: `MistralAttention`
+   Find it with `print(type(model.model.layers[0].self_attn).__name__)`.
+3. **Model type** — dense, MoE, or multimodal.
+
+Attention quantization always requires a calibration dataset.
+
+## Step 2 — Choose the template
+
+```python
+from compressed_tensors.offload import dispatch_model
+from compressed_tensors.quantization import QuantizationArgs, QuantizationScheme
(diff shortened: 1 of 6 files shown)


2imi9:
Thanks @gemini-code-assist — addressed:

- ✅ `mmlu` → `MMLU`.
- The other two are correct against current `main` (verified the files exist), so kept:
  - `qwen3_vl_example.py` does exist at `examples/autoround/quantization_w4a4_fp4/` — Qwen3 is in the repo.
  - "GPT-OSS" refers to the openai/gpt-oss model (`examples/quantization_w4a8/gpt_oss_20b_example.py`), not a typo for GPTQ — reworded to the explicit filename to avoid the confusion.

2imi9:
Corrected the `w4a8` skill: W4A8 and W4AFP8 are genuinely distinct, so it now has two separate templates rather than one. Both schemes use *dynamic* activations, so the calibration requirement comes from the **algorithm**, not the scheme:

- **W4AFP8** (int4 + fp8 dynamic acts) — `GPTQModifier`, calibration required (mirrors `examples/quantization_w4a8_fp8/`).
- **W4A8** (int4 + int8 dynamic acts) — `QuantizationModifier` (RTN), **data-free** (mirrors `examples/quantization_w4a8/gpt_oss_20b_example.py`).

My earlier single template wrongly said W4A8 needed GPTQ + calibration. Both paths now verified end-to-end on a 5090 (→ `pack-quantized`).

2imi9:
Added two more from the gap list, both data-free and verified on a 5090:

- **embedding** — weight-only int4 of the `Embedding` table (→ `pack-quantized`), mirrors `examples/quantization_embedding/`.
- **model-free-ptq** — data-free PTQ straight from safetensors for very large models (→ `float-quantized` with FP8_DYNAMIC), mirrors `examples/model_free_ptq/`, incl. the microscale `reindex_fused_weights` step.

That covers the remaining clear gaps in the format/technique axes (`transform` and the big-model run-strategies like sequential-onloading/disk-offload are the only ones left, and those are more 'how to run' than 'which scheme').

PR #2888, opened 2026-07-02
validate quantization for fused layers

HDCharles:
Summary:

first pass proof of concept

### DIFF
--- a/src/llmcompressor/modifiers/quantization/scheme_consistency_validation.py
+++ b/src/llmcompressor/modifiers/quantization/scheme_consistency_validation.py
@@ -1,0 +1,196 @@
+"""
+Validate that logically-related layers receive consistent quantization schemes.
+
+Fused layer groups (QKV attention projections, gate/up MLP projections) and MoE
+expert replicas should normally share the same quantization configuration.
+Mismatches usually indicate a recipe misconfiguration — e.g. a target pattern
+that accidentally covers only some projections in a group.
+
+This module warns (rather than errors) because advanced users may intentionally
+assign different schemes to related layers.
+"""
+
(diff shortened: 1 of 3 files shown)


src/llmcompressor/modifiers/quantization/scheme_consistency_validation.py
```diff
@@ -0,0 +1,196 @@
...
+    :return: list of (experts_fqn, sub_layer_name, {expert_idx: scheme_signature})
+    """
+    mismatches = []
+
+    for parent_name, parent_module in model.named_modules():
+        if not (
+            isinstance(parent_module, torch.nn.ModuleList)
+            and parent_name.endswith("experts")
+        ):
+            continue
```
kylesayrs, line 118:
```suggestion
    for parent_name, parent_module in `matched_named_modules(model, ["LinearExperts2D"]):
        assert isinstance(parent_module, LinearExperts2D)
```

src/llmcompressor/modifiers/quantization/scheme_consistency_validation.py
```diff
@@ -0,0 +1,196 @@
...
+
+        sublayer_schemes: dict[str, dict[str, QuantizationScheme]] = defaultdict(dict)
+
+        for expert_idx, expert_module in enumerate(parent_module):
```
kylesayrs, line 122: Technically, iterating through the experts will also get the activation function https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modeling/moe/linear_experts.py#L217

We should make this better by overriding `LinearExperts2D.__iter__` to only include experts

PR #2889, opened 2026-07-03
[Agents] Add NVFP4 skills for the `QuantizationModifier` and `GPTQModifier`

dsikka:
SUMMARY:
- Consolidate shared logic for model handling for NVFP4 and FP8 into a shared filed
- Add logic to handle NVFP4 for the QuantizationModifier and GPTQModifier


brian-dellabetta (approved):
cool, i like the folder layout

PR #2890, opened 2026-07-06
[Tests] [Bugfix] Skip MoE tests for models which cannot be imported

kylesayrs:
## Purpose ##
* Skip `test_linearize_moe` if model is not supported by local transformers version
* Skip `test_load_quantizable_moe` if model is not supported by local transformers version
* Reduce memory usage for `deepseek_v4`

## Testing ##
* Tested that all tests pass, with 3 skipped when `transformers==5.9.0`
```
tests/llmcompressor/modeling/test_linearize.py ..s..............s....s...........                                [100%]
====================================== 31 passed, 3 skipped, 14 warnings in 1.79s ======================================
```

### DIFF
--- a/tests/llmcompressor/modeling/test_linearize.py
+++ b/tests/llmcompressor/modeling/test_linearize.py
@@ -35,6 +35,11 @@ CONFIG_OVERRIDES = {
     "hy_v3": {"hidden_size": 256, "moe_intermediate_size": 256, "num_experts": 16},
     "jamba": {"hidden_size": 256, "intermediate_size": 256, "num_experts": 16},
     "nemotron_h": {"hidden_size": 32, "moe_intermediate_size": 64},
+    "deepseek_v4": {
+        "hidden_size": 512,
+        "moe_intermediate_size": 64,
+        "n_routed_experts": 16,
+    },
 }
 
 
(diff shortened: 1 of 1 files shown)


tests/llmcompressor/modeling/test_linearize.py
```diff
@@ -84,6 +85,11 @@ def patch_deepseek_fp32_modules():
 def test_load_quantizable_moe(
     model_stub, exp_keys, tmp_path, patch_deepseek_fp32_modules
 ):
+    try:
+        AutoConfig.from_pretrained(model_stub)
+    except StrictDataclassError:
+        pytest.skip("Could not import model, please upgrade your transformers version")
```
kylesayrs, line 96: We can generalize this later

brian-dellabetta (approved):

dsikka (commented):
Wouldn’t it be easier to skip just based on version?

kylesayrs:
@dsikka I don't want to have to figure out which transformers version each of these model architectures were added

dsikka (approved):

kylesayrs:
Tests are failing because need https://github.com/vllm-project/compressed-tensors/pull/767 or https://github.com/vllm-project/compressed-tensors/pull/768

PR #2900, opened 2026-07-08
[GHA] Remove `ready-label-check.yaml`

kylesayrs:
The check ready label check is no longer needed, as a PR cannot be merged unless buildkite runs.

Note, the test still runs on this PR because the GHA is a `pull_request_target`, meaning that the workflow triggers from the base (main), where the action still exists


kylesayrs:

dsikka (changes requested):
We still need this

kylesayrs:
@dsikka Right, but we no longer need this because a pending buildkite will block merging now; This does nothing but make it harder to check if a PR is failing at a glance

dsikka (approved):

PR #2906, opened 2026-07-09
Add MoNe pruning support

ieBoytsov:
## Summary

The PR adds [MoNE](https://arxiv.org/pdf/2507.00390) (Mixture of Experts and Novices) pruning support for MoE models. MoNe has an original codebase https://github.com/zxgx/mode-pd from paper authors, we rely on it in our integration and provide a benchmark comparison. 

The MoNe method is close to [REAP](https://arxiv.org/html/2510.13999v1) but instead of removing redundant experts it replaces them with constant vectors that are their unbiased estimators. In our extensive experiments MoNE outperforms REAP on a wide range of benchmarks.

The implementation is tested on `allenai/OLMoE-1B-7B-0924` following the paper experiments and achieves comparable average score across 9 benchmarks from the paper. We also add support for a larger model, particularly minimax-m2.5.

Details:

- Adds a `MoNEPruningModifier` that collects calibration statistics, ranks experts, preserves the top experts, and replaces pruned experts with approximate constant experts.
- Adds generic MoNE model support/registry utilities plus MiniMax-M2 support including native FP8 expert handling.
- Wires MoNE into the pruning modifier registry, pipeline registry, and model loading path for saved MoNE checkpoints.
- Adds checkpoint export validation and fixes cross-device embedding re-tying during save.
- Adds unit coverage for MoNE modifier behavior, MiniMax MoNE layout/export logic, linearized loading, pipeline registry integration, and save utilities.

## Testing

- `pytest tests/llmcompressor/modifiers/pruning/mone/test_base.py tests/llmcompressor/modeling/test_minimax_mone.py tests/llmcompressor/modeling/test_linearize.py::test_olmoe_uses_olmoe_experts_for_linearized_loading -q`

## Benchmarking

OLMoE model scores using original codebase (mode-pd) vs llm-compressor implementation. The scores on individual benchmarks fluctuate a bit but the average across 9 tasks is comparable. 

| Benchmark     | Metric   | mode-pd | llm-compressor | Delta   |
|---------------|----------|--------:|---------------:|--------:|
| ARC-Challenge | acc_norm | 0.4266  | 0.4454         | +0.0188 |
| ARC-Easy      | acc_norm | 0.6465  | 0.7256         | +0.0791 |
| BoolQ         | acc      | 0.6703  | 0.6388         | -0.0315 |
| COPA          | acc      | 0.8600  | 0.8500         | -0.0100 |
| MMLU          | acc      | 0.4029  | 0.3864         | -0.0165 |
| OpenBookQA    | acc_norm | 0.4080  | 0.4080         | +0.0000 |
| PIQA          | acc_norm | 0.7786  | 0.7813         | +0.0027 |
| RTE           | acc      | 0.6462  | 0.6859         | +0.0397 |
| Winogrande    | acc      | 0.6740  | 0.6598         | -0.0142 |
| **Average**   | **mean** | **0.6126** | **0.6201** | **+0.0076** |

## Example command

Run 25% pruning of OlMoE model using zyda2 dataset and 100 calibration steps.

```
from llmcompressor.modifiers.pruning.mone import MoNEPruningModifier

recipe = MoNEPruningModifier(
    preserve_n_experts=48,      # OLMoE has 64 experts/layer; prune 25%
    ranking_metric="fusion",
    ranking_scope="layer",
    fusion_io_weight=0.5,
    stats_device="cpu",
)

oneshot(
    model="allenai/OLMoE-1B-7B-0125",
    recipe=recipe,
    dataset="Zyphra/Zyda-2",
    dataset_config_name="sample-100BT",
    splits="train[:512]",
    text_column="text",
    concatenate_data=True,
    max_seq_length=4096,
    num_calibration_samples=100,
    shuffle_calibration_samples=False,
    batch_size=1,
    moe_calibrate_all_experts=False,
    pipeline="basic",
    precision="bfloat16",
    save_compressed=False,
    output_dir="./OLMoE-1B-7B-0125-mone-pruned-48",
)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for loading, pruning, and exporting MoNE/MiniMax-M2 mixture-of-experts models.
  * Added a new pruning workflow that can replace selected experts with lightweight “novice” experts while preserving model behavior.
  * Improved save/load handling for compressed models, including safer checkpoint export and local checkpoint support.

* **Bug Fixes**
  * Improved handling for distributed/offloaded models during conversion and saving.
  * Fixed pipeline selection so MoNE pruning uses the appropriate calibration pipeline.
  * Added better processor/source detection for MoNE checkpoints.


src/llmcompressor/modeling/moe/minimax_mone.py
Ryfernandes, line 1: Is this for the purpose of loading MoNE checkpoints into LLM Compressor for further compression? I'm a bit confused about the purpose of such a large file. Would separate helper files be needed for different MoNE-pruned checkpoints, depending on the model type?

src/llmcompressor/modifiers/pruning/mone/base.py
```diff
@@ -0,0 +1,430 @@
...
+
+        return True
+
+    def on_calibration_start(self, state: State, event: Event, **kwargs):
+        if get_calibrate_all_experts_flag():
+            raise ValueError(
+                "MoNEPruningModifier requires moe_calibrate_all_experts=False. "
+                "MoNE's output means and fluctuations must be collected over "
+                "routed tokens only to match the reference implementation."
+            )
```
Ryfernandes, line 125: In REAP, we are able to support both values of moe_calibrate_all_experts by explicitly filtering the top_k experts of each token during accumulation. Could something similar be done rather than this ValueError?

src/llmcompressor/modifiers/pruning/mone/base.py
```diff
@@ -0,0 +1,430 @@
...
+                        expert,
+                        partial(self._expert_hook, layer_name, idx),
+                        "forward",
+                    )
+
+            if tracker.needs_routing_stats:
+                self.register_hook(
+                    experts,
+                    partial(self._experts_block_hook, layer_name),
+                    "forward",
```
Ryfernandes, line 168: Why forward for this hook?

src/llmcompressor/modifiers/pruning/mone/base.py
```diff
@@ -0,0 +1,430 @@
...
+
+    def on_calibration_end(self, state: State, event: Event, **kwargs):
+        for adapter in self._fp8_adapters.values():
+            adapter.remove_calibration()
+        self.remove_hooks()
+        self._apply_mone(state.model)
+
+    def on_finalize(self, state: State, **kwargs) -> bool:
+        for adapter in self._fp8_adapters.values():
+            adapter.remove_calibration()
```
Ryfernandes: Why the duplication of lines 207-208?

src/llmcompressor/modifiers/pruning/mone/base.py
```diff
@@ -0,0 +1,430 @@
...
+    MoNE keeps the original router dimensionality and logical expert ids. Experts
+    selected for pruning are replaced by a constant-output module whose value is
+    the calibration-set mean output of the original expert.
+
+    :param preserve_n_experts: number of full experts to preserve per layer when
+        ``ranking_scope="layer"``.
+    :param sparsity: optional fraction of experts to replace. Exactly one of
+        ``preserve_n_experts`` and ``sparsity`` must be set.
+    :param ranking_metric: expert importance metric. ``fusion`` matches the main
+        MoNE recipe: output fluctuation multiplied by router access score.
```
Ryfernandes, line 42: List the other options for ranking metric here if we are to make it configurable

src/llmcompressor/modifiers/pruning/mone/utils.py
```diff
@@ -0,0 +1,310 @@
...
+        baseline.add_(out_float.sum(dim=0).to(baseline.device) / new_tokens)
+
+        if old_tokens > 0:
+            fluc = self.fluc_out[expert_idx]
+            fluc.mul_((old_tokens - 1) / (new_tokens - 1))
+            out_baseline = baseline.to(out_float.device).unsqueeze(0)
+            fluc.add_(
+                torch.sum((out_float - out_baseline).pow(2), dim=0).to(fluc.device)
+                / new_tokens
+            )
```
Ryfernandes: Is this the correct way to update the fluctuation/variance metric? I get that it is a batched approximation, not the actual variance, but it seems that you are skipping the first batch entirely and also computing the variance for each batch against the global/accumulated mean, not against the mean of the batch

src/llmcompressor/modifiers/pruning/mone/utils.py
```diff
@@ -0,0 +1,310 @@
...
+            src=topk_weights,
+        )
+
+        batch_size = routing_weights.shape[0]
+        count = torch.full_like(self.routing_sum, float(self.routing_count))
+        new_count = count + batch_size
+        routing_sum = torch.sum(routing_weights, dim=0).to(self.routing_sum.device)
+        self.routing_sum.mul_(count / new_count)
+        self.routing_sum.add_(routing_sum / new_count)
+        self.routing_count += batch_size
```
Ryfernandes, line 179: From the paper, aren't routing scores supposed to be averaged over the tokens for which the expert is in the top k? This seems to use the same count for all experts, so it may punish experts that are rarely in the top k in a way that the original paper did not outline

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -23,6 +23,45 @@ def copy_from_experts_module(self, experts: FusedExpertsProtocol, index: int):
...
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        if self.acc_tokens > 0:
+            with torch.no_grad():
+                total = self.acc_tokens + hidden_states.shape[0]
+                self.approx_value.mul_(self.acc_tokens / total)
+                self.approx_value.add_(
+                    hidden_states.float().sum(dim=0).to(self.approx_value.device)
+                    / total
+                )
+                self.acc_tokens = total
```
Ryfernandes: Why does this change the approximate value based on the inputs?

Ryfernandes (commented):
Overall, the core implementation in modifiers/pruning/mone looks sound to me, with a few questions/concerns about the accuracy of the MonE tracker. However, I am unsure about the purpose/scope of the modifications made in adapters.py, modeling/moe and elsewhere for FP8 adapters, the minimax model, patches, etc. Further explanation about why these are necessary and how the scope might grow with this approach in the future (the need to custom-support more model types) would be appreciated. Also, can you please indicate a specific model/model architecture for which the FP8 adapter is necessary and why?

src/llmcompressor/modifiers/pruning/mone/utils.py
ieBoytsov: I mirror the implementaion fro authors from here https://github.com/zxgx/mode-pd/blob/master/modepd/pruning/expert_prune.py#L478

but I agree with you it can be fixed to be more precise.

src/llmcompressor/modifiers/pruning/mone/base.py
ieBoytsov: You mean it can be `forward_pre` instead?

src/llmcompressor/modifiers/pruning/mone/utils.py
ieBoytsov: This logic is also mirrored from the implementation https://github.com/zxgx/mode-pd/blob/master/modepd/pruning/expert_prune.py#L526

I think the idea is that averaging only over tokens where the expert appears in top-k removes the access-frequency signal.

src/llmcompressor/modifiers/pruning/mone/base.py
Ryfernandes: Yes, I believe that since you are just grabbing the router inputs, it can be forward_pre. I used forward in the REAP PR because I also needed the expert forward passes to have already occurred for the router accumulation.

src/llmcompressor/modeling/moe/linear_experts.py
ieBoytsov: Mirrored from the original code but agree it is weird and not needed. we don't want to have some changing value during inference. In my exps it was not used, it is controlled by `enable_novice_evolving=False` flag which is false by default. Will remove this complexity.

src/llmcompressor/modeling/moe/minimax_mone.py
ieBoytsov: Let me think about it. I had issues with minimax mostly because it is a fp8 quantized checkpoint by default https://huggingface.co/MiniMaxAI/MiniMax-M2.5 and it was not easy to unpack experts for pruning there. 

I will follow up with thoughts. Want to be think to what extent it generalises to new models.

src/llmcompressor/modifiers/pruning/mone/utils.py
Ryfernandes: Got it, interesting that they implemented it this way. If we want to diverge from their implementation, maybe it would be worth it to do a short set of ablations on different ways to accumulate the fluctuation metric and see if there is any significant impact on recovery

src/llmcompressor/modifiers/pruning/mone/utils.py
Ryfernandes: Ah got it. If I'm not mistaken, it seems like the other way of averaging is commented out in that file. Maybe they got better results with the way they ended up keeping?

src/llmcompressor/modifiers/pruning/mone/base.py
ieBoytsov: Agree, will fix.

src/llmcompressor/modeling/moe/minimax_mone.py
ieBoytsov: Also, the PR is huge, if you prefer to merge it without minimax logic leaving only basic functionality tested with olmoe, i am fine with it. But in any case I want it to work with SOTA models in future so i will spend time on thinking about it anyway.

ieBoytsov: The key idea of the adapter is that minimax models store experts as `transformers.integrations.finegrained_fp8.FP8Experts` where experts are packed FP8 tensor rows rather than individual expert modules. So this module is reusable for those models that use FP8Experts container. I think deepseek-v4 models do use it too. For other models we need to check how to handle it and can need custom code. wdyt @Ryfernandes ?

ieBoytsov: From quick look should also generalise to GLM-5.2 (not sure i expect it work out of the box but i believe there will be reusable components)

Ryfernandes: Got it, thank you for all of the research on this! Might be worth it to get input from @kylesayrs since he standardized a lot of our model loading, particularly with MoE experts, in a recent release. Kyle, I'm curious to see if you tried any of these FP8 packed tensor row modules when doing the transformers v5 changes? Wondering if this is something that should be supported generally as a part of the LinearExperts2D framework

src/llmcompressor/entrypoints/utils.py
```diff
@@ -90,7 +95,13 @@ def pre_process(
 
     # if the model was loaded with accelerate offloading, convert to CT offloading
     if hasattr(model_args.model, "hf_device_map"):
-        from_accelerate(model_args.model)
```
HDCharles, line 93: whats going on with mone/offloading?

HDCharles:
is MoNe supported in vllm? how are you running these models?

src/llmcompressor/modeling/moe/minimax_mone.py
ieBoytsov: will follow up soon guys, on a corporate trip this week, less active but plan to keep working on the task

HDCharles: yeah agree with @Ryfernandes , this doesn't seem MoNe specific unless i'm not understanding something, it seems like some combination of converters to convert out of fp8 and LinearExperts2D are the cannonical ways to solve these problems, can you clarify why those don't work? we have an example for model free PTQ for minimax https://github.com/vllm-project/llm-compressor/blob/57ed2be4b9169f07c8ee065757158bd66b0925a6/examples/model_free_ptq/minimax_m27_w4a16.py#L9 so i think we should be able to handle this

src/llmcompressor/entrypoints/utils.py
```diff
@@ -138,6 +149,13 @@ def initialize_model_from_path(
     # The .from_pretrained methods guarantee that only one local process can
     # concurrently download model & vocab.
     model_path = model_args.model
+    model_path = (
+        model_path.as_posix() if isinstance(model_path, PosixPath) else model_path
+    )
+
+    if _is_local_mone_checkpoint(model_path):
```
HDCharles: what is a mone checkpoint? something that had MonE applied already?

src/llmcompressor/modeling/moe/mone_builtins.py
```diff
@@ -0,0 +1,14 @@
+"""
+Built-in architecture support modules for MoNE.
+"""
+
+import importlib
+
+_BUILTIN_MONE_SUPPORT_MODULES = (
+    "llmcompressor.modeling.moe.minimax_mone",
```
HDCharles: why can't we support things generally, what specific handling do we need for each model?

ieBoytsov:
> is MoNe supported in vllm? how are you running these models?

FYI @HDCharles  @Ryfernandes  it is not supported in vllm. We implement it by ourselves in vllm (will be happy to upstream and provide recipes too). 

The key detail is that MoNe modifies model architecture by replacing pruned expert's MLP layers with constant vectors, so in order to support it in vllm one needs to write custom kernels (v0 for minimax m2.5 was quite easy though).

The original paper code shows a way to run it via transformers which we also did and compared with our vllm implementation.

src/llmcompressor/pipelines/registry.py
```diff
@@ -55,6 +55,12 @@ def from_modifiers(
 
     @staticmethod
     def _infer_pipeline(modifiers: list[Modifier]) -> str:
+        modifier_names = {modifier.__class__.__name__ for modifier in modifiers}
+        if modifier_names == {"MoNEPruningModifier"}:
+            # MoNE's reference flow collects expert stats during full-model
+            # forwards, not layerwise sequential subgraphs.
```
HDCharles: i'm not sure i understand the difference here? the expert statistics within a subgraph will be the same for the whole forward, right? If this is intended to work on large models, using the basic pipeline is nearly impossible

src/llmcompressor/modifiers/pruning/mone/utils.py
```diff
@@ -0,0 +1,310 @@
...
+            return self.routing_score
+
+        if self.ranking_metric == "output_fluctuation":
+            return self.output_fluctuation
+
+        if self.ranking_metric == "fusion":
+            return (
+                self.fusion_io_weight
+                * self.output_fluctuation
+                * (1.0 - self.fusion_io_weight)
```
HDCharles: this doesn't seem to do anything? 

self.fusion_io_weight = .75 will give the same results as self.fusion_io_weight = .25

src/llmcompressor/modeling/moe/minimax_mone.py
ieBoytsov: You mean that we can use FP8BlockDequantizer and do pruning in bf16 /fp16 instead of directly in fp8 as of now?

I think then it should work and we can get rid of minimax adapters code then.

src/llmcompressor/pipelines/registry.py
ieBoytsov: @HDCharles you are right, making changes so that it worked with sequential pipeline.

src/llmcompressor/modifiers/pruning/mone/utils.py
ieBoytsov: agree. will fix, it mirrors original code.

src/llmcompressor/entrypoints/utils.py
ieBoytsov: refactored and simplified

src/llmcompressor/modeling/moe/mone_builtins.py
ieBoytsov: refactored and simplified

src/llmcompressor/modifiers/pruning/mone/base.py
ieBoytsov: adapter was removed

src/llmcompressor/pipelines/registry.py
ieBoytsov: fixed

src/llmcompressor/modifiers/pruning/mone/utils.py
ieBoytsov: fixed

Issue #2914, opened 2026-07-12
[Bug]: IMatrixMSEObserver crashes in sequential pipeline when IMatrixGatherer is followed by GPTQModifier


verdverm:
## Environment Information ###
Operating System: `Linux-6.17.0-1026-nvidia-aarch64-with-glibc2.39`
Python Version: `3.12.13 (main, Jun 23 2026, 15:19:07) [Clang 22.1.3 ]`
llm-compressor Version: `0.12.0`
compressed-tensors Version: `0.17.1`
transformers Version: `5.10.1`
torch Version: `2.13.0`
CUDA Devices: `['NVIDIA GB10']`
AMD Devices: `None`
NPU Devices: `None`

### 🐛 Describe the bug

New to this cool project, hit this error and sent `opencode + kimi-2.7-code` on it and this is what it came back with. The code appears to work, minus correctness because I do not know enough to validate that yet. Working towards getting the evals setup as a basic sanity check.

## Summary
When using `IMatrixGatherer` followed by `GPTQModifier` with `pipeline="sequential"`, calibration crashes on the second subgraph with:

```text
AttributeError: 'Linear' object has no attribute '_imatrix_sum'. Did you mean: '_imatrix_hook'?
```

The crash originates in `src/llmcompressor/observers/imatrix.py:108` inside the `_hook` closure registered by `IMatrixMSEObserver.attach`.

## Reproduction
Run the following recipe against a model such as `Qwen/Qwen3.6-27B`:

```python
import torch
from compressed_tensors.quantization import preset_name_to_scheme
from datasets import load_dataset
from transformers import AutoProcessor, Qwen3_5ForConditionalGeneration

from llmcompressor import oneshot
from llmcompressor.modifiers.gptq import GPTQModifier
from llmcompressor.modifiers.transform.imatrix import IMatrixGatherer
from llmcompressor.utils import load_context

MODEL_ID = "Qwen/Qwen3.6-27B"
SCHEME = "NVFP4A16"
NUM_CALIBRATION_SAMPLES = 512
MAX_SEQUENCE_LENGTH = 8192
IGNORE = [
    "re:.*lm_head.*",
    "re:.*embed_tokens.*",
    "re:model.visual.*",
    "re:.*mlp.gate.*",
    "re:.*linear_attn.*",
]

scheme = preset_name_to_scheme(SCHEME, ["Linear"])
scheme.weights.observer = "imatrix_mse"

[...]

PR #2916, opened 2026-07-12
fix(observers/imatrix): remove stale gatherer hook on second-pass attach

Anai-Guo:
## Summary

`IMatrixMSEObserver.attach()` can leave a **stale forward-pre hook** on a module, which then crashes on the next forward with:

```text
AttributeError: 'Linear' object has no attribute '_imatrix_sum'
```

Reported in #2914 (triggered by `IMatrixGatherer` → `GPTQModifier` with `pipeline="sequential"`).

## Root cause

The iMatrix lifecycle spans two passes:

1. **`IMatrixGatherer`** — `on_initialize` calls `observer.attach(module)`, which creates `module._imatrix_sum` / `module._imatrix_count` and registers `module._imatrix_hook` (a forward-pre hook that accumulates `E[x²]`). The gatherer's `on_calibration_end` later calls `observer.detach(module)`, which **removes the hook but intentionally leaves the accumulators** on the module.
2. **Quantization modifier** (`GPTQModifier` / `QuantizationModifier`) — its `imatrix_mse` observer's `attach()` hits the second-pass branch, **steals the accumulators and `del`s them** from the module, then returns.

When the two modifiers share a single calibration run, the quantization observer's `attach()` (step 2) runs **before** the gatherer's `detach()` (`on_calibration_end`). So the accumulators get deleted while the gatherer's `_imatrix_hook` is **still registered**. The next forward fires that hook, which dereferences the now-deleted `mod._imatrix_sum` → `AttributeError`.

## Fix

When the second-pass `attach()` steals and deletes the accumulators, also remove any lingering `_imatrix_hook` so it can't fire against deleted state. The removal is guarded by `hasattr`, so it's a no-op on the normal path where `detach()` already removed the hook — no behavior change for the happy path.

## Test

Adds a CPU-only regression test (`TestSecondPassHookRemoval`) that reproduces the exact failure: attach → forward → second attach (steals accumulators) → forward. Without the fix the final forward raises the reported `AttributeError`; with it, the hook is removed and the forward is a clean no-op.

---

### DIFF
--- a/src/llmcompressor/observers/imatrix.py
+++ b/src/llmcompressor/observers/imatrix.py
@@ -78,6 +78,16 @@ class IMatrixMSEObserver(Observer):
             self._imatrix_count = module._imatrix_count
             del module._imatrix_sum
             del module._imatrix_count
+            # A gatherer's accumulation hook may still be registered on this
+            # module: its ``detach()`` runs at ``on_calibration_end``, which can
+            # be *after* this second-pass ``attach()`` when the gatherer and the
+            # quantization modifier share a single calibration run (e.g.
+            # ``IMatrixGatherer`` composed with ``GPTQModifier``). Remove it so
+            # it cannot fire against the accumulators we just deleted, otherwise
+            # the next forward raises ``AttributeError`` on ``_imatrix_sum``.
+            if hasattr(module, "_imatrix_hook"):
+                module._imatrix_hook.remove()
+                del module._imatrix_hook
             return
 
         if not hasattr(module, "in_features"):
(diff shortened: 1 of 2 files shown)


brian-dellabetta:
Hi @Anai-Guo , we recently landed a PR to remove the IMatrixGatherer (as a modifier in the recipe) and have it be an Observer instead:
* #2920

Can you please try on main and confirm your issue is resolved?

Anai-Guo:
Thanks @brian-dellabetta — confirmed. #2920 reworked the IMatrixGatherer into an `IMatrixObserver`, and the new `attach()` already removes any existing `_imatrix_hook` before registering a fresh one (`observers/imatrix.py` L76-78), so a second `attach()` no longer leaves a stale hook. That's exactly what this PR was fixing, so it's now obsolete. Closing in favor of #2920.

PR #2920, opened 2026-07-13
Roderick wu/remove imatrixgatherer

### LINKED ISSUE #2914: [Bug]: IMatrixMSEObserver crashes in sequential pipeline when IMatrixGatherer is followed by GPTQModifier

Roderick-Wu:
SUMMARY:

Original issue:
https://github.com/vllm-project/llm-compressor/issues/2698

Tested + cleaned-up this old pr:
https://github.com/vllm-project/llm-compressor/pull/2716

Also uses the compressed-tensors version from this pr:
https://github.com/vllm-project/compressed-tensors/pull/710

Fixed references to _imatrix_count.item() outside observers, fixed reference to module._imatrix_hook in IMatrixMSEObserver.attach since those features have been moved to the observer. 

| Variant | PPL | Delta vs BF16 |
|---|---|---|---|---|
| Old (IMatrixGatherer) single-GPU | 5.8877 | +0.318 |
| Old (IMatrixGatherer) DDP 3-GPU | 5.8872 | +0.317 |
| New (single-pass) single-GPU | 5.8872 | +0.317 |
| New (single-pass) DDP 3-GPU | 5.8870 | +0.317 |

## Summary by CodeRabbit

- **New Features**
  - iMatrix importance-weighted quantization now works directly through quantization configuration, without a separate setup step.
  - Added standalone and distributed examples for NVFP4 and iMatrix quantization workflows.

- **Bug Fixes**
  - Improved distributed quantization so completion output, generation, and saving occur only once.
  - Added safer device handling for post-quantization generation.

- **Documentation**
  - Clarified calibration behavior, uniform-MSE fallback options, distributed workflows, and fused-layer observer scaling.


brian-dellabetta:
Resolves #2698 
Resolves #2914

Roderick-Wu:
e2e test results using:
- meta-llama/Meta-Llama3-8B-Instruct
- NVFP4 W4A16 (`strategy=TENSOR_GROUP`, `group_size=16`, `scale_dtype=FP8_E4M3`)
- wikitext-2-raw-v1



## Perplexity

| Variant | Perplexity |
|---|---|
| Old, single GPU | 22.10 |
| **New, single GPU** | **22.89** |
| Old, DDP (4 GPU) | 25.58 |
| **New, DDP (4 GPU)** | **26.08** |



## Also ran a quick diff of the model weights since we expect very similar behavior:

### Single GPU

| Metric | Value |
|---|---|
| Common tensors | 963 |
| Exact matches | 500 (51.9%) |
| Mismatched | 463 (48.1%) |

### DDP

| Metric | Value |
|---|---|
| Common tensors | 963 |
| Exact matches | 501 (52.0%) |
| Mismatched | 462 (48.0%) |

HDCharles:
something's up with those numbers, DDP shouldn't be that far off can you compare with evals/setup from https://github.com/vllm-project/llm-compressor/pull/2585 also run the DDP smoke tests

Roderick-Wu:
Previous issues were caused by the incorrect offboarding. 

NVFP4A16, `HuggingFaceH4/ultrachat_200k`, `train_sft` split, 512 samples, max_seq_length=2048

| Variant | PPL | Delta vs BF16 |
|---|---|---|---|---|
| Old (IMatrixGatherer) single-GPU | 5.8877 | +0.318 |
| Old (IMatrixGatherer) DDP 3-GPU  | 5.8872 | +0.317 |
| New (single-pass) single-GPU     | 5.8872 | +0.317 |
| New (single-pass) DDP 3-GPU      | 5.8870 | +0.317 |

Added scripts for reproducibility as examples (llm-compressor/examples/imatrix/*nvfp4*).

examples/awq/llama_example_ddp.py
```diff
@@ -85,28 +85,30 @@ def tokenize(sample):
...
-    + str(torch.distributed.get_world_size())
-)
-model.save_pretrained(SAVE_DIR, save_compressed=True)
-tokenizer.save_pretrained(SAVE_DIR)
 
+world_size = torch.distributed.get_world_size()
+rank = torch.distributed.get_rank()
 torch.distributed.destroy_process_group()
+
+if rank == 0:
```
HDCharles: why this change?

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -305,6 +305,18 @@ def has_config(self) -> bool:
             and self.kv_cache_scheme is None
         )
 
+    def requires_weight_observer_calibration_data(self) -> bool:
```
HDCharles: this isn't a function of the modifier

src/llmcompressor/pipelines/registry.py
```diff
@@ -63,13 +63,15 @@ def _modifier_requires_calibration(modifier: Modifier):
...
                 "AutoRoundModifier",
-                "IMatrixGatherer",
                 "REAPPruningModifier",
             ):
                 return True
             elif isinstance(modifier, QuantizationModifier):
                 config = modifier.resolve_quantization_config()
-                return config.requires_calibration_data()
+                return (
+                    config.requires_calibration_data()
```
HDCharles: it makes more sense to add this to config.requires_calibration_data than to make it a function on the modifier and have that function check the config to see if it needs calibration data.

examples/awq/llama_example_ddp.py
Roderick-Wu: multiple processes/ranks writing to the same file causes issues. At best its wasteful disk i/o. For me the scripts kept on crashing when trying to save at the end

src/llmcompressor/pipelines/registry.py
Roderick-Wu: yes, this idea is much cleaner. Moved to the quant config:
https://github.com/vllm-project/compressed-tensors/pull/787

examples/awq/llama_example_ddp.py
brian-dellabetta: Does calling `model.save_pretrained` cause race conditions for ddp? I did some searching, the `if rank ==0` check seems to be best practice for this case.

maybe we should include a guard on the function itself? This could probably be handled with a decorator like

```python
@run_rank0
def save_pretrained(...
```

that skips execution if it finds distributed and non-zero rank

examples/awq/qwen3_moe_example_ddp.py
```diff
@@ -90,28 +90,34 @@ def tokenize(sample):
...
+    print("==========================================\n\n")
+
+    # Save to disk compressed.
+    SAVE_DIR = (
+        MODEL_ID.rstrip("/").split("/")[-1]
+        + "-awq-sym-DDP"
+        + str(torch.distributed.get_world_size())
+    )
+    model.save_pretrained(SAVE_DIR, save_compressed=True)
+    tokenizer.save_pretrained(SAVE_DIR)
```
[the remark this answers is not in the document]
brian-dellabetta: Yeah @Roderick-Wu i think you want to move `torch.distributed.destroy_process_group()` down to the bottom, for both examples?

examples/imatrix/llama3_nvfp4_imatrix_example.py
```diff
@@ -0,0 +1,70 @@
...
+#   * use imatrix_mse observer to weight quantization error by channel importance
+scheme = preset_name_to_scheme("NVFP4A16", ["Linear"])
+scheme.weights.observer = "imatrix_mse"
+
+recipe = [
+    QuantizationModifier(
+        config_groups={"group_0": scheme},
+        ignore=["lm_head"],
+    ),
+]
```
brian-dellabetta: We have some [helper code on QuantizationMixin](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/quantization/quantization/mixin.py#L135-L137) to avoid having to manually configure configs like this when using non-default observers. this would work(?)
```suggestion
recipe = [
    QuantizationModifier(
        scheme="NVFP4A16",
        targets="Linear",
        ignore=["lm_head"],
        weight_observer="imatrix_mse"
    ),
]
```

examples/imatrix/llama3_nvfp4_imatrix_example_ddp.py
```diff
@@ -0,0 +1,99 @@
...
+#   * use imatrix_mse observer to weight quantization error by channel importance
+scheme = preset_name_to_scheme("NVFP4A16", ["Linear"])
+scheme.weights.observer = "imatrix_mse"
+
+recipe = [
+    QuantizationModifier(
+        config_groups={"group_0": scheme},
+        ignore=["lm_head"],
+    ),
+]
```
brian-dellabetta:
```suggestion
recipe = [
    QuantizationModifier(
        scheme="NVFP4A16",
        targets="Linear",
        ignore=["lm_head"],
        weight_observer="imatrix_mse"
    ),
]
```

brian-dellabetta (commented):
Things are looking good! A couple nits on the examples

examples/imatrix/llama3_imatrix_example.py
brian-dellabetta, line 1: nit -- consider removing imatrix from file name for these examples, to align with other example naming conventions and since it is in imatrix folder already

brian-dellabetta (approved):
This looks fine to me, pending @HDCharles review since i know he worked closely on initial implementation and on this work with you

HDCharles (changes requested):
looks like the stats aren't going to be cleaned up correctly unless I'm missing something.

see: 

https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/observers/base.py#L121 - calls delete

https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/observers/base.py#L164 - method that deletes stuff

https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/observers/base.py#L43 - how to specify what should be deleted

can you add those imatrix stats to the _stats_attrs and  a unit test to verify that those stats get deleted on get_qparams.

@kylesayrs anything i'm forgetting? i feel like its probably ok to assume that if we're deleting the observation stats that its ok to delete the imatrix stats but its a little different than normal.

HDCharles (approved):

examples/awq/llama_example_ddp.py
HDCharles: it handles the single rank internally

Issue #2699, opened 2026-05-12
[Help Wanted] Transformers Load of 3D Packed MoE Checkpoints


Sekri0:
Hi, I want to export onnx model from llmcompressor quantized model. Our onnx export tool relies on "huggingface transformers", so it would be very helpful if quantized model can be loaded by hf transformers.  I got a few questions about "moe calibration design" in llmcompressor.

Some model such as glm_moe_dsa set "is_permanent = True". After quantization, can this model be properly loaded by huggingface transfomrers? (my attempts failed because of missing weight and unexpected weight)
If I want to restore the original structure of a quantized model (such as glm_moe_dsa), what should I do? According to my understanding, models such as glm_moe_dsa and qwen3.5_moe have changed the model structure when replacing MoE with CalibrationMoE (It can also be said that a new weight layer or tensor is created). This means that simply returning "origin" in the restore function is not feasible, because "origin" does not include information such as "scales" and "zero point". Is there an example of llmcompressor correctly restoring the original structure after such modifications?

brian-dellabetta:
Hi @Sekri0 , no I don't think there's an easy way to do this. We generally only support running llmcompressor quantized models on vllm, which doesn't have onnx support

Can you explain your use case a bit more? What are you hoping to do with the onnx-exported model?

Sekri0:
Deploy LLM on NPU chip needs onnx-exported model. We have a LLM-onnx-export tool, which is dependent on hf transformers ecosystem. This means if a model can be loaded by hf transformers, we can get onnx-model with little efforts. My goal is to load a llmcompressor-quanted model using hf transforms. Therefore, I can easily export an quanted onnx model, which has the same accuracy with llmcompressor quanted model. 
According to my attemptation, models whose structures remain unchanged after quantization can be loaded correctly by hf transformers. But I still encounter issues when loading models such as qwen3.5_moe and glm_moe_dsa due to model structure miss match.

brian-dellabetta:
Got it, thanks for the information. I'm surprised llm-compressor compressed models work in onnx. We wrap the huggingface transformers logic to run "fake quantized" forward passes, i.e. convert the bfloat16 values to what they would be in quantized format while retaining the bfloat16 dtype. This is not at all performant, but provides a dev pathway needed during compression. Do you see speedups in onnx? Custom logic is needed to unpack parameters to be used with performant kernels, like in vllm and in llm-compressor's MoE calibration context.

[...]

PR #2926, opened 2026-07-15
Feat/qwen3 vl moe 3d save mappings

GOavi101:
### Summary
Adds save-time backwards _weight_conversions for qwen3_vl_moe, repacking linearized 2D expert weights and quantization parameters into the native 3D gate_up_proj / down_proj checkpoint format.

This allows Transformers to reload quantized MoE checkpoints after linearization and addresses #2699.

The mappings are configured through linearize_moe and load_quantizable_moe. This also adds regression tests and a smoke-test example.

### Test Plan

- pytest tests/llmcompressor/modeling/test_3d_packed_save.py -v
5/5 tests passed
Validates packed shapes, saved keys, and Transformers reload

- python examples/quantizing_moe/qwen3_vl_moe_3d_save.py
Confirmed packed gate_up_proj / down_proj keys
Confirmed no experts.0.* keys remain

- Tested with inference-optimization/Qwen3-VL-1.0B-A0.4B-Instruct
Successfully saved the packed checkpoint
Successfully reloaded fused Qwen3VLMoeTextExperts with Transformers


src/llmcompressor/modeling/moe/conversion_mappings.py
```diff
@@ -199,6 +205,78 @@
...
+        WeightConverter(
+            source_patterns=[
+                "mlp.experts.*.gate_proj.weight_scale$",
+                "mlp.experts.*.up_proj.weight_scale$",
+            ],
+            target_patterns="mlp.experts.gate_up_proj_scale",
+            operations=[MergeModulelist(dim=0), Concatenate(dim=-1)],
+        ),
+        WeightConverter(
+            source_patterns="mlp.experts.*.down_proj.weight_scale$",
```
kylesayrs: Please use [this](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/quantization/quant_metadata.py#L25) rather than manually writing scale, etc.

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -124,6 +130,11 @@ def linearize_moe(model: PreTrainedModel):
         linear_moe = linear_experts_cls.from_experts_module(module, config)
         model.set_submodule(name, linear_moe)
 
+    # Architectures with native 3D packed checkpoints need save-time repacking
+    maybe_set_3d_packed_save_mappings(model)
```
kylesayrs: Why can't we just hook into the `backwards_mapping`? Ie just add an extra entry to the conversion mapping tuple which specifies what to add to the backwards mapping

examples/quantizing_moe/qwen3_vl_moe_3d_save.py
```diff
@@ -0,0 +1,120 @@
+import os
```
kylesayrs: I don't think we should include this example in source

src/llmcompressor/modeling/moe/linearize.py
GOavi101: Good point — refactored so the 3D pack converters live as the third entry in ARCH_TO_MOE_MAPPINGS (remove_targets, load_mappings, save_pack_converters) instead of a separate hook. get_moe_save_conversion_mappings() / get_linearize_load_mappings() now pull backwards save mappings from that tuple, and linearize_moe just calls set_moe_save_conversion_mappings(model). Pushed in latest commit.

examples/quantizing_moe/qwen3_vl_moe_3d_save.py
GOavi101: Agreed, removed the smoke script from source and kept the coverage in the regression tests only.

kylesayrs:
Spoke offline. Something like this would require changes like these: https://github.com/huggingface/transformers/pull/47387, supporting many-to-one saving. However, getting transformers to adopt this would be very difficult. In the meantime, we shoudl go with https://github.com/vllm-project/llm-compressor/pull/3017

Issue #2929, opened 2026-07-15
[Refactor] Unify microscale and standard code paths in model_free_ptq


soyr-redhat:
## Summary

`model_free_ptq` currently maintains two code paths. One for standard schemes and one for microscale (NVFP4, MXFP4). 
There exists a TODO at [`microscale.py:104-108`](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/entrypoints/model_free/microscale.py#L104-L109) that describes ideal cleanup as follows:

- Promote `MicroscaleConverter` from an inline class to a top-level class fulfilling the full [Converter Protocol](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/entrypoints/convert/converters/base.py)
- Remove `build_microscale_inverse_weight_maps` entirely, replacing it with the standard `build_inverse_weight_maps`
- Merge `process_file` and `process_file_microscale_scheme` into a single function

This was left as a follow-up after #2498 introduced the inverse_weight_map approach, and #2737 completed the deprecation of `reindex_fused_weights`.

### 1. Promote `MicroscaleConverter` to a top-level `Converter`

Go with moving `MicroscaleConverter` out of the `build_microscale_inverse_weight_maps` function body and implement the full Protocol:

### 2. Remove `build_microscale_inverse_weight_maps`

Once `MicroscaleConverter` is a proper `Converter`, `_build_jobs` can instantiate it and pass it in the converters list to the standard `build_inverse_weight_maps` which is the same code path as non-microscale.

### 3. Merge `process_file` and `process_file_microscale_scheme`

A single `process_file` function that conditionally enables fused-group handling when the scheme uses `TENSOR_GROUP` compression strategy

### 4. Simplify `_build_jobs`

Remove the `if is_microscale_scheme` branch. Single `build_inverse_weight_maps` call, single `process_file` function, microscale converter passed via the converters list.

Happy to take a stab at it, but would love feedback on gameplan if needed/necessary

Related: #2919 (Q3 Roadmap — model_free_ptq is called out under "Custom Pytorch Definition Support")
Predecessor PRs: #2498, #2737, #2545

brian-dellabetta:
Thanks @soyr-redhat ! Yeah, the process_file and validate_file functions can be moved to methods directly on the converter, which we could have as a ModelFreePtqConverter, which can make use of an `is_microscale` flag to determine what dependencies tensors exist etc.

[...]

PR #2935, opened 2026-07-16
refactor: unify model-free ptq into ModelFreePtqConverter class

soyr-redhat:
### SUMMARY:
addresses #2929

Unifies the dual microscale/standard code paths in model_free_ptq into a
single ModelFreePtqConverter class. 

This eliminates the unnecessary branching in `__init__.py` and consolidates `process_file`, `validate_file`, and
`get_dependencies` as methods on the converter, gated by an `is_microscale` flag.

Changes:
- Add ModelFreePtqConverter class in process.py with Converter methods plus process_file/validate_file under parent class
- Remove build_microscale_inverse_weight_maps and inline MicroscaleConverter from microscale.py
- Simplify __init__.py to construct one converter, no branching
- Update test_reindexing_elimination.py and test_model_free_validation.py along with fixed pytest mfptq init for standardized testing

### TEST PLAN:
- pytest tests/llmcompressor/entrypoints/model_free/ -v (all pass, GPU
tests need CI)
- ruff check + ruff format clean

### DIFF
--- a/src/llmcompressor/entrypoints/model_free/__init__.py
+++ b/src/llmcompressor/entrypoints/model_free/__init__.py
@@ -18,15 +18,7 @@ from compressed_tensors.utils.safetensors_load import (
 )
 from loguru import logger
 
-from llmcompressor.entrypoints.model_free.microscale import (
-    build_microscale_inverse_weight_maps,
-    is_microscale_scheme,
-)
-from llmcompressor.entrypoints.model_free.process import (
-    process_file,
-    process_file_microscale_scheme,
-    validate_file,
-)
+from llmcompressor.entrypoints.model_free.process import ModelFreePtqConverter
 from llmcompressor.entrypoints.model_free.save_utils import (
     update_config,
 )
@@ -85,13 +77,14 @@ def model_free_ptq(
             logger.info(f"Copying {file_path} -> {save_path}")
             shutil.copyfile(resolved_path, save_path)
 
+    # Construct the mfptq converter
+    mfptq = ModelFreePtqConverter(scheme, ignore, converter)
+
     # build quantization jobs
-    jobs = _build_jobs(
-        model_files, save_directory, scheme, ignore, resolved_devices, converter
-    )
+    jobs = _build_jobs(model_files, save_directory, mfptq, resolved_devices)
 
     # 1. validate quantizable tensors — fail fast before long-running quantization
-    validate_jobs = [(validate_file, *job[1:]) for job in jobs]
+    validate_jobs = [(mfptq.validate_file, *job[1:]) for job in jobs]
     exec_jobs(validate_jobs, max_workers, desc="Validating")
 
     # 2-5. quantize and compress weights
@@ -136,10 +129,8 @@ def _resolve_devices(
 def _build_jobs(
     model_files: dict[str, str],
     save_directory: str | os.PathLike,
-    scheme: QuantizationScheme,
-    ignore: Iterable[str],
+    mfptq: ModelFreePtqConverter,
     devices: list[torch.device],
-    converter: Converter | None,
 ) -> list[tuple]:
     """
     Build jobs with precomputed inverse_weight_map per shard.
@@ -150,22 +141,19 @@ def _build_jobs(
     process function and eliminates redundant tensor reads.
 
     :returns: list of jobs tuples
-        (job_fn, inverse_weight_map, save_path, scheme, ignore, device, converter)
+        (mfptq.process_file, inverse_weight_map, save_path, device)
         Shards are distributed round-robin across the given devices.
     """
     weight_map = get_weight_map(model_files)
 
-    if is_microscale_scheme(scheme):
-        job_fn = process_file_microscale_scheme
-        build_inverse_weight_maps_fn = build_microscale_inverse_weight_maps
-    else:
-        job_fn = process_file
-        build_inverse_weight_maps_fn = build_inverse_weight_maps
+    converters = [mfptq]
+    if mfptq.converter is not None:
+        converters.append(mfptq.converter)
 
-    inverse_weight_maps = build_inverse_weight_maps_fn(
+    inverse_weight_maps = build_inverse_weight_maps(
         weight_map=weight_map,
         model_files=model_files,
-        converters=[converter] if converter is not None else [],
+        converters=converters,
     )
 
     shard_names = [name for name in model_files if name.endswith("safetensors")]
@@ -186,15 +174,7 @@ def _build_jobs(
         device = devices[i % len(devices)]
 
         jobs.append(
-            (
-                job_fn,
-                inverse_weight_maps[shard_name],
-                save_path,
-                scheme,
-                ignore,
-                device,
-                converter,
-            )
+            (mfptq.process_file, inverse_weight_maps[shard_name], save_path, device)
         )
 
     return jobs
--- a/src/llmcompressor/entrypoints/model_free/__init__.py
+++ b/src/llmcompressor/entrypoints/model_free/__init__.py
@@ -22,7 +22,7 @@ from compressed_tensors.utils.safetensors_load import (
 from loguru import logger
 from safetensors.torch import save_file
 
-from llmcompressor.entrypoints.model_free.process import ModelFreePtqConverter
+from llmcompressor.entrypoints.model_free.converter import ModelFreePtqConverter
 from llmcompressor.entrypoints.model_free.save_utils import update_config
 from llmcompressor.entrypoints.model_free.scheduler import estimate_job_memory
 from llmcompressor.entrypoints.model_free.validate import (
@@ -93,8 +93,8 @@ def model_free_ptq(
 
     # validate first (runs on meta device)
     validate_jobs = [
-        (_validate_shard, iwm, sp, cfg, torch.device("meta"), conv)
-        for _, iwm, sp, cfg, conv in jobs
+        (_validate_shard, iwm, sp, convs, torch.device("meta"))
+        for _, iwm, sp, convs in jobs
     ]
     exec_jobs(validate_jobs, max_workers, desc="Validating")
 
@@ -102,12 +102,8 @@ def model_free_ptq(
     total_size = 0
     weight_map = dict()
     callable_jobs = [
-        (
-            lambda dev, fn=fn, iwm=iwm, sp=sp, cfg=cfg, conv=conv: fn(
-                iwm, sp, cfg, dev, conv
-            )
-        )
-        for fn, iwm, sp, cfg, conv in jobs
+        (lambda dev, fn=fn, iwm=iwm, sp=sp, convs=convs: fn(iwm, sp, convs, dev))
+        for fn, iwm, sp, convs in jobs
     ]
     quantize_results = exec_jobs_dynamic(
         jobs=callable_jobs,
@@ -161,14 +157,11 @@ def _build_jobs(
     — no separate build_microscale_inverse_weight_maps needed.
 
     :returns: (jobs, memory_estimates) where each job is
-        (_process_shard, inverse_weight_map, save_path, config, converter)
+        (_process_shard, inverse_weight_map, save_path, converters)
         and each memory estimate is in bytes.
     """
     weight_map = get_weight_map(model_files)
 
-    # Build the converter chain upfront for dependency resolution only.
-    # The chain is reconstructed per shard inside _process_shard / _validate_shard
-    # so the scheduler can inject the right device.
     mfptq = ModelFreePtqConverter(config)
     all_converters = ([converter] if converter is not None else []) + [mfptq]
     inverse_weight_maps = build_inverse_weight_maps(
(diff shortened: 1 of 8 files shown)


tests/llmcompressor/entrypoints/model_free/test_reindexing_elimination.py
```diff
@@ -8,13 +8,17 @@
...
 )
+
 from llmcompressor.entrypoints.model_free.process import (
-    process_file_microscale_scheme,
+    ModelFreePtqConverter,
 )
 
+@pytest.fixture
+def mfptq():
+    return ModelFreePtqConverter(scheme=_make_nvfp4_scheme(), ignore=[])
```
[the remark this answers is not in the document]
soyr-redhat: sounds good boss

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -34,206 +36,139 @@
...
+
+        if self.converter is not None:
+            self.converter.validate(tensors)
+
+        for _, name in match_quantizable_tensors(
+            tensors, self.ignore, self.scheme.targets
+        ):
+            validate_weight_for_quantization(tensors[name], self.scheme, name)
+
+    def process_file(
```
brian-dellabetta, line 82: can this logic live in process instead?

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -34,206 +36,139 @@
...
+    def process(self, tensors):
+        return tensors
+
+    def validate(self, tensors):
+        pass
+
+    def create_config(self):
+        return None
+
+    def validate_file(
```
brian-dellabetta, line 65: can this logic live inside validate instead?

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -150,22 +141,19 @@ def _build_jobs(
...
     weight_map = get_weight_map(model_files)
 
-    if is_microscale_scheme(scheme):
-        job_fn = process_file_microscale_scheme
-        build_inverse_weight_maps_fn = build_microscale_inverse_weight_maps
-    else:
-        job_fn = process_file
-        build_inverse_weight_maps_fn = build_inverse_weight_maps
+    converters = [mfptq]
+    if mfptq.converter is not None:
```
brian-dellabetta, line 150: rather than including a converter field on ModelFreePtqConverter, could we instead chain converters together? i forgot we still need to land the changes to support that. But this way the converters will just process in the order they're defined

https://github.com/vllm-project/compressed-tensors/pull/690/changes

brian-dellabetta (commented):
Thanks @soyr-redhat for trying this out. I left a few comments, rather than having process_file/validate_file as methods, ideally it would just have process and validate like the interface, and for model free ptq we would just be able to call into convert_checkpoint with the ModelFreePtqConverter appended to the list of converters.

I forgot we never merged in the multi-converter support:
https://github.com/vllm-project/compressed-tensors/pull/690

We will have to get that in before we can append ModelFreePtqConverter to the list of converters

soyr-redhat:
## Dependencies

**Merged**
- vllm-project/compressed-tensors#805: Multi-converter chaining support. This PR's `ModelFreePtqConverter` implements the Converter protocol (`process`, `validate`, `get_dependencies`, `update_config`) introduced there. The tests also use the updated `convert_file(iwm, path, [converters])` list API and `load_tensors_from_inverse_weight_map(..., device="meta")` from that PR , so CI needs the local CT workspace installed, not the published package.

**Pending**
- vllm-project/compressed-tensors#849: ports `exec_jobs_dynamic` into CT as a reusable building block. Once that lands, the local `scheduler.py` here can be removed and the import swapped to CT.
- vllm-project/llm-compressor#3075: the llmc-side of that port, removes `scheduler.py` here and wires up the CT version.

The production code in this PR does not depend on #849/#3075: it still imports `exec_jobs_dynamic` from the local `scheduler.py`. Only the test helpers depend on the CT#805 APIs.

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -1,35 +1,37 @@
...
-from llmcompressor.entrypoints.model_free.process import (
-    process_file,
-    process_file_microscale_scheme,
-    validate_file,
-)
-from llmcompressor.entrypoints.model_free.save_utils import (
-    update_config,
-)
+from llmcompressor.entrypoints.model_free.process import ModelFreePtqConverter
 from llmcompressor.entrypoints.model_free.scheduler import (
```
soyr-redhat: this is not long for this world

src/llmcompressor/entrypoints/model_free/process.py
soyr-redhat: you got it!

src/llmcompressor/entrypoints/model_free/process.py
soyr-redhat: for sure!

soyr-redhat:
bk will likely fail because tests depend on 805 (require version pin)

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -210,9 +200,40 @@ def _build_jobs(
...
+
+def _process_shard(
+    inverse_weight_map: InverseWeightMap,
+    save_path: str | os.PathLike,
+    config: QuantizationConfig,
+    device: torch.device,
+    converter: Converter | None,
+) -> tuple[int, dict[str, str]]:
+    mfptq = ModelFreePtqConverter(config)
+    converters = ([converter] if converter is not None else []) + [mfptq]
```
brian-dellabetta: this looks duplicated from above, why not just pass in `converters:list[Converter]` directly?

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -210,9 +200,40 @@ def _build_jobs(
...
+
+def _validate_shard(
+    inverse_weight_map: InverseWeightMap,
+    save_path: str | os.PathLike,
+    config: QuantizationConfig,
+    device: torch.device,
+    converter: Converter | None,
+) -> None:
+    mfptq = ModelFreePtqConverter(config)
+    converters = ([converter] if converter is not None else []) + [mfptq]
```
brian-dellabetta: same q as above

src/llmcompressor/entrypoints/model_free/microscale.py
brian-dellabetta, line 1: so nice!

src/llmcompressor/entrypoints/model_free/converter.py
brian-dellabetta, line 1: can we rename this to converter.py?

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -59,186 +54,155 @@ def _match_tensors_to_schemes(
...
+            for template in partner_templates:
+                deps.add(template.format(**match.groupdict()))
+        return deps
+
+    def validate(self, tensors: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
+        """
+        Validate that each quantizable tensor can be quantized under its scheme.
+        Operates on meta tensors — no real computation.
+        """
+        meta_tensors = {k: v.to("meta") for k, v in tensors.items()}
```
brian-dellabetta: are they not already meta tensors?

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -59,186 +54,155 @@ def _match_tensors_to_schemes(
...
-
-
-def process_file_microscale_scheme(
-    inverse_weight_map: InverseWeightMap,
-    save_path: str | os.PathLike,
-    config: QuantizationConfig,
-    device: str | torch.device,
-    converter: Converter | None = None,
-) -> tuple[int, dict[str, str]]:
+class ModelFreePtqConverter:
```
brian-dellabetta: this should inherit the Converter base class, is there any reason not to?

src/llmcompressor/entrypoints/model_free/process.py
```diff
@@ -59,186 +54,155 @@ def _match_tensors_to_schemes(
...
+        return meta_tensors
+
+    def process(self, tensors: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
+        """Quantize and compress all tensors. Device is inferred from the tensors."""
+        tensors = split_fused_moe_experts(tensors)
+        if has_microscale_scheme(self.config):
+            return self._process_microscale(tensors)
+        return self._process_standard(tensors)
+
+    def update_config(self, incoming: QuantizationConfig | None) -> QuantizationConfig:
```
brian-dellabetta: this should match the signature of the base Converter class (incoming should be named config)

brian-dellabetta (commented):
Things are looking good! Several nits and one concern -- shouldn't ModelFreePtqConverter inherit the converter base class? it looks like the function signature doesn't match

src/llmcompressor/entrypoints/model_free/__init__.py
soyr-redhat: Good point, I'll update it to do that

src/llmcompressor/entrypoints/model_free/__init__.py
soyr-redhat: same as above, making updates for that now!

src/llmcompressor/entrypoints/model_free/process.py
soyr-redhat: Not that I can think of, makes sense to me! thx brian

src/llmcompressor/entrypoints/model_free/process.py
soyr-redhat: Yep, I had kept it in as an extra safety net layer but i think that it's unnecessary

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -210,9 +193,32 @@ def _build_jobs(
...
+    total_size = sum(t.nbytes for t in tensors.values())
+    weight_map = {key: os.path.basename(save_path) for key in tensors.keys()}
+    return total_size, weight_map
+
+
+def _validate_shard(
+    inverse_weight_map: InverseWeightMap,
+    save_path: str | os.PathLike,
+    converters: list[Converter],
+    device: torch.device,
```
brian-dellabetta: don't we want this to always be meta? Seems like it should be hard-coded instead of passed in?

src/llmcompressor/entrypoints/model_free/converter.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+def _match_tensors_to_schemes(
+    tensors: dict[str, torch.Tensor],
+    config: QuantizationConfig,
+) -> Iterator[tuple[str, str, QuantizationScheme]]:
+    """Match each quantizable tensor to its scheme from the config.
+
+    Iterates config groups in order. A tensor matched by an earlier group
+    is not re-yielded for later groups (first match wins).
```
brian-dellabetta, line 46: i think this is not the same as what happens with `match_targets`, will follow up on this in slack

brian-dellabetta (commented):
One more nit and then the question on how targets should be resolved when there are multiple config groups

src/llmcompressor/entrypoints/model_free/__init__.py
soyr-redhat: sure, sounds reasonable to me!

soyr-redhat: addressed in most recent commit :)

src/llmcompressor/entrypoints/model_free/converter.py
```diff
@@ -0,0 +1,268 @@
...
+    """
+    matched: set[str] = set()
+    for scheme in config.config_groups.values():
+        for module_name, name in match_quantizable_tensors(
```
brian-dellabetta, line 50: this is desired behavior but does not match what currently happens in apply_quantization_config. This pathway is preferred, as the other will cause an edge case bug that can't be loaded in vllm.

Wanted to call this out as an FYI, probably don't want to change here.

more information in https://github.com/vllm-project/compressed-tensors/issues/862

brian-dellabetta (approved):
The behavior is different than `apply_quantization_config`, see comment below. I will mark as approved, but please don't merge until @kylesayrs reviews and we are in agreement that we should switch to this scheme selection logic (more information in https://github.com/vllm-project/compressed-tensors/issues/862)

src/llmcompressor/entrypoints/model_free/converter.py
soyr-redhat: great callout, thank you brian!

soyr-redhat:
@brian-dellabetta sounds good, thank you for taking the time to review!

kylesayrs (approved):

kylesayrs:
Follow ups from me:
1. Since `model_free_ptq` is essentially duplicating `convert_checkpoint` entrypoint, we should have it call into `convert_checkpoint` for consistency and code-duplication
2. We should explore using something like [PTTP](https://github.com/kylesayrs/PTTP/tree/main/src/pttp) to measure how much memory each job requires during the validation step.

PR #2938, opened 2026-07-18
Add Cohere2MoE mappings (AWQ + SmoothQuant) and DeepseekV2 AWQ registration

robertlangdonn:
## Summary

Add mapping registrations for `Cohere2MoeForCausalLM` (AWQ + SmoothQuant) and `DeepseekV2ForCausalLM` (AWQ).

## Problem

- `Cohere2MoeForCausalLM` (e.g. `CohereLabs/North-Mini-Code-1.0`) gained SpinQuant support in #2867 but is absent from both the AWQ and SmoothQuant registries. Falling back doesn't work for this architecture: it keeps Cohere's parallel block (a single `input_layernorm` feeds attention and the MLP; there is no `post_attention_layernorm`), so the default mappings' patterns don't resolve against it, and the existing `_cohere_mappings` only cover a dense MLP while Cohere2MoE has a mixed dense/MoE stack.
- `DeepseekV2ForCausalLM` is registered in the SmoothQuant registry but missing from AWQ.

## Solution

**Cohere2MoE (AWQ + SmoothQuant):** new mappings balance every `input_layernorm` consumer across both layer kinds — `self_attn.{q,k,v}_proj` plus unscoped `gate_proj$`/`up_proj$`, which match the dense first layer's MLP (`mlp.gate_proj`/`up_proj`, `first_k_dense_replace`) and every expert (`mlp.experts.N.gate_proj`/`up_proj`) after MoE linearization. The router (`mlp.gate`) is omitted per the Glm4MoeLite precedent: it does not exist on the dense layer, and including it triggers the per-layer grouping collapse fixed for GlmMoeDsa in #2923. The tradeoff this shares with the Glm4MoeLite/GlmMoeDsa mappings (routing logits see the smoothed activations uncompensated) is documented in-code — for this architecture it can't be avoided by skipping MLP smoothing (the Mixtral/Qwen-MoE convention), because the parallel block's shared norm would lose attention smoothing too.

**DeepseekV2 (AWQ):** registered to `_glm4_moe_lite_mappings` rather than `_deepseek_mappings`. V2 has the same mixed dense/MoE MLA shape as Glm4MoeLite (`first_k_dense_replace` dense layers without a router), so `_deepseek_mappings` — whose MLP balance list includes `mlp.gate$` — would reproduce the #2923 collapse on the dense layer. `_glm4_moe_lite_mappings` already handles the `q_proj`/`q_a_proj` MLA variants.

## Testing

- `ruff check` passed on all touched files.
- `test_new_architecture_mappings_resolve` (SmoothQuant registry test) extended with the Cohere2MoE entry.
- Pattern-closure verified against synthetic dense and MoE Cohere2MoE layer module lists: every smooth/balance pattern in both new mappings matches ≥1 module on each layer kind (the #2923 failure mode), including a negative control confirming `_deepseek_mappings` does **not** close on a V2-style dense layer (motivating the `_glm4_moe_lite_mappings` choice).
- Pure registry data — no GPU tests applicable.


robertlangdonn:
Code complete and locally tested (ruff + registry resolution + pattern-closure checks against dense/MoE layer shapes) — ready for review. I don't have permissions to add the `ready` label myself.

robertlangdonn:
Ran the empirical validation requested — real oneshot AWQ (both mappings this PR adds are AWQ-registered) against actual public tiny checkpoints, not just the synthetic module-tree pattern checks already in the test suite.

**Cohere2MoE**: `hf-tiny-v2/tiny-random-Cohere2MoeForCausalLM`, `AWQModifier()` (mapping auto-inferred from `AWQ_MAPPING_REGISTRY`) + `QuantizationModifier` (W4A16, channelwise to fit the tiny model's dims). Full pipeline ran clean: calibration → smoothing (all layers, including both the dense-layer MLP and every MoE expert + shared expert) → quantization → save (compressed) → reload as `Cohere2MoeForCausalLM`. No NaN/Inf in any parameter post-quantization, no `Skipping smooth_layer` warnings (i.e. no mapping silently failed to find activations).

**DeepseekV2**: `hf-tiny-v2/tiny-random-DeepseekV2ForCausalLM`, same setup, `_glm4_moe_lite_mappings` auto-inferred. Same result: clean full pipeline, correctly smoothed both `mlp.experts.N.up_proj` and `mlp.shared_experts.up_proj`, no NaN/Inf, no silent skips, reload OK as `DeepseekV2ForCausalLM`.

Generation output on both is gibberish as expected — these are `tiny-random` checkpoints with untrained random weights, not real models, so coherent text isn't the signal here; the signal is that the real resolution/smoothing/quantization/save/reload path executes cleanly end-to-end with these mappings on the actual model classes, which the registry-only unit tests couldn't exercise.

Happy to attach the validation scripts if useful for review.

brian-dellabetta (approved):
thanks @robertlangdonn

HDCharles (approved):

PR #2943, opened 2026-07-20
fix DDP obs stat deletion

HDCharles:
Summary:

in https://github.com/vllm-project/llm-compressor/pull/2865 we made it so obs clean up their stats to improve memory footprint during quantization. However certain modifiers weren't deleting statistics on DDP because of the job assignment, deletion was happening on obs.get_qparams which only gets called on a subset of the modules for each rank. 

This affected both GPTQ and QuantizationModifier since they both use job assignment, i took this opportunity to unify their broadcast implementations so we have a single helper and added the observer deletion to this new helper.

Test Plan:
https://gist.github.com/HDCharles/6f98c5be0cd51b8351f1ae10c97a7e8a
image

before
28948 Addr: b'7f0e9174b800_3, Size: 2.6MiB (2715648 bytes) allocation, Total memory used after allocation: **>>140.1<<**MiB (146889205 bytes), 
30878 Addr: b'7f0e663ea000_2, Size: 2.6MiB (2715648 bytes) allocation, Total memory used after allocation: **>>142.1<<**MiB (148986357 bytes), 
32804 Addr: b'7f0e64868000_4, Size: 2.6MiB (2715648 bytes) allocation, Total memory used after allocation: **>>144.1<<**MiB (151083509 bytes), 

after
28948 Addr: b'7fac5174b800_3, Size: 2.6MiB (2715648 bytes) allocation, Total memory used after allocation: **>>120.6<<**MiB (126441973 bytes), 
30878 Addr: b'7fac263ea000_2, Size: 2.6MiB (2715648 bytes) allocation, Total memory used after allocation: **>>120.6<<**MiB (126441973 bytes), 
32804 Addr: b'7fac24868000_4, Size: 2.6MiB (2715648 bytes) allocation, Total memory used after allocation: **>>120.6<<**MiB (126441973 bytes), 

CI: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/29769647748

### DIFF
--- a/src/llmcompressor/modifiers/quantization/quantization/base.py
+++ b/src/llmcompressor/modifiers/quantization/quantization/base.py
@@ -13,6 +13,7 @@ from llmcompressor.modifiers.quantization.calibration import (
     update_qparams,
 )
 from llmcompressor.modifiers.quantization.quantization.mixin import QuantizationMixin
+from llmcompressor.observers import ACTIVATION_OBS
 from llmcompressor.utils.dist import broadcast_qparams_and_cleanup
 
 __all__ = ["QuantizationModifier"]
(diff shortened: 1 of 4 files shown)


src/llmcompressor/modifiers/quantization/quantization/base.py
```diff
@@ -107,31 +104,12 @@ def on_sequential_epoch_end(
 
         observe(rank_to_modules[rank], "weight")
         update_qparams(rank_to_modules[rank], "weight")
-        dist.barrier()  # wait for async offload updates to finish
```
HDCharles, line 110: not needed, those ranks can start broadcasting as soon as they're done, they don't have to wait, thats what wait for comms is for.

kylesayrs (approved):
Thank you!

brian-dellabetta (approved):

PR #2950, opened 2026-07-22
MSE Observer Enhancement

Roderick-Wu:
Improve the use of mse observer for NVFP4. 
-- Grid search expansion factor added as observer kwargs -- there are settings that enable a search that is equivalent to a superset of fouroversix. By expanding the grid search range, we cover fouroversix as well as potentially better ranges. 
-- Adds new nvfp4 observer with good defaults, which also disables rounding during grid search. 

fouroversix (https://arxiv.org/abs/2512.02010) chooses between two quantization ranges: the full absmax or absmax\*1.5. The reasoning is that in fp4 the largest values jump from 4 --> 6 which is a significatn gap compared to previous increments. By instead mapping xmin and xmax to +-4 (aka mapping 1.5xmin/xmin to +-6) we use fewer values (thereby losing some granularity) but in exchange we ensures that rounding error for those values large values will be smaller. This is better for some blocks. 

We can go beyond this and start the grid search at a higher max -- beyond the maximum observed value -- and gradually shrink the scale. This includes searching over the two points used in fouroversix, as well as any scale range between them. 

| Config | Llama-8B | Qwen3-8B | Qwen3-14B | Qwen3-32B | Llama-70B | MoE-30B | Avg Delta |
|---|---|---|---|---|---|---|---|
| minmax (nvfp4 default) | +0.333 | +0.284 | +0.138 | +0.099 | +0.201 | +0.208 | +0.211 |
| fouroversix | +0.230 | +0.098 | +0.128 | +0.183 | +0.261 | +0.300 | +0.200 |
| 1x + 1.5x, no fp8 rounding | +0.223 | +0.168 | +0.124 | **+0.159** | +0.256 | +0.363 | +0.216 |
| 1x + 1.5x, global-scale prior determined locally (1x) | +0.233 | +0.255 | +0.086 | +0.167 | +0.353 | +0.323 | +0.236 |
| default mse observer | +0.299 | +0.270 | +0.137 | +0.135 | +0.193 | +0.250 | +0.214 |
| Expanded search with default settings | +0.242 | +0.028 | +0.093 | +0.171 | +0.149 | +0.149 | **+0.139** |

We also include this expand kwarg for the imatrix observer. Using the same default mse params, we find that it performs slightly worse but within reasonable error. Tuning imatrix parameters with the expanded search can be done later. Both of them outperform gptq. Both imatrix and gptq use ultrachat dataset. 

| Config | Llama-8B | Qwen3-8B | Qwen3-14B | Qwen3-32B | Llama-70B | MoE-30B | Avg Delta |
|---|---|---|---|---|---|---|---|
| mse-expanded (same as above) | +0.242 | +0.028 | +0.093 | +0.171 | +0.149 | +0.149 | +0.139 |
| imatrix-expanded | +0.242 | +0.056 | +0.138 | +0.135 | +0.264 | +0.120 | +0.159 |
| gptq-nvfp4 | +0.238 | +0.176 | +0.158 | +0.123 | +0.340 | +0.096 | +0.189 |

## Justification

Starting at very high range and shrinking down, empirically observe all blocks follow this distribution:
image
Register new observer `"nvfp4_mse"` with defaults
```
        self.expand = observer_kwargs.get("expand", 1.8)
        self.maxshrink = observer_kwargs.get("maxshrink", 1 - 0.8 / 1.8) # From 1.8x to 0.8x
        self.grid = observer_kwargs.get("grid", 200.0)
        self.patience = observer_kwargs.get("patience", 1000) # Get the whole range
```

Fouroversix does set the global scale beforehand. The mse observer performs grid search without a global scale. We consider using a global-scale prior guess calculated from observed absmax values. When we include good guesses for the global scale ahead of time we can also improve over fouroversix. 

| Config | Llama-8B | Qwen3-8B | Qwen3-14B | Qwen3-32B | Llama-70B | MoE-30B | Avg Delta |
|---|---|---|---|---|---|---|---|
| expanded + global-scale determined at start (1.8x) + fused statistics | +0.257 | +0.017 | +0.132 | +0.185 | +0.139 | +0.222 | +0.159 |
| expanded (1.5x) + global-scale determined at start (1.5x) + no fusion | +0.239 | +0.090 | +0.165 | +0.167 | +0.389 | +0.194 | +0.207 |
| expanded (1.8x) + global-scale prior (1.25x, gs recalculated at end) + no fusion  | +0.142 | +0.302 | **+0.043** | +0.051 | +0.186 | - | +0.145 |
| expanded (1.8x) + global-scale prior (1.75x, gs recalculated at end) + no fusion | +0.170 | +0.284 | +0.123 | +0.045 | +0.171 | - | +0.159 |

These are improvements over fouroversix and existing mse implementation. However, still worse than just dropping global scale from grid search and disabling roudning. 

Code for tests + compression and eval scripts:
https://github.com/vllm-project/llm-compressor/pull/3042


src/llmcompressor/observers/mse_quant.py
```diff
@@ -3,12 +3,12 @@
...
 import torch._dynamo.decorators
-from compressed_tensors.quantization import QuantizationArgs
+from compressed_tensors.quantization import QuantizationArgs, QuantizationStrategy
 from compressed_tensors.quantization.lifecycle import fake_quantize
-from compressed_tensors.quantization.utils import calculate_qparams
+from compressed_tensors.quantization.utils import calculate_qparams, generate_gparam
 
 from llmcompressor.core import active_session
-from llmcompressor.observers.base import MinMaxTuple
+from llmcompressor.observers.base import MinMaxTuple, _CustomFP8ScaleData
```
brian-dellabetta: Consider making this public if you are importing it to other files in the module. you don't have to export it from the module
```suggestion
from llmcompressor.observers.base import MinMaxTuple, CustomFP8ScaleData
```

brian-dellabetta (approved):
This looks good to me pending @HDCharles 's thoughts, since I know you've been working closely on the 4over6 stuff with him. usage of observer kwargs seems like the right place to plug this in

HDCharles:
a few questions, comments

1) why is hte expand value 2x rather than 1.5?
2) the FP8 scale max thing makes no sense. They need to change that value because they pick their global scale first then choose whether they're doing 4 or 6 for each group and want both to be valid. We can just do our thing and pick global scale at the end. If you're seeing benefits its likely due to something else i.e. we never do actual fp8 rounding for the scale in MSE, something like that makes way more sense as an option. I think the reason why this helps needs to be looked into. 
2a) this is also entangled with the global scalemax, or are those different?
3) i'm unsure about the defaults being used, how were these evals and quantizations done? can you provide a link, by default the maxshrink is .2, if you're setting expand to be 2x i think that means at most you're shrinking things to (1-.2)*2=1.6 which is not what we should be doing as a comparison. It should be set so that we exactly hit the 1.5x and 1.0 points at a minimum. what min/max fractions are you seeing when you run these evals?
4) also about defaults, what patience are you using? we shouldn't cut the search off before we hit the 1.5 and 1.0 points at a minimum, how often is that happening?

HDCharles (changes requested):
see comment

src/llmcompressor/observers/mse.py
```diff
@@ -21,11 +21,12 @@ class MemorylessMSEObserver(Observer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         observer_kwargs = self.args.observer_kwargs
-        self.maxshrink = observer_kwargs.get("maxshrink", 0.20)
+        self.maxshrink = observer_kwargs.get("maxshrink", 0.6)
         self.patience = observer_kwargs.get("patience", 5)
         self.grid = observer_kwargs.get("grid", 100.0)
         self.norm = observer_kwargs.get("norm", 2.4)
         self.chunk_size = observer_kwargs.get("chunk_size", 5)
+        self.expand = observer_kwargs.get("expand", 2.0)
```
HDCharles: leave the defaults so that they work as normal, we should probably do something like creating an alias which has the defaults set, i'd recommend something like

@Observer.register("nvfp4_expand")
NVFP4ExpandObserver -> defaults that search from 1.6 -> .8 or something like that, can probably do 
some ablations about the defaults for this, it doesn't seem like starting at 2 was the optimal, right?

basically the UX we want for this is e.g. https://github.com/vllm-project/llm-compressor/blob/50d89b1993b54470a9d3d4c905caa95f945220aa/examples/quantization_w4a4_fp4/llama3_gptq_example.py#L81 where we're using a string to pick the observer

src/llmcompressor/observers/mse.py
```diff
@@ -96,3 +100,28 @@ def update_statistics_from_observed(self, observed: torch.Tensor) -> None:
...
+        QuantizationArgs(
+            ...
+            observer="nvfp4_mse",
+        )
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        observer_kwargs = self.args.observer_kwargs
+        self.expand = observer_kwargs.get("expand", 1.8)
```
HDCharles, line 126: probably need to set patience to a higher value for this

src/llmcompressor/observers/mse.py
```diff
@@ -96,3 +100,28 @@ def update_statistics_from_observed(self, observed: torch.Tensor) -> None:
...
+        self.maxshrink = observer_kwargs.get("maxshrink", 0.56)
```
HDCharles: why this value?

src/llmcompressor/observers/mse_quant.py
```diff
@@ -44,9 +45,14 @@ def _grid_search_mse(
...
+    :param expand: factor to scale the initial min/max range before searching.
+        Values > 1.0 let the search explore ranges wider than the observed
+        values (e.g. expand=2.0 starts at 2x the observed range).
     """
-    min_val = torch.amin(observed, dim=(0, -1))
-    max_val = torch.amax(observed, dim=(0, -1))
+    original_min = torch.amin(observed, dim=(0, -1))
+    original_max = torch.amax(observed, dim=(0, -1))
+    min_val = original_min * expand
+    max_val = original_max * expand
```
HDCharles, line 55:
```suggestion
    min_val = torch.amin(observed, dim=(0, -1)) * expand
    max_val = torch.amax(observed, dim=(0, -1)) * expand
```

src/llmcompressor/observers/imatrix.py
```diff
@@ -49,6 +49,8 @@ def __init__(self, *args, **kwargs):
         self.grid = kw.get("grid", 20)
         self.norm = kw.get("norm", 3.0)
         self.strict = kw.get("strict", False)
+        self.expand = kw.get("expand", 1.0)
+        self._search_args = self.args
```
HDCharles: it doesn't look like this is fixing the MSE bug, normal MSE observer is still going to do fp8 rounding, right?

I would also say this is a bit awkard of an approach, better to just edit the args when you pass them into the grid search such that the scale_dtype is None rather than storing a full copy

src/llmcompressor/observers/mse.py
```diff
@@ -96,3 +104,31 @@ def update_statistics_from_observed(self, observed: torch.Tensor) -> None:
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        observer_kwargs = self.args.observer_kwargs
+        self.expand = observer_kwargs.get("expand", 1.8)
+        self.maxshrink = observer_kwargs.get("maxshrink", 1 - 0.8 / 1.8)
+        self.grid = observer_kwargs.get("grid", 200.0)
+        self.patience = observer_kwargs.get("patience", 1000)
+        self._search_args = self.args.model_copy(update={"scale_dtype": None}) # disable rounding
```
HDCharles: this change is supposed to be for all mse obs, probably better to put it in grid search then.

src/llmcompressor/observers/imatrix.py
```diff
@@ -317,3 +326,28 @@ def _grid_search(
...
+        )
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        kw = self.args.observer_kwargs
+        self.expand = kw.get("expand", 1.8)
+        self.maxshrink = kw.get("maxshrink", 1 - 0.8 / 1.8)
+        self.grid = kw.get("grid", 200)
+        self.patience = kw.get("patience", 1000)
```
HDCharles: leave this for a followup PR where you have results

HDCharles (approved):
good, the imatrix thing is an experiment so we add functionality (expand) but not the UX helper, once that's removed this will be ready to land

examples/quantization_w4a16_fp4/nvfp4/llama3_nvfp4_expanded_mse_example.py
```diff
@@ -0,0 +1,62 @@
...
+                dynamic=False,
+                group_size=16,
+                scale_dtype=FP8_E4M3_DATA.dtype,
+                zp_dtype=FP8_E4M3_DATA.dtype,
+                observer="nvfp4_expanded_mse",
+            ),
+        )
+    },
+    ignore=["lm_head"],
+)
```
kylesayrs, line 28:
```suggestion
recipe = QuantizationModifier(
    scheme="NVFP4",
    weight_observer="nvfp4_expanded_mse",
    ignore=["lm_head"],
)
```

kylesayrs (approved):

kylesayrs (approved):

PR #2957, opened 2026-07-23
Add logging to silently swallowed exceptions in ModifierFactory

arijitroy003:
## Summary

Multiple exception handlers in ModifierFactory silently swallow errors without logging, making debugging difficult when modifiers fail to load. This adds debug-level logging to trace import and loading failures.

## Changes

- Added logger import and initialization in factory.py
- Added logger.debug calls with exc_info=True to 3 bare exception handlers that were silently passing
- Added logger.debug calls to 2 conditional skip paths that had TODO comments about logging
- No control flow changes — errors still stored in _errors dict and print statements preserved

## Testing

Verified by grep that all exception handlers now log before passing/continuing.

### DIFF
--- a/src/llmcompressor/modifiers/factory.py
+++ b/src/llmcompressor/modifiers/factory.py
@@ -1,7 +1,10 @@
 import importlib
+import logging
 import pkgutil
 
 from llmcompressor.modifiers.modifier import Modifier
+
+logger = logging.getLogger(__name__)
 
 __all__ = ["ModifierFactory"]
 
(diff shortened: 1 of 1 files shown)


kylesayrs (changes requested):
This codebase uses `from loguru import logger; logger.debug`. Otherwise lgtm

arijitroy003:
Updated — switched from stdlib logging to loguru as requested. Using `logger.opt(exception=True).debug(...)` for exception tracebacks and removed the stray `print()`. Thanks for the review!

PR #2971, opened 2026-07-24
Enable applying AWQ and SmoothQuant with GPTQ and the QuantizationModifier

dsikka:
Also adds support for seleting prebaked datasets


examples/.claude/skills/shared_quantization.md
```diff
@@ -72,6 +72,205 @@ Some deepseek-like architectures use an `attn.indexer` and `attn.indexer.compres
...
+```
+
+## Transforms
+
+A transform can improve quantization accuracy by redistributing quantization difficulty before the quantization step. Transforms work with any quantization scheme and compose with either `QuantizationModifier` or `GPTQModifier`.
+
+Ask the user if they want to apply a transform (default: No):
+
+- **SmoothQuant** — Migrates quantization difficulty from activations to weights via channel-wise scaling. Key parameter: `smoothing_strength` (default: `0.5`, range 0–1; higher values shift more difficulty to weights).
+- **AWQ** — Activation-aware Weight Quantization. Identifies salient weights based on activation magnitudes and applies channel-wise scaling to preserve them. Key parameter: `duo_scaling` (default: `"both"`).
```
soyr-redhat, line 216: Hey! Was just reading through. I believe in [base.py:148](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/transform/awq/base.py#L148), duo_scaling defaults to True, not "both".  Would this be of value to call  out here?

dsikka: The default: `duo_scaling: bool | Literal["both"] = True` includes the literal `both`

soyr-redhat: Understood, makes total sense. Sorry for stepping on your toes!

examples/.claude/skills/shared_quantization.md
```diff
@@ -72,6 +72,229 @@ Some deepseek-like architectures use an `attn.indexer` and `attn.indexer.compres
...
+    AWQModifier(duo_scaling="both"),
+    QuantizationModifier(
+        targets="Linear",
+        scheme="<SCHEME>",
+        ignore=["lm_head"],
+    ),
+]
+```
+
+Transforms compose with GPTQ as well — replace `QuantizationModifier` with `GPTQModifier` in the recipe list and update the save directory suffix accordingly (e.g. `-NVFP4-GPTQ-SmoothQuant`).
```
kylesayrs, line 250: I think adding transforms is risky, as their support in vLLM is limited.

kylesayrs (approved):
Nice reuse. Unrelated to this PR but related to skill reuse in general, I've found that the claude code harness is not set up for skills to call subskills, so I'm glad we avoid that pattern in this case.

examples/.claude/skills/shared_quantization.md
dsikka: Yeah just SmoothQuant and AWQ are the only two the skill is set-up to look into

examples/.claude/skills/shared_quantization.md
```diff
@@ -72,6 +72,229 @@ Some deepseek-like architectures use an `attn.indexer` and `attn.indexer.compres
...
+
+- Keep the default (`ultrachat_200k`, manual processing) — **recommended default**
+- `perfectblend` — general-purpose text blend
+- `open_platypus` — instruction
+- Or the user can type any other registered dataset name.
+
+Registered datasets are passed to `oneshot` **by name** and handle loading, preprocessing, and tokenization internally. The full list of registered names lives in `src/llmcompressor/transformers/data/` (each file has a `@TextGenerationDataset.register(name=...)` decorator) — point the user there rather than enumerating every option.
+
+Also determine:
+- **Number of calibration samples** — how many samples to use (default: `256` for QuantizationModifier alone, `512` for GPTQModifier or when using a transform; MoE models may benefit from more samples)
```
soyr-redhat, line 274: Just a nit but maybe it's worth mentioning open_platypus genrally has smaller samples, so would actually require 2-3 times as many samples as ultrachat or perfect blend. 

LG otherwise!

examples/.claude/skills/templates/oneshot_with_data.py
soyr-redhat, line 19: nit: could also add 

`Use 512 when applying GPTQ or a transform (AWQ, SmoothQuant). 
`
here to reinforce what's being said below

soyr-redhat (commented):
just a couple tiny nits, otherwise, LGTM!

PR #2973, opened 2026-07-25
fix: runtime bugs in recipe validation, autoround, and awq

soyr-redhat:
The goal of this pr is to just fix a couple of bugs that arise at runtime

**1. truncated error message in AWQ recipe validation**
the error message at `recipe/recipe.py:225` ends mid-sentence with "AWQ must be run with " and never finishes. THere's also a typo, "modifer" -> "modifier" 

**2. duplicate `_unwrapper_quantized_layer` in AutoRoundModifier**
identical method defined at lines 394 and 435 of `autoround/base.py`. the second silently overwrites the first. I just simply removed the duplicate.

**3. division by zero in AWQ `_log_error_metrics`**
if `_error_metrics` is empty (e.g. all mappings skipped), `sum(reductions) / len(reductions)` raises `ZeroDivisionError` and `min()`/`max()` raise `ValueError` on empty sequence. added early return guard. This was mitigated by doing an early check for if not _error_metrics ahead of time.

no new tests should be required...these are fixes to existing error paths and dead code removal.


src/llmcompressor/recipe/recipe.py
```diff
@@ -224,7 +224,8 @@ def validate_model_after(model: "Recipe") -> "Recipe":
             ):
                 raise ValueError(
                     f"Recipe includes AWQModifier with no subsequent quantization "
-                    f"modifer: {model.modifiers}. AWQ must be run with "
+                    f"modifier: {model.modifiers}. AWQ must be run with "
+                    f"a subsequent quantization modifier"
```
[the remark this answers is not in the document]
soyr-redhat: I can see that being reasonable, not something that we touched, but could reduce verbosity for sure

src/llmcompressor/modifiers/transform/awq/base.py
```diff
@@ -847,6 +847,9 @@ def _log_error_metrics(self):
         # Save to disk
         logger.debug(f"AWQ per-mapping error metrics: {metrics_data}")
 
+        if not self._error_metrics:
+            return
```
brian-dellabetta, line 859: Hey @soyr-redhat , did you hit this? Having it silently move on is different behavior than erroring out, which can act as an assertion

soyr-redhat: Hey @brian-dellabetta ! I found it by reading through the code, didn't hit it at runtime. there's a division by zero possibility if _error_metrics is empty when on_finalize calls _log_error_metrics.

I was going back and forth over a logger.warning/early return, do you think one is better than the other

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -430,19 +430,6 @@ def _move_inputs_to(
     def _is_decoding_layer(self, module: torch.nn.Module) -> bool:
         return module.__class__.__name__ in self._sequential_targets
 
-    def _unwrapper_quantized_layer(self, model: torch.nn.Module):
```
brian-dellabetta, line 435: note -- this method is duplicated exactly on the class definition, safe to remove

src/llmcompressor/modifiers/transform/awq/base.py
brian-dellabetta: ok, yeah I think it's worth adding a warning that only logs once. example [here](https://github.com/vllm-project/compressed-tensors/pull/780/changes#diff-5b0cf029f7720c4b1589d9c775e0585384e0976a898e7211fc1c75bcbfc6ec28R144-R147)

soyr-redhat: Okay great! Took care of that, Just added that warning as requested!

brian-dellabetta (approved):

yiliu30 (approved):
Thanks @soyr-redhat, LGTM!

PR #2773, opened 2026-05-30
feat: multi gpu support for model free ptq

JINO-ROHIT:
add support to distribute the jobs across multiple gpus if present

fixes #2761

PR #2976, opened 2026-07-28
feat: dynamic memory-aware GPU scheduling for model-free PTQ

### LINKED ISSUE #2975: [ModelFreePTQ] Memory issues with multi-gpu
## Background ##
https://github.com/vllm-project/llm-compressor/pull/2773 introduced multi-gpu model free PTQ jobs. It works by statically assigning jobs to gpus round-robin style. However, this can lead to bad memory allocation, as slower gpus will accumulate more jobs than faster gpus, leading to OOM errors.

I have observed this behavior when using 6 gpus.

The better approach is to use a thread pool which dynamically assigns jobs to the gpu with the most resources. The best solution involves estimating job memory size (to account for hetereogenous gpu/job sizes), perhaps folding this into the validation step. Whenever the thread pool has an unassigned thread, the main thread checks which gpus can accommodate the job size and assigns it to the gpu with the most capacity.

This mechanism would also prevent OOMs caused by users picking too many workers

## Proposed Changes ##
1. Make GPU assignment dynamic and scheduled via the main thread
2. Add a multi-gpu test on a small model with many safetensors (you can adjust the safetensors file size when saving models with transformers)

rohan9446:
## SUMMARY:

Fixes #2975,  replaces the static round-robin GPU assignment in model-free PTQ with a dynamic, memory-aware scheduler.

### Problem

PR #2773 introduced multi-GPU support by statically assigning shards to GPUs via `devices[i % len(devices)]`. When shards differ in size or GPUs differ in speed, the slower GPU accumulates pending work and eventually OOMs. This was observed on 6-GPU setups.

### Solution

New `scheduler.py` module with a capacity-first scheduling loop:

- **Dynamic device selection**: before submitting each job, the main thread calls `torch.get_device_module(dev).mem_get_info()` and picks the GPU with the most free VRAM
- **Reservation tracking**: submitted-but-not-yet-allocated memory is tracked per device, preventing multiple jobs from stacking on the same GPU in a single scheduling pass
- **Forced fallback**: if no device can fit any pending job and nothing is in flight, the smallest job is forced onto the most-free device with a warning (instead of silently dropping work)
- **Caching allocator flush**: `empty_cache()` is called after harvesting each completed future so `mem_get_info` reflects actually-free memory
- **`concurrent.futures.wait(FIRST_COMPLETED)`** replaces a hand-rolled `threading.Event` callback, eliminating a lost-wakeup race

`_build_jobs()` no longer takes a `devices` parameter — device assignment is deferred entirely to the scheduler. Validation still uses upstream `exec_jobs` (it runs on meta device, so no scheduling needed).

### Behavioral change

`max_workers` now sets an upper bound on concurrency rather than a fixed thread count. Effective concurrency may be lower when GPU memory is tight. This also prevents OOMs caused by users picking too many workers.

### Unit tests (`test_scheduler.py`, no GPU required — runs on all CI):
- `estimate_job_memory`: verifies sum-of-file-sizes × multiplier
- `_pick_device`: most-free selection, returns `None` when nothing fits, respects reservation accounting, CPU always eligible
- `exec_jobs_dynamic` CPU path: all jobs complete, empty input, order preservation

### Integration tests (`test_multi_gpu.py`, `@requires_gpu(2)`):
- `test_multi_gpu_matches_single_gpu`: saves Qwen3-0.6B with `max_shard_size="1MB"` (~198 shards), quantizes with FP8_dynamic on 1 GPU vs N GPUs, asserts bitwise tensor equality
- `test_multi_gpu_more_workers_than_shards`: verifies graceful handling when `max_workers > len(jobs)`

### Validated on:
- 2× NVIDIA RTX PRO 6000 Blackwell (homogeneous), Ubuntu 24.04, PyTorch 2.13.0+cu130
- `make quality` passes (ruff check + ruff format + `tools/lint_cuda.py`)
- All 11 tests pass


tests/llmcompressor/entrypoints/model_free/test_dynamic_multi_gpu.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.multi_gpu
+@requires_gpu(2)
+def test_multi_gpu_more_workers_than_shards(tmp_path):
+    """Scheduler should be fine when there are more threads than shards."""
+    model_id = "Qwen/Qwen3-0.6B"
+    scheme = "FP8_dynamic"
+    ignore = ["model.embed_tokens", "lm_head"]
+
+    out = tmp_path / "output"
+    n = torch.accelerator.device_count()
```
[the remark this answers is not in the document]
rohan9446: Thanks for the suggestion. This repository enforces accelerator-agnostic APIs through tools/lint_cuda.py, which flags torch.cuda.device_count() and causes make quality to fail. Therefore, I’m keeping torch.accelerator.device_count() to follow the project’s linting and portability requirements.

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,188 @@
...
+    """Rough memory estimate for a shard job, based on source file sizes."""
+    total = sum(os.path.getsize(p) for p in inverse_weight_map)
+    return int(total * _MEMORY_MULTIPLIER)
+
+
+def _free_bytes(dev: torch.device, reserved: dict[torch.device, int]) -> int:
+    """Driver-reported free VRAM minus bytes already promised to queued
+    jobs."""
+    free, _ = torch.get_device_module(dev).mem_get_info(dev)
+    return free - reserved.get(dev, 0)
```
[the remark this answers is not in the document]
rohan9446: Addressed in the latest update: _free_bytes now handles CPU devices before accessing the device module, checks whether mem_get_info is available, and raises a clear RuntimeError for unsupported accelerator backends. It also clamps the calculated available memory to zero.

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,188 @@
...
+                    key=lambda d: _free_bytes(d, reserved),
+                )
+                logger.warning(
+                    f"Shard {i} exceeds estimated capacity on all "
+                    f"devices, falling back to {dev}"
+                )
+            logger.debug(
+                f"Shard {i} -> {dev} " f"(~{memory_estimates[i] / 1e9:.2f} GB)"
+            )
+            out.append(fn(iwm, sp, sch, ign, dev, conv))
```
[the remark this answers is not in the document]
rohan9446: The single-worker path now calls _empty_cache(dev) after each completed job so that subsequent memory checks reflect memory released by the caching allocator.

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,188 @@
...
+
+                for f in done:
+                    i = inflight.pop(f)
+                    dev = fut_device.pop(f)
+                    reserved[dev] -= memory_estimates[i]
+                    results[i] = f.result()  # propagates exceptions
+                    bar.update(1)
+                    # flush the caching allocator so mem_get_info
+                    # reflects actually-free memory next pass
+                    torch.get_device_module(dev).empty_cache()
```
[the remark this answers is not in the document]
rohan9446: I added an _empty_cache helper that no-ops for CPU devices and only calls the backend’s empty_cache method when it is available. Both the single-worker and multi-worker paths now use this helper.

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+# (scales, zero-points, compressed output buffers).  Conservative on purpose
+# — a slight overestimate just means we serialize a bit more; an underestimate
+# means OOM.
+_MEMORY_MULTIPLIER = 3.0
+
+
+def estimate_job_memory(inverse_weight_map: InverseWeightMap) -> int:
+    """Rough memory estimate for a shard job, based on source file sizes."""
+    total = sum(os.path.getsize(p) for p in inverse_weight_map)
+    return int(total * _MEMORY_MULTIPLIER)
```
kylesayrs, line 36: This is a good enough heuristic, nice job

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -63,12 +67,14 @@ def model_free_ptq(
     :param scheme: weight quantization scheme or preset scheme name
     :param ignore: modules to ignore. Modules ending with "norm" are
         automatically ignored
-    :param max_workers: number of worker threads to process files with
-    :param device: gpu devices to accelerate quantization with.
+    :param max_workers: maximum number of concurrent worker threads.
+        Effective concurrency may be lower when GPU memory is tight.
+    :param device: device(s) for quantization. Accepts a single device
+        string/object or a list. When multiple devices are given, shards
+        are dynamically assigned based on real-time GPU memory.
```
kylesayrs, line 74: Perfect

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -86,24 +92,41 @@ def model_free_ptq(
...
+            validate_file,
+            job[1],
+            job[2],
+            job[3],
+            job[4],
+            torch.device("meta"),
+            job[5],
+        )
+        for job in jobs
+    ]
```
kylesayrs, line 105: Consider just changing the function signatures of `validate_file` and `process_file` to make this cleaner

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+    """Driver-reported free VRAM minus bytes already promised to queued
+    jobs.  Returns ``inf`` for CPU devices (unbounded capacity)."""
+    if dev.type == "cpu":
+        return float("inf")
+    module = torch.get_device_module(dev)
+    mem_get_info = getattr(module, "mem_get_info", None)
+    if mem_get_info is None:
+        raise RuntimeError(f"{dev.type!r} backend does not support mem_get_info")
+    free, _ = mem_get_info(dev)
+    return max(0, free - reserved.get(dev, 0))
```
kylesayrs: `torch.accelerator.memory.get_memory_info()[0]`

See: https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.memory.get_memory_info.html#torch.accelerator.memory.get_memory_info

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def _empty_cache(dev: torch.device) -> None:
+    """Flush the caching allocator if the backend supports it."""
+    if dev.type == "cpu":
+        return
+    module = torch.get_device_module(dev)
+    empty_cache = getattr(module, "empty_cache", None)
+    if empty_cache is not None:
+        empty_cache()
```
kylesayrs: https://docs.pytorch.org/docs/2.13/generated/torch.accelerator.memory.empty_cache.html#torch.accelerator.memory.empty_cache

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+
+                for f in done:
+                    i = inflight.pop(f)
+                    dev = fut_device.pop(f)
+                    reserved[dev] -= memory_estimates[i]
+                    results[i] = f.result()  # propagates exceptions
+                    bar.update(1)
+                    # flush the caching allocator so mem_get_info
+                    # reflects actually-free memory next pass
+                    _empty_cache(dev)
```
kylesayrs: I think we should avoid calling `empty_cache` and `mem_get_info` for each scheduler pass which can lead to slowdowns

I think it'd be better to find the total memory for each gpu once at the beginning, and then subtract/add `memory_estimates[i]`

```python3
non_reserved_memory[dev] = torch.accelerator.get_mem_info()[1]
non_reserved_memory[dev] -= memory_estimates[i]
non_reserved_memory[dev] += memory_estimates[i]
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+                        key=lambda d: _free_bytes(d, reserved),
+                    )
+                    logger.warning(
+                        f"Shard {idx} "
+                        f"(~{memory_estimates[idx] / 1e9:.2f} GB) "
+                        f"exceeds estimated capacity on all devices; "
+                        f"forcing onto {dev}"
+                    )
+                    fn, iwm, sp, sch, ign, conv = jobs[idx]
+                    fut = pool.submit(fn, iwm, sp, sch, ign, dev, conv)
```
kylesayrs:
```suggestion
                    fut = pool.submit(*jobs[idx])
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+                submitted = []
+                for idx in pending:
+                    if len(inflight) >= max_workers:
+                        break
+                    dev = _pick_device(devices, memory_estimates[idx], reserved)
+                    if dev is None:
+                        continue
+
+                    fn, iwm, sp, sch, ign, conv = jobs[idx]
+                    fut = pool.submit(fn, iwm, sp, sch, ign, dev, conv)
```
kylesayrs:
```suggestion
                    fut = pool.submit(*jobs[idx])
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+    the estimated footprint.
+    """
+    n = len(jobs)
+
+    # CPU path — nothing to schedule
+    if all(d.type == "cpu" for d in devices):
+        out = []
+        for job in tqdm.tqdm(jobs, desc=desc):
+            fn, iwm, sp, sch, ign, conv = job
+            out.append(fn(iwm, sp, sch, ign, devices[0], conv))
```
kylesayrs, line 107:
```suggestion
            out.append(job[0](*job[1:]))
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+                logger.warning(
+                    f"Shard {i} exceeds estimated capacity on all "
+                    f"devices, falling back to {dev}"
+                )
+            logger.debug(
+                f"Shard {i} -> {dev} " f"(~{memory_estimates[i] / 1e9:.2f} GB)"
+            )
+            res = fn(iwm, sp, sch, ign, dev, conv)
+            out.append(res)
+            _empty_cache(dev)
```
kylesayrs:
```suggestion
        device_memory = {d: _free_bytes(d, reserved) for d in devices}
        device = max(device_memory, key=lambda k: device_memory[k])
        for i, job in enumerate(tqdm.tqdm(jobs, desc=desc)):
            if memory_estimates[i] > device_memory[device]:
                logger.warning(
                    f"Shard {i} exceeds estimated capacity"  # consider decreasing concurrency of adjusting estimation multiplier
                )
            out.append(job[0](*job[1:]))
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+                    submitted.append(idx)
+                    logger.debug(
+                        f"Shard {idx} -> {dev} "
+                        f"(~{memory_estimates[idx] / 1e9:.2f} GB)"
+                    )
+
+                for idx in submitted:
+                    pending.remove(idx)
+
+                # --- nothing running and nothing fits: force it ----------
```
kylesayrs: In this case, I think it's best to not force a job (which could lead to oom). Instead, we should give a warning about adjusting `_MEMORY_MULTIPLIER` or reducing max concurrency, and wait for the next job to finish.

If there are no pending jobs, then clearly the gpus are too small (or our estimate is bad) and we should error

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+                    dev = _pick_device(devices, memory_estimates[idx], reserved)
+                    if dev is None:
+                        continue
+
+                    fn, iwm, sp, sch, ign, conv = jobs[idx]
+                    fut = pool.submit(fn, iwm, sp, sch, ign, dev, conv)
+                    inflight[fut] = idx
+                    fut_device[fut] = dev
+                    reserved[dev] += memory_estimates[idx]
+                    submitted.append(idx)
```
kylesayrs: Remove `submitted` variable

```suggestion
                    pending.remove(idx)
```

```
for idx in list(pending):
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+    fut_device: dict = {}  # future -> device, for releasing reservations
+
+    with tqdm.tqdm(total=n, desc=desc) as bar:
+        with ThreadPoolExecutor(max_workers=max_workers) as pool:
+            inflight: dict = {}  # future -> job index
+
+            while pending or inflight:
+                # --- try to fill idle workers ----------------------------
+                submitted = []
+                for idx in pending:
```
kylesayrs: It seems like an infinite loop occurs if all of the jobs are larger than the gpus available.

Consider adding an `else` statement to this `for` loop which triggers if no jobs could be assigned

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Thanks!

src/llmcompressor/entrypoints/model_free/__init__.py
rohan9446: Thanks!

src/llmcompressor/entrypoints/model_free/__init__.py
rohan9446: Cleaned up the validation job construction using tuple unpacking. I kept the existing validate_file and process_file signatures to avoid expanding the scope of this PR, but I’m happy to refactor them if you’d prefer.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Done, switched to torch.accelerator.memory.get_memory_info(dev) and removed the backend specific getattr handling.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Addressed, if no pending job can be assigned and no jobs are currently in flight, the scheduler now raises a RuntimeError instead of continuing the loop indefinitely. I also added a unit test covering this case.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Done, changed the loop to iterate over list(pending) and remove each submitted job from pending inline, so the separate submitted list is no longer needed.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Done, removed the forced-scheduling fallback. When jobs are already in flight, the scheduler waits for one to complete and retries. If nothing is in flight and no device can accommodate a remaining shard, it now raises a clear RuntimeError instead of risking an OOM.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Done, the single worker path now selects the device with the most available memory once upfront and processes all jobs on that device. It also warns when a shard’s estimated memory exceeds the device’s available capacity.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: The device is intentionally omitted from the job tuple and inserted before the converter argument when the function is called. Using job[0](*job[1:]) would omit the required device argument, so I kept the explicit unpacking for clarity.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: I initially replaced the backend-specific helper with the generic accelerator API. Based on your follow-up performance suggestion below, I’ve now removed the repeated empty_cache() calls entirely.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Done, the scheduler now queries each device’s total memory once during initialization and maintains its own available-memory accounting. It subtracts a job’s estimated memory when the job is submitted and adds it back when the job completes, so it no longer calls empty_cache() or queries device memory on every scheduling pass.

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,185 @@
...
+                    raise RuntimeError(
+                        "No device has enough estimated free memory "
+                        "for any remaining shard. Consider reducing "
+                        "max_workers or adjusting _MEMORY_MULTIPLIER."
+                    )
+
+                # --- wait for at least one job to finish ---
+                done, _ = wait(
+                    inflight.keys(),
+                    timeout=2.0,
```
kylesayrs: Is this timeout actually necessary? Afaict, forward progress is only possible once/if a job finishes

kylesayrs (approved):
Nice job! Pinging @brian-dellabetta for review

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Good catch, removed. Forward progress only happens when a job finishes, so blocking indefinitely is correct.

kylesayrs (approved):

tests/llmcompressor/entrypoints/model_free/test_dynamic_multi_gpu.py
brian-dellabetta, line 1: we already have some multi-gpu tests in this folder. can you rename to `test_dynamic` etc to differentiate?

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,184 @@
...
+    """Rough memory estimate for a shard job, based on source file sizes."""
+    total = sum(os.path.getsize(p) for p in inverse_weight_map)
+    return int(total * _MEMORY_MULTIPLIER)
+
+
+def _snapshot_free(devices: list[torch.device]) -> dict[torch.device, int]:
+    """Query free VRAM once per device.  CPU devices are skipped."""
+    free = {}
+    for d in devices:
+        if d.type != "cpu":
```
brian-dellabetta: avoid a little redundancy
```suggestion
        if d.type != "cpu" and d not in free:
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,184 @@
...
+        return out
+
+    # Multi-worker: main thread schedules, workers execute
+    reserved = {d: 0 for d in devices}
+    results = [None] * n
+    pending = list(range(n))
+    fut_device: dict = {}  # future -> device, for releasing reservations
+
+    with tqdm.tqdm(total=n, desc=desc) as bar:
+        with ThreadPoolExecutor(max_workers=max_workers) as pool:
```
brian-dellabetta: should be able to consolidate these
```suggestion
    with tqdm.tqdm(total=n, desc=desc) as bar, ThreadPoolExecutor(max_workers=max_workers) as pool:
```

src/llmcompressor/entrypoints/model_free/scheduler.py
```diff
@@ -0,0 +1,184 @@
...
+    return free
+
+
+def _free_bytes(
+    dev: torch.device,
+    initial_free: dict[torch.device, int],
+    reserved: dict[torch.device, int],
+) -> int | float:
+    """Available VRAM for *dev*: initial snapshot minus promised reservations.
+    Returns ``inf`` for CPU devices (unbounded capacity)."""
```
brian-dellabetta: won't this assign all jobs at once to cpu? we want to make sure this pathway still works for `devices=[]`

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -86,24 +92,33 @@ def model_free_ptq(
...
+        for _, iwm, sp, sch, ign, conv in jobs
+    ]
     exec_jobs(validate_jobs, max_workers, desc="Validating")
 
-    # 2-5. quantize and compress weights
+    # quantize with dynamic GPU scheduling
     total_size = 0
     weight_map = dict()
-    quantize_results = exec_jobs(jobs, max_workers, desc="Quantizing")
+    quantize_results = exec_jobs_dynamic(
```
brian-dellabetta, line 111: why a new helper `exec_jobs_dynamic` instead of updating the pre-existing `exec_jobs`?

brian-dellabetta (commented):
Hi @rohan9446 , thanks for preparing this. i have a handful of comments/questions, particularly whether we want this just in the model-free-ptq pathway and not the convert_checkpoint pathway in compressed-tensors. We have some work underway to consolidate the 2 entirely:

* #2935

tests/llmcompressor/entrypoints/model_free/test_dynamic_multi_gpu.py
rohan9446: renamed test_multi_gpu.py to test_dynamic_multi_gpu.py to distinguish the dynamic-scheduler integration tests. I also renamed test_scheduler.py to test_dynamic_scheduler.py for consistency.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: applied the suggested d not in free guard so duplicate device entries are queried only once.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Done, consolidated the tqdm and ThreadPoolExecutor context managers into a single with statement.

src/llmcompressor/entrypoints/model_free/scheduler.py
rohan9446: Good point. The public model_free_ptq path does not pass a literal empty device list: _resolve_devices(None) returns [torch.device("cpu")] when no accelerator is available, while an explicitly empty list remains a ValueError, preserving the existing behavior.

I removed the CPU-as-infinite-capacity behavior from _free_bytes and _pick_device. The resolved CPU-only case now takes the sequential path before any scheduling logic, and the CPU-path tests cover that behavior.

If you intended devices=[] itself to mean CPU fallback rather than remain invalid, I can update that contract explicitly.

src/llmcompressor/entrypoints/model_free/__init__.py
rohan9446: exec_jobs is imported from compressed_tensors, so changing it directly would require a cross-repository change and broaden the scope of this PR. The dynamic scheduler also needs jobs without a device baked into each tuple, so I kept the implementation local to the model-free pathway for now.

Since #2935 is working toward consolidating model-free PTQ with the shared conversion pathway, I’m happy to migrate the scheduling logic into the shared implementation once that work lands, or adjust the approach now if you would prefer a cross-repository change.

src/llmcompressor/entrypoints/model_free/scheduler.py
brian-dellabetta: Great, thanks. General flow for a user would be `devices=[]` means just use CPU, as long at that works the same way as `devices=[torch.device("cpu")]` (and the number running concurrently is equal to max_workers) we should be good. These pathways are runnable on CPU in a reasonable amount of time, especially for smaller models, so i want to make sure we are considering it

src/llmcompressor/entrypoints/model_free/__init__.py
brian-dellabetta: the `convert_checkpoint` entrypoint in compressed-tensors follows a very similar structure and would stand to benefit from this dynamic scheduling as well, and would make resolving merge conflicts in #2935 much more straightforward, but would increase scope of this work. @kylesayrs WDYT?

rohan9446: Makes sense, happy to move it upstream into compressed_tensors if that's the preferred direction. Waiting on @kylesayrs  take.

kylesayrs: @brian-dellabetta @rohan9446 I've assigned @soyr-redhat to take the features from this PR and unify with the `convert_checkpoint` endpoint.

kylesayrs (approved):
Nice work!

kylesayrs (approved):

brian-dellabetta (approved):
thanks for the contribution @rohan9446 !

PR #2987, opened 2026-07-30
Fix GPTQ weight writeback crash on DeepSeek V3.2 forward

### LINKED ISSUE #2745: [Bug] `sequential_epoch_end` qparam writeback crashes with `Inplace update to inference tensor outside InferenceMode` on speculative-decode heads (DeepSeek-V4 MTP)
## Summary

Calibrating a DeepSeek-V4-class model with the multi-token-prediction (MTP) block in the recipe's quantization scope crashes deterministically at the END of the MTP block's calibration pass:

```
RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed.
You can make a clone to get a normal tensor before doing inplace update.
See https://github.com/pytorch/rfcs/pull/17 for more details.
```

The crash fires inside `compressed_tensors.distributed.utils.wait_for_comms` → `comm.wait()` during the `sequential_epoch_end` callback for the MTP block. Layers 0-42 (the main MoE blocks) calibrate cleanly with the same recipe shape; only the MTP block trips this. The MTP forward path in DeepSeek-V4-Flash includes a shared-embedding lookup (`mtp.0.embed.weight` aliases the main model's `embed.weight`) which is the most likely source of an inference-mode-marked tensor downstream — but the precise tensor genesis would need maintainer investigation.

**Two independent workstreams hit this same bug with the same root cause and the same workaround.** This issue plus the H200 W4A16 sibling repo's path (`pasta-paul/dsv4-flash-w4a16-fp8-mtp`) both converged on excluding MTP from the recipe (Option Y) as the fix. That convergence is part of why I'm filing — it's not a one-off.

## Reproducer

- Model: 671B-parameter DeepSeek-V4-Flash, BF16 source, 43 main MoE blocks + 1 MTP block, 256 routed experts per block.
- Hardware: 1× B300 SXM6 AC, on-demand (verified via IMDS).
- `llm-compressor` pin: `0.10.1.dev123+gf2aa32e2` (branch `kylesayrs/transformers-v5`).
- `compressed-tensors`: `0.15.1.a20260515`.
- Recipe (NVFP4 experts + FP8_BLOCK attn + MTP `e_proj`/`h_proj` quantized via `float-quantized`):

```python
QuantizationModifier(
    config_groups={
        "attention": QuantizationScheme(
            targets=[
                r"re:.*\.attn\.(wq_a|wq_b|wkv|wo_a|wo_b)$",
                r"re:.*mtp\.\d+\.(e_proj|h_proj)$",  # <— include MTP in scope
            ],
            format="float-quantized",
            **FP8_BLOCK,
        ),
        "experts": QuantizationScheme(
            targets=[r"re:.*\.ffn\.experts\.\d+\.(w1|w2|w3)$"],  # also matches mtp.0.ffn.experts.*
            format="nvfp4-pack-quantized",
            **NVFP4,
        ),
    },
    ignore=[
        "head", "embed",
        r"re:.*norm.*",
        r"re:.*\.ffn\.gate$",
        r"re:.*\.ffn\.gate\..*",
        r"re:.*\.ffn\.shared_experts\..*",
        r"re:.*\.hc_.*",
        r"re:hc_.*",
        r"re:.*\.attn\.attn_sink$",
        r"re:.*\.attn\.(compressor|indexer)\..*",
    ],
)
```

Launched as `python script.py --samples 64 --max-seq-len 512 --batch-size 1` (single-process; the multi-rank offload-cache deadlock from #2743 is orthogonal and not relevant here).

Sequence:
- Subgraphs 1-43 (main MoE blocks) calibrate and propagate cleanly. Per-subgraph wall ~60-120 s.
- Subgraph 44 (the MTP block) completes its calibration AND propagation passes — both tqdm bars hit 100%.
- `LifecycleCallbacks.sequential_epoch_end(subgraph)` then fires for subgraph 44 and crashes.

## Traceback (verbatim)

```
File ".../llmcompressor/pipelines/sequential/pipeline.py", line 163, in __call__
File ".../llmcompressor/core/session_functions.py", line 165, in sequential_epoch_end
    return cls.event(EventType.SEQUENTIAL_EPOCH_END, subgraph=subgraph, **kwargs)
File ".../llmcompressor/core/session_functions.py", line 91, in event
    return active_session().event(event_type, **kwargs)
File ".../llmcompressor/core/session.py", line 181, in event
    mod_data = self._lifecycle.event(...)
File ".../llmcompressor/core/lifecycle.py", line 204, in event
    data = mod.update_event(state=self.state, event=event, **kwargs)
File ".../llmcompressor/modifiers/modifier.py", line 122, in update_event
    self.on_event(state, event, **kwargs)
File ".../llmcompressor/modifiers/quantization/quantization/base.py", line 104, in on_event
    QuantizationMixin.sync_activation_observers(self, state.model)
File ".../llmcompressor/modifiers/quantization/quantization/mixin.py", line 288, in sync_activation_observers
    wait_for_comms(pending_comms)
File ".../compressed_tensors/distributed/utils.py", line 116, in wait_for_comms
    comm.wait()
RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed.
```

Note that `wait_for_comms` is being called with a **non-empty** `pending_comms` list even though our run had `Observer.synchronize` monkey-patched to return `[]` (see #2734 for that workaround's context). Something else queues a `dist.Work` that survives until `wait_for_comms`. Maintainer guidance on which observer/path that is would help isolate the root cause.

## Hypothesis on root cause

The MTP block's forward path in the vendored upstream model creates intermediate tensors under `torch.inference_mode()` — almost certainly because of the shared-embedding lookup (`mtp.0.embed.weight` aliases the main model's `embed.weight`, and the embedding `forward` is wrapped in `inference_mode` somewhere). When the quantization observer state propagates to a `weight_scale` / `weight_zero_point` writeback during `sequential_epoch_end`, the target parameter has been marked as an inference tensor, and the in-place update (the `comm.wait()` finalizing an `all_reduce`) trips the safety check.

This is consistent with how `inference_mode()` marks tensors and how the error message points at PyTorch RFC #17.

## Workaround (Option Y)

Both workstreams excluded MTP from the recipe entirely:

```python
ignore=[
    ...,
    r"re:.*mtp\..*",  # skip ALL MTP modules
]
```

The MTP weights ship BF16 (unquantized) in the artifact. They are still PRESENT (not dropped like stock HF transformers does via `_keys_to_ignore_on_load_unexpected`), and load cleanly in vLLM with `--speculative_config method=mtp`. For the artifacts I'm shipping, the differentiator is **presence**, not **precision**, so this is acceptable.

For users who actually need a quantized MTP block, this workaround doesn't apply.

## Asks

- **Documented:** "Speculative-decode heads (MTP-style) with shared-embedding lookups may need to be excluded from the quantization recipe via `ignore` until upstream resolves the inference-tensor interaction." A docs line + maybe a runtime warning when a recipe targets `*mtp*` / `*medusa*` / similar would catch users before they burn an hour of calibration.
- **Investigated:** trace which observer creates the lingering `dist.Work` that survives the no-op `synchronize` monkey-patch, and trace which write in the qparam-writeback path lands on an inference-mode tensor. The clone-before-update fix the error message suggests would live somewhere in `compressed_tensors.quantization.lifecycle.forward.update_offload_parameter` or its callee.
- **Tested:** a small integration test that runs `oneshot` against a 2-layer toy model with one speculative-decode-head-style block (shared embedding, e_proj/h_proj) would catch this on future branches.

## Diagnostic data available

Full log (`phase2b_1rank_mtp_inferencemode_crash_20260521T005306Z.log`), per-subgraph progress markers, and the in-script monkey-patches for `Observer.synchronize` + `modify_save_pretrained` + device-relocation + per-subgraph progress preserved on the B300 box at `/data/nvfp4-mtp/`. Happy to share or run additional probes if helpful.

cc @kylesayrs — this is adjacent to the `propagate_error` work (#2008) but a separate code path. Filing because the convergent independent reproducers from two workstreams suggested it's worth getting on the radar.

qubeena07:
## Summary

Fixes #2745

Transformer.forward in the DeepSeek V3.2 modeling file was decorated with torch.inference_mode(), carried over from the original inference only reference implementation this file was adapted from. Any tensor produced while that decorator is active stays tagged as an inference tensor even after the call returns. When a downstream module later reuses that tensor, for example a speculative decode style head that shares the base model embedding, and GPTQ or the offload cache try to write a compressed weight back in place, the write crashes with Inplace update to inference tensor outside InferenceMode is not allowed.

Switching to torch.no_grad() keeps the same gradient disabling behavior this forward pass needs during calibration, without ever producing inference tensors in the first place, so there is nothing left for GPTQ or the offload cache to trip over later.

This was tracked down together with brian dellabetta in the issue thread. He reproduced the crash on plain CPU with a single process, which ruled out distributed or NCCL involvement entirely and pointed straight at the decorator.

## Changes

- src/llmcompressor/modeling/deepseekv32/model.py: swap torch.inference_mode() for torch.no_grad() on Transformer.forward
- tests/llmcompressor/modeling/test_deepseekv32.py: guards against this exact regression by checking the forward source no longer references inference_mode
- tests/llmcompressor/modifiers/gptq/test_inference_mode_writeback.py: a small self contained toy model that reproduces the general pattern, a forward decorated for inference only whose output feeds a downstream module that GPTQ later needs to write a compressed weight into, and confirms the full calibration flow completes without crashing

## Test plan

- [x] Ran both new tests against the real package, they pass
- [x] Confirmed the underlying mechanism directly with plain PyTorch: a tensor created under torch.inference_mode() cannot be mutated in place afterward and raises the exact reported error, while a tensor created under torch.no_grad() can be mutated in place with no issue
- [x] Reproduced the original crash on CPU with a toy model that mirrors this same decorator pattern, then confirmed removing the decorator resolves it


qubeena07:
Hi @brian-dellabetta, this is code complete and I have tested it, both against the real package on a fresh install and through the two new test files, so it should be ready for review. I do not have permission to add the ready label myself, could you add it when you get a chance?

tests/llmcompressor/modeling/test_deepseekv32.py
brian-dellabetta, line 1: this test can be pruned

src/llmcompressor/modeling/deepseekv32/model.py
```diff
@@ -842,7 +842,7 @@ def __init__(self, args: ModelConfig):
         self.norm = RMSNorm(args.dim)
         self.register_buffer("freqs_cis", precompute_freqs_cis(args), persistent=False)
 
-    @torch.inference_mode()
+    @torch.no_grad()
```
kylesayrs: I feel that `no_grad` shouldn't be included at all in modeling code? This creates incompatibility with gradient-based algorithms like AutoRound.

The only exception to this rule is the computation of MoE router weights

brian-dellabetta: @yiliu30 does AutoRound require model defs to not have `@torch.no_grad` forward methods?

yiliu30: Hi @brian-dellabetta @kylesayrs, yes, AutoRound requires gradients, so the model’s forward methods should not be decorated with `@torch.no_grad()`.

I think a cleaner approach would be to manage gradient behavior at the algorithm level rather than the modeling level. This allows each algorithm to choose the behavior it needs. For example:

* AutoRound explicitly enables gradients:
  https://github.com/vllm-project/llm-compressor/blob/2d7a7ea058793447faa40b75d285c7ce2111c11f/src/llmcompressor/modifiers/autoround/base.py#L302-L303

* GPTQ explicitly disables gradients:
  https://github.com/vllm-project/llm-compressor/blob/2d7a7ea058793447faa40b75d285c7ce2111c11f/src/llmcompressor/modifiers/gptq/base.py#L327-L328

brian-dellabetta: Thanks @yiliu30 for the detail. Yeah, the `torch.enable_grad`/`torch.no_grad` contexts will override so this shouldn't be an issue, but @qubeena07 could you just remove the decorator entirely? I think it is better for readability, so users don't think that's a requirement

qubeena07:
Thanks @kylesayrs, that is a fair point and I agree, hardcoding no_grad into modeling code has the same underlying issue as inference_mode, just less severe. I removed the decorator entirely instead. During regular calibration this still avoids inference tensors, since calibration_forward_context already wraps the forward pass in no_grad, but it no longer takes gradients away from something like AutoRound that calls this model directly and actually needs them.

Also pruned the source inspection test per Brian's comment, the other test already covers the real behavior end to end so it was not adding much and was flagged as fragile anyway.

Rebased on main and pushed both changes, verified the updated test against the real package before pushing.

tests/llmcompressor/modifiers/gptq/test_inference_mode_writeback.py
brian-dellabetta, line 1: Hi @qubeena07 , now that we know how inference_mode works and breaks GPTQ, we probably don't need to add a per-commit test validating the behavior of non-inference mode models.

However, since we know inference_mode can cause issues well into a compression run, maybe we can just add `@torch.inference_mode(mode=False)` on one of the top-level GPTQ functions and/or one of the lower-level functions that is doing the in-place update, and override that? Do you think you could try this out and modify the test to confirm oneshot gptq works on a toy model with inference_mode on one of its submodules' forward method? Then we can keep that as a unit test here

brian-dellabetta (commented):
Thanks @qubeena07 for the update. I have one last request to include an early warning now that we know inference_mode can cause issues

qubeena07:
Done, and I tried it out end to end.

I went with an isinstance check on the closure instead of the string matching version, since it turned out torch's decorator context managers capture a bound method, something like self.clone, as the free variable rather than the context manager instance itself. So the check unwraps __self__ from whatever is in the closure cell and checks if that is an instance of torch.inference_mode, which should be more reliable than searching cell contents as a string.

Wired it into QuantizationMixin.start_calibration right next to the existing embeddings check, so it runs for any modifier using that mixin, not just GPTQ.

Rewrote the test to confirm this actually fires. It builds a toy model with a submodule forward decorated with torch.inference_mode, runs it through on_initialize and on_calibration_start the same way a real modifier would, and checks the warning shows up naming that submodule. Ran it against a fresh install from the branch and it passes, warning correctly names the decorated submodule.

src/llmcompressor/utils/transformers.py
```diff
@@ -94,3 +101,43 @@ def get_embeddings(
         output_embed = None
 
     return input_embed, output_embed
+
+
+def warn_inference_mode_forwards(model: torch.nn.Module):
```
brian-dellabetta, line 106: ah, sorry @qubeena07 ! I updated [my comment](https://github.com/vllm-project/llm-compressor/pull/2987/changes/360b325fb7706b9af933ea1aace6113ade6b4f3b..22b2ba487f4e424777f3539f5df8682a4d16cc47#r3715087086) below after discussing internally, that a more robust solution would be to wrap the calls that are in question with a `@torch.inference_mode(mode=False)` decorator to cancel out the `torch.inference_mode()` decorator on the module's forward method. Does that make sense? We can still keep your unit test to instead make sure oneshot runs end-to-end with the changes

qubeena07:
Before implementing this I tried it out directly with plain PyTorch, and it looks like mode=False does not actually fix this, in either place I could think to put it.

Wrapping the in place write itself in torch.inference_mode(mode=False) still crashes with the exact same error. What does work is wrapping that same write in torch.inference_mode(mode=True) instead, since the check is specifically that in place mutation on an inference tensor is disallowed outside an active InferenceMode, so being inside one, not outside, is what the tensor needs.

I also tried wrapping the call to the model itself in inference_mode(mode=False), hoping that would cancel out the forward method's own decorator before it takes effect. That does not work either, the inner decorator on forward always wins for its own execution regardless of what the caller is doing, so the output still comes back tagged as an inference tensor either way.

Small repro of both attempts:

```python
import torch

with torch.inference_mode():
    x = torch.zeros(4)

# mode=False around the write, still crashes
with torch.inference_mode(mode=False):
    x.copy_(torch.ones(4))
```

```python
class Foo(torch.nn.Module):
    @torch.inference_mode()
    def forward(self, x):
        return x + 1

f = Foo()
with torch.inference_mode(mode=False):
    out = f(torch.zeros(4))
print(out.is_inference())  # still True
```

Did you mean wrapping the write side specifically in mode=True rather than mode=False, or something else entirely, maybe stacking the decorator differently on the forward definition itself. Happy to implement whichever direction actually holds up, just did not want to push something based on a name that turned out not to do what it sounds like it does.

brian-dellabetta (approved):
Hi @qubeena07 , thanks for the detailed report and for helping us work through this. Given your snippets above, I think we are just stuck with inference_mode being finicky. Your warning and unit test look good to me. I vote we get this in and see if this arises again in future

kylesayrs (commented):
LGTM, I'll note that `model.named_modules()` can take a noticeable amount of time for large models, so something that we should consider speeding up in the future, maybe using something like

```
def calibration_context():
    ...
    patch_attr(torch, “inference_mode”, noop)
```

kylesayrs (approved):

PR #2990, opened 2026-07-31
[Bugfix] Do not recurse into container modules in observe/update_qparams

Ar4ikov:
SUMMARY:

`observe()` and `update_qparams()` in `modifiers/quantization/calibration.py` decide whether their `module` argument is one module or a collection of modules with `isinstance(module, collections.abc.Iterable)`. Container modules are both: `nn.ModuleDict` is a `Module` *and* is `Iterable`, and iterating it yields its keys, which are strings. From there `observe("weight", ...)` iterates the string, `observe("w", ...)` iterates `"w"` back into `"w"`, and the recursion never terminates:

```
  File ".../llmcompressor/modifiers/quantization/calibration.py", line 99, in observe
    observe(m, base_name)
  [Previous line repeated 992 more times]
  File ".../llmcompressor/modifiers/quantization/calibration.py", line 97, in observe
    if isinstance(module, Iterable):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen abc>", line 119, in __instancecheck__
RecursionError: maximum recursion depth exceeded in comparison
```

Any model using `torch.nn.utils.parametrize` has such a `ModuleDict`, because torch stores parametrizations at `<module>.parametrizations`. The wav2vec2 family all do: wav2vec2, HuBERT, WavLM and UniSpeech-SAT weight-norm `encoder.pos_conv_embed.conv`.

Reproducer on `main` (28c9c76b), no downloads:

```python
import torch
from llmcompressor.modifiers.quantization.calibration import observe

model = torch.nn.Sequential(
    torch.nn.utils.parametrizations.weight_norm(torch.nn.Conv1d(4, 4, 1)),
    torch.nn.Linear(4, 4),
)
observe(list(model.modules()), "weight")  # RecursionError
```

**Scope, honestly.** On `main` neither in-tree caller trips this. `QuantizationModifier.on_sequential_epoch_end` and `GPTQModifier.on_sequential_epoch_end` both filter with `is_module_quantized` first, and a `parametrizations` `ModuleDict` has no `quantization_scheme`, so it never reaches `observe`. That filter arrived with #2784. On 0.12.0.x, where `on_event` passed `get_modules(parents)` straight through, it is a hard crash: `oneshot` on `microsoft/wavlm-large` with `QuantizationModifier(scheme="W8A16", targets=["Linear"], ignore=["classifier"])` dies with the traceback above (verified on 0.12.0.1). So on `main` this is a latent defect rather than a live crash — but the helpers are public and `QuantizationMixin` tells modifier authors to call them directly ("Modifier must explicitly call `observe(modules, base_name="weight")` then `update_qparams(modules, base_name="weight")`"), and the failure mode is a blown stack rather than an error message. The guard that hides it today lives in the callers, not in the helpers.

**Fix.** Check `Module` before `Iterable` so a container module is treated as a single module. `update_qparams` has the same dispatch and the same defect, so it gets the same treatment.

Two things worth calling out:

- This is a behaviour change for one case: passing an `nn.ModuleList`/`nn.Sequential` of quantized modules used to iterate it and now treats it as a single module, which is a no-op since containers carry no observers. No in-tree caller does that — callers pass a `list`, `dict_values`, or a single module — and the pipelines flatten with `.modules()`, so a container's children already arrive in the same list as separate entries.
- I fixed the helpers rather than filtering containers where the module lists are built (`Subgraph.submodules`, `list(model.modules())` in the basic/data-free pipelines), because there are several producers and only two consumers.

I did not add a `not isinstance(module, str)` guard: with the `Module` check in place nothing in-tree can produce a string here. Happy to add it if you would rather have the recursion hard-stopped regardless of caller.

Found while quantizing speech emotion classifiers (WavLM/HuBERT encoders with a classification head) to INT8/FP8/INT4. Our workaround on the released version is to fold the weight-norm parametrization away before `oneshot` and reapply it before saving.

TEST PLAN:

New test `tests/llmcompressor/modifiers/calibration/test_observers.py::test_observe_skips_container_modules`, which builds a weight-normed `Conv1d` next to a quantized `Linear` and runs `observe` + `update_qparams` over `list(model.modules())`, as the pipelines do. It is a real regression test on `main`:

```
$ git stash push -- src/llmcompressor/modifiers/quantization/calibration.py
$ pytest tests/llmcompressor/modifiers/calibration/test_observers.py::test_observe_skips_container_modules -q
E   RecursionError: maximum recursion depth exceeded in comparison
!!! Recursion detected (same locals & position)
1 failed

$ git stash pop
$ pytest tests/llmcompressor/modifiers/calibration/ tests/llmcompressor/observers/ \
    tests/llmcompressor/modifiers/quantization/test_sequential_observation.py -q
153 passed, 4 skipped
```

`make quality` is clean.

`make test` on my box gives 503 passed / 21 failed, none of which I can attribute to this change. I only have 2x RTX 3090 with about 11 GB free per card, and all but one failure is `torch.OutOfMemoryError` or an assertion downstream of one (`test_linearize_moe[*]`, `test_linearize_offload`, `test_handling_shared_embeddings`, `test_quantization_ddp`, `test_cache`). Re-running the same selection on a clean tree fails *more* (28 vs 20 of the same 76 tests), and the two failure sets differ in both directions, which is what OOM contention looks like. The one non-OOM failure, `test_model_free_ptq_matches_oneshot[Qwen/Qwen3-0.6B-NVFP4A16]`, fails identically on a clean tree (`AssertionError: ('model.layers.0.mlp.down_proj.weight_scale', False, True)`) and so is pre-existing here, not caused by this change. Nothing in either set touches `observe`/`update_qparams`. These need to be judged on your H100 runners.

Environment: Python 3.11.15, torch 2.13.0+cu130, transformers 5.14.1, compressed-tensors 0.17.2a. Originally hit on llmcompressor 0.12.0.1 with torch 2.12.0+cu130, transformers 5.10.1, compressed-tensors 0.17.1.


src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -83,6 +83,18 @@ def initialize_observer(
...
+def _is_module_collection(value: Any) -> bool:
+    """
+    Whether a value is a collection of modules rather than a single module.
+
+    Container modules such as `ModuleDict` are both `Module` and `Iterable`, so the
+    module check has to come first: iterating a `ModuleDict` yields its keys, which
+    are strings, and iterating a one character string yields itself, so recursing
+    into a container module never terminates.
+    """
+    return isinstance(value, Iterable) and not isinstance(value, Module)
```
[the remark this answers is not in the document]
Ar4ikov: Taken for `str`, pushed in f500c774. I had flagged this one in the PR description and you are right that leaving it out was the weaker choice.

Left `bytes` out, though, because it cannot recurse: iterating `b"weight"` yields `int`s, and `int` is not `Iterable`, so the next call falls straight through. Verified on both variants:

| call | `Module` only | `(Module, str)` |
|---|---|---|
| `observe(list(model.modules()), "weight")` | ok | ok |
| `observe(model.conv.parametrizations, "weight")` | ok | ok |
| `observe("weight", "weight")` | RecursionError | ok |
| `observe("weight", linear)` (args swapped) | RecursionError | ok |
| `update_qparams("weight", linear)` (args swapped) | `TypeError: 'Linear' object is not iterable` | same |
| `observe(b"weight", "weight")` | ok | ok |

Happy to add `bytes` anyway if you prefer the familiar `not isinstance(x, (str, bytes))` idiom — it costs nothing, it just does not buy anything here.

One tradeoff worth naming, for whoever reviews this next: the guard turns the swapped-argument case into a silent no-op instead of a loud `RecursionError`. That is consistent with how `observe` already returns silently for a module with no observer attached, so I went with it, but if you would rather it fail loudly I can `raise TypeError` on a non-`Module` non-collection instead. That is a one line change either way.

brian-dellabetta:
Hi @Ar4ikov , can you explain how you came across this? I am trying to understand if this is something you hit during a normal compression flow for one of these wave2vec models or if this usage of `torch.nn.utils.parametrizations.weight_norm(torch.nn.Conv1d(4, 4, 1))` is not expected. For example, this runs fine

```python
import torch
from llmcompressor.modifiers.quantization.calibration import observe
 
model = torch.nn.Sequential(
    torch.nn.Linear(4, 4),
    torch.nn.Linear(4, 4),
)
observe(list(model.modules()), "weight")
```

Ar4ikov:
Hi @brian-dellabetta. A normal flow, not a synthetic case.

 I hit it quantizing speech emotion classifiers (WavLM/HuBERT/wav2vec2/UniSpeech-SAT encoders + a classification head) to INT8/FP8/INT4 with `llmcompressor==0.12.0.1`. Every wav2vec2-family model on the Hub weight-norms `encoder.pos_conv_embed.conv`, so `torch` keeps a `parametrizations` ModuleDict on it. On 0.12.0.x `QuantizationModifier.on_event` passed `get_modules(parents)` straight into `observe`, that ModuleDict reached it, and `oneshot` died with the traceback in the description. Public reproducer, verified on 0.12.0.1:

 ```python
 from transformers import AutoModelForAudioClassification

 model = AutoModelForAudioClassification.from_pretrained("microsoft/wavlm-large")
 oneshot(model=model, recipe=[QuantizationModifier(
     scheme="W8A16", targets=["Linear"], ignore=["classifier"])])
 ```

 The `weight_norm(Conv1d(4, 4, 1))` in the test is just that reduced to something that runs in milliseconds — it's the same shape torch produces for the real model, not an unusual API use.

 Your example passes because it has no parametrizations at all, so no ModuleDict ever enters `observe`. Add weight norm and it recurses on `main` too:

 ```python
 import torch
 from llmcompressor.modifiers.quantization.calibration import observe

 model = torch.nn.Sequential(
     torch.nn.utils.parametrizations.weight_norm(torch.nn.Conv1d(4, 4, 1)),
     torch.nn.Linear(4, 4),
 )
 observe(list(model.modules()), "weight")  # RecursionError
 ```

 To restate the scope from the description: on `main` no in-tree caller trips this, because `on_sequential_epoch_end` filters with `is_module_quantized` since #2784 and a `parametrizations` ModuleDict has no `quantization_scheme`. So `main` has a latent defect in two public helpers rather than a live crash — the guard lives in the callers, `QuantizationMixin`'s docstring tells modifier authors to call `observe`/`update_qparams` directly, and the failure mode is a blown stack. Happy to close this if you'd rather keep the invariant in the callers; I'd just note the released version does hard-crash on this whole model family.

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -94,7 +107,7 @@ def observe(
     :param module: module or iterable of modules with observer attributes
     :param base_name: substring used to fetch the observer and value to observe
     """
-    if isinstance(module, Iterable):
+    if _is_module_collection(module):
```
brian-dellabetta, line 110: could this instead be?
```suggestion
    if isinstance(module, Iterable) and not isinstance(module, torch.nn.ModuleDict):
```
rather than guarding against the failure modes, like a string being passed into `def observe`, wondering if we can be more explicit at the top level to know the edge cases we need to guard against

brian-dellabetta (commented):
Thanks @Ar4ikov for the extra detail. If this is specifically related to a torch.nn.ModuleDict being passed into observe, I am wondering if we could be more explicit with the check. Please see comment below

src/llmcompressor/modifiers/quantization/calibration.py
kylesayrs: @brian-dellabetta This doesn't cover other iterable module types like ModuleLists

kylesayrs (approved):
This seems fine and future-proofing useful. Need review from @HDCharles

src/llmcompressor/modifiers/quantization/calibration.py
brian-dellabetta: ModuleLists should work fine, they are iterable so the children will all be observed. ModuleDict is the true culprit, because the keys are strings. Guaranteeing strings aren't passed into `observe` seems a more suitable safeguard to me than checking for strings in observe.

Will ask @HDCharles for his thoughts since this is touching code he added

kylesayrs (approved):

PR #2997, opened 2026-08-03
Avoid CUDA indexing failures in iMatrix grid search updates

yasu-oh:
## Summary

Replace boolean-indexed in-place updates in the iMatrix grid search with shape-validated torch.where selections.

## The change:

- validates that all update tensors and the boolean mask match best_error.shape
- avoids the CUDA advanced-indexing path used by the masked assignments
- preserves the existing no-improvement and patience behavior

## Problem

The current update path can fail during CUDA quantization runs:

python
best_error[improved] = err[improved]
best_min[improved] = shrink_min[improved]
best_max[improved] = shrink_max[improved]

The following failures were observed with GPTQ, NVFP4, and the imatrix_mse observer:

- Gemma 4 E2B: inconsistent selection sizes when the same boolean mask was used on both sides of an assignment
- Gemma 4 E4B: CUDA IndexKernel.cu out-of-bounds device assertions

The exact lower-level cause has not been confirmed, so this change does not assume a specific PyTorch defect. The added shape validation did not trigger during successful patched GPU runs.

## Validation

The development environment used to prepare this change is CPU-only.

CPU validation:

- pytest -q tests/llmcompressor/observers/test_imatrix.py
  - 34 passed
- ruff check
  - passed
- ruff format --check
  - passed
- git diff --check
  - passed
- added a CPU regression test for explicit grid-search shape validation

GPU validation was performed separately using GPTQModifier with NVFP4 imatrix_mse, 1,024 calibration samples, max_seq_length=8192, FP8 static KV-cache calibration, and the basic calibration pipeline for Gemma 4 E models:

- Gemma 4 E2B failed repeatedly before the change and completed on the first run after the change
- Gemma 4 E4B hit CUDA IndexKernel.cu assertions before the change and completed on the first run after the change
- the shape validation did not trigger in either successful run
- quantization and compressed-model saving completed successfully

The same patched environment was used to produce the Gemma 4 QAT NVFP4 collection covering 31B, 26B-A4B, 12B, E4B, and E2B:

https://huggingface.co/collections/yasu-oh/gemma-4-qat-nvfp4

## Trade-off

torch.where materializes full output tensors, so it may use slightly more temporary memory and memory bandwidth than masked in-place assignment.


brian-dellabetta (commented):
Hi @yasu-oh , the changes look fine to me, but could you provide a minimal reproducible example script that triggers this "CUDA IndexKernel.cu out-of-bounds device assertion" error you are reporting? I've not seen it before

yasu-oh:
Thanks. I was able to reproduce the E4B failure again, although it appears to be intermittent rather than deterministic.

On the same system and configuration, it reproduced once in six fresh-process runs. In the failed run, the CUDA IndexKernel.cu out-of-bounds assertion was followed by a traceback that surfaced at:

    best_max[improved] = shrink_max[improved]

in llmcompressor/observers/imatrix.py, ending with a CUDA device-side assert.

Because CUDA operations may be asynchronous, I am not yet claiming that this exact assignment is definitively the original source of the failure. I am currently preparing a standalone reproduction script and collecting the relevant logs, including runs with CUDA_LAUNCH_BLOCKING=1, and will share them once they are ready.

yasu-oh:
A quick update on the E4B case: although I observed the IndexKernel assertion once, I have not been able to reproduce it again after more than 20 additional fresh-process runs under the same conditions.

Given the lack of reliable reproducibility, please disregard the E4B-specific report for now. I do not want it to distract from the consistently reproducible E2B failure. I will revisit the E4B case separately if I can reproduce it reliably and collect stronger evidence.

brian-dellabetta:
Hi @yasu-oh , so you're seeing it pretty frequently with E2B? do you have a minimal reproduction i can try on my side?

yasu-oh:
A further update: I can no longer reproduce the E2B failure either with my current environment:

- compressed-tensors 0.18.1a20260806
- llmcompressor 0.12.1.dev101+gded34c1f3

I also noticed that compressed-tensors recently disabled the Triton `_quantize` backend in vllm-project/compressed-tensors#817. Since the llm-compressor iMatrix grid-search code itself has not changed between the version where I originally observed the failure and my current environment, one possibility is that the CUDA error originated from the quantization backend and only surfaced later at the boolean-indexed assignment due to asynchronous CUDA execution.

Given that I can no longer reliably reproduce either the E2B or E4B issue, I don't want to overstate the evidence that the masked assignments themselves were the root cause.

That said, I still think the `torch.where` change may have some value as defensive hardening: it avoids the advanced-indexing update path while preserving the intended element-wise update semantics, and the added shape check provides a clearer failure mode if an unexpected tensor shape ever reaches the grid search.

I appreciate you taking the time to review and investigate this. If you think this defensive change is still worthwhile, I would be happy to keep the PR open; if you would prefer not to carry it without a currently reproducible failure, I completely understand and am also happy to close it.

brian-dellabetta:
Hi @yasu-oh , thanks for the report. We disabled the triton kernels and changed some of the logic away from torch.where, to fix a regression we were hitting in nvfp4 regression tests. We will re-enable after the releases, can we revisit at that time? It'd be good to see an MRE so we can understand root cause before merging

Issue #2981, opened 2026-07-29
[Performance] Speed up subgraph tracing for large models


kylesayrs:
## Background ##
LLM Compressor uses [Sequential Onloading](https://docs.vllm.ai/projects/llm-compressor/en/latest/guides/big_models_and_distributed/sequential_onloading/#implementation) to avoid loading the entire model into GPU memory. This relies on [`trace_subgraphs`](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/helpers.py#L84) which performs a virtual pass through the model to create a graph which is later partitioned into subgraphs

This process can take upwards of 10min for very large models like `moonshotai/Kimi-K3`

## Proposed Changes ##
Because of the complexity of the tracing and partitioning logic within LLM Compressor, this issue only focuses on small diff + high impact changes which can reduce this runtime.

A few places I've noticed already are:
1. Changing [`partitions`](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/helpers.py#L339C36-L339C46) into a set, not a list for O(1) lookup
2. Changing [this code](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/transformers_helpers.py#L1356-L1368) to cache an inverse dict of params and buffers, rather than iterating through params and buffers each time, leading to O(N^2) runtime for parameter proxy making

**PRs which fundamentally refactor subgraph/ tracing logic will not be accepted**. Only well thought-out improvements from DSA principles will be accepted (thank you!).

Please include a benchmark script on a dummy model, see [`test_models`](https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/transformers/tracing/test_models.py) for regression testing.

Ping @kylesayrs for any follow up questions

wanadzhar913:
Hi @kylesayrs, would love to help with this. Can I take No. 1?

For benchmarking scripts, what metrics would you be keen to look at? I presume Sequential Onloading wall time is the most important metric?

kylesayrs:
@wanadzhar913 Sounds good! I'll assign this ticket to you, while leaving 2 (and extras) open.

W.r.t. benchmarking, the goal is to see runtime improvements on a trace_subgraphs task with around the magnitude of 100K matched sequential targets.

This is relevant to the Kimi-K3 model, but you can create a dummy model to trace and put the benchmark results in the benchmarks folder.

[...]

PR #2999, opened 2026-08-04
[Performance] Speed up subgraph tracing - add direct dict lookup for O(1) node membership checks

wanadzhar913:
# SUMMARY

- Replace linear searches through partition node lists with average O(1) set membership checks and direct dictionary lookups for node partition indices.
- Preserve ordered partition lists for deterministic, topologically valid FX graph construction.
- Avoid temporary list allocation when calculating node indegrees.
- Add a configurable synthetic benchmark supporting approximately 100K matched sequential targets, including a shared `get_attr` node with many consumers to exercise delayed partition placement.

# ISSUE
- Related to item 1 in #2981

# BENCHMARKS

The benchmark was run on a WSL2 with Python 3.13.4 and PyTorch 2.12.1 on CPU.

The workload contains 100,000 matched targets grouped into 335 subgraphs with up to 300 targets each. The synthetic model reads a shared model buffer each layer, producing one shared `get_attr` node with many consumers. Grouping targets creates sufficiently large partitions to exercise the membership-check optimization without constructing 100,001 single-target subgraphs.

Only `trace_subgraphs` was timed. Results are medians of three runs.

| Revision | Run 1 (s) | Run 2 (s) | Run 3 (s) | Median (s) |
| --- | ---: | ---: | ---: | ---: |
| Baseline (`2d7a7ea0`) | 587.682 | 574.621 | 543.853 | 574.621 |
| This PR | 37.253 | 34.790 | 33.020 | 34.790 |

This provides a **16.5x speedup** on the benchmark workload. The baseline regresses heavily on the `get_attr` case because it repeatedly scans partition lists to locate each consumer; this PR avoids that with `node_to_partition` lookups.

The baseline was measured by exporting commit `2d7a7ea058793447faa40b75d285c7ce2111c11f` to an isolated directory and pointing `PYTHONPATH` at its `src/` tree. The complete workload and reproduction instructions are documented in `benchmarks/trace_subgraphs/README.md`.

# TEST PLAN

```bash
python3 -m pytest -q tests/llmcompressor/pipelines/sequential/
CUDA_VISIBLE_DEVICES="" python benchmarks/trace_subgraphs/benchmark.py --smoke
```

# ADDITIONAL

Happy to standardize on benchmarks by https://github.com/vllm-project/llm-compressor/pull/2992 since those came first.

EDIT:
Closes https://github.com/vllm-project/llm-compressor/pull/2999


src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -272,16 +272,17 @@ def topological_partition(
         )
 
     partitions: list[list[Node]] = [[]]
+    partition_sets: list[set[Node]] = [set()]
```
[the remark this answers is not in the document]
kylesayrs: This is not a bad suggestion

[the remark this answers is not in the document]
wanadzhar913: Hi Kyle! Yeahp, let me undraft this the soonest. 🙏🏻

kylesayrs:
@coderabbit review

benchmarks/trace_subgraphs/benchmark.py
```diff
@@ -0,0 +1,257 @@
...
+        "--smoke",
+        action="store_true",
+        help=f"Run a small smoke benchmark with {SMOKE_NUM_TARGETS} targets",
+    )
+    parser.add_argument(
+        "--validate-graphs",
+        action="store_true",
+        help="Run graph_is_well_formed on every subgraph (slow at large scale)",
+    )
+    return parser.parse_args()
```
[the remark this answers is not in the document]
kylesayrs: Not worried about this

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -352,6 +353,8 @@ def partition_graph(model: Module, partitions: list[list[Node]]) -> list[Subgrap
 
     # create subgraphs
     for partition_nodes in partitions:
+        partition_set = set(partition_nodes)
```
kylesayrs, line 356: Would it be better to have `partitions` just be a `partitions: list[set[Node]]` from the beginning?

kylesayrs (approved):
Looks great, thanks for the PR! Awesome jobs measuring speedups. Because https://github.com/vllm-project/llm-compressor/pull/2992 was submitted first, I'll merge the other and then merge this after

src/llmcompressor/pipelines/sequential/helpers.py
wanadzhar913: Oh, I wanted the partitions (from `topological_partition`) to stay topologically ordered, so I maintained the list data structure. `partition_graph` relies on that ordering, since `node_copy` resolves each node's in-partition inputs through `node_map`, which requires producers to be copied first. A set data structure wouldn't guarantee this.

`partition_set = set(partition_nodes)`'s mainly there for `O(1)` membership checks.

kylesayrs: > Oh, I wanted the partitions (from topological_partition) to stay topologically ordered

What I suggested would keep partitions ordered, but the nodes within partitions would become unordered.

Good either way!

kylesayrs (approved):
Looks great, nice job!

benchmarks/trace_subgraphs/benchmark.py
brian-dellabetta, line 1: isn't this a duplicate of what was merged in #2992 (benchmarks/bench_trace_subgraphs.py)?

HDCharles (approved):
Looks good but need to address Brian's comment with rebase or reaponse

benchmarks/trace_subgraphs/benchmark.py
wanadzhar913: Hi @brian-dellabetta, you're right, there's duplicate work here. I was thinking of a way to make this cleaner. Please let me know if this approach is okay:
- delete `benchmarks/trace_subgraphs/*`
- add one flag in `benchmarks/bench_trace_subgraphs.py` that switches the dummy model's workload e.g. `--num-shared-buffer-consumers N (defaut is 0 as of today's behaviour).
- add the new bits in my script into `benchmarks/bench_trace_subgraphs.py` e.g., Resident Set Size reporting, an assertion on the `get_attr` node count, etc.
- rerun benchmarks relative to a newer baseline commit and add them to `benchmarks/README.md` (as well as PR description). We should see a reduced speedup gain (instead of16.5x) now that #2992 is merged.

Let me know if this works!

PR #3004, opened 2026-08-07
[Tests] Nightly lmeval tests

kylesayrs:
## Purpose ##
* More frequent lm eval tests to catch failures earlier

## Co-requisites ##
* https://github.com/neuralmagic/llm-compressor-testing/pull/331

## Changes ##
* Set cadence of all tests to nightly
* Add fp8 block test

## Testing ##
* https://github.com/neuralmagic/llm-compressor-testing/actions/runs/31545546132
* https://github.com/neuralmagic/llm-compressor-testing/actions/runs/32224295402

configs (h100-solo): 1h 35m 18s
vl_configs (h100-solo): 47m 21s
distributed_configs (h100-duo): 29m 15s


tests/lmeval/nightly_configs/fp8_block.yaml
```diff
@@ -0,0 +1,7 @@
+cadence: "nightly"
+model: meta-llama/Meta-Llama-3-8B-Instruct
+scheme: FP8_BLOCK
+lmeval:
```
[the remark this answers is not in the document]
kylesayrs: This recipe is data-free

dsikka (commented):
FYI, based on the benchmarking we’ve done + conversations with Michael, fp8 static per tensor may have better performance than dynamic per token and should maybe be a path we focus on more

dsikka (commented):
I would make two changes:

1. Just remove the disabled one 
2. Keep the configs in the same `configs` folder - just add both cadences, weekly and nightly and update the cadence handling logic to consider multiple options so that we can be sure that the tests run both during a nightly run or a weekly run

dsikka (approved):

dsikka (approved):

tests/e2e/test_vllm.py
```diff
@@ -79,7 +79,8 @@ class TestvLLM:
     def set_up(self, test_data_file: str):
         eval_config = yaml.safe_load(Path(test_data_file).read_text(encoding="utf-8"))
 
-        if os.environ.get("CADENCE", "commit") != eval_config.get("cadence"):
+        cadence = os.environ.get("CADENCE", "commit")
```
dhuangnm, line 82: It's not correct to add this logic here and tests will be skipped when we run RELEASE job. We should update the workflows instead and run the correct cadence (nightly, weekly) for e2e and lm eval tests when we run RELEASE job. Please feel free to schedule a time for discussion. Thanks.

dhuangnm (changes requested):

tests/e2e/test_vllm.py
dhuangnm: I'm sorry that I misunderstood the code. After a second look, I think it'll actually work here to run all tests when cadence is set to release.

dhuangnm (approved):
Looks good, thanks @kylesayrs!

PR #3006, opened 2026-08-07
[not for land] Trigger test: src/ change

rashmigottipati:
**Test PR to verify if_changed triggers**

This PR modifies a file in `src/` which should **TRIGGER** transformers tests.

Part of verifying the if_changed include/exclude pattern in https://github.com/vllm-project/llm-compressor/pull/2983


brian-dellabetta:
change in src/ is triggering transformers tests to run here, so things are working


Screenshot 2026-08-18 at 11 51 33 AM

PR #3012, opened 2026-08-09
Explain calibration levers in the sequential OOM message

### LINKED ISSUE #3011: Untruncated text calibration produces a misleading 16 GiB CUDA OOM in mask expansion; warn or guard when max_seq_length is unset

- llmcompressor 0.12.1.dev100+g170a2e19 (traceback below); the same 16.00 GiB OOM also killed earlier runs on released 0.12.0.1 (both `pipeline="basic"` and CPU-offload variants)
- transformers 5.10.1, torch 2.11.0+cu130
- 1x RTX 5090 32 GB (sm_120), driver 580.173.02, Ubuntu 24.04
- Dataset: `HuggingFaceH4/ultrachat_200k` train_sft, 128 samples, rendered through the model chat template. Longest sample 3,994 tokens; only 10 of 128 exceed 2,048.

### 🐛 Describe the bug

Calling `oneshot()` with a text dataset and no `max_seq_length` reliably OOMs a 32 GB card during calibration of Qwen3-8B, with the allocation landing in transformers' attention-mask expansion:

```
File ".../transformers/masking_utils.py", line 52, in and_mask
    result = result & mask(batch_idx, head_idx, q_idx, kv_idx).to(result.device)
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 16.00 GiB.
```

Setting `max_seq_length=2048` and changing nothing else makes the identical run complete with headroom. We have not pinned down which broadcast produces the 16.00 GiB request (calibration `batch_size` defaults to 1, so it is not simple cross-sample padding); full logs available on request.

What makes this worth an issue is how misleading the failure is. The error points at GPU capacity, not sample length, so the natural responses are `pipeline="basic"`, CPU offload via device_map, or concluding the model does not fit the card. None of them help, and each failed attempt costs a full calibration cycle. It took us several such cycles to find the actual cause.

Related but distinct prior work: #2649 fixed truncation for pre-tokenized datasets, and #2917 fixed truncation state leaking into the saved tokenizer. This report is about the untruncated-text path and the quality of the failure when it bites.

Proposal, any of which would have saved the debugging time:

1. Log a prominent warning when tokenizing calibration text with no `max_seq_length` set, including the longest sample length found.
2. Or apply a documented default truncation for text datasets (matching what most recipes in `examples/` opt into).
3. Or catch OOM during calibration and re-raise with a hint naming `max_seq_length` when long samples are present.

Happy to submit a small PR for option 1, or whichever direction maintainers prefer.

### 🛠️ Steps to reproduce

```python

MODEL = "Qwen/Qwen3-8B"
tok = AutoTokenizer.from_pretrained(MODEL)
ds = load_dataset("HuggingFaceH4/ultrachat_200k", split="train_sft") \
    .shuffle(seed=3407).select(range(128))
ds = ds.map(lambda ex: {"text": tok.apply_chat_template(ex["messages"], tokenize=False)},
            remove_columns=ds.column_names)
recipe = QuantizationModifier(targets="Linear", scheme="NVFP4", ignore=["lm_head"])
oneshot(model=MODEL, dataset=ds, recipe=recipe,
        num_calibration_samples=128, output_dir="out-nvfp4")   # no max_seq_length -> OOM on 32 GB
```

Add `max_seq_length=2048` to the `oneshot()` call and it completes.

Hit while producing the quantization benchmark at https://github.com/Rodder5/sm120-quant-bench.

rishabhsinha17:
SUMMARY:
Resolves #3011

When `oneshot()` runs with a text dataset and no `max_seq_length`, tokenization uses `max_length=tokenizer.model_max_length`, so long samples reach calibration untruncated. The resulting CUDA OOM points at GPU capacity, not sequence length, sending users toward `pipeline="basic"` or CPU offload, which do not help (#3011).

Per review, this is now a message-only change: the sequential pipeline's OOM re-raise explains the two calibration levers inline. `max_seq_length`: when unset, calibration samples are not truncated, so a few long samples can dominate memory. `num_calibration_samples`: scoped to modifiers that cache activations (such as AWQ), where memory scales with sample count. No dataset inspection, no behavior change outside the message text.

Carries over #3020's direction (closed in favor of this PR) with co-author credit.

TEST PLAN:
One unit test asserts the message names both levers through `handle_sequential_oom`. `pytest tests/llmcompressor/pipelines/sequential/test_helpers.py tests/llmcompressor/datasets/` passes (22 tests). `ruff check` and `ruff format --check` clean on the changed files with the repo-pinned ruff.


src/llmcompressor/datasets/utils.py
```diff
@@ -179,6 +181,51 @@ def format_calibration_data(
...
+
+    column_names = dataset.column_names or []
+    if "input_ids" in column_names:
+        feature_name = "input_ids"
+    elif "decoder_input_ids" in column_names:
+        feature_name = "decoder_input_ids"
+    else:
+        return
+
+    lengths = [len(sample) for sample in dataset[feature_name]]
```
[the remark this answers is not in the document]
rishabhsinha17: Done in the amended commit: lengths now come from pyarrow list_value_length on the arrow column, with the indices mapping applied for shuffled/selected datasets and the list comprehension kept as fallback. Verified the arrow path returns identical lengths to the fallback on an indexed dataset.

rishabhsinha17:
@brian-dellabetta #3020 looks like the right fix for the failure path, and I agree the error site beats a warning for anyone who never hits this. Flagging where this PR sits relative to it so you can take one, both, or neither.

They fire at different points and in different places:

- #3020 edits the re-raise in `handle_sequential_oom` (`pipelines/sequential/helpers.py`), which is applied at `pipelines/sequential/pipeline.py:54`. It tells you which levers exist once the allocation has already failed.
- This PR warns in `format_calibration_data` (`datasets/utils.py`), the path `get_calibration_dataloader` goes through to build the loader, so it runs before calibration starts and independently of which pipeline follows.

The part that is not duplicated: the warning reports the measured longest sample and how many exceed the threshold, so the user sees which lever applies to their data and by how much, rather than a list of levers to try. It also covers pre-tokenized datasets, which skip the tokenization path entirely.

If you would rather carry a single mechanism, say so and I will close this. If both are wanted I am happy to rebase on #3020 once it lands and trim anything the error message makes redundant.

brian-dellabetta:
Hi @rishabhsinha17 , yes the error is favorable to the warning. logs are inevitably noisy with warnings from transformers and ourselves, and the error will display it at the proper location in logs. if you'd like to update your PR to look more like mine -- raising an error, fixing the merge conflict, and displaying max sequence length in error message -- we can close mine in favor of yours

rishabhsinha17:
@brian-dellabetta Done, reworked as you described. `format_calibration_data` now raises a ValueError when `max_seq_length` is unset and the tokenized dataset has samples over 2048 tokens. The message keeps the measured data (how many samples exceed the threshold, longest length) and shows `max_seq_length` both ways: set it to truncate, or set it to at least the longest length to keep full-length samples deliberately. Also carried your #3020 message addition into the sequential OOM re-raise, with co-author credit, so closing #3020 loses nothing. Branch is rebased on latest main and conflict-free; tests updated to assert the error.

rishabhsinha17:
@brian-dellabetta the rework you outlined is in as of Aug 14: ValueError instead of warning, `max_seq_length` shown both ways in the message, your #3020 change carried over with co-author credit, and the branch is rebased onto the transformers 5.15 pin. Could you add the ready label so the testing suite runs, and take a look when you get a chance?

src/llmcompressor/datasets/utils.py
```diff
@@ -159,6 +160,7 @@ def format_calibration_data(
     tokenized_dataset: Dataset,
     processor: Processor,
 ) -> DataLoader:
+    _check_untruncated_sequences(args, tokenized_dataset)
```
brian-dellabetta: This triggers a separate validation check, we only want to raise if an OOM error is hit

brian-dellabetta (changes requested):
See https://github.com/vllm-project/llm-compressor/pull/3020/changes#diff-4ce075d0f430bf77f0b7aa31d0c1e2a1b5f032e064062b324d19ba1ce88949b4R484 for where the error messaging should be included

src/llmcompressor/datasets/utils.py
```diff
@@ -23,6 +23,7 @@
 from llmcompressor.typing import Processor
 
 BS_WARNING_THRESHOLD = 16
+SEQ_LEN_ERROR_THRESHOLD = 2048
```
brian-dellabetta: we don't want to make these assumptions

tests/llmcompressor/datasets/test_max_seq_length.py
```diff
@@ -43,3 +59,58 @@ def test_uniform_lengths_with_max_seq_length(self):
         result = collator(_make_features([10, 10]))
         assert result["input_ids"].shape[1] == 5
         assert result["attention_mask"].shape[1] == 5
+
+
+class TestUntruncatedSequenceError:
```
brian-dellabetta: these tests can be pruned

src/llmcompressor/datasets/utils.py
rishabhsinha17: Removed in 7a2d7f74. format_calibration_data no longer checks or raises, so the success path is unchanged. Lengths are measured only inside handle_sequential_oom after torch.OutOfMemoryError is caught, and only appended to the re-raised message.

src/llmcompressor/datasets/utils.py
rishabhsinha17: The threshold no longer gates anything in 7a2d7f74. Renamed to SEQ_LEN_REPORT_THRESHOLD and used only to phrase the OOM report, "N sample(s) longer than 2048 tokens (longest is Y tokens)". No raise, no behavior change outside the OOM handler.

tests/llmcompressor/datasets/test_max_seq_length.py
rishabhsinha17: Pruned in 7a2d7f74. What remains covers the measurement helper directly (untruncated_sequence_summary, including the arrow indices mapping), plus one test in tests/llmcompressor/pipelines/sequential/test_helpers.py asserting the enriched message through handle_sequential_oom.

rishabhsinha17:
@brian-dellabetta reworked in 7a2d7f74 per your review: no pre-flight validation and no length assumptions. format_calibration_data no longer raises; handle_sequential_oom appends measured sample lengths (count over 2048, longest) to the re-raised OOM message at the spot #3020 touched, computed only after an OOM is hit. Reject-path tests pruned. I cannot re-request review from a fork, so flagging here. Ready for another look.

Rodder5:
As the reporter of #3011, this is the message that would have ended my investigation on night one. The stock error pointed at GPU capacity and sent me toward pipeline="basic" and CPU offload, neither of which helps, and I burned a full evening on a misdiagnosis before rereading the log. A line reporting how many samples ran long and the longest length would have made the cause obvious immediately. Appending it to the re-raised OOM rather than validating up front also seems right: the success path stays untouched and the cost is only paid when the failure already happened.

src/llmcompressor/datasets/utils.py
```diff
@@ -179,6 +180,60 @@ def format_calibration_data(
     )
 
 
+def untruncated_sequence_summary(
```
brian-dellabetta: merge this helper with the `_untruncated_sequence_summary` helper in sequential pipeline

tests/llmcompressor/datasets/test_max_seq_length.py
```diff
@@ -43,3 +59,66 @@ def test_uniform_lengths_with_max_seq_length(self):
         result = collator(_make_features([10, 10]))
         assert result["input_ids"].shape[1] == 5
         assert result["attention_mask"].shape[1] == 5
+
+
+class TestUntruncatedSequenceSummary:
```
brian-dellabetta: your unit test `test_handle_sequential_oom_reports_long_untruncated_samples` should be sufficient, we can prune all this

src/llmcompressor/datasets/utils.py
```diff
@@ -179,6 +180,60 @@ def format_calibration_data(
...
+    # column as Python lists can spike CPU memory on large datasets
+    try:
+        import pyarrow.compute as pc
+
+        column = dataset.data.column(feature_name)
+        if getattr(dataset, "_indices", None) is not None:
+            column = column.take(dataset._indices.column(0))
+        lengths = pc.list_value_length(column).to_pylist()
+    except Exception:
+        lengths = [len(sample) for sample in dataset[feature_name]]
```
brian-dellabetta: pyarrow is fine because it doesn't require materialization, but i'd rather avoid this just for the summary. if we hit an exception, let's just return None here

src/llmcompressor/datasets/utils.py
```diff
@@ -179,6 +180,60 @@ def format_calibration_data(
...
+    longest = max(lengths, default=0)
+    if longest <= SEQ_LEN_REPORT_THRESHOLD:
+        return None
+
+    num_long = sum(length > SEQ_LEN_REPORT_THRESHOLD for length in lengths)
```
brian-dellabetta: `SEQ_LEN_REPORT_THRESHOLD` will bias users towards a smaller dataset, which is not clearly desired, especially for long-context evals. Instead, have the summary include num samples, median and max sequence length. that should be sufficient

brian-dellabetta (commented):
Thanks @rishabhsinha17 , some more changes requested

src/llmcompressor/datasets/utils.py
rishabhsinha17: Merged in 415681ba: the measurement now lives entirely in _untruncated_sequence_summary in the sequential pipeline; the datasets-side helper is gone.

tests/llmcompressor/datasets/test_max_seq_length.py
rishabhsinha17: Pruned in 415681ba; the two handle_sequential_oom tests are the remaining coverage.

src/llmcompressor/datasets/utils.py
rishabhsinha17: Done in 415681ba: pyarrow only, and any failure in the summary path returns None so the base message is unchanged.

src/llmcompressor/datasets/utils.py
rishabhsinha17: Changed in 415681ba: no threshold anywhere; the summary reports sample count, median, and max length, e.g. "2 sample(s) with median length 2508 and max length 5000 tokens".

src/llmcompressor/pipelines/sequential/helpers.py
```diff
@@ -482,15 +483,73 @@ def wrapper(*args, **kwargs):
...
+            )
+            summary = _untruncated_sequence_summary(args, kwargs)
+            if summary is not None:
+                message = f"{message} {summary}"
+            raise torch.OutOfMemoryError(message) from e
 
     return wrapper
+
+
+def _untruncated_sequence_summary(args: tuple, kwargs: dict) -> str | None:
```
brian-dellabetta:
```suggestion
def _get_untruncated_sequence_summary(args: tuple, kwargs: dict) -> str | None:
```

brian-dellabetta (commented):
Thanks @rishabhsinha17 , things look good. one nit, will share internally to make sure people are on board

src/llmcompressor/pipelines/sequential/helpers.py
rishabhsinha17: Applied in c9c5762b.

kylesayrs (changes requested):
This is quite a lot of code change to give a log which shows information that the user can already read from the top-level example script.

Imho the advice doesn't really make sense either. We offload samples in between subgraphs anyways (unless sequential_offload_device="cuda"), so a cuda oom is usually caused by
* Too much model memory
* Too large of a batch size
* Too long of a sequence length

Cuda memory usage doesn't scale with num calibration samples

brian-dellabetta:
@rishabhsinha17 I spoke to @kylesayrs offline. I can't think of a better way to infer dataset by inspecting args/kwargs, but it's a bit hacky and an error message to set num_calibration_samples and/or max_seq_length to a small value should suffice without listing the dataset statistics. Sorry for all the extra legwork, can you update the PR to simply update the error message with a few additional lines?

Note on above -- AWQ requires memory proportional to dataset size, each activation gets cached. So Kyle's comment on advice not making sense is not comprehensive of all modifiers in our library.

rishabhsinha17:
@brian-dellabetta @kylesayrs done in 5444194f. All dataset inspection is gone; the PR is now the message string plus a 12-line test. The message explains both levers inline: `max_seq_length` (unset means samples are untruncated, so a few long ones can dominate memory) and `num_calibration_samples` scoped to activation-caching modifiers per the AWQ note, so it does not overclaim for modifiers where sample count does not scale memory. No behavior change anywhere else.

brian-dellabetta (approved):
Thanks @rishabhsinha17 !

kylesayrs (approved):

PR #3017, opened 2026-08-11
Add explicit LinearExperts2D -> 3D MoE repack for HF-native saves

GOavi101:
- Add `repack_moe()` to restore native fused 3D expert modules after 2D linearization / calibration
- Packs expert weights (+ `weight_*` qparams) so `save_pretrained` writes HF-native keys (`gate_up_proj` / `down_proj`)
- Avoids transformers one-to-many `WeightConverter` support (see #2699 )

### Test plan

- pytest tests/llmcompressor/modeling/test_moe_repack.py (4 passed)

- Smoke on inference-optimization/Qwen3-VL-1.0B-A0.4B-Instruct:
load via load_quantizable_moe → LinearExperts2D
light W4A16 oneshot (16 synthetic samples)
repack_moe → Qwen3VLMoeTextExperts
save has packed keys (experts.gate_up_proj / experts.down_proj, no experts.0.*)
Transformers reload OK


src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -198,13 +245,122 @@ def from_experts_module(
...
+            expert: ExpertMLP = self[index]
+            expert.copy_to_experts_module(fused, index)
+
+        self._pack_weight_qparams(fused)
+
+        offload_kwargs = get_cache_init_kwargs(self)
+        offload_module(fused, **offload_kwargs)
+        return fused
+
+    def _pack_weight_qparams(self, fused: FusedExpertsProtocol) -> None:
```
kylesayrs, line 296: I think there might be a simpler way to write this

kylesayrs (commented):
Looks good! Pending review from @HDCharles @Roderick-Wu

src/llmcompressor/modeling/moe/linear_experts.py
GOavi101: Now stacks once per projection via a small helper, then a single stack/cat for gate+up, and uses self.has_gate instead of the isinstance check. Still indexes by num_experts so we don't pick up act_fn on the ModuleList.

kylesayrs (approved):
Looks great! I think one thing to keep in mind is that, when we do integrate this, we need to make sure the repacking only gets called on checkpoints with 3d weights. Repacking checkpoints which originally started as 2D will break loading and hf's conversion mapping logic.

Roderick-Wu (approved):
Looks good. Note to self to modify `copy_to_experts_module` to the view stuff we wanted to change

PR #3021, opened 2026-08-11
[transformers] Update tests to use new compressed-tensors api, pin transformers

kylesayrs:
## Purpose ##
* Fix unit tests which were failing for latest transformers version

## Changes ##
* Use new `dequantize` & `optimized_inference` api which replaces `run_compressed`
* When a model is decompressed, transformers no longer attaches the qconfig to `model.config.quantization_config`. Update test and oneshot load expectations to reflect this

## Testing ##
* All tests which load compressed models now pass

### DIFF
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -24,10 +24,10 @@ from compressed_tensors.utils import getattr_chain
 from loguru import logger
 from torch.utils.data import DataLoader
 from transformers import (
+    AutoConfig,
     PreTrainedModel,
     PreTrainedTokenizerBase,
     ProcessorMixin,
-    AutoConfig,
 )
 
 from llmcompressor.args import parse_args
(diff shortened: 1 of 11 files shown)


src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -290,13 +290,7 @@ def validate_model(model: PreTrainedModel):
             " -- https://github.com/vllm-project/compressed-tensors/blob/"
             "main/examples/convert_checkpoint/kimi_k26_example.py"
         )
-        if quant_method == QUANTIZATION_METHOD:
-            logger.warning(
```
brian-dellabetta, line 294: We discussed making this a warning due to it being very brittle based on user's configuration. do you think that's no longer an issue?

tests/llmcompressor/transformers/kv_cache/test_kv_cache.py
```diff
@@ -205,6 +209,8 @@ def test_kv_cache_gptq_config_format(kv_cache_fixture, tmp_path):
     assert counts > 0
 
 
+@requires_gpu
+@requires_version("transformers", ">5.15")
```
brian-dellabetta: is this intended to be inclusive of transformers 5.15.0?

kylesayrs: It is not, this failure was caused by 5.15

src/llmcompressor/entrypoints/oneshot.py
kylesayrs: Thinking a little more about it, I actually don't think it's an issue if the model is decompressed?

If you can think of a consistent way of detecting ct decompressed models lmk and we can add it back. This was removed because the ct config now removed after decompress.

In the longer term, we need to support layerwise decompression fully (hopefully by this release), so removing this warning is in the direction we're headed anyways.

brian-dellabetta: Discussed offline to use the quant config directly and keep warning, looks good to me!

brian-dellabetta (approved):
LGTM!

PR #3024, opened 2026-08-12
fix(autoround): prevent input_capture_hook from accumulating GPU memory during optimization

### LINKED ISSUE #3023: [Bug] AutoRoundModifier: input_capture_hook accumulates GPU memory during optimization when gradient_accumulate_steps > 1
# [Bug] AutoRoundModifier: input_capture_hook accumulates GPU memory during optimization when gradient_accumulate_steps > 1

## Environment

- llm-compressor: current main (1677fece)
- auto-round: latest
- PyTorch: 2.x, CUDA
- Hardware tested: 8 × RTX 5880 (44 GB), `Qwen/Qwen3.5-35B-A3B` (MoE)

## Bug Description

`AutoRoundModifier` registers `input_capture_hook` as a `forward_pre` hook on
all decoding layers in `on_calibration_start()`.  This hook is **never disabled
before the SignSGD optimization loop**, so every mini-batch forward pass during
optimization also triggers it.

With `gradient_accumulate_steps=N`, the hook appends N copies of
`(args, kwargs)` — each containing the full `hidden_states` tensor — into
`_all_module_input` per outer iteration.  These tensors are not released until
`post_autoround_cleanup()` calls `_all_module_input.clear()` after the entire
optimization run for that layer finishes.

## Reproduction

```python

# Monitor torch.cuda.memory_allocated() during calibration:
# it grows by ~gradient_accumulate_steps × hidden_states_bytes each outer iteration.

recipe = AutoRoundModifier(
    iters=200,
    batch_size=1,
    gradient_accumulate_steps=8,  # paper default — triggers the bug
    scheme="W4A16",
)
oneshot(model="<any model>", recipe=recipe, dataset="<calibration data>")
```

## Measured Impact

On `Qwen/Qwen3.5-35B-A3B` (hidden_size=2048, seq_len=6144, bf16):

| Parameter | Value |
|---|---|
| `gradient_accumulate_steps` | 8 |
| `hidden_states` size per sample | 24 MB |
| Memory growth per outer iter | **+204 MB** |
| Total accumulation at ITERS=200 | **~40 GB** |
| Result | OOM on 44 GB GPUs |

```
[MEM-ITER]   1  Δalloc=+204.4MB  alloc=13234MB
[MEM-ITER]   2  Δalloc=+204.4MB  alloc=13438MB
[MEM-ITER]   3  Δalloc=+204.4MB  alloc=13642MB
...
```

## Root Cause

```
on_calibration_start()
  → register_hook(module, input_capture_hook, "forward_pre")

apply_autoround()
  → ar.quantize_block(block, inputs=ar_inputs, ...)     # ITERS optimization steps
      → for each outer iter:
          for each mini-batch (gradient_accumulate_steps times):
            BlockForwardRunner.forward()
              → decoding_layer.forward()                # triggers forward_pre hook!
                  → input_capture_hook()
                      → _all_module_input[name].append((args, kwargs))
                                                 ↑
                                    hidden_states kept alive
```

The hook was intended only for the calibration pass but fires during optimization too.

## Proposed Fix

The fix requires distinguishing two states that look identical from the dict alone:
**"this layer has never been seen"** (calibration phase, hook should capture) vs
**"this layer's inputs were already consumed by apply_autoround()"** (optimization
phase, hook should be a no-op).

Checking `if name not in _all_module_input` does not work: after `pop()` the key
is absent, so the calibration-phase first fire is also blocked — causing `KeyError`
when the next layer's inputs are requested.

The correct approach is a separate "consumed" tracking structure alongside
`_all_module_input` so that the hook can distinguish the two cases.  The fix
touches only `AutoRoundModifier` in `base.py` and produces a diff of ~+17/-4 lines
against `main`.

## Verification

After fix, on the same setup:

```
[MEM-ITER]   1  Δalloc=+0.0MB  alloc=11276MB
[MEM-ITER]   2  Δalloc=+0.0MB  alloc=11276MB
[MEM-ITER]   3  Δalloc=+0.0MB  alloc=11276MB
...
[MEM-PTR] iter=2: 2 new tensors (+48.0MB new, -48.0MB freed)  ← legitimate pred/ref, freed same iter
```

Memory is completely flat across all 8 ranks for 20 iterations.

## PR Ready

I have a clean fix on branch `fix/autoround-input-capture-hook-leak` (forked from
`vllm-project/llm-compressor` at `1677fece`) with a single-file diff of
+17/-4 lines against `main`.  Happy to submit a PR once the approach is confirmed.

xesdiny:
## Summary

Fixes #3023

`AutoRoundModifier.input_capture_hook` is registered on all decoding layers during
`on_calibration_start()` but is never disabled before the SignSGD optimization loop.
During optimization, `BlockForwardRunner.forward()` calls `decoding_layer.forward()`
on every mini-batch, triggering the hook each time. With `gradient_accumulate_steps=N`
this retains N hidden_states tensors per outer iteration, causing
~`N × hidden_states_size` MB of GPU memory growth per iteration.

At `gradient_accumulate_steps=8`, `hidden_size=2048`, `seq_len=6144` (bf16):
**+204 MB/iter → ~40 GB accumulated at ITERS=200 → OOM on 44 GB GPUs.**

## Changes

- Add `_consumed_layers: set` PrivateAttr to track layers processed by `apply_autoround()`
- `input_capture_hook`: use `setdefault()` for calibration-phase initialization; exit immediately for consumed layers
- `apply_autoround()`: mark layer consumed before `pop()`-ing its inputs
- `on_calibration_end()`: clear `_consumed_layers`

## Verification

After fix, on 8 × RTX 5880 (44 GB), `Qwen/Qwen3.5-35B-A3B`, `gradient_accumulate_steps=8`:

```
Before: Δalloc = +204 MB/iter across all 8 ranks
After:  Δalloc = +0.0 MB/iter across all 8 ranks (20 iterations verified)
```


xesdiny:
@gemini-code-assist review

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -313,7 +320,10 @@ def apply_autoround(self, state, modules):
             ar.batch_dim = 0
             first_param = next(decoding_layer.parameters())
             device = first_param.device
-            cur_inputs = self._all_module_input[decoding_layer._tmp_name]
+            # Mark consumed before pop so the hook immediately becomes a no-op
+            # for any forward call triggered during the optimization loop.
+            self._consumed_layers.add(decoding_layer._tmp_name)
```
yiliu30: Hi @xesdiny, good catch! Rather than adding type information to `_consumed_layers`, I’d prefer to remove the input-capture hook for each layer as soon as its calibration inputs have been collected, before AutoRound tuning begins.

The hooks for later layers can remain active until their own calibration pass. This way, we avoid retaining optimization inputs without introducing additional state for consumed layers.

xesdiny:
Thanks for the reviews! Updated the implementation in cb3cbc09:

**Per @yiliu30**: replaced the `_consumed_layers` guard approach with early hook removal. Each layer's `forward_pre` hook is now stored by name in `_capture_hooks` at registration time, then removed via `self.remove_hooks({handle})` immediately before `quantize_block` is called — so the hook is gone before the SignSGD optimization loop starts. Hooks for later layers remain active until their own calibration pass. This eliminates the extra tracking state entirely.


**Local validation** on 8× L20 (Qwen3.5-35B-A3B, `gradient_accumulate_steps=8`, `iters=200`, `LIMIT_TO_LAYERS=2`):
```
[MEM-ITER]   1  Δalloc=+0.0MB  alloc=9846MB
[MEM-ITER]  50  Δalloc=+0.0MB  alloc=9846MB
[MEM-ITER] 100  Δalloc=+0.0MB  alloc=9846MB
[MEM-ITER] 150  Δalloc=+0.0MB  alloc=9846MB
[MEM-ITER] 200  Δalloc=+0.0MB  alloc=9846MB
```

yiliu30:
Thanks @xesdiny! This fix LGTM. The failing UTs are unrelated to this change, so please wait for the fix to land on the main branch.

yiliu30 (approved):

xesdiny:
@gemini-code-assist review

xesdiny:
@yiliu30 Thanks for the review and approval! The PR now needs a second approving review to satisfy the merge protections. Could you help ping another maintainer (e.g. @HDCharles, @dsikka, or @kylesayrs) when you get a chance? Appreciate it!

brian-dellabetta (approved):

Issue #2922, opened 2026-07-13
FP4 Baseline Benchmarking - Small (non-MoE) Models (No Rotations)


dsikka:
# Llama-3.1-8B-Instruct NVFP4 Evaluation Results

Baseline: `RedHatAI/Llama-3.1-8B-Instruct` (FP16)  
Seeds: 1234, 2345, 3456  
Generation params: temperature=0.6, top_p=0.9  
Serving: vLLM v0.25.1, TP=1, max-model-len=20480  

## Summary
- GTPQ models were calibrated with Ultrachat and Perfect Blend
- Autoround models were calibrated using the autoround default dataset NeelNanda/pile-10k and a second set of models were calibrated using Ultrachat for an easier comparison
- An attempt at MRCR results are below; note: the task was missing from MLR's fork so claude added it based on other longcontext evals available 

Scripts: https://github.com/vllm-project/llm-compressor/compare/nvfp4_scripts?expand=1

| Model | Top-2 finishes | Tasks |
|-------|---------------|-------|
| NVFP4-GPTQ-damp05-perf-blend | 4 / 5 | gsm8k_platinum_cot_llama, mmlu_pro_chat, ifeval, math_500 |
| NVFP4-GPTQ | 2 / 5 | mmlu_cot_llama, ifeval |
| NVFP4-GPTQ-perf-blend-more-data | 2 / 5 | gsm8k_platinum_cot_llama, math_500 |
| NVFP4-GPTQ-AWQ-imatrix-perf-blend | 2 / 5 | mmlu_pro_chat, math_500 |
| NVFP4-GPTQ-imatrix-perf-blend | 2 / 5 | gsm8k_platinum_cot_llama, mmlu_cot_llama |
| NVFP4-GPTQ-perf-blend | 1 / 5 | mmlu_pro_chat |
| NVFP4-AutoRound-AWQ | 1 / 5 | ifeval |

**GPTQ calibration data:** Perf-blend variants all match or beat base GPTQ (94.2%), while fineweb (93.2%) and red-pajama (93.0%) hurt recovery due to math_500 drops. damp05-perf-blend is the new best overall GPTQ variant (95.6%), +1.0pp over the original perf-blend. perf-blend-more-data is second (95.0%). MSE loss and lower damping are neutral.

[...]

PR #3025, opened 2026-08-12
docs: model-size guidance for scheme selection, language alignment for calibration data

spped2000:
Two documentation gaps I ran into while quantizing Thai models, each backed by paired measurements.

## 1. `choosing-scheme.md` — model size, not just GPU

The scheme table answers *what can my GPU run*. It does not answer *what can my model absorb*, and that turned out to matter more than hardware for us. Same recipe (NVFP4 W4A4 via `QuantizationModifier`), same 512-sample calibration set, same machine, paired per-item against each model's own BF16 source with exact McNemar:

| Model size | Outcome |
|---|---|
| ~30B | multiple-choice recovery within noise (−0.5 pt, n.s.) |
| ~7B | −2.4 pt knowledge (p=0.02), **−10.2 pt instruction following** (p=0.0009) |

Re-running the same 7B as **W4A16** recovered both (−0.7 and −3.7 pt, neither significant) at the same footprint and speed — hence the added rule of thumb to prefer weight-only below ~10B.

The two halves of W4A4 also fail on different task types, which the two arms isolate cleanly: dropping activation quantization fixed instruction following and knowledge, but **mathematical reasoning was unchanged (−5.4 vs −5.6 pt, both significant)** — that damage comes from the 4-bit weights and weight-only does not rescue it.

## 2. `choosing-dataset.md` — language alignment

"Domain alignment" covers general/instruct/code, and all four suggested datasets are English. There is currently no guidance that calibration data language matters, even though calibration is what sets the activation scales. Added a short subsection: use a mixed set (we used 50/50 target-language/English, 512 samples), a snippet for building one, and a note to evaluate in the target language too.

## Framing

These are one lab's numbers on one model family (Thai-capable Qwen2/Qwen3 derivatives, NVIDIA GB10). I wrote both sections as a starting prior with an explicit "validate on your own workload" caveat rather than as universal rules — happy to soften or cut anything that reads as over-claiming. All runs were pre-registered before measurement and the artifacts are public: https://github.com/spped2000/thaillm-nvfp4-dgx-spark

---

*Disclosure: prepared with assistance from Claude (Anthropic). Every figure quoted comes from paired runs on the hardware described.*

### DIFF
--- a/docs/steps/choosing-dataset.md
+++ b/docs/steps/choosing-dataset.md
@@ -48,7 +48,7 @@ the scales to English activation ranges, and the languages you actually serve
 inherit whatever error is left over.
 
 A mixed set is usually enough: half target-language, half English preserves
-English behaviour while covering the target language's activation ranges. In
+English behavior while covering the target language's activation ranges. In
 one NVFP4 run on a Thai-capable 30B model, a 50/50 Thai/English calibration
 set (512 samples) left Thai multiple-choice accuracy statistically unchanged
 against the BF16 source, while English tasks moved within the same small
--- a/docs/steps/choosing-dataset.md
+++ b/docs/steps/choosing-dataset.md
@@ -40,6 +40,37 @@ Some popular datasets include:
 | `wikitext-2-raw-v1` | General language models | Clean Wikipedia text for broad language understanding |
 | `c4` | General pre-training | Large-scale web text for general-purpose models |
 
+### Language alignment
+If the model will serve non-English traffic, the calibration set should
+contain that language. Calibration determines the activation scales, and
+every dataset listed above is English — so an English-only calibration fits
+the scales to English activation ranges, and the languages you actually serve
+inherit whatever error is left over.
+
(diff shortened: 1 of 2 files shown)


spped2000:
Thanks — both points taken, pushed:

- `behaviour` → `behavior`.
- The speed claim was overstated. My measurement was single-stream decode, which is bandwidth-bound, so the two schemes looked equal there; that does not generalize. Reworded to say W4A16 matches W4A4 only in the bandwidth-bound regime and that W4A4 pulls ahead on compute throughput at larger batch or longer context via 4-bit tensor cores, with a note to measure your own serving shape.

Also adding the `ready` label now that the PR is code-complete.

spped2000:
Note: I cannot add the `ready` label myself (no write access as an outside contributor). The PR **is** code-complete and the review feedback above is addressed — could a maintainer add `ready` so the test suite runs?

dsikka (commented):
Hi @spped2000 much of this is outdated. See https://github.com/vllm-project/llm-compressor/issues/2922

PR #3026, opened 2026-08-12
Update review rules

dsikka:
- Require one review from a maintainer, unless the two-reviews label has been applied


kylesayrs (approved):
Thanks

PR #3028, opened 2026-08-13
HIGGS: ILP-based mixed-precision quantization

### LINKED ISSUE #2690: [RFC] HIGGS Integration into llm-compressor
# [RFC] Native HIGGS-style mixed-precision allocation pipeline in `llm-compressor` (NVFP4 / FP8_Dynamic / FP16)

## Background

We implemented a practical HIGGS-style workflow externally around `llm-compressor` and `vLLM`, inspired by the paper *Pushing the Limits of Large Language Model Quantization via the Linearity Theorem*.

Today, the workflow is split into three custom scripts:

1. `create_quantized_layers.py`  
   - Runs `llmcompressor.oneshot(...)` to export per-layer/group weights for multiple precisions (currently `FP16`, `FP8_Dynamic`, `NVFP4`).
2. `HIGGS.py`  
   - Builds an error DB from exported layer banks, estimates sensitivity coefficients (`alpha`) with heuristic/noise/hybrid modes, and solves an LP bit allocation problem under a target average bit budget.
3. `create_mixed_precision_models.py`  
   - Reads per-layer assignments and runs one final mixed quantization pass to generate the final compressed checkpoint.

This has been exercised across model families including:
- `meta-llama/Llama-3.2-{1B,3B}-Instruct`
- `meta-llama/Llama-3.1-8B-Instruct`
- `Qwen/Qwen3-8B`
- `Qwen/Qwen3-30B-A3B`
- `Qwen/Qwen3-30B-A3B-Instruct-2507`

## What we have today (working, but external)

- A reproducible optimization-first pipeline for mixed precision allocation.
- Architecture-aware grouping (e.g., `q/k/v -> qkv`, `gate/up -> gate_up`) to avoid incoherent per-subprojection assignment.
- Error modeling from measured relative Frobenius distortion per candidate precision.
- Alpha modes:
  - `heuristic` (cheap)
  - `noise` (loss-perturbation probing)
  - `hybrid`
- LP solve with constraints:
  - one precision choice per layer/group
  - weighted bit budget target
  - optional anti-degenerate mixedness behavior
- Final checkpoint generation via one-shot quantization using chosen assignments.
- Evaluation utilities (KL/perplexity) with `vLLM` prompt logprobs.

## Problem / pain points

The current setup requires substantial glue code and multiple passes around `oneshot`:

- custom artifact export + parsing (`.pth` banks, metadata, layer size JSON)
- custom grouping/expansion conventions duplicated outside compressor internals
- external optimization script and solver dependencies
- custom config handoff into final quantization pass
- hard to standardize and maintain as first-class `llm-compressor` UX

In short: the approach works, but orchestration is outside `llm-compressor`, so users must reimplement pipeline logic to reproduce HIGGS-like dynamic allocation.

## Proposal

Add first-class HIGGS-style mixed-precision allocation support in `llm-compressor` so users can run this as a native pipeline/modifier flow.

### Proposed capabilities

1. **Candidate precision scoring pass**
   - Native way to collect per-layer/group distortion for multiple candidate quant schemes.
   - Reuse existing quantization infrastructure instead of external layer-bank scripts.

2. **Layer/group abstraction**
   - First-class grouping policy for transformer projections (e.g. `qkv`, `gate_up`) and singletons (`o_proj`, `down_proj`), with extensibility for architectures.

3. **Sensitivity (`alpha`) estimation plugins**
   - Built-in modes equivalent to current prototype:
     - `heuristic`
     - `noise`
     - `hybrid`
   - Noise mode should integrate cleanly with sequential/offload-friendly pipelines.

4. **Global allocator**
   - Built-in LP/MIP allocator over candidate precisions under weighted bit budget.
   - Output should include assignment map + summary stats + feasibility diagnostics.

5. **Materialization stage**
   - Apply assignment map directly to produce one final compressed checkpoint.
   - Integrate with current `QuantizationModifier` / config groups and serialization conventions.

6. **Artifacts and reporting**
   - Optional persisted artifacts for:
     - per-layer error tables
     - alpha tables
     - allocator stats
     - final assignment JSON

## Sketch of desired user experience

Something close to:

```yaml
HIGGSMixedPrecisionModifier:
  targets: ["Linear"]
  candidate_schemes: ["NVFP4", "FP8_DYNAMIC", "FP16"]
  target_avg_bits: [5.0, 5.5, 6.0, 6.5, 7.0]
  grouping:
    policy: "transformer_default"  # qkv, gate_up, o_proj, down_proj
  alpha:
    method: "hybrid"               # heuristic | noise | hybrid
    calibration_dataset: "HuggingFaceH4/ultrachat_200k"
    calibration_split: "train_sft"
    calibration_samples: 32
  optimizer:
    solver: "scip"
    force_mixed: true
  output:
    save_assignment: true
    save_stats: true
```

Or equivalent CLI flags under a dedicated pipeline entrypoint.

## Scope / non-goals (for initial RFC)

**In scope**
- Weight precision allocation over a defined candidate set (`NVFP4` / `FP8_Dynamic` / `FP16`)
- Group-aware assignment and budget-constrained optimization
- Exporting reproducible assignment/stats artifacts

**Out of scope (initial)**
- New low-level quantization kernels
- Runtime scheduling/autotuning in serving stack
- Full automatic search over arbitrary precision families

## Why this should live in `llm-compressor`

- Uses existing compressor quantization primitives already validated in production paths.
- Converts a proven external workflow into reusable core functionality.
- Reduces user-side script maintenance and mismatch risk.
- Aligns with optimizer-first compression roadmap: quality-budget tradeoffs become explicit and reproducible.

## Requested feedback

1. Preferred integration point: new modifier vs dedicated pipeline stage vs hybrid.
2. Best place to host alpha estimation APIs (modifier-level, observer-level, pipeline utility).
3. Solver expectations and optional dependency policy for LP/MIP backends.
4. Recommended artifact format for assignment + diagnostics (JSON schema expectations).
5. Interest in upstreaming with an initial constrained scope (`NVFP4/FP8_Dynamic/FP16`) and expanding later.

## Acceptance criteria for an initial implementation

- User can run one command/recipe and obtain:
  - per-layer/group precision assignment satisfying target average bits
  - a final compressed checkpoint materialized from that assignment
  - saved stats/diagnostics for reproducibility
- Works on representative Llama/Qwen decoder-only transformer models.
- Produces stable assignment behavior across multiple target bit budgets and alpha modes.

## Repro context from current prototype
Reference implementation and scripts are available here: https://github.com/neuralmagic/HIGGS

HDCharles:
## Overview

HIGGS is a technique from [this](https://arxiv.org/pdf/2411.17525) paper which can be used to determine the theory-based optimal mixed-precision quantization configuration to use to quantize a model given a constraint on the average quantization bitwidth the user wants. Note in the paper there are 2 methods for calculating how sensitive each layer is to quantization, this PR only implements the heuristic method.

HIGGS has 2 phases, it first collects a bunch of MSE data and layer location information to estimate layer sensitivity to quantization, it then converts this information to an ILP and then uses PuLP to solve the ILP and generate an optimal quantization config. This config can then be passed into oneshot or model_free_ptq to actually quantize the model.

All the above functionality lives in the HiggsMSECollectorConverter which is supported by the get_higgs_config entrypoint since our converter architecture makes a number of assumptions about the converter actually applying quantization to the model. In the future we could get rid of get_higgs_config. 

## Results

| Target Bits | NVFP4A16 Layers | FP8_DYNAMIC Layers | word_perplexity |
|-------------|-----------------|---------------------|-----------------|
| 4.0 | 224 | 0 | 9.9381 |
| 4.5 | 124 | 100 | 9.8493 |
| 5.0 | 92 | 132 | 9.7695 |
| 5.5 | 75 | 149 | 9.7209 |
| 6.0 | 61 | 163 | 9.6805 |
| 6.5 | 46 | 178 | 9.6384 |
| 7.0 | 30 | 194 | 9.5992 |
| 7.5 | 15 | 209 | 9.5674 |
| 8.0 | 0 | 224 | 9.4934 |
| 16.0 (base) | 0 | 0 | 9.4236 |

As you can see, a straightforward bitwidth sweep from 4-8 yields consistently monotonic  PPL improvement as the target bitwidth grows. While such a growth is not strictly guaranteed given that after all HIGGS doesn't take into account activation quantization accuracy and only has a local estimate of weight quantization accuracy, it performs extremely well (especially for such a fast technique). These results were generated using the higgs_nvfp4a16_fp8_model_free.py example. Note we use the nvfp4_expanded_mse observer to get a better than RTN quantized model both for error estimates for HIGGS and the final quantization.

As mentioned above though HIGGS can compose with oneshot for quantization schemes that require calibration:

| Example | Schemes | Method | HIGGS Time | Quant Time | PPL |
|---------|---------|--------|------------|------------|-----|
| `higgs_nvfp4_fp8_qmod` | NVFP4 + FP8_DYNAMIC | QuantizationModifier | 8.05s | 1:21 | 9.7502 |
| `higgs_nvfp4_fp8_gptq` | NVFP4 + FP8_DYNAMIC | GPTQModifier | 7.64s | 3:24 | 9.7486 |

As you can see we're able to apply GPTQ or just use a normal QuantizationModifier after HIGGS, you can do things like AWQ as well though that may invalidate the HIGGS scheme's theoretical optimality.

Also note the speed of this HIGGS implementation, since we use a model_free method to collect MSE values without writing anything, we are able to do this incredibly quickly, for all the models i've tested HIGGS takes up a minority of the time compared to the actual application of quantization. This massively outperforms other designs in terms of speed even if you were to pre-quantize the model to each candidate scheme. The process of reading N different model weights into memory in order to calculate MSE is much slower than just calculating it all from scratch, even though some work is duplicated due to weights needing to be quantized both to calculate MSE and to save the final quantized model.

these results were generated using the /home/HDCharles/repos/llm-compressor/examples/quantization_higgs/higgs_nvfp4_fp8_gptq.py example.

## Files

llmcompressor/entrypoints/higgs/base.py - contains MSE collection code and general orchestation (HiggsMSECollectorConverter and get_higgs_config

llmcompressor/entrypoints/higgs/ilp_solver.py - our ILP setup/solver

llmcompressor/entrypoints/higgs/utils.py - all helpers/utils includes the code that calculates the heuristic alpha (layer sensitivity based on location in model)

### DIFF
--- a/examples/quantization_higgs/llama3_higgs_example.py
+++ b/examples/quantization_higgs/llama3_higgs_example.py
@@ -1,0 +1,91 @@
+"""
+HIGGS: ILP-Based Mixed-Precision Quantization Example
+
+This example demonstrates how to use HIGGS to automatically select optimal
+quantization schemes for each layer in Llama-3-8B.
+"""
+
+import os
+import time
+import torch
+
+from llmcompressor.transformers.compression.higgs import ilp_quantize
+
+# Select model
+MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
+
+# Define candidate quantization schemes
+# HIGGS will evaluate each scheme on each layer and select the optimal assignment
+# Note: Use schemes with different quality/size tradeoffs for mixed-precision
+CANDIDATE_SCHEMES = ["W4A16", "W8A16"]  # 4-bit vs 8-bit weights
+
+# Run HIGGS quantization
+# Phase 1: Collect MSE data and solve ILP
+# Phase 2: Apply selected schemes and save
+print("Starting HIGGS quantization...")
+
+# Reset peak memory stats
+if torch.cuda.is_available():
+    torch.cuda.reset_peak_memory_stats()
+
+start_time = time.time()
+
+optimal_config = ilp_quantize(
+    model_stub=MODEL_ID,
+    save_directory=os.path.expanduser("~/hf_hub/Meta-Llama-3-8B-Instruct-HIGGS"),
+    candidate_schemes=CANDIDATE_SCHEMES,
+    targets="Linear",  # Quantize all Linear layers
+    ignore=["lm_head", "model.embed_tokens"],  # Skip output projection and embeddings
+    enforce_fused_layer_constraints=True,  # gate_proj+up_proj, q+k+v same scheme
+    target_avg_bitwidth=6.0,  # Target 6-bit average
+    max_workers=8,  # Parallel workers for phase 2
+    device="cuda:0",  # Use GPU for faster processing
+)
+
+end_time = time.time()
+elapsed_time = end_time - start_time
+
+# Get peak memory
+if torch.cuda.is_available():
+    peak_memory_gb = torch.cuda.max_memory_allocated() / 1024**3
+    print(f"\nPeak CUDA memory: {peak_memory_gb:.2f} GB")
+
+# Print summary
+print("\n" + "=" * 80)
+print("HIGGS Quantization Complete!")
+print("=" * 80)
+print(f"Total time: {elapsed_time:.2f} seconds ({elapsed_time/60:.2f} minutes)")
+print(f"Generated {len(optimal_config.config_groups)} config groups:")
+for group_name, scheme in optimal_config.config_groups.items():
+    num_layers = len(scheme.targets)
+    bitwidth = scheme.weights.num_bits if scheme.weights else "N/A"
+    print(f"  {group_name}: {num_layers} layers @ {bitwidth}-bit")
+print(f"\nQuantized model saved to: ~/hf_hub/Meta-Llama-3-8B-Instruct-HIGGS")
+print("=" * 80)
+
+# Verify the quantized model was saved
+print("\nVerifying quantized model files...")
+import os
+output_dir = os.path.expanduser("~/hf_hub/Meta-Llama-3-8B-Instruct-HIGGS")
+if os.path.exists(output_dir):
+    files = os.listdir(output_dir)
+    safetensor_files = [f for f in files if f.endswith('.safetensors')]
+    print(f"  Found {len(safetensor_files)} safetensors files")
+    print(f"  Total files: {len(files)}")
+
+    # Check config
+    config_path = os.path.join(output_dir, "config.json")
+    if os.path.exists(config_path):
+        import json
+        with open(config_path) as f:
+            config = json.load(f)
+        if "quantization_config" in config:
+            print(f"  Quantization config present: {config['quantization_config']['format']}")
+print("=" * 80)
+
+"""
+lm_eval --model vllm \
+    --model_args "pretrained=/home/HDCharles/repos/llm-compressor/Meta-Llama-3-8B-Instruct-HIGGS,dtype=auto,max_model_len=4096,add_bos_token=True,gpu_memory_utilization=0.85" \
+    --tasks "wikitext" \
+    --batch_size auto
+"""
--- a/setup.py
+++ b/setup.py
@@ -145,6 +145,7 @@ setup(
             if BUILD_TYPE == "release"
             else "compressed-tensors>=0.18.1a2"
         ),
+        ("pulp>=2.7.0,<=2.10.0" if BUILD_TYPE == "release" else "pulp>=2.7.0"),
     ],
     extras_require={
         "dev": [
--- a/examples/quantization_higgs/ALPHA_HEURISTIC_ANALYSIS.md
+++ b/examples/quantization_higgs/ALPHA_HEURISTIC_ANALYSIS.md
@@ -1,0 +1,311 @@
+# Alpha Heuristic Analysis for HIGGS Mixed-Precision Quantization
+
+## Overview
+
+This document summarizes the empirical analysis of per-layer quantization sensitivity
+across multiple LLM architectures. The goal is to derive a heuristic alpha formula
+that predicts how much each layer's quantization error (MSE) contributes to overall
+perplexity degradation, enabling ILP-based optimal bitwidth allocation without
+expensive per-model calibration.
+
+The HIGGS paper models expected perplexity as:
+
+    E[PPL(W_q)] ~ PPL(W) + sum_l( alpha_l * MSE_l )
+
+where `alpha_l` is the per-layer importance weight. This analysis fits the form:
+
+    alpha_l = base_func(size; type) * depth_factor(d) * scale
+
+with:
+- `base_func`: a function of layer parameter count, selected per layer type
+- `depth_factor(d) = 1 + c1*d + c2*d^2`: a quadratic in layer depth
+- `scale`: a global multiplier (absorbed into the ILP objective, does not affect allocation)
+
+## Methodology
+
+### Data Collection
+
+1. **Bitwidth sweep**: For each model, quantize at 15 target average bitwidths
+   (4.01 to 7.5 in 0.25 steps) using ILP with W4A16/W8A16 candidate schemes.
+2. **MSE extraction**: Record per-layer MSE for each quantization scheme directly
+   from the model safetensors (no calibration data needed).
+3. **Perplexity evaluation**: Measure wikitext word_perplexity via lm_eval + vLLM
+   for each quantized checkpoint and the baseline model.
+4. **Regression fitting**: Grid search over 6 candidate base functions per layer type,
+   fit depth quadratic coefficients via SLSQP with non-negativity constraints.
+
+### Base Function Candidates
+
+Six candidates are evaluated for each layer type:
+- `log(s+1)` where s = parameter count
+- `s` (linear)
+- `s^2` (quadratic)
+- `s^(-1)` (inverse)
+- `sqrt(s)` (square root)
+- `1` (constant / size-independent)
+
+### Depth Curve Constraints
+
+The depth quadratic `f(d) = 1 + c1*d + c2*d^2` is constrained to be non-negative
+for all valid depths d in [0, max_depth]. This is enforced by checking:
+- `f(0) >= 0` (always 1.0, trivially satisfied)
+- `f(max_depth) >= 0`
+- `f(-c1/(2*c2)) >= 0` if the vertex falls within [0, max_depth]
+
+### Layer Type Classification
+
+Layers are classified into types based on name patterns:
+- **attention**: contains `self_attn` (q/k/v/o_proj)
+- **mlp**: contains `mlp` but not `experts` or `shared_expert`
+- **moe**: contains `experts.\d+.` (MoE expert layers)
+- **moe-shared-expert**: contains `shared_expert`
+- **embedding**: contains `embed`
+- **default**: anything else (e.g., Gemma4's `per_layer_projection`)
+
+## Models Evaluated
+
+| Model | Params | Architecture | Layers | Quantizable Layers | Layer Types | Eval Points |
+|-------|--------|-------------|--------|-------------------|-------------|-------------|
+| Llama-3-8B | 8B | Dense | 32 | 224 | attention, mlp | 15/15 |
+| Llama-3.1-8B | 8B | Dense | 32 | 224 | attention, mlp | 15/15 |
+| Qwen2.5-7B | 7B | Dense | 28 | 196 | attention, mlp | 15/15 |
+| Mistral-7B | 7B | Dense | 32 | 224 | attention, mlp | 15/15 |
+| Qwen3-30B-A3B | 30B | MoE (128 experts) | 48 | 18624 | attention, moe | 8/15 |
+| Gemma-4-E4B | 12B | Multimodal | 42 | 379 | attention, mlp, default | 0/15 |
+
+Notes:
+- Qwen3-30B-A3B has no shared experts; all MLP layers are routed expert layers.
+- Gemma-4-E4B evaluation failed: baseline PPL=251.84 (instruction-tuned multimodal model)
+  and all quantized checkpoints produce NaN in vLLM inference (compatibility issue).
+- Qwen3-30B-A3B intermediate bitwidths (0.25-step) fail on vLLM with TP>=4
+  (engine initialization failure); only the original 0.5-step evals succeeded.
+
+## Fitting Results
+
+### Best Fit Per Model
+
+| Model | Residual | Base(attention) | Base(mlp/moe) | Scale |
+|-------|----------|-----------------|---------------|-------|
+| Mistral-7B | **3.07e-4** | sqrt(s) | sqrt(s) | 4.28 |
+| Llama-3.1-8B | 2.00e-3 | 1 | log(s+1) | 2.59e2 |
+| Llama-3-8B | 2.38e-3 | 1 | log(s+1) | 4.46e2 |
+| Qwen3-30B-A3B | 4.12e-3 | log(s+1) | 1 | 4.51 |
+| Qwen2.5-7B | 1.63e-2 | log(s+1) | log(s+1) | 1.02e2 |
+
+Mistral-7B has the best fit by a wide margin, suggesting its PPL response to
+quantization is the most predictable from MSE alone.
+
+### Raw Depth Coefficients
+
+| Model | Type | c1 | c2 | Max Depth |
+|-------|------|-----|-----|-----------|
+| Llama-3-8B | attention | +0.054884 | -0.002811 | 31 |
+| Llama-3-8B | mlp | -0.093039 | +0.002972 | 31 |
+| Llama-3.1-8B | attention | +0.201046 | -0.007526 | 31 |
+| Llama-3.1-8B | mlp | -0.076475 | +0.002589 | 31 |
+| Qwen2.5-7B | attention | -0.096514 | +0.002536 | 27 |
+| Qwen2.5-7B | mlp | -0.151138 | +0.007692 | 27 |
+| Mistral-7B | attention | -0.058253 | +0.001573 | 31 |
+| Mistral-7B | mlp | -0.077680 | +0.002658 | 31 |
+| Qwen3-30B-A3B | attention | -0.050264 | +0.000632 | 47 |
+| Qwen3-30B-A3B | moe | -0.091133 | +0.002076 | 47 |
+
+### Normalized Depth Curves
+
+To compare across models with different layer counts, we normalize depth to
+`t = d / max_depth` so t ranges from 0 (first layer) to 1 (last layer):
+
+    f(t) = 1 + a*t + b*t^2    where a = c1*D, b = c2*D^2
+
+#### Normalized Attention Coefficients
+
+| Model | a | b | f(0) | f(0.5) | f(1) | min | t_min |
+|-------|-----|-----|------|--------|------|-----|-------|
+| Llama-3-8B | +1.70 | -2.70 | 1.00 | 1.18 | 0.00 | 0.00 | 1.00 |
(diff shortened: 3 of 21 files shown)


examples/quantization_higgs/results_bitwidth_sweep.md
```diff
@@ -0,0 +1,45 @@
...
+baselines at 4-bit, 8-bit, and 16-bit.
+
+## Configuration
+
+- **Model**: meta-llama/Meta-Llama-3.1-8B-Instruct
+- **Schemes**: NVFP4A16 (4-bit weight-only) + FP8_DYNAMIC (8-bit dynamic)
+- **Method**: convert_checkpoint + HiggsQuantizationConverter (data-free)
+- **Eval**: lm_eval wikitext word_perplexity via vLLM
+
+## Results
```
brian-dellabetta, line 14: that's cool, it seems like there's a pretty consistent linear drop from 8.0 target bits on down to 4.0

examples/quantization_higgs/llama3_higgs_example.py
```diff
@@ -0,0 +1,89 @@
...
+
+from llmcompressor.entrypoints.higgs import get_higgs_config
+
+# Select model
+MODEL_ID = "meta-llama/Meta-Llama-3-8B-Instruct"
+
+# Define candidate quantization schemes
+# HIGGS will evaluate each scheme on each layer and select the optimal assignment
+# Note: Use schemes with different quality/size tradeoffs for mixed-precision
+CANDIDATE_SCHEMES = ["W4A16", "W8A16"]  # 4-bit vs 8-bit weights
```
brian-dellabetta, line 20: this is because humming is broken on vllm? would be very curious to see how `CANDIDATE_SCHEMES = ["W2A16", ... , "W8A16"]` differs per bit budget relative to this

examples/quantization_higgs/results_convert_vs_oneshot.md
```diff
@@ -0,0 +1,66 @@
...
+`group_0`, `group_1`), but `convert_checkpoint` preserves the original
+names from the `QuantizationConfig` object.
+
+vLLM failed to correctly match layers to quantization schemes when
+config group names contained suffixes, resulting in degraded inference
+quality despite the quantized weights being byte-identical.
+
+## Fix
+
+Changed `generate_config_groups()` to use plain `group_{idx}` names
```
brian-dellabetta, line 21: strange, i wouldn't think the group name would matter as long as order was preserved. worth raising an issue if we have an MRE

examples/quantization_higgs/results_method_comparison.md
```diff
@@ -0,0 +1,29 @@
...
+- **Calibration** (where applicable): ultrachat_200k, 256 samples, 2048 seq len
+- **Eval**: lm_eval wikitext word_perplexity via vLLM
+
+## Results
+
+| Example | Schemes | Method | HIGGS Time | Quant Time | PPL |
+|---------|---------|--------|------------|------------|-----|
+| higgs_nvfp4_fp8_qmod | NVFP4 + FP8_DYNAMIC | QuantizationModifier | 3.5s | 1:21 | 9.3860 |
+| higgs_nvfp4_fp8_gptq | NVFP4 + FP8_DYNAMIC | GPTQModifier | 3.5s | 10:28 | 9.3760 |
+| higgs_nvfp4a16_fp8_model_free | NVFP4A16 + FP8_DYNAMIC | convert_checkpoint | 3.5s | 0:22 | 9.0381 |
```
brian-dellabetta, line 19: surprising to me that this outperforms GPTQ

src/llmcompressor/entrypoints/higgs/README.md
```diff
@@ -0,0 +1,168 @@
...
+- **Fusion-aware**: Respects hardware constraints (fused layers get same scheme)
+- **Flexible**: Supports any set of candidate quantization schemes
+
+## Quick Start
+
+```python
+from llmcompressor.transformers.compression.higgs import ilp_quantize
+
+# Quantize a model with automatic mixed-precision selection
+config = ilp_quantize(
```
brian-dellabetta, line 22: i really like this interface! really gives us a lot of flexibility.

one nit -- include higgs in the name rather than ilp, like `higgs_create_quant_config` or something?

src/llmcompressor/entrypoints/higgs/README.md
```diff
@@ -0,0 +1,168 @@
...
+    fusion_detector=detect_fused_groups,
+)
+
+# Process shards manually...
+# (see implementation in ilp_quantize() for details)
+
+optimal_config = collector.create_config()
+
+# Phase 2: Apply quantization
+quantizer = HiggsQuantizationConverter(
```
brian-dellabetta, line 145: i think this is effectively model-free-ptq? you pass in a quant config and it applies it. The work to convert model-free-ptq to the Converter interface is underway (cc @soyr-redhat )

setup.py
```diff
@@ -143,6 +143,9 @@ def localversion_func(version: ScmVersion) -> str:
         ),
     ],
     extras_require={
+        "higgs": [
```
brian-dellabetta, line 146: agree, makes sense as an extra

src/llmcompressor/entrypoints/higgs/README.md
```diff
@@ -0,0 +1,168 @@
...
+from llmcompressor.transformers.compression.higgs import (
+    HiggsMSECollectorConverter,
+    HiggsQuantizationConverter,
+    compute_heuristic_alphas,
+    detect_fused_groups,
+)
+from compressed_tensors.entrypoints.convert import convert_checkpoint
+
+# Phase 1: Collect MSE and solve ILP
+collector = HiggsMSECollectorConverter(
```
brian-dellabetta, line 132: still trying to figure how this fits in. You're contrasting this 2-phase approach to the well-lit `ilp_quantize` pathway?

src/llmcompressor/entrypoints/higgs/base.py
```diff
@@ -0,0 +1,544 @@
...
+    def get_dependencies(self, weight_name: str) -> set[str]:
+        return set()
+
+
+# ---------------------------------------------------------------------------
+# Phase 2: Quantization application
+# ---------------------------------------------------------------------------
+
+
+class HiggsQuantizationConverter(Converter):
```
brian-dellabetta, line 220: This looks to be the same as https://github.com/vllm-project/llm-compressor/pull/2935

src/llmcompressor/entrypoints/higgs/base.py
```diff
@@ -0,0 +1,544 @@
...
+        tensors = split_fused_moe_experts(tensors)
+        logger.info(f"Collecting MSE data from shard with {len(tensors)} tensors")
+
+        layers_processed = 0
+        for module_name, tensor_name in match_quantizable_tensors(
+            tensors, self.ignore, self.targets
+        ):
+            self.layer_sizes[module_name] = tensors[tensor_name].numel()
+            if module_name not in self.mse_matrix:
+                self.mse_matrix[module_name] = {}
```
brian-dellabetta, line 145: i see, so you're using it as a converter, but process is a pass-through just to set `self.mse_matrix`. Interesting

brian-dellabetta (commented):
This is cool! Just leaving some comments for now. I really like the UX for this and our idea to have it use convert_checkpoint looks pretty clean

src/llmcompressor/entrypoints/higgs/base.py
soyr-redhat: Yep! From what I can see, _match_tensors_to_schemes already handles the per-layer config groups the ILP emits, so once #2935 lands we can just do convert_checkpoint(ModelFreePtqConverter(config))!

examples/quantization_higgs/llama3_higgs_example.py
HDCharles: also because the humming ones are slow relative to 4/8 per bit.

examples/quantization_higgs/results_convert_vs_oneshot.md
HDCharles: it changed the eval numbers significantly

examples/quantization_higgs/results_method_comparison.md
HDCharles: yeah good point idk what happened here, i didn't notice this at the time.

src/llmcompressor/entrypoints/higgs/base.py
HDCharles: yeah i was trying to use as much existing scaffolding as possible

src/llmcompressor/entrypoints/higgs/base.py
HDCharles: note: split_fused_moe_experts i think was a necessary fix compared to model_free_ptq

src/llmcompressor/entrypoints/higgs/README.md
HDCharles: oh the readme is out of date, the PR description has the correct one, it should be get_higgs_config

src/llmcompressor/entrypoints/higgs/README.md
HDCharles: this is kind of what i meant above about the api and having 3 pieces but we could get rid of some of them if we want.

we basically just have 1 converter (HiggsMSECollectorConverter) that gets MSE, solves the ILP, generates a config, this is all of higgs and is theoretically the only thing we need for higgs

the rest is scaffolding, get_higgs_config basically just runs the converter if we have general scaffolding, we don't need it.
then as you mentioned HiggsQuantizationConverter is just model free PTQ but handles multiple configs and has some fixes i think.

so as of this PR i needed 3 pieces, it seems soon HiggsQuantizationConverter can go away, that just leaves get_higgs_config as something we can generalize/remove

src/llmcompressor/entrypoints/higgs/README.md
HDCharles: yes i think there are some fixes but its model free PTQ + handling multiple config groups + a few minor fixes

examples/quantization_higgs/results_convert_vs_oneshot.md
brian-dellabetta: Yes, this actually matters, my understanding was wrong. config groups are sorted by key before saving, so if it rearranges it can change the order of target resolution to match a given layer to scheme. Creating config group keys that are already alphabetized is important. More information at https://github.com/vllm-project/compressed-tensors/issues/862#issuecomment-5531118902

src/llmcompressor/entrypoints/higgs/README.md
brian-dellabetta: ok cool. Model free PTQ should now allow for multiple configs

examples/quantization_higgs/higgs_nvfp4_fp8_gptq.py
```diff
@@ -0,0 +1,116 @@
...
+    ds = ds.map(tokenize, remove_columns=ds.column_names)
+
+    recipe = GPTQModifier(
+        config_groups=config.config_groups,
+        ignore=config.ignore,
+    )
+
+    oneshot(
+        model=model,
+        dataset=ds,
```
kylesayrs: Consider using perfectblend

src/llmcompressor/entrypoints/higgs/base.py
```diff
@@ -0,0 +1,332 @@
...
+        tensors = load_tensors_from_inverse_weight_map(
+            inverse_weight_maps[shard_name], device
+        )
+        collector.process(tensors)
+
+    mse_time = time.time() - mse_start
+    logger.info(f"MSE collection completed in {mse_time:.2f}s")
+
+    ilp_start = time.time()
+    optimal_config = collector.create_config()
```
HDCharles, line 324: key line: make config from collected info

src/llmcompressor/entrypoints/higgs/base.py
```diff
@@ -0,0 +1,332 @@
...
+    for shard_name in shard_names:
+        if shard_name not in inverse_weight_maps:
+            logger.warning(f"Shard {shard_name} not in inverse_weight_maps, skipping")
+            continue
+
+        logger.info(f"Processing shard: {shard_name}")
+        tensors = load_tensors_from_inverse_weight_map(
+            inverse_weight_maps[shard_name], device
+        )
+        collector.process(tensors)
```
HDCharles, line 318: key line: where MSE's are calculated/collected

src/llmcompressor/entrypoints/higgs/ilp_solver.py
```diff
@@ -0,0 +1,261 @@
...
+        Dict mapping layer_name -> selected_scheme_name
+
+    Raises:
+        ValueError: If ILP is infeasible or unbounded
+    """
+    if pulp is None:
+        raise ImportError(
+            "HIGGS ILP solver requires the 'pulp' package. "
+            "Install it with: pip install pulp"
+        )
```
kylesayrs:
```suggestion
    if pulp is None:
        raise ImportError(
            "HIGGS ILP solver requires the 'pulp' package. "
            "Install with: pip install llmcompressor[higgs]"
        )
```

examples/quantization_higgs/higgs_nvfp4a16_fp8_model_free.py
```diff
@@ -0,0 +1,73 @@
...
+
+1. get_higgs_config(): model-free ILP selects optimal per-layer schemes
+2. model_free_ptq(): applies the generated mixed-precision config
+   directly to safetensors without loading the full model into GPU memory
+
+NVFP4A16 (W4A16) and FP8_DYNAMIC are data-free schemes, so no calibration
+dataset is needed. This path is faster and uses less memory.
+
+Usage:
+    python higgs_nvfp4a16_fp8_convert.py \
```
soyr-redhat: I think this name needs updated

src/llmcompressor/entrypoints/higgs/base.py
```diff
@@ -0,0 +1,334 @@
...
+                    )
+                else:
+                    key = f"scheme_{len(resolved)}"
+                resolved[key] = scheme
+            else:
+                raise ValueError(f"Invalid scheme type: {type(scheme)}")
+        return resolved
+
+    def validate(self, tensors: Dict[str, torch.Tensor]):
+        """Check that there are quantizable tensors in this shard. Note: this is currently
```
soyr-redhat: pretty sure this is over the line len limit

src/llmcompressor/entrypoints/higgs/utils.py
```diff
@@ -0,0 +1,251 @@
...
+        initialize_module_for_quantization(module, scheme, force_zero_point=False)
+        initialize_observer(module, "weight")
+        apply_calibration_status(module)
+        observe(module, base_name="weight")
+        update_qparams(module, base_name="weight")
+        freeze_module_quantization(module)
+
+        scale = module.weight_scale
+        zp = getattr(module, "weight_zero_point", torch.zeros_like(scale))
+        quantized = fake_quantize(
```
soyr-redhat: Not sure if this is intended, but [ct ](https://github.com/vllm-project/compressed-tensors/blob/6d852ba7ef22b7b212c4f9eea6ad8d423e779e73/src/compressed_tensors/quantization/lifecycle/forward_helpers.py#L182-L202) computes the effective block scale as 
`weight_scale / weight_global_scale`
when global_scale is passed. 

Here we only pass weight_scale to fake_quantize, so [TENSOR_GROUP](https://github.com/vllm-project/compressed-tensors/blob/6d852ba7ef22b7b212c4f9eea6ad8d423e779e73/src/compressed_tensors/quantization/lifecycle/initialize.py#L184-L190) scale MSE appears to use weight_scale directly.

I would imagine that might mess things up right?

soyr-redhat (commented):
just a couple of tiny things + one question. Excited for this!

PR #3029, opened 2026-08-14
[Feature] Add block-wise QAD with shared data and weight re-observation

BeichenHuang:
## Summary

Adds block-wise local quantization-aware distillation after a preceding weight quantization method. During the existing calibration forward, QAD hooks cache a decoder block's inputs and unquantized outputs. At `sequential_epoch_end`, the preceding method prepares quantization, QAD reconstructs the cached outputs, and the existing pipeline propagates the final quantized outputs.

- Share one calibration dataset and dataloader. No changes to the pipeline, public lifecycle, Modifier base class, or oneshot entrypoint.
- Compose RTN (`QuantizationModifier`) or `GPTQModifier` followed by `QADModifier`. The tested AWQ composition is `AWQModifier` → `QuantizationModifier` → `QADModifier`; AWQ's transform still needs a weight quantization modifier. Other methods use the same lifecycle contract; QAD does not inspect GPTQ-specific state or quantizer class names.
- Follow the weight re-observation schedule in #3051: before training, after every epoch, and before final materialization. Observe all modules before updating any qparams to preserve fused NVFP4 global scales.
- Retain masked MSE, FP32 optimizer master weights for FP16/BF16 execution, dynamic FP16 loss scaling, gradient accumulation/clipping, deterministic validation, early stopping, and best-checkpoint restoration. Restore matching weight qparams with the best weights. `reobserve_weights=False` retains a fixed-scale option.
- Use one complete target module per subgraph, local teacher targets, single-process calibration, and `propagate_error=True`. Ignore the output LM head. Full teacher streams, separate QAD datasets, and arbitrary multi-target graph reconstruction are outside this version.
- Keep `LayerwiseQADModifier` as an alias. Usage and quantizer composition requirements are documented in `examples/qad/`.

## Validation

- CPU: QAD tests plus `tests/llmcompressor/core/test_session_functions.py` — 39 passed, 4 accelerator cases skipped. Regression cases cover long-sequence FP16 gradient underflow, unscaling before clipping, complete/partial accumulation groups, overflow retries, and persistent nonfinite-gradient failure. The experiment training/export/reload entrypoint also passes a tiny CPU check with source/input provenance validation.
- H200 on the corrected source: 41 passed, 2 CPU-only cases skipped. Covers all QAD tests, including FP16 long-sequence gradients, overflow recovery, exact teacher-input propagation, RTN/GPTQ/AWQ+RTN composition with NVFP4A16/W4A16, and RTN/GPTQ FP16 NVFP4 save/reload/generation.
- Changed Python files pass Ruff lint/format and CUDA portability checks. Full `make quality` reaches two existing import-order errors in unchanged sparsity tests.
- An additional check on main `8ada4c40` with compressed-tensors `698967f` encounters a `weight_global_scale` Tensor/Parameter error in the dependency's post-save offload conversion; the same error reproduces with GPTQ alone. The existing experiment environment passes save/reload.
- Paired Llama-3.1-8B-Instruct experiments use shared UltraChat budgets 512/1024/2048, evaluation seeds 1234/2345/3456, five benchmarks, and WikiText-2 PPL (context 2048, stride 512). All six checkpoints passed quantization, export and reload validation; 18/18 full evaluations are verified, including all nine complete GPTQ/QAD budget/seed pairs. Each condition has one trained checkpoint (training seed 42); the three seeds are evaluation repeats.
- The three GPTQ-only checkpoints and nine baseline evaluations are reused. Comparing archived runtime sources confirms that only QAD training code changed. Model/data revisions, calibration token IDs, hyperparameters and the evaluator are pinned. Accepted corrected QAD evaluations pass the same complete protocol, sample and paired-input checks.

## Corrected 8B results — 512, 1024 and 2048 samples

**All six checkpoints and 18 full evaluations are verified.** These results use the corrected FP16 loss-scaling implementation in `cdf35548` and replace the earlier, pre-correction QAD measurements. Llama-3.1-8B-Instruct, NVFP4A16, shared nested UltraChat calibration data, training seed 42. Entries are mean ± sample standard deviation over evaluation seeds 1234/2345/3456, using one saved checkpoint per condition.

| Samples | Method | GSM8K (%) | MMLU CoT (%) | MMLU-Pro (%) | IFEval instruction strict (%) | MATH-500 (%) | WikiText-2 PPL ↓ |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 512 | GPTQ | 85.829 ± 1.034 | 71.312 ± 0.032 | 44.401 ± 0.211 | 80.416 ± 0.591 | 46.933 ± 2.043 | 6.683308 ± 0.000000 |
| 512 | GPTQ + QAD | 85.360 ± 0.975 | 71.322 ± 0.108 | 43.645 ± 0.317 | 81.535 ± 0.729 | 46.333 ± 0.577 | 6.680964 ± 0.000000 |
| 1024 | GPTQ | 86.270 ± 0.573 | 71.108 ± 0.199 | 44.354 ± 0.349 | 80.216 ± 1.568 | 43.800 ± 1.039 | 6.664566 ± 0.000000 |
| 1024 | GPTQ + QAD | 86.187 ± 0.414 | 71.032 ± 0.029 | 44.637 ± 0.331 | 81.175 ± 1.144 | 46.333 ± 1.419 | 6.653724 ± 0.000000 |
| 2048 | GPTQ | 85.911 ± 0.563 | 71.027 ± 0.248 | 44.656 ± 0.274 | 81.335 ± 0.916 | 46.733 ± 1.301 | 6.648578 ± 0.000000 |
| 2048 | GPTQ + QAD | 85.580 ± 0.191 | 71.524 ± 0.201 | 44.631 ± 0.172 | 80.096 ± 1.024 | 46.200 ± 0.346 | 6.665703 ± 0.000000 |

Positive paired improvement means higher accuracy in percentage points or lower absolute PPL.

| Samples | GSM8K | MMLU CoT | MMLU-Pro | IFEval | MATH-500 | PPL decrease |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 512 | -0.469 | +0.009 | -0.756 | +1.119 | -0.600 | +0.002345 |
| 1024 | -0.083 | -0.076 | +0.283 | +0.959 | +2.533 | +0.010843 |
| 2048 | -0.331 | +0.496 | -0.025 | -1.239 | -0.533 | -0.017126 |

Results are mixed. At 1024 samples, QAD improves MMLU-Pro, IFEval, MATH-500 and PPL means, with small decreases in GSM8K and MMLU CoT. At 512 samples, IFEval improves but MMLU-Pro and MATH-500 decrease; at 2048, MMLU CoT improves while IFEval and PPL worsen. These measurements do not establish a consistent quality improvement or robustness across training seeds.

All eighteen evaluations passed protocol, checkpoint and sample validation: 28,324 unique task documents, 29,533 scoring records and 75 task/filter groups each. All nine GPTQ/QAD budget/seed pairs have identical document, prompt and target hashes for every scoring record. WikiText-2 scores exactly 289,076 target tokens with context 2048 and stride 512; its repeated scores are identical for each checkpoint under this deterministic scoring procedure.

Evaluation tasks are `gsm8k_platinum_cot_llama`, `mmlu_cot_llama`, `mmlu_pro_chat`, `ifeval` (instruction-level strict accuracy), and the documented custom MATH-500 task adaptation. Generation uses the same vLLM 0.25.1 environment, temperature 0.6, top-p 0.9 and maximum 4096 generated tokens for every condition. H100/H200 evaluation workers are mixed.

### Resource usage

| Samples | Method | Quantize/save/reload (min) | Quantization GPU allocated peak (GiB) | Worker host sampled peak (GiB) |
| --- | --- | ---: | ---: | ---: |
| 512 | GPTQ | 18.8 | 19.67 | 8.99 |
| 512 | GPTQ + QAD | 64.9 | 21.15 | 20.73 |
| 1024 | GPTQ | 26.8 | 19.67 | 15.02 |
| 1024 | GPTQ + QAD | 114.1 | 21.15 | 36.59 |
| 2048 | GPTQ | 41.4 | 19.67 | 24.90 |
| 2048 | GPTQ + QAD | 227.8 | 21.15 | 68.55 |

All six completed quantization workers used H200 GPUs. Times include model loading, saving and reload validation and reflect separate scheduled runs, not controlled speed measurements. Host RSS sampling can double-count shared pages or miss brief peaks. GPU allocated peaks use PyTorch counters through quantization and exclude saving/reloading. Each exported model archive is approximately 5.63 GiB.

Three evaluations required recovery after earlier processes printed scores but did not exit. The recovery retains the frozen evaluator body, model, data, seeds, generation and scoring, and appends bounded cleanup after all scientific outputs are written. All three recovered evaluations exited successfully, returned from vLLM shutdown, and passed the complete validation and paired-input checks. Incomplete and untransferred earlier scores are excluded.


examples/layerwise_qad/README.md
```diff
@@ -0,0 +1,53 @@
...
+      gradient_accumulation_steps: 4
+      max_grad_norm: 1.0
+      validation_fraction: 0.1
+      early_stopping_patience: 3
+      validation_relative_min_delta: 0.001
+"""
+
+model = oneshot(
+    model="student-model-or-path",
+    distill_teacher="full-precision-teacher-model-or-path",
```
HDCharles: don't think we need this

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -24,6 +24,34 @@
 __all__ = ["SequentialPipeline"]
 
 
+def _configure_modifier_pipeline(modifiers, dataset_args):
```
HDCharles: i'm unsure why these pipeline changes are necessary can you explain?

examples/layerwise_qad/README.md
```diff
@@ -0,0 +1,53 @@
+# Layerwise MSE QAD
+
+`LayerwiseQADModifier` performs block-local quantization-aware distillation.
```
HDCharles: maybe just QAD modifier, this makes it sound like it works on one linear layer at a time, (ideally this would work on a whole subgraph forward/backward)

src/llmcompressor/modifiers/layerwise_qad/base.py
```diff
@@ -0,0 +1,725 @@
...
+        self._capture_enabled = True
+        for module in self._student_to_teacher:
+            self.register_hook(
+                module,
+                self._capture_input,
+                "forward_pre",
+                with_kwargs=True,
+            )
+
+    def _capture_input(self, module, args, kwargs):
```
HDCharles: lets try to use the same architecture as AWQ which also has the pattern of recording inputs to the subgraph and then running the subgraph. Since we're hoping to do subgraph gradients rather than layerwise we don't need intermediates for every layer, just the whole subgraph. There's a lot of features like prefetching and offloading and stuff that this can have for free.

examples/layerwise_qad/README.md
brian-dellabetta: Yeah, wouldn't we just always use the passed in `model` as the teacher? we would run forward passes through the subgraph (with quantization disabled) and the subgraph with quantization enabled, and the loss function would be the MSE difference?

examples/layerwise_qad/README.md
brian-dellabetta: Agree, but awesome that we can do this entirely through modifiers and the sequential pipeline

examples/layerwise_qad/README.md
BeichenHuang: Agreed. The updated implementation no longer requires a separate teacher model. Before quantizing each subgraph, it uses the original weights with quantization disabled to cache the teacher targets. I also removed the distill_teacher argument from oneshot.

examples/layerwise_qad/README.md
BeichenHuang: Renamed this to QADModifier, with the old name retained as a compatibility alias. It now jointly optimizes the quantized weights within the traced subgraph, including subgraphs containing multiple sequential targets, branches, and residual connections.

src/llmcompressor/modifiers/layerwise_qad/base.py
BeichenHuang: I refactored QAD to consume the traced subgraph and the pipeline’s IntermediatesCache, replacing the decoder-block input hooks. It now reconstructs subgraph outputs without collecting per-linear intermediate targets. The QAD replay loop still caches inputs and teacher targets itself and does not yet use the cache’s prefetch iterator, so that part could be improved further.

src/llmcompressor/pipelines/sequential/pipeline.py
BeichenHuang: I revisited this and removed the QAD-specific checks and the one-block-per-subgraph restriction from the pipeline. Teacher capture and optimization now live in QADModifier.

src/llmcompressor/core/events/event.py
```diff
@@ -50,6 +51,7 @@ class EventType(Enum):
 
     # calibration lifecycle
     CALIBRATION_START = "calibration_start"
+    SEQUENTIAL_EPOCH_START = "sequential_epoch_start"
```
HDCharles, line 54: why is this needed?

HDCharles: changing the lifecycle is a big step and its not clear why this is needed in general and what alternatives there are to get around this, it seems like its making the fx graph accessible but its unclear what problems that solves compared to what i'm doing in my PR

src/llmcompressor/modifiers/qad/base.py
```diff
@@ -0,0 +1,627 @@
...
+
+def _quantized_modules(modules):
+    return [
+        module
+        for module in dict.fromkeys(modules)
+        if getattr_chain(module, "quantization_scheme.weights", None) is not None
+    ]
+
+
+def _distillation_graph(model, subgraph, quantized_modules, include_propagation=False):
```
HDCharles, line 99: whats going on here? why is this necesssary?

src/llmcompressor/modifiers/qad/base.py
```diff
@@ -0,0 +1,627 @@
...
+
+    ``num_epochs`` and early stopping apply separately to each subgraph. Cached
+    batches are split into training/validation sets. By default these are the PTQ
+    calibration batches. Pass ``qad_dataset`` and ``qad_dataset_args`` to oneshot
+    to configure an independent stream, including its microbatch size. Gradient
+    accumulation combines microbatches.
+    """
+
+    requires_calibration_data: bool = True
+    teacher_mode: Literal["local", "full"] = "local"
```
HDCharles, line 186: whats the impact of this?

HDCharles:
still probably need to do re-observe at the end as far as activation qparams

HDCharles:
do you have any ablations to show the impact of things like the train/validation split, teacher mode, qad specific data?

in general i'd like this to be self contained and i feel like each of those things except maybe qad_data can be. But i'm not sure if qad data matters tbh, you are just going to do gptq -> qad or quant -> qad and in either case the bottleneck is QAD, i don't think it'd make much difference to just put the whole QAD set into oneshot and run it through gptq.

PR #3034, opened 2026-08-16
Update What's New with Nemotron 3.5 Lightning and Kimi-K3 checkpoints

dsikka:
- Remove old entries as well


examples/quantization_w8a8_fp8/nemotron_3_5_lightning_example.py
```diff
@@ -0,0 +1,86 @@
...
+DATASET_ID = "HuggingFaceH4/ultrachat_200k"
+DATASET_SPLIT = "train_sft"
+
+# Select number of samples. 512 samples is a good place to start.
+# Increasing the number of samples can improve accuracy.
+NUM_CALIBRATION_SAMPLES = 512
+MAX_SEQUENCE_LENGTH = 2048
+
+# Load dataset and preprocess.
+ds = load_dataset(DATASET_ID, split=f"{DATASET_SPLIT}[:{NUM_CALIBRATION_SAMPLES}]")
```
HDCharles, line 24: we should standard to always use get_rank_partition probably just to avoid footguns for people that try to move to DDP.

e.g.

https://github.com/vllm-project/llm-compressor/blob/d9cd9885d9f5e606e3a1f87159dc7e015224cad1/examples/awq/llama_example_ddp.py#L39-L42

this will work in DDP and not DDP

HDCharles (approved):

PR #3045, opened 2026-08-18
[REAP] DDP Support

kylesayrs:
## Purpose ##
* Support DDP REAP
* Support better logging for REAPed experts

## Changes ##
* Reduce saliency stats across ranks before computing on rank 0. Broadcast rank 0's result
  * While doing the computation on each rank would avoid broadcasting, it also increases the risk that numerical precision leads to rank divergence

## Testing ##
* WIP: testing reap qwen3.8 model
* Added (DDP is same as non-ddp) test


src/llmcompressor/modifiers/pruning/reap/base.py
```diff
@@ -60,6 +70,7 @@ class REAPPruningModifier(Modifier):
 
     sparsity: float
     ignore: list[str] = Field(default_factory=list)
+    report_path: Optional[str] = Field(default=None)
```
brian-dellabetta, line 73: can't this just be tracked in-memory and shared between processes with a broadcast? or are we constrained to file because that is that not supported for arbitrary dicts?

kylesayrs: What's the value of sharing between processes? This is just a user debug tool, only one rank needs to write to disk.

brian-dellabetta: nm, I thought this was needed for DDP

brian-dellabetta (approved):

PR #3046, opened 2026-08-18
[Autowrapper] Skip dead branches when tracking local names

kylesayrs:
## Summary
- When an `if`/`elif` condition resolves to a constant, the autowrapper now only visits the live branch for name tracking. Previously, names from dead branches (e.g. `_` in `elif False: x, _ = self.self_attn(...)`) were incorrectly added to `_local_names`, causing the autowrapper to generate wrapper signatures that reference unbound variables — leading to `UnboundLocalError` at trace time.
- Adds a regression test covering the dead-branch name leakage scenario.

## Test plan
- [x] New unit test `test_dead_branch_names_not_tracked` verifies the fix
- [x] Run existing autowrapper test suite: `pytest tests/llmcompressor/pipelines/sequential/ast_utils.py/test_auto_wrapper.py`


HDCharles (approved):
makes sense

PR #3050, opened 2026-08-18
Extend ARCH_TO_2D_MAPPINGS to Qwen3.5-MoE

### LINKED ISSUE #3037: Qwen3.5-MoE misses ARCH_TO_2D_MAPPINGS, forcing a 2D→3D→2D expert round-trip that doubles expert memory at load
## Summary

For Qwen3.5-MoE, `has_linearize_load_mappings()` returns `False`, so `load_quantizable_moe` takes its fallback branch: load the model normally, then call `linearize_moe()`. Because the checkpoint **already stores experts as 2D per-expert tensors**, and transformers **fuses them to 3D on load**, that fallback performs a full 2D → 3D → 2D round-trip and **leaves one extra copy of every expert resident in RSS for the lifetime of the process**.

For a 35B-A3B Qwen3.5-MoE (256 experts × 40 layers, 32.21B expert params = **60.0 GiB** in bf16), the round-trip needs ~120 GiB for experts plus ~6 GiB for everything else ≈ **126 GiB**. On a 124 GB host that is fatal, and it is fatal **during loading — before calibration ever starts**, so no reduction in `num_calibration_samples` or `max_seq_length` avoids it. It presents as a bare `Out of memory: Killed` with anon-rss ~118 GB in `journalctl -k`.

`linearize_moe()` already logs a warning about exactly this: *"this may be inefficient if the model checkpoint is already linearized (2D -> 3D -> 2D)."* On a host where 2× the expert weights exceeds RAM, that is not a speed warning — it is a fatal memory warning.

The library already documents the pathway this model should be taking — `llmcompressor/modeling/moe/linearize.py`, `load_quantizable_moe`:

> Two loading pathways are supported:
> 1. **Direct loading**: If the model checkpoint contains 2D weights and conversion mappings are registered for the model type, weights are loaded directly in linearized format.
> 2. **Post-load conversion**: If no conversion mappings exist, the model is loaded normally and then linearized via `linearize_moe`.

Qwen3.5-MoE satisfies the *substance* of pathway 1 — its checkpoints store 2D per-expert tensors — and is denied it only by a missing dictionary key, so it takes the expensive pathway for a checkpoint that never needed converting.

## Why the lookup misses

```python
def has_linearize_load_mappings(model_type: str) -> bool:
    remapped_type = _MODEL_TO_CONVERSION_PATTERN.get(model_type, model_type)
    return model_type in ARCH_TO_IMPORT_PATHS and remapped_type in ARCH_TO_2D_MAPPINGS
```

Evaluated against `llmcompressor` 0.13.0 / `transformers` 5.14.1:

| model_type | in `ARCH_TO_IMPORT_PATHS` | remaps to | remap in `ARCH_TO_2D_MAPPINGS` | result |
|---|---|---|---|---|
| `qwen2_moe` | True | `qwen2_moe` | True | **True** |
| `qwen3_moe` | True | `qwen2_moe` | True | **True** |
| `qwen3_next` | True | `qwen2_moe` | True | **True** |
| `qwen3_vl_moe` | True | `qwen3_vl_moe` | False | **False** |
| `qwen3_5_moe` | True | `qwen3_5_moe` | False | **False** |
| `qwen3_5_moe_text` | **False** | `qwen3_5_text` | False | **False** |

`ARCH_TO_2D_MAPPINGS` on `main` today holds only `deepseek_v4`, `qwen2_moe`, `hy_v3`. `qwen3_moe` and `qwen3_next` reach it by remapping onto `qwen2_moe`.

Two Qwen arches miss the table, but **only Qwen3.5-MoE is harmed by it**, and the distinction matters: `qwen3_vl_moe`'s conversion rules are *identity* (`['mlp.experts.gate_up_proj'] -> ['mlp.experts.gate_up_proj']`), i.e. its checkpoints are already 3D, so there is no 2D → 3D fusion to undo and no round-trip. Qwen3.5-MoE's rules genuinely fuse, so it is the one arch that pays a full extra copy.

## The fusion rules are identical to `qwen2_moe`

`get_checkpoint_conversion_mapping()`:

```
qwen3_5_moe_text:
  WeightConverter: ['mlp.experts.*.gate_proj.weight', 'mlp.experts.*.up_proj.weight'] -> ['mlp.experts.gate_up_proj']
  WeightConverter: ['mlp.experts.*.down_proj.weight']                                 -> ['mlp.experts.down_proj']

qwen2_moe:
  WeightConverter: ['mlp.experts.*.gate_proj.weight', 'mlp.experts.*.up_proj.weight'] -> ['mlp.experts.gate_up_proj']
  WeightConverter: ['mlp.experts.*.down_proj.weight']                                 -> ['mlp.experts.down_proj']
```

Byte-for-byte the same, so the existing `ARCH_TO_2D_MAPPINGS["qwen2_moe"]` mapping *body* should apply to Qwen3.5-MoE unchanged.

## Measured

The reproducer below builds a scale model in-process — 8 layers × 64 experts, 1.50 GiB of experts — so no checkpoint download is needed, and it walks the GC for live torch storages rather than trusting RSS. On `llmcompressor` 0.13.0:

| | live tensor storage | RSS |
|---|---|---|
| after build | 1.50 GiB | 0.97 GiB |
| after `linearize_moe()` | **1.50 GiB** | **2.48 GiB** |
| after `malloc_trim(0)` | 1.50 GiB | 2.48 GiB |

RSS grows by **exactly one full copy of the experts** and never comes back, while *live* storage never moves. So this is **not a reference leak** — the 3D tensors are genuinely dropped, but their pages are retained by the allocator and never returned to the OS, and `malloc_trim(0)` does not recover them.

### What does not fix it

Dropping references eagerly inside the linearize loop — popping from the list, `del`, `gc.collect()` per layer — measures **identical** peak RSS (2.48 GiB). The reproducer's `streaming` variant does exactly that, so it is easy to confirm. Worth knowing before anyone spends time there: the fix is to avoid the round-trip, not to free faster.

### Current workaround

Forcing weight offloading during load (`device_map` with an offload folder and a low per-device memory cap) works, but costs substantial wall-clock, and should not be necessary for a checkpoint that is *already* in the target layout.

## Which spelling arrives, and why this may not be a one-line fix

The call site uses the **top-level** config type:

```python
config = AutoConfig.from_pretrained(*args, **kwargs)
model_type = config.model_type
if not has_linearize_load_mappings(model_type):
    model = original_from_pretrained(*args, **kwargs)
    linearize_moe(model)      # <- the expensive path
    return model
```

Both released checkpoint shapes miss, on *different* clauses:

| checkpoint | top-level `model_type` | `text_config.model_type` | fails on |
|---|---|---|---|
| multimodal (`Qwen3_5MoeForConditionalGeneration`) | `qwen3_5_moe` | `qwen3_5_moe_text` | clause 2 — not in `ARCH_TO_2D_MAPPINGS` |
| text-only (`Qwen3_5MoeForCausalLM`) | `qwen3_5_moe_text` | – | clause 1 — not in `ARCH_TO_IMPORT_PATHS` |

I first assumed adding `ARCH_TO_2D_MAPPINGS["qwen3_5_moe"]` would be enough. It probably is not: `get_linearize_load_mappings()` calls `get_checkpoint_conversion_mapping(model_type)`, and that returns **`None`** for `qwen3_5_moe` — transformers registers the rules under `qwen3_5_moe_text` (itself remapping to `qwen3_5_text`). So the two libraries key this model at different levels: llm-compressor at the top-level wrapper, transformers at the text tower. Adding the 2D entry alone would satisfy the predicate and then hand `None` to the mapping consumer.

Reconciling that is a design choice I would rather not guess at. Options appear to be resolving `text_config.model_type` when the top-level type has no conversion mapping, registering both spellings, or normalizing the `_text` suffix inside `has_linearize_load_mappings()`. Happy to send a PR once you indicate which convention you prefer — the mapping body itself can be copied from `qwen2_moe`.

## Prior art in this repo

- **#2482** ("[Qwen3.5] Calibration support and NVFP4 Example") added Qwen3.5 support, and its summary notes it works by *"specifically unstack[ing] 3D weights"* — i.e. via pathway 2. That is what makes the model quantizable at all, and is why this is a memory/performance issue rather than a correctness one. It simply never got a mapping-table entry.
- **#2739** ("[deepseek_v4] Extend `ARCH_TO_2D_MAPPINGS` for MTP block") establishes this fix shape — extending `ARCH_TO_2D_MAPPINGS` for an architecture the existing entries missed. This request is the same change for a different arch.
- **#2847 / #2848** broadened MoE architecture support; **#2939** reports a separate Qwen3.6-A3B problem (silent corruption of fused-expert MoE — a correctness bug, not this one).

Also flagging in case it affects the fix: #2739 found `ARCH_TO_2D_MAPPINGS` regexes anchored at `^layers\.`, which excludes an `mtp.` prefixed block. Qwen3.5-MoE checkpoints can carry an `mtp.layers.0.mlp.experts.*` block too, so a new entry should probably cover both prefixes rather than repeat that gap.

## Environment

- `llmcompressor` 0.13.0, `transformers` 5.14.1, torch 2.11.0
- verified the same `ARCH_TO_2D_MAPPINGS` keys (`deepseek_v4`, `qwen2_moe`, `hy_v3`) on `main` today
- 124 GB host RAM. **The OOM is host RAM, not VRAM**, so this is not accelerator-specific — any host where 2× the expert weights exceeds RAM will hit it.

## Reproducer

<details>
<summary><code>linearize_probe.py</code> — no checkpoint required</summary>

```python
"""Reproducer: llm-compressor's linearize_moe fallback doubles expert RSS at load.

Qwen3.5-MoE misses ARCH_TO_2D_MAPPINGS, so load_quantizable_moe falls back to
"load normally, then linearize_moe()". The checkpoint already stores experts as 2D
per-expert tensors and transformers fuses them to 3D on load, so that fallback is a
2D -> 3D -> 2D round-trip and leaves one extra copy of every expert in RSS forever.

Walks the GC for live torch storages rather than trusting RSS, to distinguish a
reference leak (live storage grows) from allocator retention (it does not).

  usage: python linearize_probe.py {upstream|streaming}

    upstream  - call linearize_moe() as llm-compressor does
    streaming - drop references eagerly per layer. Measures IDENTICAL peak RSS.
"""
import ctypes, gc, resource, sys
import torch

VARIANT = sys.argv[1]

from transformers.models.qwen3_5_moe.configuration_qwen3_5_moe import Qwen3_5MoeTextConfig
from transformers.models.qwen3_5_moe.modeling_qwen3_5_moe import Qwen3_5MoeExperts
from llmcompressor.modeling.moe.linearize import linearize_moe, get_non_linearized_moes
from llmcompressor.modeling.moe.linear_experts import LinearExperts2D

N_LAYERS, N_EXPERTS, HIDDEN, INTER = 8, 64, 1024, 512
cfg = Qwen3_5MoeTextConfig(
    hidden_size=HIDDEN, moe_intermediate_size=INTER, num_experts=N_EXPERTS,
    num_experts_per_tok=8, num_hidden_layers=N_LAYERS, intermediate_size=INTER,
    vocab_size=1000, dtype=torch.bfloat16,
)

def rss_gb():
    with open("/proc/self/status") as f:
        for line in f:
            if line.startswith("VmRSS:"):
                return int(line.split()[1]) / 1024**2

def live_tensor_gb():
    """Sum unique untyped storages of every live tensor the GC can see."""
    gc.collect()
    seen, total = set(), 0
    for o in gc.get_objects():
        try:
            if torch.is_tensor(o):
                s = o.untyped_storage()
                if s.data_ptr() not in seen:
                    seen.add(s.data_ptr()); total += s.nbytes()
        except Exception:
            pass
    return total / 2**30

def trim():
    try: ctypes.CDLL("libc.so.6").malloc_trim(0)
    except Exception: pass

class Container(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.config = cfg
        self.layers = torch.nn.ModuleList()
        for _ in range(N_LAYERS):
            m = torch.nn.Module(); m.experts = Qwen3_5MoeExperts(cfg)
            self.layers.append(m)

torch.set_default_dtype(torch.bfloat16)
model = Container()
print(f"variant={VARIANT}")
print(f"  before : live_tensors={live_tensor_gb():.2f} GiB  rss={rss_gb():.2f} GiB")

def linearize_streaming(model):
    moes = get_non_linearized_moes(model)
    for i in range(len(moes)):
        name, module = moes[i]
        moes[i] = (None, None)
        conf = getattr(module, "config", model.config)
        cls_ = LinearExperts2D.get_linear_experts_cls(module.__class__)
        linear = cls_.from_experts_module(module, conf)
        model.set_submodule(name, linear)
        del module, linear
        gc.collect()
    return model

if VARIANT == "upstream":
    linearize_moe(model)
elif VARIANT == "streaming":
    linearize_streaming(model)

print(f"  after  : live_tensors={live_tensor_gb():.2f} GiB  rss={rss_gb():.2f} GiB")
trim()
print(f"  +trim  : live_tensors={live_tensor_gb():.2f} GiB  rss={rss_gb():.2f} GiB")
print(f"  PEAK RSS = {resource.getrusage(resource.RUSAGE_SELF).ru_maxrss/1024**2:.2f} GiB")
```

</details>

kgt392:
SUMMARY:

Fixes #3037. Qwen3.5-MoE checkpoints already store 2D per-expert tensors, but `has_linearize_load_mappings` returned False for them, so `load_quantizable_moe` fell back to loading normally and calling `linearize_moe` — a 2D → 3D → 2D round trip that leaves an extra copy of every expert resident for the process lifetime. On a memory-constrained host that is fatal during load, before calibration starts.

Registers both released spellings, reusing the Qwen2-MoE 2D body (its fusing rules are identical):

- `qwen3_5_moe` — multimodal wrapper (`Qwen3_5MoeForConditionalGeneration`)
- `qwen3_5_text` — what `_MODEL_TO_CONVERSION_PATTERN` remaps the text-only checkpoint's `qwen3_5_moe_text` to, plus its `ARCH_TO_IMPORT_PATHS` entry

The issue notes that adding a 2D entry alone is not sufficient, and that is confirmed: transformers registers the rules on the text tower, so `get_checkpoint_conversion_mapping("qwen3_5_moe")` returns `None` while llm-compressor keys on the top-level type. Of the three options raised in the issue this takes the "resolve the text spelling" one, as the least invasive — `_resolve_checkpoint_conversion_mapping` falls back to `f"{model_type}_text"`, and `has_linearize_load_mappings` now additionally requires a resolvable mapping, so a future 2D entry cannot satisfy the predicate and then hand `None` to `get_linearize_load_mappings`. Happy to switch to registering both spellings or normalising the suffix if you prefer a different convention.

One thing worth flagging: taking the mapping from the text tower also preserves its `model.language_model.*` prefix rule, which the Qwen2-MoE rules do not carry — copying that body verbatim would have silently dropped it. There is a test pinning it.

TEST PLAN:

New CPU-only unit tests in `tests/llmcompressor/modeling/test_moe_conversion_mappings.py` (15 cases, transformers 5.15.0): the predicate for both Qwen3.5-MoE spellings, no `WeightConverter` surviving in the load/save mappings (a survivor means fusion on load, i.e. the round trip), the per-expert renames, the `language_model` prefix rule, the shared 2D body, and guards that `qwen2_moe` / `qwen3_moe` / `qwen3_next` / `deepseek_v4` / `hy_v3` still resolve while `qwen3_vl_moe` keeps using post-load conversion (its rules are identity, so it has no round trip to avoid).

- 15 passed with the change; 9 of them fail on `main`, and the 6 that pass either way are the regression guards.
- Swept `has_linearize_load_mappings` over all 32 `ARCH_TO_IMPORT_PATHS` keys before and after: exactly `qwen3_5_moe` and `qwen3_5_moe_text` flip to True, no architecture regresses.
- `ruff check` and `ruff format --check` clean on both files.
- Not verified: an end-to-end load of a real Qwen3.5-MoE checkpoint, which needs hardware I do not have. The memory claim in the issue is the reporter's measurement, not mine — what I verified is the mapping resolution.

### DIFF
--- a/src/llmcompressor/modeling/moe/conversion_mappings.py
+++ b/src/llmcompressor/modeling/moe/conversion_mappings.py
@@ -270,7 +270,13 @@ def get_linearize_load_mappings(
     _config_paths, expert_paths = ARCH_TO_IMPORT_PATHS[model_type]
     experts_cls = import_or_none(expert_paths)
 
-    mapping: list[WeightTransform] = _resolve_checkpoint_conversion_mapping(model_type)
+    mapping = _resolve_checkpoint_conversion_mapping(model_type)
+    if mapping is None:
+        raise ValueError(
+            "No checkpoint conversion mapping is registered for model type "
+            f"`{model_type}`, so linearized load mappings cannot be derived. "
+            "Gate this call on `has_linearize_load_mappings`."
+        )
     model_type = _MODEL_TO_CONVERSION_PATTERN.get(model_type, model_type)
     remove_targets, new_mappings = ARCH_TO_2D_MAPPINGS[model_type]
 
--- a/src/llmcompressor/modeling/moe/conversion_mappings.py
+++ b/src/llmcompressor/modeling/moe/conversion_mappings.py
@@ -147,6 +147,12 @@ ARCH_TO_IMPORT_PATHS: dict[str, tuple[str | list[str], str | list[str]]] = {
         "transformers.models.qwen3_5_moe.configuration_qwen3_5_moe.Qwen3_5MoeTextConfig",
         "transformers.models.qwen3_5_moe.modeling_qwen3_5_moe.Qwen3_5MoeExperts",
     ),
+    # Text-only checkpoints (`Qwen3_5MoeForCausalLM`) report `qwen3_5_moe_text`
+    # as their top-level model type, and share the multimodal expert classes.
+    "qwen3_5_moe_text": (
+        "transformers.models.qwen3_5_moe.configuration_qwen3_5_moe.Qwen3_5MoeTextConfig",
+        "transformers.models.qwen3_5_moe.modeling_qwen3_5_moe.Qwen3_5MoeExperts",
+    ),
     "qwen3_moe": (
         "transformers.models.qwen3_moe.configuration_qwen3_moe.Qwen3MoeConfig",
         "transformers.models.qwen3_moe.modeling_qwen3_moe.Qwen3MoeExperts",
@@ -216,9 +222,45 @@ ARCH_TO_2D_MAPPINGS = {
 }
 
 
+# Qwen3.5-MoE stores 2D per-expert tensors and fuses them with the same rules
+# as Qwen2-MoE, so it reuses that 2D body. Both released spellings are
+# registered: `qwen3_5_moe` for the multimodal wrapper, and `qwen3_5_text` for
+# the text-only checkpoint, which `_MODEL_TO_CONVERSION_PATTERN` remaps
+# `qwen3_5_moe_text` to.
+ARCH_TO_2D_MAPPINGS["qwen3_5_moe"] = ARCH_TO_2D_MAPPINGS["qwen2_moe"]
+ARCH_TO_2D_MAPPINGS["qwen3_5_text"] = ARCH_TO_2D_MAPPINGS["qwen2_moe"]
+
+
+def _resolve_checkpoint_conversion_mapping(
+    model_type: str,
+) -> list[WeightTransform] | None:
+    """
+    Resolve the transformers checkpoint conversion mapping for a model type.
+
+    `ARCH_TO_IMPORT_PATHS` is keyed on the top-level model type, but transformers
+    may register the conversion rules on the text tower instead: a multimodal
+    Qwen3.5-MoE config reports `qwen3_5_moe`, while the rules live under
+    `qwen3_5_moe_text`. Fall back to the text spelling so those architectures
+    resolve instead of returning `None`.
+
+    :param model_type: top-level model type reported by the config
+    :return: conversion mapping, or None if neither spelling is registered
+    """
+    mapping = get_checkpoint_conversion_mapping(model_type)
+    if mapping is None and not model_type.endswith("_text"):
+        mapping = get_checkpoint_conversion_mapping(f"{model_type}_text")
+    return mapping
+
+
 def has_linearize_load_mappings(model_type: str) -> bool:
     remapped_type = _MODEL_TO_CONVERSION_PATTERN.get(model_type, model_type)
-    return model_type in ARCH_TO_IMPORT_PATHS and remapped_type in ARCH_TO_2D_MAPPINGS
+    return (
+        model_type in ARCH_TO_IMPORT_PATHS
+        and remapped_type in ARCH_TO_2D_MAPPINGS
+        # a 2D entry alone is not enough: without conversion rules to strip the fused
+        # targets from, `get_linearize_load_mappings` has nothing to build on
+        and _resolve_checkpoint_conversion_mapping(model_type) is not None
+    )
 
 
 def get_linearize_load_mappings(
@@ -228,7 +270,7 @@ def get_linearize_load_mappings(
     _config_paths, expert_paths = ARCH_TO_IMPORT_PATHS[model_type]
     experts_cls = import_or_none(expert_paths)
 
-    mapping: list[WeightTransform] = get_checkpoint_conversion_mapping(model_type)
+    mapping: list[WeightTransform] = _resolve_checkpoint_conversion_mapping(model_type)
     model_type = _MODEL_TO_CONVERSION_PATTERN.get(model_type, model_type)
     remove_targets, new_mappings = ARCH_TO_2D_MAPPINGS[model_type]
 
--- a/tests/llmcompressor/modeling/test_moe_conversion_mappings.py
+++ b/tests/llmcompressor/modeling/test_moe_conversion_mappings.py
@@ -74,6 +74,19 @@ def test_existing_architectures_still_resolve(model_type):
     assert has_linearize_load_mappings(model_type)
 
 
+def test_model_type_without_conversion_mapping_raises_clearly():
+    """
+    `gpt_oss` is in `ARCH_TO_IMPORT_PATHS` but transformers registers no conversion
+    mapping for it, so `has_linearize_load_mappings` is False and the direct-load
+    pathway is never taken. Called directly it should say so, rather than failing on
+    a downstream lookup or iterating `None`.
+    """
+    assert not has_linearize_load_mappings("gpt_oss")
+
+    with pytest.raises(ValueError, match="No checkpoint conversion mapping"):
+        get_linearize_load_mappings("gpt_oss")
+
+
 def test_qwen3_vl_moe_still_uses_post_load_conversion():
     """
     Qwen3-VL-MoE's conversion rules are identity, so its checkpoints are already
--- a/tests/llmcompressor/modeling/test_moe_conversion_mappings.py
+++ b/tests/llmcompressor/modeling/test_moe_conversion_mappings.py
@@ -1,0 +1,83 @@
+import pytest
+from transformers.core_model_loading import WeightConverter
+
+from llmcompressor.modeling.moe.conversion_mappings import (
+    ARCH_TO_2D_MAPPINGS,
+    get_linearize_load_mappings,
+    has_linearize_load_mappings,
+)
+
+# Qwen3.5-MoE checkpoints store 2D per-expert tensors, so they must take the direct-load
+# pathway. Both released spellings are covered: `qwen3_5_moe` is reported by
+# the multimodal wrapper config, `qwen3_5_moe_text` by the text-only checkpoint.
+QWEN3_5_MOE_TYPES = ["qwen3_5_moe", "qwen3_5_moe_text"]
+
+
+@pytest.mark.parametrize("model_type", QWEN3_5_MOE_TYPES)
+def test_qwen3_5_moe_has_linearize_load_mappings(model_type):
+    assert has_linearize_load_mappings(model_type)
+
+
+@pytest.mark.parametrize("model_type", QWEN3_5_MOE_TYPES)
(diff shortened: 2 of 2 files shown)


kgt392:
Fixed in 0df550c — the Optional return being assigned straight to a `list[WeightTransform]` was a real type mismatch on my part.

Small correction to the failure mode, for the record: with the current tables the crash is a `KeyError` on `ARCH_TO_2D_MAPPINGS` rather than a `TypeError` from iterating `None`, because the 2D lookup happens before the mapping is iterated. Every model type that resolves to no mapping today (`gpt_oss`, `llama4`, `gemma4`, `openai_privacy_filter`) is absent from `ARCH_TO_2D_MAPPINGS`, so it trips that lookup first. The `TypeError` becomes reachable only once a type has a 2D entry but no conversion mapping, which is the combination `has_linearize_load_mappings` now rejects.

Either way the error pointed away from the cause, so the guard raises a `ValueError` naming the model type and pointing at `has_linearize_load_mappings`. Test added using `gpt_oss`, asserting both that the predicate rejects it and that the direct call now explains why. 16 tests passing, and the sweep over all 32 `ARCH_TO_IMPORT_PATHS` keys still shows only `qwen3_5_moe` / `qwen3_5_moe_text` changing, with no regressions.

kylesayrs:
See comment: https://github.com/vllm-project/llm-compressor/issues/3037#issuecomment-5329639607

kylesayrs:
https://github.com/vllm-project/llm-compressor/issues/3037#issuecomment-5333653577

PR #3053, opened 2026-08-19
[ModelFreePTQ] Support quantization configs

kylesayrs:
## Purpose ##
* Support mixed-precision model free PTQ
* Support KV cache quantization for model free PTQ

### DIFF
--- a/src/llmcompressor/entrypoints/model_free/validate.py
+++ b/src/llmcompressor/entrypoints/model_free/validate.py
@@ -1,7 +1,10 @@
 import json
+from typing import Iterable
 
 from compressed_tensors.quantization import (
+    QuantizationConfig,
     QuantizationScheme,
+    QuantizationStatus,
     preset_name_to_scheme,
 )
 from compressed_tensors.utils import getattr_chain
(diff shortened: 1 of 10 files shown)


src/llmcompressor/entrypoints/model_free/validate.py
```diff
@@ -11,33 +14,57 @@
...
     else:
-        scheme_name = "config_group_0"
+        if ignore:
+            existing = list(config.ignore) if config.ignore else []
+            config = config.model_copy(update={"ignore": existing + list(ignore)})
+
+    for scheme in config.config_groups.values():
+        _validate_scheme(scheme)
+
+    return config
```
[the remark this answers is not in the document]
kylesayrs: I'm not sure what a dictionary would really mean. This can be future work

src/llmcompressor/entrypoints/model_free/validate.py
```diff
@@ -11,33 +14,57 @@
...
+        config = QuantizationConfig(
+            config_groups={scheme_name: scheme},
+            quantization_status=QuantizationStatus.INITIALIZED,
+            ignore=list(ignore),
+        )
     else:
-        scheme_name = "config_group_0"
+        if ignore:
+            existing = list(config.ignore) if config.ignore else []
+            config = config.model_copy(update={"ignore": existing + list(ignore)})
```
brian-dellabetta, line 47: nit -- rather than concatenating the ignore lists, i think it would be more clear to an end user if 
- either scheme+ignore should be set
- or only config (which has an ignore field) should be set

brian-dellabetta (approved):
good stuff, one nit

src/llmcompressor/entrypoints/model_free/validate.py
kylesayrs: Yeah totally fair, let me change this

kylesayrs: After some investigation, I think that this might be worth kicking down the road. Many processors rely on the global ignore list, so making this change would make it impossible to create recipes where use MFPTQ and a converter which requires ignores.

In the future, we should move away from the global ignore list

PR #3055, opened 2026-08-19
[AutoRound] Offload calibration inputs to CPU to prevent VRAM OOM with large N

### LINKED ISSUE #3054: [Bug] AutoRoundModifier: calibration inputs retained on GPU exhaust VRAM before optimization
## Environment

- llm-compressor: main (post-#3024)
- auto-round: 0.14.2
- PyTorch: 2.13.0+cu129
- Hardware tested: 8 × NVIDIA H20 (140 GB), 122B MoE model (BF16)

## Bug Description

`AutoRoundModifier.input_capture_hook` is registered as a `forward_pre` hook on
each decoding layer during `on_calibration_start()`.  When the sequential pipeline
runs the calibration forward passes, this hook captures `(args, kwargs)` — which
contain the full `hidden_states` tensor — and appends them **as-is** to
`_all_module_input`.

Because the forward pass executes on GPU, every captured tensor stays in GPU VRAM
until `apply_autoround()` calls `_all_module_input.pop()` before starting
optimization.  For large calibration datasets this amounts to:

```
GPU pressure = (N / world_size) × seq_len × hidden_dim × dtype_bytes
```

which can easily exceed total VRAM before a single optimization step runs.

## Reproduction

Run AutoRound on a large MoE model with a large calibration set:

```python

recipe = AutoRoundModifier(scheme="W4A16", iters=200)
oneshot(
    model="<large MoE model>",
    recipe=recipe,
    dataset="<calibration data>",
    num_calibration_samples=8192,
    max_seq_length=6144,
)
```

OOM occurs at `setup_ddp_if_needed_` → `DDP(block, ...)` before the first
optimization iteration, once the per-block calibration forward pass completes
and `apply_autoround()` attempts to wrap the block.

## Measured Impact

On `122B MoE` (hidden_size=3072, seq_len=6144, BF16, N=4096), 8 × H20 (140 GB):

| Metric | Value |
|---|---|
| Peak VRAM / rank — before fix (measured) | **~134 GB** |
| Peak VRAM / rank — after fix (measured) | **~51 GB** |
| `cur_inputs` offload contribution (512/rank × 6144 × 3072 × 2B) | ~18 GB |
| `fp_outputs` + optimizer savings (`low_gpu_mem_usage=True`) | ~65 GB |

Observed at OOM:

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 38.00 MiB.
GPU 0 has a total capacity of 139.80 GiB of which 21.00 MiB is free.
Including non-PyTorch memory, this process has 139.77 GiB memory in use.
Of the allocated memory 138.57 GiB is allocated by PyTorch.
```

Call stack:
```
apply_autoround()
  → ar.quantize_block(block, inputs=ar_inputs, ...)
      → sign_round/quantizer.py: quantize_block()
          → setup_ddp_if_needed_()
              → DDP(block, device_ids=device_list, ...)   ← OOM here
```

## General VRAM Savings Formula

```
ΔV_VRAM = (N / W - 1) × L × H × sizeof(dtype)
```

| Symbol | Meaning |
|---|---|
| N | total calibration samples |
| W | world_size (number of GPUs) |
| L | max_sequence_length |
| H | model hidden_dim |
| sizeof(dtype) | 2 for BF16/FP16, 4 for FP32 |

Representative values:

| Scenario | N | W | L | H | ΔV / GPU |
|---|---|---|---|---|---|
| 122B MoE, H20 × 8 | 4096 | 8 | 6144 | 3072 | **≈ 18 GB** |
| 35B MoE, RTX 5880 × 8 | 4096 | 8 | 6144 | 4096 | ≈ 26 GB |
| 7B dense, A100 × 1 | 512 | 1 | 2048 | 4096 | ≈ 4 GB |

N越大、模型越宽、卡数越少，收益越显著。对超大 MoE 模型这是量级上的改善。

## Root Cause

```python
# base.py — hook appends GPU tensors as-is
def input_capture_hook(self, module, args, kwargs):
    if module._tmp_name not in self._all_module_input:
        self._all_module_input[module._tmp_name] = []
    self._all_module_input[module._tmp_name].append((args, kwargs))  # ← GPU tensors retained

# apply_autoround() — then tries to move them again (redundant and OOM-prone)
cur_inputs = self._move_inputs_to(cur_inputs, device)   # ← moves N inputs to GPU at once
```

`auto_round` already has `low_gpu_mem_usage=True` which streams `fp_outputs` from CPU
per mini-batch, but `cur_inputs` bypass this because llm-compressor uploads them
before `auto_round` ever sees them.

## Why `low_gpu_mem_usage=True` Does Not Fix This

A natural question is: does setting `low_gpu_mem_usage=True` on the `AutoRound`
constructor already solve the problem?  **No**, because it operates on a completely
different data path:

```
input_capture_hook (llm-compressor)          ← cur_inputs captured here, on GPU
    ↓
apply_autoround
    cur_inputs already ~18 GB on GPU         ← still on GPU before AutoRound can act
    _move_inputs_to(cur_inputs, device)      ← old code: redundant bulk re-upload
    ↓
ar.quantize_block(block, ar_inputs, ...)     ← AutoRound enters here
    → low_gpu_mem_usage takes effect HERE
      (controls cache_device for fp_outputs and best_params only)
```

`low_gpu_mem_usage` is an AutoRound-internal parameter.  It controls where
`fp_outputs` (reference outputs) and `best_params` are stored during the
200-iter optimization loop — setting it `True` moves those to CPU.  It has
no visibility into `_all_module_input`, which was populated by llm-compressor's
hook before `quantize_block` was ever called.

In other words: by the time `low_gpu_mem_usage` can act, `cur_inputs` have
already consumed the `(N/world_size) × seq × hidden × dtype` of VRAM (~18 GB for the test config).
The fix must happen at the hook capture site, which is llm-compressor's
responsibility — not a parameter AutoRound can tune away.

## Fix

Offload tensors to CPU immediately in the hook.  `auto_round`'s `block_forward`
(`compressors/utils.py`) already contains a device check:

```python
if input_ids.device != device:
    input_ids = to_device(input_ids, device)
    input_others = to_device(input_others, device)
```

so it handles per-batch CPU→GPU transfer automatically — no change to `auto_round`
is required.

```python
def input_capture_hook(self, module, args, kwargs):
    name = module._tmp_name
    # Immediately offload to CPU so captured inputs don't accumulate on GPU.
    # auto_round's block_forward (compressors/utils.py) handles per-batch CPU->GPU
    # transfer, so no device mismatch occurs during optimization.
    cpu_args = tuple(
        x.detach().cpu() if isinstance(x, torch.Tensor) else x for x in args
    )
    cpu_kwargs = {
        k: v.detach().cpu() if isinstance(v, torch.Tensor) else v
        for k, v in kwargs.items()
    }
    self._all_module_input.setdefault(name, []).append((cpu_args, cpu_kwargs))
```

The redundant `_move_inputs_to()` call is removed; `block_forward` now owns
per-batch device placement entirely.

**Total diff**: +14 / -6 lines, single file (`autoround/base.py`).

**Speed impact**: per-batch PCIe transfer (~36 MB at batch_size=1, seq=6144,
hidden=3072) at ~32 GB/s ≈ 1.1 ms.  Across 200 iters × 8 accum_steps = 1600
transfers ≈ 1.8 s overhead vs. >25 min per block total — **< 0.1%**.

## Correctness Argument

`.detach()` removes the autograd reference (the calibration forward was never
in the training graph anyway).  `.cpu()` changes storage location but not
values.  `block_forward` restores the tensor to the correct device before any
computation.  The gradient, SignSGD update, and final quantized weights are
therefore **numerically identical** to the pre-fix behavior.

## Verification

8 × H20, 122B MoE model, **N=4096**, seq_len=6144, LIMIT_TO_LAYERS=4 (smoke test, 4 decoder layers × 49 subgraphs):

```
[Rank 0] SUCCESS elapsed=15778s
[Rank 1] SUCCESS elapsed=15778s
[Rank 2] SUCCESS elapsed=15778s
[Rank 3] SUCCESS elapsed=15778s
[Rank 4] SUCCESS elapsed=15778s
[Rank 5] SUCCESS elapsed=15778s
[Rank 6] SUCCESS elapsed=15778s
[Rank 7] SUCCESS elapsed=15777s
exit code: 0
OOM errors: 0
```

All 8 ranks completed with zero OOM errors. Full 48-layer run (N=4096, seq=6144) is in progress.

xesdiny:
## Summary

- `AutoRoundModifier.input_capture_hook` was appending captured `(args, kwargs)` tensors as-is to `_all_module_input`, retaining them on GPU for the full calibration pass duration
- For large models (122B MoE, hidden=3072, seq=6144, N=4096, 8×GPU), `cur_inputs` accumulated ~18 GB VRAM per rank (plus ~65 GB from `fp_outputs`/optimizer states without `low_gpu_mem_usage`), causing OOM at `DDP(block, ...)` inside `quantize_block`
- Fix: offload tensors to CPU immediately in the hook; `auto_round`'s `block_forward` already handles per-batch CPU→GPU transfer, so no change to `auto_round` is required
- Removed the redundant `_move_inputs_to()` bulk re-upload call
- Added `low_gpu_mem_usage=True` to keep `fp_outputs` and `best_params` on CPU during the 200-iter optimization loop
- Pre-load block to local GPU in DDP path to eliminate cold Lustre page-fault stall before `quantize_block`
- Guard `torch.cuda.empty_cache()` with `is_available()` for CPU-only environments

## Changes

Single file change (`src/llmcompressor/modifiers/autoround/base.py`), +18 / -25 lines:

```python
def input_capture_hook(self, module, args, kwargs):
    name = module._tmp_name
    # Immediately offload to CPU so captured inputs don't accumulate on GPU.
    # auto_round's block_forward handles per-batch CPU->GPU transfer.
    cpu_args = tuple(
        x.detach().cpu() if isinstance(x, torch.Tensor) else x for x in args
    )
    cpu_kwargs = {
        k: v.detach().cpu() if isinstance(v, torch.Tensor) else v
        for k, v in kwargs.items()
    }
    self._all_module_input.setdefault(name, []).append((cpu_args, cpu_kwargs))
```

## Why `low_gpu_mem_usage=True` Does Not Fix This

`low_gpu_mem_usage` only controls where `fp_outputs` and `best_params` are stored inside `quantize_block`. By the time AutoRound is invoked, `cur_inputs` have already accumulated the full `(N/world_size) × seq × hidden × dtype` of VRAM via llm-compressor's hook. The fix must happen at the capture site.

## Verification

**Smoke test** — 8 × H20 (140 GB), 122B MoE (BF16, hidden=3072), N=4096, seq=6144, LIMIT_TO_LAYERS=4:

```
[Rank 0] SUCCESS elapsed=15778s  |  OOM errors: 0
[Rank 7] SUCCESS elapsed=15777s  |  all 8 ranks ✅
```

**Full run** — 8 × H20 (140 GB), 122B MoE (BF16, hidden=3072), N=4096, seq=6144, all 48 decoder layers:

```
[Rank 0] SUCCESS elapsed=27710s  |  OOM errors: 0
[Rank 7] SUCCESS elapsed=27709s  |  all 8 ranks ✅
```

48 decoder layers completed with zero OOM errors across all 8 ranks.

Fixes #3054


xesdiny:
@gemini-code-assist review

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -363,6 +384,10 @@ def apply_autoround(self, state, modules):
 
     def post_autoround_cleanup(self):
         self._all_module_input.clear()
+        # Release cached GPU memory back to the driver so the next block starts
+        # with a clean allocator state (avoids cross-block fragmentation).
+        if torch.accelerator.is_available():
+            torch.accelerator.empty_cache()
```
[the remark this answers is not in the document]
xesdiny: Good catch. Fixed in e5c8a88b — dynamically retrieve the active accelerator's torch submodule and call `empty_cache()` only if it exists, making the cleanup portable across CUDA, XPU and future backends.

xesdiny:
@gemini-code-assist review

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -340,10 +353,18 @@ def apply_autoround(self, state, modules):
...
                 auto_offload = True
+            elif torch.distributed.is_initialized():
+                # Standard DDP (1 GPU per rank): pre-load the block onto the rank's
+                # local GPU before quantize_block is called.  auto_round 0.14.2+
+                # calls _move_block_to_device inside setup_ddp_if_needed_, which
+                # triggers a cold Lustre page-fault reload (~15 min) when the block
+                # was evicted from page cache during the sequential calibration pass.
+                # Moving it here makes that call a no-op.
+                device = get_main_device()
+                decoding_layer.to(device)
```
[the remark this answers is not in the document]
xesdiny: Good catch. Fixed in 5e7659b9 — replaced `get_main_device()` (which uses global rank as device ordinal) with `torch.accelerator.current_device_index()` to target the correct local GPU. Single-node behavior is unchanged.

xesdiny:
@gemini-code-assist review

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -340,10 +353,26 @@ def apply_autoround(self, state, modules):
...
+                # Moving it here makes that call a no-op.
+                # Use current_device_index() (local GPU index) rather than global
+                # rank to avoid invalid device ordinal errors on multi-node setups.
+                if torch.accelerator.is_available():
+                    device = torch.device(
+                        torch.accelerator.current_accelerator().type,
+                        torch.accelerator.current_device_index(),
+                    )
+                else:
+                    device = torch.device("cpu")
```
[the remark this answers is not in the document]
xesdiny: Fixed in abc6c2af — replaced `get_main_device()` with `torch.accelerator.current_device_index()`. Dropped the `torch.cuda` fallback: the project requires `torch>=2.10.0` (torch.accelerator present since 2.4), so the fallback is dead code and also violates the project's lint_cuda policy.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -228,7 +228,17 @@ def start_calibration(self, model: torch.nn.Module):
...
+        # Immediately offload to CPU so captured inputs don't accumulate on GPU.
+        # auto_round's block_forward (compressors/utils.py) handles per-batch CPU->GPU
+        # transfer, so no device mismatch occurs during optimization.
+        cpu_args = tuple(
+            x.detach().cpu() if isinstance(x, torch.Tensor) else x for x in args
+        )
+        cpu_kwargs = {
+            k: v.detach().cpu() if isinstance(v, torch.Tensor) else v
+            for k, v in kwargs.items()
+        }
```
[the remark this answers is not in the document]
xesdiny: Fixed in abc6c2af — switched to `tree_map` from `torch.utils._pytree` to recursively offload all tensors in nested structures (lists, dicts, tuples) within `args` and `kwargs`.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -363,6 +392,13 @@ def apply_autoround(self, state, modules):
 
     def post_autoround_cleanup(self):
         self._all_module_input.clear()
+        # Release cached GPU memory back to the driver so the next block starts
+        # with a clean allocator state (avoids cross-block fragmentation).
+        if torch.accelerator.is_available():
+            device_type = torch.accelerator.current_accelerator().type
+            device_module = getattr(torch, device_type, None)
+            if device_module is not None and hasattr(device_module, "empty_cache"):
+                device_module.empty_cache()
```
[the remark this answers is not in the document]
xesdiny: Fixed in abc6c2af — removed the `torch.cuda` fallback for the same reason as above (`torch>=2.10.0` minimum, lint_cuda policy). The accelerator-agnostic path via `getattr(torch, device_type)` already handles CUDA, XPU, and other backends.

xesdiny:
@gemini-code-assist review

yiliu30:
@codex review

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -340,10 +354,26 @@ def apply_autoround(self, state, modules):
...
+                else:
+                    device = torch.device("cpu")
+                decoding_layer.to(device)
 
-            # Ensure cached inputs are on the same device as the block.
-            # Calibration forward may have run on a different GPU.
-            cur_inputs = self._move_inputs_to(cur_inputs, device)
+            # cur_inputs remain on CPU; auto_round's block_forward handles
+            # per-batch CPU->GPU transfer automatically.
             ar_inputs = [((args, kwargs),) for args, kwargs in cur_inputs]
```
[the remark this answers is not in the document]
xesdiny: @yiliu30 Thanks for flagging this.   The test conflict is intentional — it reflects a deliberate shift in device-placement ownership:

**Before this PR**: `apply_autoround` called `_move_inputs_to(cur_inputs, device)` to bulk-upload all N inputs to the block's device before `quantize_block`. The test mocked `get_main_device()` as `meta` and asserted `device.type == "meta"` to verify that contract.

**After this PR**: `input_capture_hook` offloads immediately to CPU, and device placement is delegated to `auto_round`'s `block_forward` (which transfers one batch at a time via the existing `to_device` call in `compressors/utils.py`). Inputs deliberately stay on CPU all the way into `quantize_block`.

The reason `_move_inputs_to` had to go — not just the hook — is that bulk-uploading N inputs at once is structurally incompatible with large-N runs: at N=1024/rank, seq=6144, hidden=7168, BF16, that is ~91 GB transferred to GPU before any optimization begins, which is what caused the OOM in #3054. `auto_round`'s per-batch transfer (~88 MB × 1600 iterations) is the correct replacement.

The test assertion should change from `"meta"` to `"cpu"` to reflect the new contract. `auto_round` is not installed in the current CI test environment so this does not block merging, but the fix is straightforward. 

**Happy to update the test in this PR if preferred.** 🤝

xesdiny:
@gemini-code-assist review

xesdiny:
@gemini-code-assist review

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -340,10 +354,26 @@ def apply_autoround(self, state, modules):
                 device = get_main_device()
                 decoding_layer.to("cpu")
                 auto_offload = True
+            elif torch.distributed.is_initialized():
+                # Standard DDP (1 GPU per rank): pre-load the block onto the rank's
+                # local GPU before quantize_block is called.  auto_round 0.14.2+
+                # calls _move_block_to_device inside setup_ddp_if_needed_, which
+                # triggers a cold Lustre page-fault reload (~15 min) when the block
+                # was evicted from page cache during the sequential calibration pass.
```
yiliu30: > A 15-minute Lustre page-fault stall

Does this look like an environment-dependent issue? Also, which model are you testing, and how much memory does it offload and onload?

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -228,7 +229,17 @@ def start_calibration(self, model: torch.nn.Module):
 
     def input_capture_hook(self, module, args, kwargs):
         name = module._tmp_name
-        self._all_module_input.setdefault(name, []).append((args, kwargs))
+        # Immediately offload to CPU so captured inputs don't accumulate on GPU.
+        # auto_round's block_forward (compressors/utils.py) handles per-batch CPU->GPU
+        # transfer, so no device mismatch occurs during optimization.
+        # tree_map handles nested structures (lists/dicts of tensors) robustly.
+        cpu_args = tree_map(
```
yiliu30, line 236: good point!

yiliu30 (commented):
Hi @xesdiny thanks for the enhancement!
Overall LGTM, left a few questions. Could you verify this example locally, thanks!
https://github.com/vllm-project/llm-compressor/blob/main/examples/autoround/ddp/ddp_qwen3_example.py

xesdiny:
@yiliu30  
**Memory offloaded/onloaded** — the saving comes from two combined changes:
1. `input_capture_hook` offloads `cur_inputs` to CPU immediately: `(N/world_size) × seq × hidden × dtype_bytes` per rank
2. `low_gpu_mem_usage=True` offloads `fp_outputs` and `best_params` to CPU during the 200-iter optimization loop

Measured on the 122B production run (8× H20, N=4096):
- Peak VRAM before fix: **~134 GB/rank** (nearly exhausting the 140 GB H20)
- Peak VRAM after fix: **~51 GB/rank**
- Saving: **~83 GB/rank**

```
# before-fix: run from old worktree, GPUs 0,1
CUDA_VISIBLE_DEVICES=0,1 \
PYTHONPATH=${AUTOROUND_SRC}:/tmp/autoround-before/src \
torchrun --nproc_per_node=2 --master_port=29500 \
    ${MAIN_REPO}/wnd_dev/github/logs/ddp_vram_monitor.py \
    --label before --nsamples 512 --iters 10 \
    > ${MAIN_REPO}/wnd_dev/github/logs/before.log 2>&1 &
BEFORE_PID=$!
echo "before-fix PID: $BEFORE_PID"

sleep 3

# after-fix: run from current (fixed) code, GPUs 4,5
CUDA_VISIBLE_DEVICES=4,5 \
PYTHONPATH=${AUTOROUND_SRC}:${MAIN_REPO}/src \
torchrun --nproc_per_node=2 --master_port=29501 \
    ${MAIN_REPO}/wnd_dev/github/logs/ddp_vram_monitor.py \
    --label after --nsamples 512 --iters 10 \
    > ${MAIN_REPO}/wnd_dev/github/logs/after.log 2>&1 &
AFTER_PID=$!
echo "after-fix PID: $AFTER_PID"

echo "Both launched. Waiting..."
wait $BEFORE_PID && echo "before done (ok)" || echo "before done (exit $?)"
wait $AFTER_PID && echo "after done (ok)" || echo "after done (exit $?)"
```
Also verified locally on Qwen3-8B (2× L20 46 GB, N=512, seq=2048):
- Peak VRAM before fix: **29.25 GB/rank** (torch) / **~31 GB** (nvidia-smi)
- Peak VRAM after fix: **8.17 GB/rank** (torch) / **~10.5 GB** (nvidia-smi)
- Saving: **~21 GB/rank (-72%)**

Detail Log: 

[after.log](https://github.com/user-attachments/files/31303041/after.log)
[before.log](https://github.com/user-attachments/files/31303044/before.log)
[vram_comparison_summary.md](https://github.com/user-attachments/files/31303047/vram_comparison_summary.md)
[vram_poll.log](https://github.com/user-attachments/files/31303048/vram_poll.log)

src/llmcompressor/modifiers/autoround/base.py
xesdiny: Thanks for the review! I've verified the DDP example locally:
Model tested: Qwen/Qwen3.5-122B-A10B
On H20-3e(144G *8) Use nsamples =4K seqlen=6144 pack_conconcate_data.


```bash
torchrun --nproc_per_node=2 examples/autoround/ddp/ddp_qwen3_example.py \
    --model Qwen/Qwen3-8B --nsamples 512 --iters 10 --disable_torch_compile
```

Both Rank 0&1 and Rank 4&5 completed quantization and sample generation successfully.
Also ran a before/after VRAM comparison (Qwen3-8B, N=512) to quantify the memory
improvement — results included in the comment above.

When I set **nsamples=512 -> 1024** , the former resulted in `torch.OutOfMemoryError: CUDA out of memory.`, while the latter ran normally up to the auto iter stage.

xesdiny:
> Hi @xesdiny thanks for the enhancement! Overall LGTM, left a few questions. Could you verify this example locally, thanks! https://github.com/vllm-project/llm-compressor/blob/main/examples/autoround/ddp/ddp_qwen3_example.py



@yiliu30 Thanks for the review!

I have verified examples/autoround/ddp/ddp_qwen3_example.py locally — full results in [this earlier comment](https://github.com/vllm-project/llm-compressor/pull/3055#issuecomment-5369122243). Summary:

Qwen3-8B, 2× L20 (46 GB), DDP:

|Config |	Before (9804eac3)	| After (this branch) |
|---|---|---|
|nsamples=512, peak VRAM |	29.25 GB	| 8.17 GB |
| nsamples=1024 |	OOM crash (~44 GB) |	Completed (13.3 GB) |

Regarding the Lustre page-fault: yes, environment-dependent — specific to network-attached storage where evicted pages need a remote fetch. Local NVMe would not exhibit this. Our 122B MoE model lives on a Lustre mount, hence the ~15 min stall per block.

Also just pushed 83dfcc63 adding hasattr(torch, "accelerator") guards at all three call sites, addressing the bot feedback on PyTorch < 2.4 compatibility.

src/llmcompressor/modifiers/autoround/base.py
xesdiny: Yes, environment-dependent — three conditions must all be true:
1. Model weights on a network-attached filesystem (Lustre/NFS)
2. Physical RAM < total mmap footprint across all ranks (8 × ~244 GB ≈ 1.9 TB in our case)
3. OS evicts the block's pages from page cache during the sequential calibration pass

Under those conditions, `setup_ddp_if_needed_()` → `_move_block_to_device()` triggers a cold Lustre fetch per block (~24 min). Local NVMe keeps pages in kernel page cache and reloads in < 1 s.

This is orthogonal to the PR fix — the PR addresses `cur_inputs` VRAM accumulation (N × seqlen GPU tensors never freed). The Lustre stall would occur with or without this PR.

For clarification on models: the DDP example (`ddp_qwen3_example.py`) was verified with **Qwen3-8B**. The 122B MoE (`Qwen3.5-122B-A10B`) is our production model where the Lustre stall was observed.

yiliu30:
> > Hi @xesdiny thanks for the enhancement! Overall LGTM, left a few questions. Could you verify this example locally, thanks! https://github.com/vllm-project/llm-compressor/blob/main/examples/autoround/ddp/ddp_qwen3_example.py
> 
> @yiliu30 Thanks for the review!
> 
> I have verified examples/autoround/ddp/ddp_qwen3_example.py locally — full results in [this earlier comment](https://github.com/vllm-project/llm-compressor/pull/3055#issuecomment-5369122243). Summary:
> 
> Qwen3-8B, 2× L20 (46 GB), DDP:
> 
> Config	Before ([9804eac](https://github.com/vllm-project/llm-compressor/commit/9804eac351566dcdb7f7040d76dfcb8f1cc4274c))	After (this branch)
> nsamples=512, peak VRAM	29.25 GB	8.17 GB
> nsamples=1024	OOM crash (~44 GB)	Completed (13.3 GB)
> Regarding the Lustre page-fault: yes, environment-dependent — specific to network-attached storage where evicted pages need a remote fetch. Local NVMe would not exhibit this. Our 122B MoE model lives on a Lustre mount, hence the ~15 min stall per block.
> 
> Also just pushed [83dfcc6](https://github.com/vllm-project/llm-compressor/commit/83dfcc631acb4afef8d16f44a88698a56541ebc1) adding hasattr(torch, "accelerator") guards at all three call sites, addressing the bot feedback on PyTorch < 2.4 compatibility.

Thanks @xesdiny , we can ignore PyTorch < 2.4 compatibility warning.

src/llmcompressor/modifiers/autoround/base.py
yiliu30: Thanks for clarifying, could you update the comment as well, thanks!

yiliu30 (approved):
LGTM, thanks!

xesdiny:
Understood, will leave the hasattr guards as-is (they do no harm) and ignore the bot warning. Thanks!

src/llmcompressor/modifiers/autoround/base.py
xesdiny: Updated in 4a8729c7 — the comment now reads:

> triggers a slow reload when the block was evicted from page cache during the sequential calibration pass.

yiliu30 (approved):

xesdiny:
@brian-dellabetta    — could you trigger Buildkite CI for this PR? Thanks!🙏

dsikka:
@yiliu30 can you apply the ready label? You should have access. If not let me know

yiliu30:
> @yiliu30 can you apply the ready label? You should have access. If not let me know

Sure, added, the CI has started.

yiliu30:
Hi @dsikka @brian-dellabetta, could you please take a look and help get it merged if it looks good to you?

xesdiny:
@yiliu30 Rebased onto latest main, ready for merge. Thanks!

src/llmcompressor/modifiers/autoround/base.py
brian-dellabetta, line 236: The changes make sense to me, but we have an IntermediatesCache abstraction to handle this. Check out usage in AWQ [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/transform/awq/base.py#L456-L459).

Happy to merge in as is, but just FYI in case you find this helpful.

brian-dellabetta (commented):
Hey guys, the offloading addition makes sense to me, but see comment for a potential solution leveraging our IntermediatesCache abstraction to handle this.

Let me know how you'd like to proceed -- we can merge as is, refactor to IntermediatesCache at a later date, or just not use it at all, but we've found it pretty valuable for AWQ

src/llmcompressor/modifiers/autoround/base.py
xesdiny: @brian-dellabetta  Thanks for the pointer! Looking at the `IntermediatesCache` usage in AWQ was helpful.

For `cur_inputs` in this PR, the hook fires *within* the calibration forward pass — before the pipeline's `activations` cache has the subgraph outputs — so there's no direct overlap with the existing `IntermediatesCache` data flow. The `tree_map` approach here is minimal and self-contained.

Your suggestion does point toward a real next step we've been thinking about: `auto_round`'s `collect_reference` runs a full FP16 forward pass on the next subgraph to build its reference outputs, but when `propagate_error=False` the pipeline's `activations` `IntermediatesCache` already holds the identical tensors. Eliminating that redundant pass would require adding a `reference_output` parameter to `ar.quantize_block()` upstream in `intel/auto-round`, after which llm-compressor can pass the cached values directly — exactly the integration pattern you described. We're planning that as a follow-up coordinated with an upstream PR.

xesdiny: @brian-dellabetta Thanks for the suggestion on IntermediatesCache — we actually are already using that pattern in a follow-up fix for `fp_ref` pre-seeding (branch `fix/autoround-fp-ref-preseed`). For this PR, I'd prefer to keep it scoped to the immediate OOM fix and track the broader IntermediatesCache refactor for calibration inputs as a follow-up issue. Would option 1 (merge as-is) work for you, given yiliu30 has already approved?

@yiliu30 anything blocking on your end?

brian-dellabetta: sounds like a plan, thanks @xesdiny

brian-dellabetta (approved):

PR #3058, opened 2026-08-19
[Bugfix] Resolve decorated forwards without functools.wraps in autowrap_forward

### LINKED ISSUE #3057: [Bug] sequential pipeline cannot trace models whose forward is decorated without functools.wraps (Qwen3.8 / Qwen3_5GatedDeltaNet)
## Describe the bug

The sequential-pipeline calibration path (used by GPTQ, SparseGPT, and AWQ via this pipeline) cannot trace any model whose module `forward` is wrapped by a decorator that does **not** apply `functools.wraps`. This blocks quantization of such models entirely.

Concretely, in transformers >= 5.x the `Qwen3_5GatedDeltaNet.forward` method is decorated with `@force_accelerate_hooks("conv1d")` (defined in [`transformers/integrations/accelerate.py`](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/accelerate.py)). That decorator returns an inner function:

```python
def force_accelerate_hooks(...):
    def wrapped(self, *args, **kwargs):
        ...
        return fn(self, *args, **kwargs)
    return wrapped
```

and does **not** apply `functools.wraps`, so the wrapper has neither `__wrapped__` nor the original `__name__`.

## Root cause

In `src/llmcompressor/pipelines/sequential/ast_helpers.py`, `autowrap_forward` does:

```python
source = inspect.getsource(module.forward)   # returns the wrapper's source
...
exec(code, namespace)
new_forward = namespace["forward"].__get__(module)   # <-- KeyError: 'forward'
```

Because `inspect.getsource(module.forward)` returns the source of the **wrapper** (the inner `def wrapped(self, *args, **kwargs)`), the re-exec'd code defines a function named `wrapped`, not `forward`. The subsequent `namespace["forward"]` lookup therefore raises `KeyError: 'forward'`.

`inspect.unwrap` does **not** help here because the wrapper has no `__wrapped__` attribute, so there is nothing to unwrap — `inspect.unwrap(module.forward)` returns the wrapper unchanged.

## Steps to reproduce

```python
import torch
from llmcompressor.pipelines.sequential.ast_helpers import autowrap_forward


# Mirrors the shape of transformers' force_accelerate_hooks:
# a decorator that does NOT use functools.wraps.
def force_conv1d(fn):
    def wrapped(self, *args, **kwargs):
        return fn(self, *args, **kwargs)
    return wrapped


class Mod(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = torch.nn.Linear(4, 4)

    @force_conv1d
    def forward(self, x):
        return self.linear(x)


with autowrap_forward(Mod(), ignore=[]):
    pass
```

### Error

```
KeyError: 'forward'
```

at `src/llmcompressor/pipelines/sequential/ast_helpers.py`, line `new_forward = namespace["forward"].__get__(module)`.

### Real-world impact

Running a sequential GPTQ calibration of `Qwen/Qwen3.8-27B` (64-layer hybrid: 16 full-attention + 48 `Qwen3_5GatedDeltaNet` layers) fails immediately during tracing. The same failure applies to any model using a `forward` decorated this way under transformers >= 5.x.

## Expected behavior

`autowrap_forward` should resolve the decorated method to the original `forward` function before calling `inspect.getsource`, so the re-exec'd code defines `forward` and the lookup succeeds. The decorator's behaviour (e.g. accelerate hook setup) should be preserved — `inspect.getsource` on the original function includes its decorator line, so re-executing it re-applies the decorator.

## Environment

- llm-compressor: main (commit `44e78dcc`)
- transformers >= 5.x (provides `force_accelerate_hooks`)
- Model: `Qwen/Qwen3.8-27B`

A fix is forthcoming in a linked PR.

malaiwah:
SUMMARY:
Fixes #3057.

The sequential-pipeline calibration path (GPTQ, SparseGPT, AWQ via this pipeline) cannot trace any model whose module `forward` is wrapped by a decorator that does **not** apply `functools.wraps`. This blocks quantization of such models entirely.

Concretely, in transformers >= 5.x `Qwen3_5GatedDeltaNet.forward` is decorated with `@force_accelerate_hooks("conv1d")` (see `transformers/integrations/accelerate.py`), which returns an inner `def wrapped(self, *args, **kwargs)` without copying `__wrapped__` or `__name__`.

### Root cause

In `src/llmcompressor/pipelines/sequential/ast_helpers.py`, `autowrap_forward` does:

```python
source = inspect.getsource(module.forward)   # returns the wrapper's source
...
exec(code, namespace)
new_forward = namespace["forward"].__get__(module)   # <-- KeyError: 'forward'
```

Because `inspect.getsource(module.forward)` returns the source of the **wrapper** (the inner `def wrapped(...)`), the re-exec'd code defines a function named `wrapped`, not `forward`, and `namespace["forward"]` raises `KeyError: 'forward'`. `inspect.unwrap` does not help because the wrapper has no `__wrapped__` attribute to follow.

### The fix

Resolve the decorated method to the original `forward` function **before** calling `inspect.getsource`, by searching the wrapper's closure cells for a function named `forward`:

```python
target = inspect.unwrap(module.forward)
if getattr(target, "__name__", "") != "forward" and getattr(target, "__closure__", None):
    from types import FunctionType

    for cell in target.__closure__:
        try:
            contents = cell.cell_contents
        except ValueError:
            continue
        if isinstance(contents, FunctionType) and contents.__name__ == "forward":
            target = contents
            break
source = inspect.getsource(target)
```

### Why decorator behaviour is preserved

`inspect.getsource` on the original `forward` **includes its decorator line**, so re-executing the source re-applies the decorator. The accelerate hook setup (and any other wrapper side effects) therefore still take effect on the autowrapped forward. This is important: we are not bypassing the decorator, we are recovering the decorated definition so that `torch.fx` autowrapping can be applied to the real `forward` body while the decorator is re-applied around it.

There is **no behaviour change for undecorated forwards**: `inspect.unwrap` returns the method unchanged and the closure search is skipped.

TEST PLAN:
- Added regression tests in `tests/llmcompressor/pipelines/sequential/test_ast_helpers.py`:
  - `test_autowrap_forward_handles_unwrapped_decorator`: a module whose `forward` is wrapped by a no-`functools.wraps` decorator (mirroring `force_accelerate_hooks`). Previously raised `KeyError: 'forward'`; now autowraps successfully, produces correct output, and the decorator's side effect still fires (proving the decorator is re-applied).
  - `test_autowrap_forward_plain_module_unchanged`: an undecorated forward, asserting behaviour is unchanged.
- Both tests fail with `KeyError: 'forward'` on `main` and pass with this change.
- `ruff check` and `ruff format --check` pass on the touched files.
- End-to-end validation: a full GPTQ run of `Qwen/Qwen3.8-27B` (64-layer hybrid: 16 full-attention + 48 `Qwen3_5GatedDeltaNet` layers) over 482 subgraphs completed in ~45 minutes on a single RTX 5090, where on `main` it could not trace at all. This confirms tracing/calibration now completes end-to-end for this model class. (Note: this PR claims only that tracing/calibration completes; it makes no claim about the resulting checkpoint quality.)


kylesayrs (commented):
Thanks for the fix! Could you please
1. Break this functionality into a helper function (something like `get_unwrapped_forward`)
2. Make sure that imports are declared at the top module level

Thank you!

malaiwah:
Thank you both for the reviews. All feedback has been addressed in commit b9b6ff2:

1. **@gemini-code-assist** — The bound method `__closure__` issue is fixed: `inspect.ismethod(forward)` now extracts `__func__` before the closure search, so the unwrapping logic actually runs on real module instances.
2. **@kylesayrs** — Functionality extracted to `get_unwrapped_forward` helper function, and imports (`FunctionType` from `types`, `Callable` from `typing`) are at the top module level.

CI is green (quality-check ✅, markdown-link-check ✅, DCO ✅). Ready for re-review.

kylesayrs (approved):

PR #3060, opened 2026-08-19
Add "perfectblend" prebaked dataset and "ultrachat" alias

kylesayrs:
## Summary
- Add `"perfectblend"` as a registered prebaked dataset pointing to `mlabonne/open-perfectblend`, enabling `oneshot(model, dataset="perfectblend")`. Handles ShareGPT format conversion (`from`/`value` → `role`/`content`) and role mapping (`human`→`user`, `gpt`→`assistant`) automatically.
- Add `"ultrachat"` as a convenience alias for the existing `"ultrachat_200k"` dataset.

## Test plan
- [x] Verified registry resolves `"perfectblend"`, `"ultrachat"`, and `"ultrachat_200k"` correctly
- [ ] Run `oneshot(model, dataset="perfectblend", ...)` end-to-end with a small model
- [ ] Run `oneshot(model, dataset="ultrachat", ...)` end-to-end with a small model


src/llmcompressor/transformers/data/ultrachat_200k.py
```diff
@@ -10,7 +10,7 @@
     from llmcompressor.args import DatasetArguments
 
 
-@TextGenerationDataset.register(name="ultrachat_200k")
+@TextGenerationDataset.register(name="ultrachat_200k", alias="ultrachat")
```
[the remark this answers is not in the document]
kylesayrs: is supported https://github.com/vllm-project/compressed-tensors/blob/bc350799eeeee0661ccdc89bf75cd90632084ebd/src/compressed_tensors/registry/registry.py#L109

src/llmcompressor/transformers/data/perfectblend.py
```diff
@@ -0,0 +1,70 @@
...
+from llmcompressor.typing import Processor
+
+if TYPE_CHECKING:
+    from llmcompressor.args import DatasetArguments
+
+ROLE_MAP = {"human": "user", "gpt": "assistant"}
+
+
+@TextGenerationDataset.register(name="perfectblend")
+class PerfectBlendDataset(TextGenerationDataset):
```
brian-dellabetta, line 16: This is a nice addition. Eldar had messaged in a thread that perfect blend dataset needs to be shuffled because it contains specific sections. Does that get handled anywhere in TextGenerationDataset? afaik it doesn't

kylesayrs: Shuffling is handled [here](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/datasets/utils.py#L256), agnostic to the particular `TextGenerationDataset`.

A user can disable shuffling if they want, but shuffling is the default and always has been.

brian-dellabetta: I want to distinguish between 
- downloading the entire dataset and shuffling/picking out num_calibration_samples
- downloading num_calibration_samples from the dataset and then shuffling just that subset.

But it looks like it's up to the user, if we set dataset, don't set set split, and set num_calibration_samples=256 it will pick a shuffled subset of 256 samples from the entire dataset. Is that your understanding as well?

kylesayrs: Both cases are accessible through the `oneshot` interface.

download -> shuffle -> pick
```python
oneshot(..., split="train_sft", num_calibration_samples=512)
```

pick -> download -> shuffle
```python
oneshot(..., split="train_sft[:512]")
```

brian-dellabetta (approved):
good stuff!

PR #3062, opened 2026-08-19
Replace example datasets with prebaked "perfectblend" and "flickr30k"

kylesayrs:
## Summary
- Replace manual dataset loading boilerplate across ~55 example files with `dataset="perfectblend"`, removing ~1700 lines of repetitive `load_dataset` + `preprocess` + `tokenize` code.
- DDP examples are also simplified — automatic rank partitioning from #3061 means `get_rank_partition()` is no longer needed when using prebaked datasets.
- Add `examples/custom_dataset_example.py` showing how to use a custom dataset (e.g. ultrachat_200k) for users who need custom preprocessing.
- Examples with model-specific processing (custom processors/collators, audio/vision pipelines, AutoRound, DeepSeek encoding) are left unchanged.
- Stacked on #3061.

## Examples kept as-is (custom processing required)
- AutoRound examples (use `auto_round.calib_dataset.get_dataset`)
- Multimodal vision examples with custom image/processor handling (idefics3, internvl3, phi3_vision, qwen VL, etc.)
- Multimodal audio examples (whisper, qwen2_audio)
- Examples with custom data collators (qwen3_5, qwen3_6, qwen3_8 GPTQ+AWQ)
- Examples using specialized datasets (neuralmagic/calibration, codeparrot, Magpie)
- AWQ masking example (custom loss_mask tokenization)
- DeepSeek V4 examples (custom encoding)
- Benchmark scripts (complex argparse + timing)
- Data-free examples (FP8_DYNAMIC, model_free_ptq, etc.)

## Test plan
- [x] Spot-checked simplified files preserve model loading, recipe, generation, and save logic
- [x] Verified no unintended ultrachat/perfectblend HF references remain in simplified files
- [ ] Run a simplified example end-to-end with a small model
- [ ] Run a simplified DDP example with `torchrun --nproc_per_node=2`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3063, opened 2026-08-19
[Tests] Use prebaked "perfectblend" as default calibration dataset for tests

kylesayrs:
* Update straggler examples not caught by https://github.com/vllm-project/llm-compressor/pull/3062
* Update tests to use perfect blend
  * This is create a slight difference in lm_eval tests, but will mean that our lm_eval tests are more representative of our default flows


dsikka (approved):

dsikka:
@kylesayrs failing tests

brian-dellabetta (approved):

PR #3070, opened 2026-08-21
fix(trace): make llmcompressor.trace boolean flags actually toggleable

Anai-Guo:
## Summary

`llmcompressor.trace` (a published console script) declares two boolean CLI arguments with `type=bool`:

```python
parser.add_argument("--trust_remote_code", type=bool, default=False, ...)
parser.add_argument("--skip_weights", type=bool, default=True, ...)
```

`type=bool` runs the raw string through `bool(str)`, and every non-empty string is truthy. So both flags evaluate to `True` for every value a user would plausibly write to turn them off:

| command | resulting value |
|---|---|
| `--skip_weights False` | `True` |
| `--skip_weights 0` | `True` |
| `--skip_weights no` | `True` |
| `--trust_remote_code False` | `True` |
| `--skip_weights ""` | `False` (the only value that works) |

Two consequences:

* `--skip_weights` defaults to `True`, so **there is no way to trace a model with real weights from the CLI** — the flag exists but cannot be turned off.
* `--trust_remote_code False` **silently enables** remote code execution instead of declining it.

## Fix

Switch both to `argparse.BooleanOptionalAction`, which keeps the `--flag` spelling and adds the `--no-flag` counterpart:

```
--trust_remote_code / --no-trust_remote_code
--skip_weights      / --no-skip_weights
```

`BooleanOptionalAction` is stdlib since Python 3.9 and this package already requires `>=3.10`.

These two lines are the **only** `type=bool` arguments in the repository — every other boolean argument in `src/`, `examples/` and `tests/` already uses an action, so this brings them in line with the existing convention.

### Behaviour change worth calling out

`--skip_weights False` / `--trust_remote_code True` (value-style) now error as unrecognized arguments instead of being silently accepted. Previously the value was ignored and the flag was always `True`, so no working invocation changes meaning — an invocation that silently did the opposite of what it said now fails loudly.

## Testing

Added `tests/llmcompressor/transformers/tracing/test_debug_args.py` — a CPU-only parametrized test that asserts both flags are settable in either direction. It fails on `main` (the `--no-*` forms do not exist) and passes with this change.

Lint verified locally with the pinned `ruff~=0.4.8` using the repo's own config, via the exact `make quality` invocation:

```
ruff check src tests          -> All checks passed!
ruff format --check src tests -> 326 files already formatted
```

(`src/llmcompressor/transformers/tracing/` is in `extend-exclude`, so `debug.py` itself is not linted; the new test file is under `tests/` and is.)

This was found by an audit for boolean flags that cannot be turned off, not from an existing issue, so it is not tracked anywhere and nobody else is working on it.


kylesayrs (approved):
Thanks!

Anai-Guo:
Friendly ping — this is approved and merge protections report satisfied. Is there anything else needed before it can go in?

Issue #3064, opened 2026-08-20
[Bug]: tools/collect_env.py crashes on Apple Silicon (torch.mps has no get_device_name)


Isitthakkar11:
### ⚙️ Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
### Environment Information ###
Operating System: `macOS-26.5.2-arm64-arm-64bit`
Python Version: `3.12.5`
llm-compressor Version: `0.13.1.dev24+g48116705`
compressed-tensors Version: `0.18.1a20260818`
transformers Version: `5.15.0`
torch Version: `2.13.0`
CUDA Devices: `None`
AMD Devices: `None`
NPU Devices: `None`
MPS Devices: `['arm']`
```

</details>

Note: the output above required the patch described below. The unpatched script crashes before printing anything on this machine.

### 🐛 Describe the bug

`tools/collect_env.py` crashes on Apple Silicon, so macOS users cannot produce the environment block this bug report template asks for.

```
Traceback (most recent call last):
  File "tools/collect_env.py", line 60, in <module>
    collect_environment_info()
  File "tools/collect_env.py", line 40, in collect_environment_info
    cuda_devices, amd_devices, npu_devices = get_torch_hardware_info()
  File "tools/collect_env.py", line 27, in get_torch_hardware_info
    name = torch.get_device_module().get_device_name(i)
AttributeError: module 'torch.mps' has no attribute 'get_device_name'
```

Cause: on Apple Silicon `torch.accelerator.is_available()` returns `True` because MPS counts as an accelerator, so the device loop in `get_torch_hardware_info` runs. `torch.get_device_module()` then returns `torch.mps`, which does not implement `get_device_name`. The surrounding `except ImportError` only catches a missing torch, so the `AttributeError` propagates and the script exits without printing anything.

Proposed fix:

- guard with `hasattr(device_module, "get_device_name")` rather than special-casing MPS, so any future backend lacking that method degrades gracefully
- report MPS devices in their own field
- widen the except to `(ImportError, AttributeError)` — a diagnostic tool should degrade rather than crash, since anyone running it is already debugging something else

I have this working locally against `48116705` and am happy to open a PR.

### 🛠️ Steps to reproduce

On any Apple Silicon Mac with torch installed:

```bash
python3 tools/collect_env.py
```

No model, GPU, or dataset required.

brian-dellabetta:
[...]

PR #3074, opened 2026-08-21
Add WhisperForConditionalGeneration to AWQ mapping registry

robertlangdonn:
`WhisperForConditionalGeneration` had no AWQ mapping, so it silently fell back to `default_mappings`, whose balance layers (`q_proj`/`k_proj`/`v_proj`, `gate_proj`/`up_proj`, `down_proj`) never match Whisper's real layer names — its MLP is a plain `fc1`/`fc2` pair with no `gate_proj`/`up_proj`/`down_proj` at all, so smoothing was a complete no-op for this arch.

This mirrors `WHISPER_V2_SMOOTHQUANT_MAPPINGS` (`smoothquant/utils.py`), which is already validated in production, translated into the `AWQMapping` dataclass shape:
- `self_attn_layer_norm` → `k_proj`, `v_proj`, `q_proj`
- `final_layer_norm` → `fc1`

**A scope note I want to be upfront about:** I deliberately did **not** add a `v_proj → out_proj` or `activation_fn → fc2` "second matmul" chain the way most other architectures here do (see `_bloom_mappings`' analogous `gelu_impl → dense_4h_to_h` entry for its own single-path MLP). Two things stopped me: Whisper's attention output projection is `out_proj`, not `o_proj`, and there's no existing evidence (SmoothQuant precedent or otherwise) this repo has validated such a chain for Whisper specifically — `_bloom_mappings`' own comment shows this kind of chain is sometimes deliberately excluded per empirical AutoAWQ findings, not a universal requirement. Happy to add it if you'd like, but didn't want to guess at something unverified. This PR also doesn't cover the decoder's cross-attention block (`encoder_attn` / `encoder_attn_layer_norm`), matching the existing SmoothQuant mapping's scope exactly.

TEST PLAN:
Constructed a tiny `WhisperForConditionalGeneration` on the meta device (no HF Hub download, no weight allocation) and confirmed empirically that the new mapping's regex matches its real `self_attn_layer_norm`/`k_proj`/`v_proj`/`q_proj`/`final_layer_norm`/`fc1` module names.

Focused suite passes: 24/24 in `tests/llmcompressor/modifiers/transform/awq/test_dynamic_mappings.py`. `make style` / `make quality` clean on both changed files.

(`tests/llmcompressor/modifiers/transform/awq/test_base.py` fails to even collect on this machine — `torch.mps` has no `get_device_capability`, the same class of Apple Silicon gap as issue #3064 — unrelated to this change; not run here per this repo's own guidance to use focused subsets on non-GPU machines.)

### DIFF
--- a/src/llmcompressor/modifiers/transform/awq/mappings.py
+++ b/src/llmcompressor/modifiers/transform/awq/mappings.py
@@ -285,6 +285,28 @@ _example_parallel_transformer_block_mappings = [
     )
 ]
 
+# Whisper (encoder + decoder self-attention share this naming). Mechanical port of
+# WHISPER_V2_SMOOTHQUANT_MAPPINGS (smoothquant/utils.py), which is already validated
+# in production. Deliberately does NOT add a v_proj->out_proj or activation_fn->fc2
+# "second matmul" chain the way most other architectures' mappings do (see
+# _bloom_mappings above): Whisper's attention output projection is out_proj, not
+# o_proj, and there's no existing evidence (SmoothQuant precedent or otherwise) this
+# repo has validated such a chain for Whisper specifically -- Bloom's own mapping
+# shows this chain is sometimes deliberately excluded per empirical AutoAWQ findings,
+# not a universal requirement. Also does not cover the decoder's cross-attention
+# block (encoder_attn / encoder_attn_layer_norm), matching the existing SmoothQuant
+# mapping's scope exactly.
+_whisper_mappings = [
+    AWQMapping(
+        "re:.*self_attn_layer_norm$",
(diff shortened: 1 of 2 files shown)


robertlangdonn:
Good catch — fixed in 104c2ad. The decoder layer has both `self_attn` and `encoder_attn` (cross-attention) blocks with identically-named `q/k/v_proj` children, and `match_modules_set` groups by shared parent context, so the unscoped patterns were pulling `encoder_attn.{q,k,v}_proj` into the same group as `self_attn_layer_norm`. Scoped all three to `self_attn.*` and added a regression test that calls `match_modules_set` directly and asserts no `encoder_attn` module ends up in that group — confirmed it fails without the fix and passes with it.

PR #3076, opened 2026-08-21
imatrix expansion, add fouroversix observer

Roderick-Wu:
Add in imatrix observer, register fouroversix observer built from expanded mse observer. 

Note that fouroversix is NOT identical to fouroversix implementation in https://arxiv.org/abs/2512.02010. 
We get global scale after, and this performs better for most models. 

Results added to:
https://github.com/vllm-project/llm-compressor/pull/2950

Set fouroversix as default for nvfp4:
https://github.com/vllm-project/compressed-tensors/pull/850


kylesayrs (approved):
nj, please add test to verify 4/6 behavior

dsikka (approved):

PR #3077, opened 2026-08-21
MoE Linearization and Offloading Changes

Roderick-Wu:
Move moe linearization and repacking (same operation but reversed) to be handled by the pipeline. Currently linearization is handled at load time, requiring an onload + offload operation. Repacking is not currently used in any pipelines, and is also added to pipelines. 

## Implementation

We add a `disable_offloading_controlled` context which takes in a subgraph, onloads everything, then offloads at the end. Note that we have to wrap the modules in `OnloadWrapper`'s in order to keep track of onloading/offloading. The modules themselves also get a pointer to their `OnloadWrapper`. This is because when we linearize or repack, we actually generate fully new modules, so we need to maintain a way to track the most recent module. 
 * Considering having the `OnloadWrapper` attached to the model, so that it's replace function handles in swapping modules and deleting references from any hashes *

We add flags `sequential_linearize_repack` (Linearize and/or repack each subgraph during calibration onload if `True`, linearize and/or repack before/after the whole model is calibrated for `False) and `repack_moe` (Whether or not to pack experts to 3d). 

Before calibration, we trace the whole model and build a hash for fast lookup for all moe modules -- packed or not packed. This is stored on the model as `model._moe_lookup`. Generated by `get_moe_linear_status`. 

This lookup is important because using model.get_submodules(name) traverses the whole model which is a taxing operation for large models. 

From here, we have functions as follows:
`linearize_moe_model`
`linearize_moe_subgraph`
`linearize_moe_layer`

and their corresponding repacking/reverse counterparts
`repack_moe_model`
`repack_moe_subgraph`
`repack_moe_layer`

The `*_model` functions are used with `sequential_linearize_repack=False`. They iterate over moe layers, onloading/offloading each layer individually (If you can't fit an entire moe layer on gpu at once, you're out of luck). The `*_subgraph` functions are assumed to be under a `disable_offloading_controlled` context. 

Some things to note:

Weird hard-to-read code such as:
```python
    linearized = [
        (moe_lookup[module], module)
        for module in subgraph_set
        if module in moe_lookup and isinstance(module, LinearExperts2D)
    ]

    for i in tqdm.tqdm(range(len(linearized)), desc="Repacking experts in subgraph"):
        repack_moe_layer(model, linearized[i][0], linearized[i][1])
```

Under these new contexts, since we want to constantly free up memory, we need to be careful with how we put out references. Iterating naively `for name, module in linearized` makes it such that linearized maintains a reference to the module. In the case that we are linearizing/repacking and the resulting new module is a new object, we want to free the old object. To my knowledge this can only be done when the python object is garbage collected. So here we make sure that the old module's reference is deleted by the end of its iteration. This is also why we need:

```python
        gc.collect()
        if torch.accelerator.is_available():
            torch.accelerator.empty_cache()
```

At the end of the new load context. We don't generate enough objects to force automatic garbage collection (at least for the tests I've run). The cache release might have a solution but I haven't been able to dig it up yet. Without that part it seems that we constantly have allocated but unused gpu memory. 

Added tests. 

Using example under examples/sequential_linearization/gemma4_example.py:
```python

model_id = "google/gemma-4-26B-A4B"
with load_context():
        model_id,
        device_map="auto_offload",  # fit as much as possible on cpu, rest goes on disk
        #max_memory={0: 0, "cpu": 16e9},  # remove this line to use as much cpu as possible
        offload_folder="./offload_folder",  # folder to store offloaded weights
    )

...


oneshot(
    model=model,
    dataset=DATASET_ID,
    dataset_config_name=DATASET_SUBSET,
    splits=f"train[:{NUM_CALIBRATION_SAMPLES}]",
    recipe=recipe,
    sequential_linearize_repack=True,
    repack_moe_layers=True,
)

```

# Time for one shot
Main: 175s
New (Sequential linearization): 119s
New (Upfront linearization): 139s

Since the timing doesn't include load, the main branch branch takes additional time for load-time linearization. 

The timings for the old version are highly dependent on the offload memory allocation `max_memory`. If we limit gpu memory as well we can throttle to nearly infinite time. This version is always faster. 

Can use the offload examples to compare old/new performance


src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -2,11 +2,19 @@
 from typing import TYPE_CHECKING, Iterator
```
kylesayrs, line 2: Be sure to add support for the other pipelines

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -141,6 +153,11 @@ def __call__(
                 # reduce memory movement by keeping modules onloaded
                 num_batches = len(dataloader)
                 with disable_offloading():
+                    # linearize moe layers just before calibration,
+                    # deferring offloading setup until after compression
+                    linearized = linearize_moe_layer(
+                        model, subgraph.submodules(model), moe_lookup
```
kylesayrs: consider not passing `moe_lookup` into `linearize_moe_layer`, instead just materializing it in the function and caching the result. It makes the function signature simpler and easier to call later

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -58,26 +54,16 @@ def patched(cls, *args, **kwargs):
...
         model_type = config.model_type
 
-        # model is 3d (or otherwise doesn't have mappings)
-        # fall back to post-load conversion
-        if not has_linearize_load_mappings(model_type):
-            model = original_from_pretrained(*args, **kwargs)
-            linearize_moe(model)
-            return model
+        # load model in 3D format — linearization is deferred to the
+        # sequential pipeline for efficient per-subgraph conversion
```
kylesayrs: Any pipeline

src/llmcompressor/modeling/moe/linear_experts.py
kylesayrs, line 53: Ideally, we can replace `copy_from_experts_module` with `view_from_experts_module`, where we get views rather than copying. But this can be left for a later PR

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -142,3 +128,33 @@ def get_non_linearized_moes(
...
+
+    :param model: the full model, used for config fallback and set_submodule
+    :param subgraph_modules: modules in the subgraph to check for experts
+    :param moe_lookup: pre-computed mapping of MoE module -> name in model
+    :return: list of (new LinearExperts2D module, offload kwargs from original)
+    """
+    non_linearized = [(moe_lookup[m], m) for m in subgraph_modules if m in moe_lookup]
+
+    linearized = []
+    for name, module in non_linearized:
```
kylesayrs, line 162: We haven't been doing this up until now, but we should consider deleting the original unlinearized `module`. This isn't important for cpu offloading (as python will garbage collect), but no garbage collection occurs for disk offloading, so we should write a helper `delete_offload_module` which deletes all the parameters off all submodules of a module

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -58,26 +54,16 @@ def patched(cls, *args, **kwargs):
...
-            linearize_moe(model)
-            return model
+        # load model in 3D format — linearization is deferred to the
+        # pipeline for efficient per-subgraph conversion
+        model = original_from_pretrained(*args, **kwargs)
 
-        # prepare to load linearized weights
-        experts_cls, load_map, save_map = get_linearize_load_mappings(model_type)
-        linear_experts_2d_cls = LinearExperts2D.get_linear_experts_cls(experts_cls)
-        register_patch_mapping({experts_cls.__name__: linear_experts_2d_cls})
```
kylesayrs, line 71: I believe you still need this line, otherwise it will try to load a 2d checkpoint into a 3d model

src/llmcompressor/pipelines/basic/pipeline.py
```diff
@@ -54,6 +55,8 @@ def __call__(
         if use_loss_mask:
             session.state.loss_masks = []
 
+        linearize_moe(model)
```
kylesayrs, line 58: Good

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -141,6 +146,9 @@ def __call__(
                 # reduce memory movement by keeping modules onloaded
                 num_batches = len(dataloader)
                 with disable_offloading():
+                    # linearize moe layers just before calibration,
+                    # deferring offloading setup until after calibration
+                    linearized = linearize_moe_layer(model, subgraph.submodules(model))
```
kylesayrs: Spoke offline, will want to add a flag for upfront linearization. The only case we can really think of which this is required is

kimi-k3 decoder => 56Gb BF16
gptq doubles memory => ~100Gb BF16

This goes over typical 80Gb requirement, need granular subgraphs if not X200 machine.

Imho not a case I think we need to build around.

.agents/skills/create-tiny-model/scripts/finetune.py
```diff
@@ -2,18 +2,18 @@
 Fine-tune a language model on famous internet copypastas until target perplexity is reached.
 """
 
+import argparse
```
kylesayrs: These changes to the claude skills seem unrelated? Consider removing their diff

examples/sequential_linearization/gemma4_example.py
```diff
@@ -0,0 +1,63 @@
...
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+from llmcompressor.utils import load_context
+
+start_time = time.time()
+
+# Select model and load it in the `load_context` context
+# In this example, we emulate large model quantization with disk offloading by
+# restricting the theoretical size of CPU RAM to be smaller than the size of the model
```
kylesayrs, line 16: As of the latest `accelerate` version, we can now just pass `max_memory={}` and hf will offloading the entire model to disk

examples/sequential_linearization/gemma4_example.py
```diff
@@ -0,0 +1,63 @@
...
+from compressed_tensors.offload import (
+    get_device_map,
+)
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from llmcompressor import oneshot
+from llmcompressor.modifiers.quantization import QuantizationModifier
+from llmcompressor.utils import load_context
+
+start_time = time.time()
```
kylesayrs, line 12: Is this just a benchmarking script? It probably shouldn't be in the examples

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -298,6 +298,13 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
+    sequential_linearize_moe: bool = field(
+        default=True,
+        metadata={
+            "help": "When using the sequential pipeline, linearize MoE layers to 2D "
```
kylesayrs, line 304: Consider mentioning that this only applies to models which have 3d checkpoints

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -194,7 +212,10 @@ def get_linear_experts_cls(
     @classmethod
     @torch.no_grad()
     def from_experts_module(
-        cls, experts: FusedExpertsProtocol, config: PreTrainedConfig
+        cls,
+        experts: FusedExpertsProtocol,
+        config: PreTrainedConfig,
+        setup_offloading: bool = True,
```
kylesayrs: It seems like you want to pull out the offloading part from this function, which makes sense.

Can we remove the `setup_offloading` and always just set the behavior to not do offloading?

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,27 +133,96 @@ def linearize_moe(model: PreTrainedModel):
...
+            )
+            model.set_submodule(name, linear_moe)
+
+        for submodule in linear_moe.modules():
+            offload_module(submodule, **offload_kwargs)
+
+        # The caching allocator does not return its reserved pool
+        # to the driver on its own. We need to manually release
+        if torch.accelerator.is_available():
+            torch.accelerator.empty_cache()
```
kylesayrs, line 155: Did you find this to be necessary? I generally consider this to be bad practice, as the cuda caching allocator will garbage collect if memory pressure becomes too great

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -51,7 +51,8 @@ def __init__(
         self._apply_gate = _apply_gate
 
     def copy_from_experts_module(self, experts: FusedExpertsProtocol, index: int):
-        # load weights
+        # Rebind linear parameters to views into the fused expert tensors so the
+        # linearized representation can share storage with the original module.
```
kylesayrs, line 61: This gets a bit complicated. [Here](https://github.com/vllm-project/llm-compressor/pull/3077/changes#diff-7fa7c4bb4a7a6087e1af538b307f86d166cff0365bfe9977f9512fd8777df0a4R198-R200) you're performing manual offloading, that way we can be non-offloaded during sequential processing and share views.

I just want to call out that this may make things more complicated: now we have a layer for which some modules are offloaded (everything except moe), and some modules are not offloaded (moe, if the checkpoint is 3d checkpoint). 

This "mixed offloading" state is something we've never handled before. I can't think of any issues offhand, but I think it might be worth leaving the ct offloading changes for later, and instead copy as normal for now.

src/llmcompressor/modeling/moe/llama4.py
```diff
@@ -25,7 +25,12 @@ def _apply_gate(self, gate_up: torch.Tensor) -> torch.Tensor:
 
     @classmethod
     @torch.no_grad()
-    def from_experts_module(cls, experts: "Llama4TextExperts", config: Llama4Config):
+    def from_experts_module(
```
kylesayrs, line 28: Modify  GPT_OSS as well

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -103,11 +107,22 @@ def linearize_moe(model: PreTrainedModel):
...
     (as designated by the `use_experts_implementation` decorator)
 
+    Modules already in LinearExperts2D format are left as-is.
+
     :param model: model containing MoE layers to linearize
     """
-    non_linearized_moes = get_non_linearized_moes(model)
+    # Clear cached lookup to detect experts in the current state of the model
+    if hasattr(model, "_moe_lookup"):
+        delattr(model, "_moe_lookup")
```
kylesayrs, line 116: The problem with this lookup is that is stores strong references to the original, unconverted experts. This means that the memory from the unconverted experts will never be freed. Otherwise a nice idea, if it saves runtime

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,27 +133,96 @@ def linearize_moe(model: PreTrainedModel):
...
+    :param subgraph_modules: modules in the subgraph to check for experts
+    :return: list of (new LinearExperts2D module, offload kwargs from original)
+    """
+    subgraph_set = set(subgraph_modules)
+    moe_lookup = get_non_linearized_moes(model)
+
+    non_linearized = [
+        (moe_lookup[module], module)
+        for module in subgraph_set
+        if module in moe_lookup and not isinstance(module, LinearExperts2D)
```
kylesayrs, line 244: By construction, `moe_lookup` only contains non-linearized moes, right?
```suggestion
        if module in moe_lookup
```

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,27 +133,96 @@ def linearize_moe(model: PreTrainedModel):
...
+        with disable_offloading():
+            offload_kwargs = get_cache_init_kwargs(module)
+            config = getattr(module, "config", model.config)
+            linear_experts_cls = LinearExperts2D.get_linear_experts_cls(
+                module.__class__
+            )
+            linear_moe = linear_experts_cls.from_experts_module(
+                module, config, setup_offloading=False
+            )
+            model.set_submodule(name, linear_moe)
```
kylesayrs, line 147: Seems like there's code duplication between this part and `linearize_moe_layer`

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,27 +133,96 @@ def linearize_moe(model: PreTrainedModel):
...
+    (from checkpoints loaded via patch mappings), capturing offload kwargs for both.
+
+    :param model: the full model, used for config fallback and set_submodule
+    :param subgraph_modules: modules in the subgraph to check for experts
+    :return: list of (new LinearExperts2D module, offload kwargs from original)
+    """
+    subgraph_set = set(subgraph_modules)
+    moe_lookup = get_non_linearized_moes(model)
+
+    non_linearized = [
```
kylesayrs, line 241: We should also check if a (subgraph module) is a submodule of a (non-linearized expert).

That way, if we have two subgraphs of [32/64 experts], then the first subgraph linearizes the whole moe first, despite neither of them actually calling the parent `experts` module

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -141,6 +151,14 @@ def __call__(
                 # reduce memory movement by keeping modules onloaded
                 num_batches = len(dataloader)
                 with disable_offloading():
```
kylesayrs, line 153: BTW. Previously we had a discussion about targeting linear expert submodules, but not being able to since the model wasn't linearized. I'm realizing that this will become a non-issue because
1. There is essentially no reason to target linear experts directly (moe weights must all share the same quant in vLLM)
2. Once we move `apply_quantization_config` to the sequential pipeline, then we can call `apply_quantization_config` after `linearize_moe_layer`, thereby allowing targeting of the model after linearization.

src/llmcompressor/modeling/moe/linearize.py
kylesayrs: On this note, please add a test to verify that such a sequential target still works.

examples/sequential_linearization/gemma4_example.py
```diff
@@ -0,0 +1,63 @@
+import time
```
HDCharles: wouldn't make a new folder just for this, probably just take an existing example and make it X_sequential_linearization_example.py

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -298,6 +298,13 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
+    sequential_linearize_moe: bool = field(
```
HDCharles: are we set on this name, shouldn't it be like just_in_time_linearize or jit_linearize or something like that?

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -168,16 +198,70 @@ def get_non_linearized_moes(
...
+        model._moe_lookup = {
+            module: name
+            for name, module in model.named_modules()
+            if isinstance(module, FusedExpertsProtocol)
+            or LinearExperts2D.get_registration(module.__class__) is not None
+        }
+    return model._moe_lookup
+
+
+def linearize_moe_layer(
```
HDCharles: now that we have 2 functions can we name them more clearly?

linearize_moe_subgraph and linear_moe_model?

src/llmcompressor/modeling/moe/linearize.py
HDCharles: this name is bad now that we have functionality for the whole model and for just a subgraph, 

maybe linear_moe_model/linearize_moe_subgraph?

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -168,16 +198,70 @@ def get_non_linearized_moes(
...
+
+    linearized = []
+    for name, module in tqdm.tqdm(
+        non_linearized, desc="Linearizing experts in subgraph"
+    ):
+        offload_kwargs = get_cache_init_kwargs(module)
+        config = getattr(module, "config", model.config)
+        linear_experts_cls = LinearExperts2D.get_linear_experts_cls(module.__class__)
+        linear_moe = linear_experts_cls.from_experts_module(
+            module, config, setup_offloading=False
```
HDCharles: doesn't this need to be added to the base module? 

https://github.com/vllm-project/llm-compressor/blob/c7016a391fd272ead2be8d2775ee01f3e65c023d/src/llmcompressor/modeling/moe/linear_experts.py#L261

i.e. no setup_offloading argument

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -68,19 +69,30 @@ def copy_from_experts_module(self, experts: FusedExpertsProtocol, index: int):
...
-        self.down_proj.weight.copy_(down_weight)
+        self.gate_proj.weight = torch.nn.Parameter(
+            gate_weight, requires_grad=experts.gate_up_proj.requires_grad
+        )
+        self.up_proj.weight = torch.nn.Parameter(
+            up_weight, requires_grad=experts.gate_up_proj.requires_grad
+        )
+        self.down_proj.weight = torch.nn.Parameter(
+            down_weight, requires_grad=experts.down_proj.requires_grad
+        )
```
HDCharles, line 79: done this way, non contiguous tensors are going to be in the linear weights which i think can break safetensors when serializing?

probably want a unit test to make sure this is supported

src/llmcompressor/modeling/moe/linearize.py
Roderick-Wu: I'm thinking these get left as strong refs and get managed by the onloadwrapper, but weak refs might be better

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -298,13 +298,18 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
-    sequential_linearize_moe: bool = field(
+    sequential_linearize_repack: bool = field(
```
HDCharles: not sure about this name, this new name doesn't address my issue with the previous name and now removes the MoE context and introduces false impression of normal repack on save type functionality.

Roderick-Wu: how about like
`moe_jit_linearize_repack`
or something

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -298,6 +298,18 @@ class DatasetArguments(CustomDatasetArguments):
...
     )
+    sequential_linearize_repack: bool = field(
+        default=True,
+        metadata={
+            "help": "When using the sequential pipeline, linearize MoE layers to 2D "
+            "before calibration. Note that this only impacts models which have 3D "
+            "MoE layers. Default True. Set False to linearize upfront. "
+        },
+    )
+    repack_moe_layers: bool = field(
```
HDCharles, line 310: save_moe_repacked

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -93,22 +96,118 @@ def patched(cls, *args, **kwargs):
...
+    """
+
+    if hasattr(model, "_moe_lookup"):
+        delattr(model, "_moe_lookup")
+
+    model._moe_lookup = {
+        module: name
+        for name, module in model.named_modules()
+        if isinstance(module, FusedExpertsProtocol)
+        or isinstance(module, LinearExperts2D)
```
HDCharles: this needs to respect the registrations too

LinearExperts2D._registry[Llama4TextExperts] = Llama4LinearExperts

src/llmcompressor/args/dataset_arguments.py
```diff
@@ -298,6 +298,19 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
+    moe_eager_linearization_and_repack: bool = field(
```
HDCharles, line 301: i think this should be lazy, eager implies it happens immediately i.e. upfront

Roderick-Wu:
Moved to https://github.com/vllm-project/llm-compressor/pull/3208

PR #3078, opened 2026-08-22
Fix split_fused_moe_experts for transposed MoE weights (Llama-4)

HDCharles:
## Summary

- `split_fused_moe_experts` assumes fused MoE expert weights are always stored as `[E, fused_intermediate, hidden]`, but Llama-4 uses a transposed layout `[E, hidden, fused_intermediate]`
- Without this fix, the split produces wrong-shaped tensors (e.g. `[2560, 16384]` instead of `[8192, 5120]`), which silently passes through model-free PTQ but errors at vLLM load time
- Detects the transposed format by reading `intermediate_size` from model `config.json` and comparing against the actual tensor dimensions — no shape heuristic needed
- Falls back to the `shape[2] > shape[1]` heuristic only when `moe_intermediate_size` is not provided (e.g. standalone callers outside the `model_free_ptq` pipeline)

## Repro

```python
import torch
from llmcompressor.entrypoints.model_free.process import split_fused_moe_experts

# Llama-4-Scout-like dimensions
num_experts, hidden, intermediate = 16, 5120, 8192

# Llama-4 stores experts transposed: [E, hidden, fused_intermediate]
tensors = {
    "model.layers.0.mlp.experts.gate_up_proj.weight": torch.randn(
        num_experts, hidden, 2 * intermediate, dtype=torch.float16
    ),
    "model.layers.0.mlp.experts.down_proj.weight": torch.randn(
        num_experts, intermediate, hidden, dtype=torch.float16
    ),
}

result = split_fused_moe_experts(tensors, moe_intermediate_size=intermediate)

# On main: gate_proj is [2560, 16384] — WRONG
# With fix: gate_proj is [8192, 5120] — correct
gate = result["model.layers.0.mlp.experts.0.gate_proj.weight"]
assert gate.shape == (intermediate, hidden), f"got {gate.shape}, expected ({intermediate}, {hidden})"
```

## Test plan

- [x] New unit test `test_split_fused_moe_experts_transposed` — creates standard and transposed tensors, splits both with `moe_intermediate_size`, asserts identical shapes and values
- [x] Verified test fails on main and passes with fix
- [x] Existing tests still pass (standard format unaffected)

### DIFF
--- a/src/llmcompressor/entrypoints/model_free/__init__.py
+++ b/src/llmcompressor/entrypoints/model_free/__init__.py
@@ -1,3 +1,4 @@
+import json
 import os
 import shutil
 from pathlib import Path
@@ -86,6 +87,8 @@ def model_free_ptq(
     validate_safetensors_index(model_files, config)
     os.makedirs(save_directory, exist_ok=True)
 
+    moe_intermediate_size = _get_moe_intermediate_size(model_files)
+
     # copy non-safetensors files (configs, tokenizers, etc.)
     for file_path, resolved_path in model_files.items():
         if not file_path.endswith("safetensors"):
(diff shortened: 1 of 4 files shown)


kylesayrs (approved):
I'd prefer to just make this explicit (with warning and suggestion), that way we can set expectations about when this feature will be active

PR #3079, opened 2026-08-22
fix: use forward globals when autowrapping

### LINKED ISSUE #3065: [Bug]: autowrap_forward fails with NameError for models defined in __main__ under cProfile/runpy

<details>

```text
Operating System: `macOS-26.5.2-arm64-arm-64bit`
Python Version: `3.12.5`
llm-compressor Version: `0.13.1.dev24+g48116705`
compressed-tensors Version: `0.18.1a20260818`
transformers Version: `5.15.0`
torch Version: `2.13.0`
CUDA Devices: `None`
AMD Devices: `None`
NPU Devices: `None`
MPS Devices: `['arm']`
```

</details>

Note: `tools/collect_env.py` currently crashes on Apple Silicon, so the output above required a local patch. Filed separately as #3064.

### 🐛 Describe the bug

`autowrap_forward` builds the namespace for the recompiled forward from `sys.modules[module.__class__.__module__].__dict__` in `src/llmcompressor/pipelines/sequential/ast_helpers.py`.

For a model class defined in a script, `__module__` is `"__main__"`. Runners such as `cProfile` and `runpy` exec the script into a fresh globals dict without installing it as `sys.modules["__main__"]`, so that lookup returns the runner's own module namespace instead of the script's. Names from the script are then missing, and the subsequent `exec` raises:

```
RuntimeError: --- <Autowrapped ParameterAccessWrapper ...>:1 ---
name 'torch' is not defined
```

Minimal demonstration of the cause:

```python
import sys, torch

class Wrapper(torch.nn.Module):
    def forward(self, x):
        return torch.relu(x)

mod = sys.modules.get(Wrapper.__module__)
print(mod.__file__, "torch" in mod.__dict__)
```

- direct run: `/tmp/diag.py True`
- under `python3 -m cProfile`: `.../cProfile.py False`

I hit this trying to profile `benchmarks/bench_trace_subgraphs.py`, whose `ParameterAccessWrapper` is defined in the benchmark script.

Proposed fix: autowrap_forward already calls inspect.getsource(module.forward), and inspect.getsourcelines calls unwrap internally — so the parsed source belongs to the unwrapped function. Taking the namespace from that same function keeps source and globals consistent by construction:

```python
forward_fn = inspect.unwrap(module.forward)
namespace = getattr(forward_fn, "__globals__", None)
if namespace is None:
    namespace = sys.modules[module.__class__.__module__].__dict__
namespace = namespace.copy()
```

The `inspect.unwrap` is load-bearing. Using `module.forward.__globals__` directly fails for HF models, since `forward` is decorated and `functools.wraps` does not copy `__globals__` — you get the decorator's namespace while `getsource` returns the original's. I confirmed this: it breaks 6 tests including `test_trace_subgraphs[1-5]`.

With `unwrap`, all 32 tests in `tests/llmcompressor/pipelines/sequential/` pass and the reproduction below completes. Happy to open a PR, with a subprocess-based regression test if you would like one.

### 🛠️ Steps to reproduce

Fails:

```bash
python3 -m cProfile -o /tmp/x.prof benchmarks/bench_trace_subgraphs.py \
  --num-targets 2000 --targets-per-subgraph 300 \
  --num-parameters 5000 --num-parameter-accesses 20 --repeat 1
```

Succeeds — identical arguments, without `-m cProfile`:

```bash
python3 benchmarks/bench_trace_subgraphs.py \
  --num-targets 2000 --targets-per-subgraph 300 \
  --num-parameters 5000 --num-parameter-accesses 20 --repeat 1
```

Also reproduces under `runpy`, or with any model class defined in `__main__`.

mikemikimike:
Fixes #3065.

utowrap_forward previously selected the namespace from sys.modules[module.__class__.__module__]. When a model class is defined in __main__ and the script is executed through cProfile or 
unpy, that module entry can refer to the runner namespace rather than the namespace that owns the forward function. The generated function then fails with a NameError for globals such as 	orch.

This change unwraps decorated forward methods before retrieving source and globals, uses the unwrapped function's __globals__ when available, retains the existing module lookup as a compatibility fallback, and adds a regression test for the __main__ mismatch.

Verification:
- pytest focused regression: 1 passed, 13 deselected
- ruff check and format check: passed
- mypy focused check: passed
- git diff --check: passed

The test exercises the real utowrap_forward context manager without model downloads, GPU requirements, or external APIs.

### DIFF
--- a/src/llmcompressor/pipelines/sequential/ast_helpers.py
+++ b/src/llmcompressor/pipelines/sequential/ast_helpers.py
@@ -53,13 +53,18 @@ def autowrap_forward(module: torch.nn.Module, ignore: list[str]):
         )
 
     # get source code of module forward
-    source = inspect.getsource(module.forward)
+    forward_fn = inspect.unwrap(module.forward)
+    forward_fn = getattr(forward_fn, "__func__", forward_fn)
+    source = inspect.getsource(forward_fn)
     source = textwrap.dedent(source)
     tree = ast.parse(source)
 
     # construct namespace for our new code
-    defining_module = sys.modules[module.__class__.__module__]
-    namespace = defining_module.__dict__.copy()
+    namespace = getattr(forward_fn, "__globals__", None)
+    if namespace is None:
+        defining_module = sys.modules[module.__class__.__module__]
+        namespace = defining_module.__dict__
+    namespace = namespace.copy()
     namespace.update({"torch.fx.wrap": torch.fx.wrap})
     namespace.update({"self": module})
 
--- a/tests/llmcompressor/pipelines/sequential/test_helpers.py
+++ b/tests/llmcompressor/pipelines/sequential/test_helpers.py
@@ -1,4 +1,5 @@
 import math
+import sys
 
 import pytest
 import torch
@@ -6,6 +7,7 @@ import torch.fx
 from transformers import AutoModelForCausalLM
 
 from llmcompressor.args.dataset_arguments import DatasetArguments
+from llmcompressor.pipelines.sequential.ast_helpers import autowrap_forward
 from llmcompressor.pipelines.sequential.helpers import (
     Subgraph,
     get_sequential_ancestors,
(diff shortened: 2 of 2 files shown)


mikemikimike:
Thanks for catching this. The review is correct: module.forward is a bound method, so inspect.unwrap alone does not expose __globals__. I updated the implementation to retrieve __func__ after unwrapping, and strengthened the regression test by removing 	orch from the __main__ namespace. The test now proves the function globals path is used rather than passing via the old fallback. Focused test: 1 passed; Ruff and git diff --check pass.

kylesayrs (approved):
I believe this is going to be more robust anyways, thanks for the awesome contribution!

1. Please take care that `test_models` passes (this should be caught by the CI)
2. Please make sure to resolve conflicts with https://github.com/vllm-project/llm-compressor/pull/3058 once it lands.

Isitthakkar11:
Ran the full tests/llmcompressor/pipelines/sequential/ suite against this approach locally — all 32 pass, including test_trace_subgraphs[1-5]. Worth flagging that those five are the ones that catch it if the unwrap is dropped: without it you get the decorator's namespace instead of the original's, since functools.wraps doesn't copy __globals__.

kylesayrs:
@mikemikimike @Isitthakkar11 Thanks for the awesome contributions!

PR #3080, opened 2026-08-22
feat(moe): add patch_moe_mappings to override 2D load mappings per checkpoint

Anai-Guo:
SUMMARY:

Follow-up to #3037 / #3050, implementing the interface @kylesayrs sketched when closing #3050:

> I think the best course of action is to open a new PR which provides an interface to override load mappings.
> ```python
> patch_moe_mappings("qwen3_5_moe", [... "3d load mappings", ...])
>
> with load_context():
>     model = AutoModelForCausalLM.from_pretrained(...)
> ```

@kgt392 said in that thread they would not be picking it up, so this is not stepping on anyone.

**Why an override rather than another table entry.** Whether experts arrive as fused 3D weights or as 2D per-expert weights is a property of the *checkpoint*, not of the architecture. `Qwen/Qwen3.5-35B-A3B` stores 3D, while derivatives that were loaded and re-saved through transformers (`ornith-ai/Ornith-1.0-35B`, `brayniac/Qwen3.5-35B-A3B-heretic`, …) store 2D under the same `model_type`. `ARCH_TO_2D_MAPPINGS` is keyed by architecture and so has to pick one, which is exactly what made #3050 a break for the common case. Those checkpoints currently fall back to post-load linearization and pay the `2D -> 3D -> 2D` round trip that `linearize_moe` warns about.

**The change.**

- `patch_moe_mappings(model_type, mappings, remove_targets=None)` registers 2D load mappings at runtime, so `load_context()` takes the direct-load pathway for that checkpoint. Mappings are registered against the architecture's conversion pattern, matching how `has_linearize_load_mappings` / `get_linearize_load_mappings` look them up. A `model_type` with no registered experts module raises `ValueError` instead of failing later with a `KeyError` on `ARCH_TO_IMPORT_PATHS`.
- `get_linearize_load_mappings` now tolerates an architecture with no default conversion mapping (`get_checkpoint_conversion_mapping(...) is None`). This is not reachable today, but it is reachable the moment the override exists: `qwen3_5_moe` — the model type from #3037 — is the one registered architecture with no default mapping, so the first thing this feature is used for would have hit it. See the test plan for the before/after.
- Docs: a short subsection under *MoE Conversion Before Loading* covering when to override rather than contribute a table entry.

Deliberately out of scope: the override only accepts architectures already in `ARCH_TO_IMPORT_PATHS`, so it does not become a second way to register an experts module. Happy to change the signature (e.g. accept the `(remove_targets, mappings)` tuple shape of the table directly, or make it a context manager) if you would rather it look different — the shape here is the smallest thing matching the sketch.

TEST PLAN:

New CPU-only unit tests in `tests/llmcompressor/modeling/test_conversion_mappings.py`: the override turns a previously unmapped architecture into a direct-load one and its mappings come back out of `get_linearize_load_mappings`; `remove_targets` drops the named targets of the architecture's default mapping; an unregistered `model_type` raises without touching the table.

I do not have hardware to load a real 2D Qwen3.5-MoE checkpoint, so what I verified is mapping resolution, not an end-to-end load or the memory numbers in #3037.

Against `transformers==5.15.0`, resolving every one of the 31 `ARCH_TO_IMPORT_PATHS` entries through `has_linearize_load_mappings` / `get_linearize_load_mappings` and comparing the transform lists element-by-element gives identical results before and after this change — 15 direct-load, 16 fallback, same transforms:

```
upstream == patched for all 31 registered arches: True
```

And with a 2D entry registered for `qwen3_5_moe`, which is what the new function does:

```
upstream get_* with a 2D entry -> TypeError: 'NoneType' object is not iterable
patched  get_* with a 2D entry -> ok, n_load = 1
```

`ruff check` and `ruff format --check` are clean on the changed files.


src/llmcompressor/modeling/moe/conversion_mappings.py
```diff
@@ -259,6 +261,55 @@ def get_linearize_load_mappings(
...
+    which already stores its experts as 2D per-expert tensors is loaded directly in
+    linearized format rather than being fused into 3D weights on load and unfused
+    again afterwards.
+
+    Not every checkpoint of an architecture shares a layout: a model which has been
+    loaded and re-saved through transformers stores 2D per-expert weights where the
+    original release stores fused 3D weights. Which mappings apply is therefore a
+    property of the checkpoint rather than of the architecture, and cannot be decided
+    by `ARCH_TO_2D_MAPPINGS`. Use this to declare the layout of such a checkpoint:
+
```
kylesayrs, line 278: Please add context of the ornith-ai/Ornith-1.0-35B example to justify why this is needed

kylesayrs (approved):
Looks great! Feel free to add the `ornith-ai/Ornith-1.0-35B` example as well either here or in a follow up

Anai-Guo:
Friendly ping @maintainers — this one is approved and still `MERGEABLE`. Is there anything else needed before it can be merged?

yiliu30:
@codex review

Anai-Guo:
Thanks @yiliu30 for running the review, and @kylesayrs — pushed `ece72c5` addressing both threads.

**Scoping the override (codex P2, and gemini's note):** the finding is right, and it matters exactly because the mappings describe a *checkpoint*, not an architecture. Overriding `qwen3_5_moe` for a 2D checkpoint and then loading the fused original release in the same process would have left `has_linearize_load_mappings()` returning `True` and applied the 2D mappings to it.

`patch_moe_mappings` is now a context manager that restores the previous entry (or removes it, when there was none) on exit, and the docs/docstring lead with the `with` form:

```python
with patch_moe_mappings("qwen3_5_moe", [...]):
    with load_context():
        model = AutoModelForCausalLM.from_pretrained(model_id)
```

The override is still applied eagerly in `__init__`, so a bare `patch_moe_mappings(...)` call keeps working and keeps lasting for the rest of the process — existing usage is unchanged.

**The `Ornith-1.0-35B` context:** added to both the docstring and the tutorial as the motivating checkpoint. It is a `qwen3_5_moe` model, but its `model.safetensors.index.json` stores experts as 2D `experts.{i}.{gate,up,down}_proj.weight` tensors (10240 of each across its 40 layers), so without a patch it fuses every expert to 3D on load only for linearization to unfuse it again.

**Tests:** two new cases cover restoration — one for an architecture with no prior entry, one for `qwen2_moe`, which has a default conversion mapping that must come back intact. Both fail against the previous commit and pass now (`5 passed`).

yiliu30 (approved):
LGTM

PR #3018, opened 2026-08-11
[lut-b] data-free round-to-nearest example

brian-dellabetta:
Prerequisite
* https://github.com/vllm-project/compressed-tensors/pull/801

SUMMARY:
Adds an example for pure round-to-nearest quantization using the LUT-B compressor and QDQ pathway, targeting only MLP layers at the moment.

LUT-B compression is currently disconnected from Observers, which at the moment return `weight_scale`, `weight_zero_point`, and `weight_global_scale` qparams but do not yet return `weight_codebook`. It is not clear if it is worth adding to Observers throughout, as a custom `CodebookObserver`, or outside of the Observer abstraction. @HDCharles and I discussed this can be left as a TODO after first determining how valuable GPTQ is for codebook quant.

TEST PLAN:
Checkpoint -- https://huggingface.co/bdellabe/Meta-Llama-3-8B-Instruct-LUTB

- [x] safetensors file looks correct -- weight_packed and weight_codebook tensors for targeted layers
output of example:
- [x] reasonable output 
  - [x] both in example before save call
  - [x] and after loading up in transformers in a fresh session

> <|begin_of_text|>Hello my name is, and I am a
I am a 31-year-old male from the city of New York. I am a software engineer by trade, but I am also a passionate musician and music producer. I am a bit of a perfectionist and I am always striving to improve my skills and knowledge.
> I am a bit of a introvert and I prefer to spend my free time alone or with close friends and family. I am also a bit of a homebody and I prefer to stay close to

- [x] `meta-llama/Meta-Llama-3-8B-Instruct` PPL scores for LUTB are reasonably in between W4A16 and W3A16. Unfortunately, SpinQuant does not seem to improve and PPL increases monotonically with hadamard block size. Results below are for targeting only MLP layers

| Format / Configuration | Value (LUTB) |
| :--- | :--- |
| **bfloat16** | 10.0523 |
| **LUTB** | 13.3204 |
| **LUTB R1 had2** | 13.4095 |
| **LUTB R1 had8** | 13.8657 |
| **LUTB R1 had128** | 14.5493 |
| **W4A16** | 10.5297 |
| **W4A16 R1 had128** | 10.7515 |
| **W3A16** | 21.9974 |
| **W3A16 R1 had128** | 17.006 |

PR #3081, opened 2026-08-22
Add SpinQuant support for Qwen3.5 hybrid attention

machero:
## Summary

Adds SpinQuant support for Qwen3.5, which uses a hybrid attention architecture
(a mix of full self-attention and linear / Gated DeltaNet layers). SpinQuant
previously had no mappings for this layout, so the modifier could not be applied
to Qwen3.5 models.

## Background

Qwen3.5 alternates full-attention layers (`q/k/v/o_proj`) with linear-attention
layers (`linear_attn.in_proj_qkv/z/b/a` + `linear_attn.out_proj`). The existing
SpinQuant mapping inference assumes a uniform self-attention layout, so it cannot
correctly identify the residual-stream read/write projections in the linear-
attention layers.

This change mirrors the dynamic-mapping approach already used by AWQ for hybrid
attention, generalized for SpinQuant.

## Changes

- **`spinquant/dynamic_mappings.py` (new)**: runtime mapping generation for dense
  hybrid-attention models.
  - Folds `linear_attn.in_proj_*` into `mlp_in` and `linear_attn.out_proj` into
    `mlp_out`, so R1 rotates the residual stream consistently across both
    full- and linear-attention layers.
  - Layer-index-restricted `input_layernorm` norm mappings so that full- and
    linear-attention layers each group exactly one norm per layer.
- **`spinquant/base.py`**: infer mappings via
  `get_spinquant_mapping_from_model` / `get_norm_mapping_from_model`; read the
  attention `head_dim` from `text_config` for heterogeneous configs.
- **`spinquant/mappings.py` / `spinquant/norm_mappings.py`**: move mapping
  inference out of the static registries; they now hold data only.
- **`utils/hybrid_attention.py`**: move `detect_linear_attn_projections` into a
  shared util, reused by AWQ (`awq/dynamic_mappings.py`).

## Why only R1 is supported

R2 rotates the value/output head space. Qwen3.5's full-attention layers apply an
element-wise gate (`attn_output * sigmoid(gate)`) between the attention output
and `o_proj`, inside that same head space, which does not commute with R2 and
breaks logits invariance. R3/R4 are online rotations and are unaffected by these
mappings. R1 alone preserves logits invariance (measured below).

## Testing

- Added unit tests for the dynamic mappings and fallback logic
  (`spinquant/test_dynamic_mappings.py`); updated AWQ's
  `test_dynamic_mappings.py` for the shared util move.
- Added an end-to-end R1 logits-invariance test on a tiny Qwen3.5 model
  (`spinquant/test_e2e_qwen3_5.py`), plus a `head_dim`-from-`text_config` test.
  **33 tests pass.**
- Verified R1 logits invariance on Qwen3.5-27B: MSE ≈ 1e-3 (below the existing
  SpinQuant correctness tolerance of 8e-3).

## Results (Qwen3.5-27B, WikiText-2 perplexity)

| method | PPL |
|---|---|
| baseline-fp16 | 6.4063 |
| rtn-fp8 | 6.4849 |
| rtn-w8a8 | 6.5804 |
| awq-w4a16 | 6.6177 |
| **spinquant-w4a16 (R1)** | **6.6233** |
| autoround-w4a16 | 6.7211 |
| smoothquant-w8a8 | 6.7306 |
| gptq-w4a16 | 6.9394 |
| rtn-w4a16 | 6.9595 |

SpinQuant W4A16 is data-free (no calibration data required) and matches AWQ W4A16,
which uses calibration data, while outperforming the other 4-bit methods.


src/llmcompressor/modifiers/transform/spinquant/base.py
```diff
@@ -21,8 +21,12 @@
 from llmcompressor.typing import NamedModules
 from llmcompressor.utils import get_high_precision, untie_word_embeddings
 
-from .mappings import SpinQuantMapping, infer_mapping_from_model
-from .norm_mappings import NormMapping, infer_norm_mapping_from_model
+from .dynamic_mappings import (
+    get_norm_mapping_from_model,
```
brian-dellabetta: rather than change the API, why not just update the infer methods to include the mappings that need to be built from the model?

brian-dellabetta (commented):
Hi @machero , thanks for preparing this. your results that spinquant R1 outperforms GPTQ seem promising. We were unable to show this in our experiments, but we never went beyond the llama architecture. i will take some time over the next couple days validating this

brian-dellabetta:
Hi @machero , can you please provide the script and benchmark you are running? On your branch, I am able to run spinquant R1 on Qwen 3.5 and 3.8 27B, and the model output is reasonable, but I am not seeing it perform better than RTN on wikitext PPL, as you are claiming in your PR summary table:

Screenshot 2026-09-03 at 4 07 48 PM

Script i'm using:

```python
from compressed_tensors.offload import dispatch_model

from llmcompressor.modifiers.transform import SpinQuantModifier

# Select model and load it.
MODEL_ID = "Qwen/Qwen3.5-27B"

model = AutoModelForCausalLM.from_pretrained(MODEL_ID)

# NOTE: currently only rotations R1, R2, and R4 are available
# R3 and learned R1/R2 rotations will be added in a future release.
recipe = [
    SpinQuantModifier(
        rotations=["R1"],
        transform_block_size=128,
        transform_type="hadamard",
    ),
    QuantizationModifier(targets="Linear", scheme="W4A16", ignore=["lm_head"]),
]

# Apply algorithms.
oneshot(model=model, recipe=recipe, pipeline="datafree")

SAVE_DIR = MODEL_ID.split("/")[1] + "-spinquantR1had128-w4a16"
```

lm_eval command i'm using (tried vllm as well):

```
lm_eval \
  --model hf \
  --model_args pretrained=Qwen3.5-27B-rtn-w4a16,add_bos_token=True,dtype=auto \
  --tasks wikitext \
  --batch_size auto
```

machero:
Thanks for testing it — the discrepancy is almost certainly the `ignore` list, not the SpinQuant transform.

Reproduction script: https://gist.github.com/machero/ed285e2bca6ed01e58c73e325be58873

My results skip the Gated DeltaNet projections `linear_attn.in_proj_a` / `in_proj_b` (along with `lm_head` and the visual tower), matching the repo's own example `examples/quantization_w4a16/qwen3_8_gptq_awq_example.py`. Those two projections are highly sensitive to 4-bit quantization and dominate WikiText PPL; quantizing them (i.e. with `ignore=["lm_head"]`) degrades every W4A16 method and masks the SpinQuant gain.

The reproduction script now has an `--ignore` preset so both configurations can be reproduced side by side:

```
python reproduce.py all                          # --ignore full  (PR results)
python reproduce.py all --ignore lm_head         # your configuration
```

- `python reproduce.py verify-invariance` reproduces the R1 logits-invariance check (MSE ≈ 1e-3, tolerance 8e-3) on Qwen3.5-27B.
- `python reproduce.py verify-invariance --rotations R1 R2` shows the R2 drift that motivates "R1 only".
- `python reproduce.py all` reproduces the 9-method WikiText-2 PPL table (spinquant-w4a16 = 6.6233 ≈ AWQ W4A16 = 6.6177, with no calibration data).

Also note the model class (`Qwen3_5ForConditionalGeneration` vs `AutoModelForCausalLM`) and the PPL harness (custom stride-512 sliding window over wikitext-2-raw-v1 vs `lm_eval`) — both are encoded in the script. The `get_*_mapping_from_model` → `infer_*_from_model` refactor is pushed. Tests: `pytest tests/llmcompressor/modifiers/transform/spinquant/` → 11 passed; `ruff check` / `ruff format --check` clean.

brian-dellabetta:
Hi @machero , I have found that your ppl evals are flipping the values i'm seeing with lm_eval:

| eval harness | RTN | SpinQuant | winner |
|---|---|---|---|
| your gist token-PPL (stride-512, short left-context) | 6.83 | 6.67 | SpinQuant |
| lm_eval `wikitext` `word_perplexity` (full rolling context) | 7.85 | 8.51 | RTN |

You can confirm by running lm_eval on your side
```
lm_eval \
  --model hf \
  --model_args pretrained=Qwen3.5-27B-rtn-w4a16,add_bos_token=True,dtype=auto \
  --tasks wikitext \
  --batch_size auto
```

This is an interesting result. I bet if we go down to `W3A16` the actual signal will appear, that spinquant does improve (see table in pr summary [here](https://github.com/vllm-project/llm-compressor/pull/3018)). We never tested below 4-bit in our ablations

Unfortunately, this will be my last day maintaining llm-compressor 🥹 . I will loop in @kylesayrs to take over from here

machero:
Thanks @brian-dellabetta  — appreciate you running this, and good luck with what's next.
I'll re-benchmark on lm_eval wikitext word_perplexity as the standard harness and add a W3A16 comparison (W4A16 kept for completeness) to see if the signal shows up there.

PR #3091, opened 2026-08-25
Move Observer Resolution and Defaulting to LLM-Compressor

Roderick-Wu:
With
https://github.com/vllm-project/compressed-tensors/pull/853

Move defaulting and quantargs resolution from ct to lc. After recipe creation, recipe calls `ModifierFactory.create( ...` which creates `QuantizationModifiers`. From here, we construct `QuantizationScheme`, `QuantizationArgs`. `QuantizationArgs` are pydantic validated upon creation. 

We move validating args to the creation of `QuantizationModifiers`. Define `_resolve_quantization_config_args()` under the mixin, validated upon creation. 

Observers are not part of args, and are resolved seperately. We do this at the start of calibration instead.


src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -142,6 +151,104 @@ class QuantizationMixin(HooksMixin):
     _calibration_hooks: set[RemovableHandle] = PrivateAttr(default_factory=set)
     _resolved_config: QuantizationConfig | None = PrivateAttr(None)
 
+    @staticmethod
+    def _resolve_quantization_args(value: dict[str, Any]) -> dict[str, Any]:
```
kylesayrs: This kind of defaulting logic should exist directly on the `QuantizationArgs`, as doing so ensures portability outside of just LLM Compressor.

The only field that should be modified by LLM Compressor after the qargs have been is the observer, since the observer is the only field that is related to algorithm selection during compression.

src/llmcompressor/modifiers/quantization/calibration.py
```diff
@@ -33,6 +35,25 @@
...
+            warnings.warn(
+                "No observer is used for dynamic quantization; ignoring "
+                f"observer={args.observer!r}",
+                UserWarning,
+                stacklevel=2,
+            )
+        args.observer = None
+    elif args.observer is None:
+        args.observer = (
+            "minmax" if args.dynamic == DynamicType.LOCAL else "memoryless_minmax"
```
kylesayrs: This is wrong. Static FP8 schemes have `args.dynamic == False`, but need the `"minmax"` observer.

The real defaulting rule is:
If doing weight quantization, use 4over6 if nvfp4, use memorylesss_minmax otherwise
If doing input quantization, use minmax

kylesayrs (changes requested):
I strongly disagree with moving all defaulting and parsing logic to LLM Compressor. All of this functionality is already built into the pydantic validation implemented on `QuantizationArgs/Scheme`. The only work should be to have `QuantizationArgs` default to `observer=None`, and have LLM Compressor set the observer in `_apply_observer_overrides`

src/llmcompressor/modifiers/quantization/quantization/mixin.py
```diff
@@ -421,17 +421,23 @@ def _apply_observer_overrides(
...
+            (output_obs, "output_activations", "minmax"),
         ]
 
-        for obs_value, scheme_attr in update_map:
+        for obs_value, scheme_attr, default_obs in update_map:
             q_args = getattr(scheme, scheme_attr, None)
-            if obs_value is not None and q_args is not None:
-                args_dict = q_args.model_dump()
-                args_dict["observer"] = obs_value
-                setattr(scheme, scheme_attr, QuantizationArgs.model_validate(args_dict))
```
Roderick-Wu, line 437: removed because no longer set in CT

tests/llmcompressor/modifiers/calibration/test_observers.py
```diff
@@ -15,25 +15,29 @@
     [
         ((1, 1), None, False),
         ((1, 1), 1, False),
-        ((1, 1), 1, True),
+        ((1, 1), 1, "weight"),
```
Roderick-Wu, line 18: needed because of older actorder changes, `True` is now invalid

HDCharles (approved):

kylesayrs (approved):
Nice job!

PR #3096, opened 2026-08-25
[Bugfix] Fix polynomial_decay pruning scheduler for even exponents

winklemad:
## Summary

`polynomial_decay_scheduler` computes the AGP (Zhu & Gupta, 2017) sparsity decay factor as:

```python
scaled_complete = pow(per_complete - 1, exponent) + 1
```

This only equals the intended `1 - (1 - per_complete)**exponent` when `exponent` is **odd**. For **even** exponents — including the default of `2` — `(per_complete - 1)**exponent == (1 - per_complete)**exponent`, so the decay term keeps the wrong sign: sparsity starts *above* `final_sparsity` and *decreases* over training instead of rising from `init_sparsity`.

`cubic_scheduler` sets `exponent=3` (odd), so it happens to be correct — which is likely why this went unnoticed — but selecting `"polynomial_decay"` (default `exponent=2`) or any even exponent produces a broken schedule.

### Reproduce

```python
from llmcompressor.core import Event
from llmcompressor.modifiers.pruning.helpers import (
    polynomial_decay_scheduler, PruningCreateSettings,
)

s = PruningCreateSettings(start=0, end=10, update=1,
                          init_sparsity=0.0, final_sparsity=0.9, args={})
sched = polynomial_decay_scheduler(s)   # default exponent=2

def at(i):
    e = Event(); e.current_index = i
    return sched(e, None)

print(at(0), at(5), at(10))   # 1.8  1.125  0.9   -> starts at 1.8 (>1.0) and falls
# expected a rise 0.0 -> ~0.675 -> 0.9
```

## Fix

Use `1 - pow(1 - per_complete, exponent)`, which is correct for all exponents and **identical to the previous behaviour for odd exponents**, so `cubic_scheduler` (exponent=3) is byte-for-byte unchanged.

After the fix the same schedule rises `0.0, 0.171, 0.324, …, 0.9`.

## Testing

Adds `tests/llmcompressor/modifiers/pruning/test_helpers.py`:
- `polynomial_decay` is monotonic and bounded in `[init, final]` for exponents 2, 3, and 4;
- the default (even) exponent starts at `init_sparsity`;
- `cubic_scheduler` output is unchanged.

The even-exponent tests fail on `main` and pass with the fix. `ruff check` / `ruff format --check` clean.


kylesayrs (approved):
You are indeed correct, thanks for the fix!

winklemad:
Thanks for the quick review and approve, @kylesayrs!

dsikka:
@winklemad do you mind looking into the failing tests?

winklemad:
Looked into it — the red build isn't the scheduler change, it's a pytest **collection** error that aborts the whole suite before any test runs:

```
ERROR collecting tests/llmcompressor/pipelines/sequential/test_helpers.py
import file mismatch:
imported module 'test_helpers' has this __file__ attribute:
  .../tests/llmcompressor/modifiers/pruning/test_helpers.py
which is not the same as the test file we want to collect:
  .../tests/llmcompressor/pipelines/sequential/test_helpers.py
HINT: ... use a unique basename for your test file modules
```

Root cause is on me: I added `tests/llmcompressor/modifiers/pruning/test_helpers.py` but the new `pruning/` dir had no `__init__.py`. Under the default prepend import mode a test module in a dir without `__init__.py` is imported by its bare basename (`test_helpers`), which then collides with the pre-existing bare `tests/llmcompressor/pipelines/sequential/test_helpers.py`. `main` is green only because `pruning/` was the *second* bare `test_helpers` in the tree.

Fixed in 9af1153 by adding an empty `tests/llmcompressor/modifiers/pruning/__init__.py`, so it imports as the unique dotted `tests.llmcompressor.modifiers.pruning.test_helpers` — matching the sibling dirs that already have one (`modifiers/reap`, `modifiers/utils`, `observers`). Locally the full `tests/llmcompressor` collection now completes with 0 errors, and the pruning tests themselves pass (the `helpers.py` math change is unchanged). Should be green on the next run.

PR #3097, opened 2026-08-26
[GPTQ] Hoist loop invariants out of the per-column quantization loop

rishabhsinha17:
## Summary

Part of the plan in neuralmagic/compressed-tensors#766 (step 2 of the ordering proposed in [this comment](https://github.com/neuralmagic/compressed-tensors/issues/766#issuecomment-5406315144); step 1 was neuralmagic/compressed-tensors#855). The GPTQ per-column inner loop runs once per weight column (245,760 times for Qwen2.5-0.5B), and two of its operations are loop invariant.

## Root cause

In `quantize_weight`:
- The GROUP/TENSOR_GROUP branch rebuilds `altered_qargs = copy(quant_args)` and sets `strategy = CHANNEL` on every column. The pydantic copy is identical every iteration (4 `copy.copy` frames per column).
- The CHANNEL branch re-slices `scale[:, 0]` / `zero_point[:, 0]` on every column.

## Changes

- Hoist both above the block loop, guarded by strategy. No behavior change.

## Validation

CPU only (no GPU on this machine), torch 2.13.0, compressed-tensors 0.18.1.a20260824, Python 3.11.

Bit-identical outputs: fixed-seed comparison of `quantize_weight` outputs (weight, scale, zero_point, global_scale, loss) between main @ 1b0d3e223 and this branch over 8 configs covering TENSOR, CHANNEL (sym+asym), GROUP g8 (actorder None/weight/static), TENSOR_GROUP NVFP4, and BLOCK 2x4:

```
bit-identical across all 33 tensors: True
```

Bench, `quantize_weight` on a 2048x2048 Linear, W4A16 group-128, 7 reps after warmup, observer re-observed and hessian recloned outside the timed region:

| | main | this PR |
|---|---|---|
| wall clock min / median (s) | 0.578 / 0.598 | 0.586 / 0.604 |
| cProfile total function calls per `quantize_weight` | 182,487 | 141,547 |
| `copy.py:66(copy)` ncalls | 8192/2048 | 4/1 |

CPU wall clock is unchanged within run-to-run noise, as expected from the profiling discussion in #766: on CPU the leaf math dominates. The measured effect is the call-count and per-column allocation reduction; the per-column Python overhead matters most on the CUDA path discussed in #766, which I could not measure here.

Existing tests:

```
$ pytest tests/llmcompressor/modifiers/gptq/ -q
35 passed, 1 skipped, 14 warnings in 0.60s
```

(skip is the H100-gated NVFP4 e2e test)


rishabhsinha17:
GPU validation done, alongside the compressed-tensors#855 measurement I promised in vllm-project/compressed-tensors#766 (step-3 comment there has the full tables and methodology).

Measured on an NVIDIA L40S (driver 570.124.06, torch 2.11.0+cu128, CUDA 12.8), W4A16 GPTQ, Qwen2.5-0.5B-Instruct, 512 x 2048 open_platypus, 3 runs per config, this branch merged onto main @ 50d0a1c7: GPTQ compress total 68.95 s -> 68.58 s against the post-#855 baseline (-0.5% on the mean; rep spreads overlap, baseline [65.4, 72.6] vs this [68.0, 68.9], so noise-level at model scale), oneshot wall 226.0 s -> 229.0 s (also within spread). The saving matches the CPU-measured pydantic `copy()` cost almost exactly: ~1.5 us per call x 245,760 calls = 0.37 s predicted, 0.36 s measured. torch.profiler on a single `quantize_weight` window (layers.1.mlp.down_proj, 4864 columns) confirms the hoisted per-column invariant work (pydantic copy, scale/zp reslice) is out of the loop: window self-CPU 1.679 s -> 1.655 s with kernel launches identical (108,083 cudaLaunchKernel both), i.e. a pure host-side trim. Per-module GPTQ losses match the baseline bit for bit (max delta 0.0 over 168 modules, 0 RTN fallbacks), as expected from a pure hoist.

CPU smoke (2-layer model) and the unit path were checked before the GPU run; no behavior change, output parity holds on both.

This should be ready for review. I cannot add labels on this repo, so if a maintainer agrees, please apply `ready` to kick off the extended tests.

Roderick-Wu (approved):
I think this looks fine. Seems to be generally better practices even if gains are minimal. Will need some opinions from other maintainers.

rishabhsinha17:
Rebased onto main @ b5e46776 (was 16 behind, no conflicts); head is now 4932c4f9. The hoist applies unchanged on top of the `used_rtn_fallback` return from #3098, and nothing else has touched `gptq_quantize.py` since. Local `tests/llmcompressor/modifiers/gptq/`: 37 passed, 1 skipped (py3.11, torch 2.13 CPU, transformers 5.16.1); buildkite #1446 and quality-check are green on the rebased head. Mergify auto-rebase failed on the fork push because it needs a one-time Mergify dashboard login on my side to enable; I will sort that out separately so future rebases go through automatically.

@kylesayrs @brian-dellabetta: @Roderick-Wu approved on 08-31 and CI is green, but Merge Protections needs an approval from a listed maintainer. Could one of you take a look?

rishabhsinha17:
@kylesayrs @dsikka could one of you take a look when you get a chance? @Roderick-Wu approved and it carries the ready label, buildkite and quality checks are green on the current rebase, and the measured speedups are posted above. Merge protections still need an approval from the maintainer pool, and with @brian-dellabetta departing that ask now lands with you. Happy to address any comments.

kylesayrs (approved):
Thanks!

PR #3098, opened 2026-08-26
[GPTQ] Emit an end-of-run summary when modules fall back to RTN

### LINKED ISSUE #2952: GPTQ: Hessian inversion fails on every module of a looped/weight-shared architecture (NanbeigeForCausalLM) — silent RTN fallback
## Summary
Running GPTQModifier (llm-compressor 0.12.0, transformers <=5.10.1) against `Nanbeige/Nanbeige4.2-3B` — a looped transformer (`num_loops=2`: 22 physical layers, each executed twice per forward) — fails Hessian inversion on **all 154 Linear modules** ("Failed to invert hessian"), silently falling back to RTN for every one. The output artifact is RTN-equivalent while appearing to be GPTQ.

## What I ruled out
- **Calibration activations are clean** (no NaN/inf, sane ranges — probed at the hook).
- **A manually-computed H from the same activations is well-conditioned and inverts fine.**
- **The double-fire hook is not the cause mathematically**: because the layer runs twice per forward, the observer hook fires 2x per sample, but the running-mean accumulation is invocation-count-agnostic — verified against the source. The doubled sample count alone cannot make H singular.
- Dampening increases did not help.

## Repro
Looped arch + standard GPTQ oneshot recipe (any scheme, INT4/NVFP4 alike):
```python
# trust_remote_code model with num_loops=2; GPTQModifier(scheme=..., targets="Linear", ignore=["lm_head"])
```
198/198 calibration samples, default pipeline. Every module logs a Hessian inversion failure; run completes "successfully".

## Context
Looped/weight-shared (recurrent-depth) models are an emerging small-model architecture and are unusually quantization-sensitive (error compounds across passes), i.e. exactly the models where GPTQ-class compensation would help most. As far as I can tell no public report covers GPTQ on this architecture class; the known Hessian-failure issues (#1019, #966, #1003, #1152) have different causes (softcapping, rank-deficient activations).

Two asks: (1) any pointer on where inversion actually goes wrong given the ruled-out causes above — happy to run instrumented builds; (2) consider making the RTN fallback loud (summary count at end of run) — 154 silent fallbacks looked like a successful GPTQ run.

rishabhsinha17:
## Summary

Fixes #2952. When GPTQ hessian inversion fails, the module silently (in aggregate) falls back to round-to-nearest: each failure logs one warning, but nothing reports the overall outcome. In #2952, all 154 modules of a looped-transformer model fell back and the run still looked like a successful GPTQ run. This adds a single end-of-run WARNING summarizing how many modules fell back and which ones.

## Root cause

The RTN fallback introduced in #1283 was designed to keep large runs alive when inversions spontaneously fail, which is correct, but it reports per module only. `quantize_weight` does not tell its caller that the fallback happened, so `GPTQModifier` cannot aggregate.

## Changes

- `quantize_weight` returns `used_rtn_fallback` alongside the existing loss and qparam dict. The per-module warning and fallback behavior are unchanged; no retry logic.
- `GPTQModifier` counts compressed modules and records the names of those that fell back, then logs one WARNING at finalize with count, percentage, and affected module names (first 10; full list at DEBUG). In the distributed case each rank reports the modules it compressed.

Example output for the #2952 scenario (154/154 modules):

```
WARNING - Hessian inversion failed for 154/154 modules (100.0%). These modules were quantized with round-to-nearest instead of GPTQ (the quantization scheme is unchanged). Consider increasing GPTQModifier.dampening_frac, increasing the number of calibration samples, or shuffling the calibration dataset. Affected modules: model.layers.0.self_attn.q_proj, model.layers.1.self_attn.q_proj, ..., and 144 more (full list at DEBUG level)
```

## Validation

CPU, torch 2.13.0, Python 3.11. New tests: `test_quantize_weight_singular_hessian_rtn_fallback` forces the fallback with a singular (rank-1) hessian and `percdamp=0.0` and asserts the flag; `test_gptq_rtn_fallback_summary_fires` drives `GPTQModifier.compress_modules` through the same singular hessian and asserts the summary warning content (count, percentage, module name).

```
$ pytest tests/llmcompressor/modifiers/gptq/ -q
37 passed, 1 skipped, 14 warnings in 0.75s
```

(skip is the H100-gated NVFP4 e2e test)

### DIFF
--- a/src/llmcompressor/modifiers/gptq/base.py
+++ b/src/llmcompressor/modifiers/gptq/base.py
@@ -134,6 +134,8 @@ class GPTQModifier(Modifier, QuantizationMixin):
     _num_samples: dict[torch.nn.Module, torch.Tensor] = PrivateAttr(
         default_factory=dict
     )
+    _num_compressed_modules: int = PrivateAttr(default=0)
+    _rtn_fallback_module_names: list[str] = PrivateAttr(default_factory=list)
 
     def resolve_quantization_config(self) -> QuantizationConfig:
         config = super().resolve_quantization_config()
@@ -308,7 +310,7 @@ class GPTQModifier(Modifier, QuantizationMixin):
                 self._maybe_onload_hessian(module),
                 CompressionLogger(module) as comp_logger,
             ):
-                loss, q_param_dict = quantize_weight(
+                loss, q_param_dict, used_rtn_fallback = quantize_weight(
                     module=module,
                     quant_args=quant_args,
                     hessian=self._hessians.pop(module) / self._num_samples.pop(module),
@@ -316,6 +318,10 @@ class GPTQModifier(Modifier, QuantizationMixin):
                     percdamp=self.dampening_frac,
                 )
                 comp_logger.set_results(name="GPTQ", loss=loss)
+
+            self._num_compressed_modules += 1
+            if used_rtn_fallback:
+                self._rtn_fallback_module_names.append(name)
 
             for attr, val in q_param_dict.items():
                 update_offload_parameter(module, attr, val)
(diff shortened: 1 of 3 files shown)


kylesayrs (approved):
Looks great! Thanks for the improvement

PR #3100, opened 2026-08-26
Register GptOssLinearExperts for gpt-oss expert linearization

ganeshr10:
SUMMARY:
gpt-oss interleaves gate and up along the last dim of `gate_up_proj`, so the generic linearizer's half-split leaves `gate_proj` and `up_proj` each holding a mix of both projections. This registers a `GptOssLinearExperts` that splits on the even and odd
columns instead, so the per-projection weights mean what they say.
The class reuses `LinearExperts2D.__init__` and `from_experts_module` and overrides only the expert copy. `MoEConfig.from_config` already special-cases `gpt_oss` for `alpha`, `limit` and `use_bias`, so there is nothing to restate here. Selecting the per-expert class needs one hook in the base, `expert_cls_with_gate`, which defaults to today's `ExpertMLPWithGate`.

TEST PLAN:
`test_linearize_moe_gpt_oss` asserts that each expert's `gate_proj` and `up_proj` weights and biases equal the even and odd columns of the original `gate_up_proj`, then checks the linearized forward against `GptOssExperts` (MSE < 1e-10). The MSE half passes with or without this change, so the column assertions are what pin the layout down. It runs on CPU, unlike the parametrized `test_linearize_moe`, which is `@requires_gpu` and already covers `gpt_oss` for numerics only.

### DIFF
--- a/src/llmcompressor/modeling/moe/linear_experts.py
+++ b/src/llmcompressor/modeling/moe/linear_experts.py
@@ -149,6 +149,9 @@ class LinearExperts2D(torch.nn.ModuleList):
     has_gate: ClassVar[bool]
     _apply_gate: ClassVar[Callable[[torch.Tensor], torch.Tensor]]
 
+    # override to load a gate_up_proj whose halves are not gate and up
+    expert_cls_with_gate: ClassVar[type[ExpertMLP]] = ExpertMLPWithGate
+
     num_experts: int
     intermediate_size: int
 
@@ -159,6 +162,7 @@ class LinearExperts2D(torch.nn.ModuleList):
     def get_registration(
         cls, key: type[torch.nn.Module], default: Any = None
     ) -> type["LinearExperts2D"]:
+        from .gpt_oss import GptOssLinearExperts  # noqa: F401
         from .llama4 import Llama4LinearExperts  # noqa: F401
 
         return cls._registry.get(key, default)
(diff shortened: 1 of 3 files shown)


kylesayrs (changes requested):
Hey Ganesh! Are you sure that the current linearization logic doesn't already work?
The transformers model definition already uses [use_experts_implementation](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L68), and the [test_linearize_moe](https://github.com/vllm-project/llm-compressor/blob/main/tests/llmcompressor/modeling/test_linearize.py#L157) test already tests gpt_oss.

src/llmcompressor/modeling/moe/linear_experts.py
```diff
@@ -213,7 +217,9 @@ def __init__(self, config: PreTrainedConfig, *args, **kwargs):
         self.intermediate_size = moe_config.intermediate_size
         act_fn: torch.nn.Module = ACT2FN[moe_config.hidden_act]
 
-        expert_cls = ExpertMLPWithGate if self.has_gate else ExpertMLPWithoutGate
+        expert_cls = (
+            self.expert_cls_with_gate if self.has_gate else ExpertMLPWithoutGate
```
kylesayrs: Consider standardizing without gate as well
```suggestion
            self.expert_cls_with_gate if self.has_gate else self.expert_cls_without_gate
```

kylesayrs (approved):
Great contribution! Thank you!

tests/llmcompressor/modeling/test_linearize.py
kylesayrs:
```suggestion
@pytest.mark.parametrize("model_type", list(ARCH_TO_IMPORT_PATHS.keys() - {"llama4", "gpt_oss"}))
```

src/llmcompressor/modeling/moe/linear_experts.py
ganeshr10: Done

tests/llmcompressor/modeling/test_linearize.py
ganeshr10: Done

Issue #3071, opened 2026-08-21
[Bug]: Qwen3-VL-2B (dense) AWQ/GPTQ calibration fails, vision tower not FX-traceable (rot_pos_emb CUDA driver error), even with visual.* in ignore


gxaurab:
### ⚙️ Your current environment

<details>
<summary>The output of <code>python collect_env.py</code></summary>

```text
llm-compressor Version: 0.11.0
compressed-tensors Version: 0.16.0
transformers Version: 4.57.6
torch Version: 2.10.0+cu126
Python Version: 3.10.8
CUDA Devices: ['NVIDIA A100-SXM4-80GB']
CUDA Driver: 575.57.08 (CUDA 12.9)
```

</details>

### 🐛 Describe the bug

Calibrated quantization (AWQ and GPTQ, W4A16) of the **dense** Qwen3-VL-2B-Instruct fails during the sequential pipeline's FX tracing of the vision tower, even though all `visual.*` modules are in the quantization `ignore` list.

The `ignore` list correctly excludes the vision tower from *quantization*, but the sequential pipeline still *traces* through the vision tower to build the graph, and the Qwen3-VL vision code is not FX-traceable on transformers 4.57.x. Data-free RTN quantization of the same model works fine which isolates the problem specifically to the calibration/tracing path, not model loading or the environment.

**Recipe:**
`AWQModifier`, W4A16 (num_bits 4, int, group_size 128, symmetric, mse observer), `ignore=["re:.*lm_head", "re:.*visual.*"]`, `sequential_targets=["Qwen3VLTextDecoderLayer"]`. Also reproduced with `GPTQModifier(scheme="W4A16")` and the same ignore list.

**What works up to the crash:**
- Model loads; decoder-layer class confirmed as `Qwen3VLTextDecoderLayer`
- Vision modules confirmed under `model.visual.*`
- Calibration dataset builds correctly (256 samples, columns `['input_ids', 'attention_mask', 'pixel_values', 'image_grid_thw']`)
- "Preparing cache" completes; calibration starts at `(1/29)`
- AWQ mapping resolution runs ("28 mappings were skipped due to incompatible shapes")

**The crash:**\
Raised from inside `llmcompressor/pipelines/sequential/` while FX-tracing the model graph. The tracer walks `model.visual` (`get_image_features` → `self.visual` → `rot_pos_emb`), and the `.item()` call fails under tracing. Full traceback attached below.

**What I've already tried:**
[...]

PR #3107, opened 2026-08-27
[Tests] Add dense Qwen3-VL to the tracing model enumeration

rishabhsinha17:
SUMMARY:
Adds dense Qwen3-VL to the tracing model enumeration in
`tests/llmcompressor/transformers/tracing/test_models.py`, the configuration
reported broken in #3071 (AWQ/GPTQ W4A16 calibration with `visual.*` in the
quantization ignore list and `sequential_targets=["Qwen3VLTextDecoderLayer"]`).

Investigation of #3071 found the report does not reproduce on main or on the
reporter's versions (llmcompressor 0.11.0 + transformers 4.57.6): the
autowrapper wraps the `if pixel_values is not None:` branch of
`Qwen3VLModel.forward` (it cannot statically evaluate the condition), so
`get_image_features` and the vision tower run eagerly inside a
`torch.fx.wrap`ped function and never enter the fx graph. The new row pins
that behavior so future autowrapper or transformers changes cannot silently
regress it: dense Qwen3-VL traces to `num_hidden_layers + 1` subgraphs with
the vision tower opaque.

The row mirrors the Qwen2-VL and Qwen2.5-VL entries:
`Qwen3VLForConditionalGeneration`, `["Qwen3VLTextDecoderLayer"]`, vision
modality, torchvision backend, default `tracing_ignore`. Per review, this
replaces the standalone tiny-config CPU test from the first revision.

Related to #3071 (not reproducible on main; test guards the working behavior)

TEST PLAN:
* `pytest --collect-only tests/llmcompressor/transformers/tracing/test_models.py`:
  19 tests collected (18 existing + the new row); the row is
  `@requires_hf_token`, so it runs in the token-enabled suite
* `ruff check`, `ruff format --check`, `tools/lint_cuda.py` clean on the file
* VALIDATION (GPU pod, first revision): real `Qwen/Qwen3-VL-2B-Instruct`
  AWQ W4A16 end-to-end oneshot on an L40S (46GB) against main @ 50d0a1c75,
  llmcompressor 0.13.1.dev51+g50d0a1c75, transformers 5.16.1,
  compressed-tensors 0.18.1.a20260826, torch 2.11.0+cu128,
  `CUDA_LAUNCH_BLOCKING=1`, 256 flickr30k calibration samples: traces to
  29 subgraphs with the vision tower opaque, which is exactly the
  `len(target_modules) + 1` the row asserts; oneshot completes,
  save_compressed + reload clean (905M params), 6m35s wall, 5.4 GiB peak
  allocated VRAM.
* The first revision's tiny-config CPU test (2 passed in 0.23s on the same
  pod; full `tests/llmcompressor/pipelines/sequential/` 45 passed) covered
  the same trace on random weights; dropped here in favor of the enumeration.


kylesayrs (commented):
Thanks for the investigation! Do you think this test could be covered as an enumeration in `test_models.py`?

rishabhsinha17:
Yes, done in c6d772f1: added a `Qwen/Qwen3-VL-2B-Instruct` row to `test_models.py` mirroring the Qwen2-VL / Qwen2.5-VL rows (`Qwen3VLForConditionalGeneration`, `["Qwen3VLTextDecoderLayer"]`, vision, torchvision backend, default `tracing_ignore`, same as the standalone test used) and dropped `test_qwen3_vl_trace.py`. Locally: 19 tests collected, `ruff check` and `ruff format --check` clean; the row is gated on `HF_TOKEN` so it runs in the token-enabled suite, and the same config traced the real 2B to 29 subgraphs in the L40S run in the description. Also added the missing DCO sign-off. If a token-free tiny-config variant in the CPU suite is still wanted, I can add it back (a couple of dozen lines).

kylesayrs (approved):
Thanks!

PR #3111, opened 2026-08-28
Implement streaming quantization pipeline

Isotr0py:
SUMMARY:

- This PR implement streaming quantization pipeline to avoid offload/dispatch overheads for efficient quantization on extremely large models
- Firstly, we initialize model on meta device, than prefetch and h2d weights for each layer on demand similar to distributed layerwise offload (DLO): https://vllm.ai/blog/2026-08-17-distributed-layerwise-offload
- Not DLO yet, will implement it in a followup PR after this PR merge

TEST PLAN:
## Setup

- Hardware: 1 node, 4x NVIDIA GB300 (284 GiB each), 956 GiB RAM; checkpoints on
  a shared network filesystem.
- Recipe (same as `examples/quantizing_moe/glm5_example.py`): FP8_BLOCK for
  attention, NVFP4 for MLP, `lm_head` and first layers ignored.
- Calibration: 8 samples, batch size 4, seq len 1024, identical pre-tokenized
  data for both pipelines. Saving was not timed.

## Example code

Baseline and streaming differ only in model loading and the `pipeline` argument;
the recipe is identical.

<details>
<summary>Baseline (auto_offload + sequential)</summary>

```python
import torch
from compressed_tensors.offload import init_dist

from llmcompressor.utils import load_context

init_dist()  # requires torchrun
model_id = "zai-org/GLM-5.2"
with load_context():
        model_id,
        device_map="auto_offload",
        max_memory={"cpu": "500GiB"},
        offload_folder="offload_folder",
    )

oneshot(
    model=model,
    dataset="perfectblend",
    batch_size=4,
    recipe=recipe,
    num_calibration_samples=512,
)
```

</details>

<details>
<summary>Streaming (meta skeleton + checkpoint-referenced layers)</summary>

```python

from llmcompressor.utils import load_context

model_id = "zai-org/GLM-5.2"
with load_context(meta=True):
    model = AutoModelForCausalLM.from_pretrained(model_id, device_map="meta")

oneshot(
    model=model,
    dataset="perfectblend",
    batch_size=4,
    recipe=recipe,
    pipeline="streaming",  # default for meta-device models
    num_calibration_samples=512,
)
```
</details>

### Qwen3.6-35B-A3B (82 GiB bf16, 40 MoE layers x 256 experts, fused 3D expert checkpoint)

Baseline ran with `max_memory={"cpu": "20GiB"}` to force the disk-offload path.

| Phase | baseline (`auto_offload` + `sequential`) | `streaming` | Speedup |
|---|---|---|---|
| Model load | 128.2 s | 2.1 s | 61x |
| `oneshot` (calibration + layer dispatch) | 328.5 s | 198.5 s | 1.65x |
| **Total (no save)** | **460.0 s** | **207.8 s** | **2.2x** |

### Reduced GLM-5.3-BF16 (12 layers, 176 GiB, per-expert checkpoint, local NVMe)

| Phase | baseline (`auto_offload` + `sequential`) | `streaming` | Speedup |
|---|---|---|---|
| Model load | 20.4 s | 2.8 s | 7.3x |
| `oneshot` | 146.6 s | 112.0 s | 1.3x |
| **Total (no save)** | **171.7 s** | **118.1 s** | **1.45x** |

### DIFF
--- a/src/llmcompressor/args/dataset_arguments.py
+++ b/src/llmcompressor/args/dataset_arguments.py
@@ -297,6 +297,17 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
+    streaming_release: str = field(
+        default="cpu",
+        metadata={
+            "help": "Only relevant for the streaming pipeline. Device to release "
+            "weights to after each subgraph is calibrated. 'cpu' keeps (possibly "
+            "modifier-updated) weights in host memory; 'meta' drops "
+            "checkpoint-backed weights, keeping host memory flat for models "
+            "larger than RAM. 'meta' is only valid when no modifier mutates "
+            "weights. Default is 'cpu'."
+        },
+    )
     enable_compile: bool = field(
         default=False,
         metadata={
--- a/src/llmcompressor/entrypoints/oneshot.py
+++ b/src/llmcompressor/entrypoints/oneshot.py
@@ -50,6 +50,12 @@ if TYPE_CHECKING:
 TOKENIZERS_PARALLELISM_ENV = "TOKENIZERS_PARALLELISM"
 
 
+def _is_meta_model(model: PreTrainedModel) -> bool:
+    """True if all model parameters live on the meta device (no weights loaded)."""
+    params = list(model.parameters())
+    return len(params) > 0 and all(p.device.type == "meta" for p in params)
+
+
 class Oneshot:
     """
     Class responsible for carrying out one-shot calibration on a pretrained model.
@@ -262,6 +268,13 @@ class Oneshot:
             session.state.enable_compile = self.dataset_args.enable_compile
 
             user_pipeline = self.dataset_args.pipeline
+            if user_pipeline in (None, "independent") and _is_meta_model(self.model):
+                logger.info(
+                    "Model was loaded on the meta device; defaulting to the "
+                    "'streaming' pipeline, which materializes weights directly "
+                    "from the original checkpoint"
+                )
+                user_pipeline = "streaming"
             pipeline = CalibrationPipeline.from_modifiers(
                 session.lifecycle.recipe.modifiers, user=user_pipeline
             )
--- a/src/llmcompressor/modeling/moe/linear_experts.py
+++ b/src/llmcompressor/modeling/moe/linear_experts.py
@@ -196,8 +196,7 @@ class LinearExperts2D(torch.nn.ModuleList):
         # the checkpoint later, so no data is copied and no offload cache is
         # attached
         src_is_meta = any(
-            param.device.type == "meta"
-            for param in experts.parameters(recurse=False)
+            param.device.type == "meta" for param in experts.parameters(recurse=False)
         )
 
         if src_is_meta:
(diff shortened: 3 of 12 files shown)


kylesayrs:
@Isotr0py as I understand it, the main pros/cons of this pipeline when compared to the existing pipeline are as follows:

| Pros | Cons |
| - | - |
| Subgraph prefetching (load subgraph[i + 1] on cpu while processing subgraph[i]) | Does not implement checkpoint conversion mappings |
| Group reads by shard file (can greatly improve when combined with instanttensor) | |
| | Does not support writing modified weights/qparams to disk (must be cpu) |

These cons would mean that this pipeline does not support the following models:
* Models which have conversion mappings (weight fusions, transposes, etc.)
* Models which are too large to fit in cpu memory

kylesayrs:
I've mocked up some examples of how we can implement these performance benefits without reducing model coverage [here](https://github.com/vllm-project/llm-compressor/issues/3126)

Isotr0py:
> Does not implement checkpoint conversion mappings

I think only `WeightConverter` is not supported now, but other conversion like `WeightRenaming` and `WeightTransform` are supported. Since there's no zero-copy code path yet, I think this can be fixed it in followup.

> Does not support writing modified weights/qparams to disk (must be cpu)

Yea, we didn't include streaming model saving in this PR since it's not compatible with Transformers `save_pretrained` API natively. We can solve it with some hacks, but I'm still thinking some workaround to avoid hacky patching.

> These cons would mean that this pipeline does not support the following models

I think these cons can be addressed within streaming pipeline as followup improvements, because current streaming pipeline implementation in this PR only contains the core part, and we will have more followup features based on it in the futrue.

For #3126, it considers the overlap idea from this pipeline, but _still relys on `accelerate`'s DiskCache_, which is most expensive in offload pipeline for extreme large models and it's also the keypoint we want to avoid it.

In practice, to quantize GLM-5.3 on sequential pipeline, it costs ~1hr to load and dispatch model offload, and it can be much slower if we uses NFS (we usually store large checkpoints on NFS to share it in different racks). Therefore, we want to use meta device to skip the DiskCache construction and use checkpoint itself as `DiskCache` for a faster iteration.

kylesayrs:
@Isotr0py 

> For https://github.com/vllm-project/llm-compressor/issues/3126, it considers the overlap idea from this pipeline, but still relys on accelerate's DiskCache, which is most expensive in offload pipeline for extreme large models and it's also the keypoint we want to avoid it.

Yeah, but some level of disk offload writing is going to be necessary if the model is too large to fit on CPU memory.

> In practice, to quantize GLM-5.3 on sequential pipeline, it costs ~1hr to load and dispatch model offload, and it can be much slower if we uses NFS (we usually store large checkpoints on NFS to share it in different racks). Therefore, we want to use meta device to skip the DiskCache construction and use checkpoint itself as DiskCache for a faster iteration.

Yeah, the initial "dispatching" of the model during loading is a real cost, especially on NFS file systems. During dispatching, [symlinks which point to the original checkpoint](https://github.com/vllm-project/compressed-tensors/blob/099fa98fea7f3533a8e304a081795a1136ef67c0/src/compressed_tensors/offload/convert/from_accelerate.py#L160) are created and [ranks share metadata about the offloaded tensors](https://github.com/vllm-project/compressed-tensors/blob/main/src/compressed_tensors/offload/cache/dist_disk.py#L44). The cost could either the creation of the symlinks or lots of unbatched collective ops.

PR #3112, opened 2026-08-29
fix: Rename Recipe.dict() to to_dict() to avoid shadowing builtin dict

Asthenia0412:
### Description

`Recipe.dict()` shadows the builtin `dict` name, making `llmcompressor` unimportable on Python 3.14. PEP 649 lazy annotations evaluate the string annotation `"dict[str, Any]"` against the class namespace, finding the `dict` method instead of the builtin, resulting in `TypeError: function object is not subscriptable`.

Renaming to `to_dict()` unshadows the builtin and resolves the Python 3.14 compatibility issue. The method is only called in tests, so the two test call sites are updated accordingly.

### Related Issues

Closes #2871

### Changes

- `src/llmcompressor/recipe/recipe.py`: `Recipe.dict()` → `Recipe.to_dict()`
- `tests/llmcompressor/recipe/test_recipe.py`: update the two call sites

### Testing

- Syntax verified with `ast.parse`
- No production code references `Recipe.dict()`

### DIFF
--- a/src/llmcompressor/recipe/recipe.py
+++ b/src/llmcompressor/recipe/recipe.py
@@ -232,7 +232,7 @@ class Recipe(BaseModel):
 
         return model
 
-    def dict(self, *args, **kwargs) -> dict[str, Any]:
+    def to_dict(self, *args, **kwargs) -> dict[str, Any]:
         """
         :return: A dictionary representation of the recipe
         """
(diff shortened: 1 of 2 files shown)


Asthenia0412:
Good catch. Removed the unused `*args` and `**kwargs` from `to_dict()`.

Roderick-Wu (commented):
change looks good, maybe add a warning for deprecated in the case that `Recipe.dict()` is called for versions <3.14?

Roderick-Wu (approved):

kylesayrs:
Thanks for the finding! Closing as a duplicate
https://github.com/vllm-project/llm-compressor/pull/2871

PR #3114, opened 2026-08-31
Fix sequential tracer for decorated/forwards with complex conditionals

hammertoe:
Fixes vllm-project/compressed-tensors#832 (Saved Qwen3-Omni GPTQ W4A16 checkpoints contain NaN/Inf weight_scale tensors).

The NaN/Inf scales are a *symptom* of an upstream llm-compressor tracing failure: `Qwen3OmniMoeThinkerForConditionalGeneration.forward` cannot be traced by FX end-to-end, so zero decoder-layer `call_module` nodes reach calibration, and every `weight_scale` parameter is left at its `torch.empty()` initialisation values (frequently NaN/Inf in BF16).

This PR contains three verified autowrap fixes, each with a passing local unit test:

1. `ast_helpers.autowrap_forward`: unwrap `module.forward` with `inspect.unwrap` so the AST reflects the raw forward body. The Qwen3-Omni Thinker forward is decorated with `@can_return_tuple`, which `inspect.getsource` cannot parse.

2. `ast_helpers.autowrap_forward`: patch both `module.forward` *and* `type(module).forward`. FX resolves forward via `type(root).forward`, so patching only the instance leaves the decorated class forward in place and FX cannot trace through decorators like `@capture_outputs` or `@merge_with_config_defaults` on ancestor modules (`Qwen3OmniMoeThinkerTextModel`).

3. `ast_utils.AutoWrapper`: force any `if`-statement whose body only raises to be wrapped with `body=pass`, and force any `BoolOp` test into a `@torch.fx.wrap` helper. FX cannot constant-fold across `if` branches that depend on values produced by the model itself, so rewriting `(a is None) ^ (b is not None)` to `if True: raise ValueError(...)` kills the trace before any `call_module` nodes are emitted.

These fixes are *necessary* but I have not been able to confirm they are *sufficient* to resolve the full Qwen3-Omni Thinker tracing on local CPU. The complex multimodal forward in `Qwen3OmniMoeThinkerForConditionalGeneration.forward` continues to truncate the FX graph at the rope-deltas block, and reaching `self.model(...)` from there likely requires either an architectural change in llm-compressor or additional autowrap logic.

Three regression tests are included:
- `test_trace_subgraphs_unwraps_decorated_forward` — minimal model with `@can_return_tuple`-like decorated forward
- `test_trace_subgraphs_wraps_combined_condition` — XOR-style combined-condition test like the one in the Thinker forward
- `test_trace_subgraphs_wraps_raise_condition` — `if … raise` block

Local test status:
- All 12 `tests/llmcompressor/pipelines/sequential/test_helpers.py` tests pass (including the 3 new ones and the existing `test_trace_subgraphs[1-5]` that exercises the full Qwen3-0.6B model)
- All other `tests/llmcompressor/pipelines/` tests pass except `test_device_handling`, which is a pre-existing CUDA-only failure unrelated to this PR

Reproduction confirmation on GPU was via H200 jobs `6a7e679d8747785b3ad53546` and `6a7e6e868747785b3ad5361b` documented in the linked issue thread. The diagnostic job `6a7e6bc473ce7897b8939fd1` showed the FX graph has 4 placeholders + 2 `call_function` + 8 `getitem` = 14 nodes total, zero `call_module`, zero target nodes — the graph truncates after the rope-deltas wrapper before reaching `self.model(...)`.

For users needing to ship a Qwen3-Omni GPTQ artifact before the upstream fix is fully validated, a 30-line user-side guard that fails fast on any non-finite `weight_scale` after calibration has been deployed in `hammertoe/bimomni` at `src/bimomni/publish/quantize_gptq.py:runtime.validate_weight_scales`.

cc `@brian-dellabetta` per the comment thread on the linked issue.

### DIFF
--- a/src/llmcompressor/pipelines/sequential/ast_helpers.py
+++ b/src/llmcompressor/pipelines/sequential/ast_helpers.py
@@ -52,8 +52,9 @@ def autowrap_forward(module: torch.nn.Module, ignore: list[str]):
             "`oneshot`. For example, `oneshot(model.thinker, ...)`"
         )
 
-    # get source code of module forward
-    source = inspect.getsource(module.forward)
+    # get source code of module forward, unwrapping decorators (e.g.
+    # ``@can_return_tuple``) so the AST reflects the raw forward body
+    source = inspect.getsource(inspect.unwrap(module.forward))
     source = textwrap.dedent(source)
     tree = ast.parse(source)
 
@@ -82,9 +83,15 @@ def autowrap_forward(module: torch.nn.Module, ignore: list[str]):
         filename,
     )
 
-    # patch forward with autowrapped forward
+    # patch forward with autowrapped forward on both the instance and the
+    # class. FX's tracer resolves forward via ``type(module).forward``, so
+    # patching only the instance leaves the decorated class forward in place
+    # and FX cannot trace through decorators like ``@capture_outputs``.
     new_forward = namespace["forward"].__get__(module)
-    with patch_attr(module, "forward", new_forward):
+    with (
+        patch_attr(module, "forward", new_forward),
+        patch_attr(type(module), "forward", namespace["forward"]),
+    ):
         yield
 
 
--- a/src/llmcompressor/pipelines/sequential/ast_utils/auto_wrapper.py
+++ b/src/llmcompressor/pipelines/sequential/ast_utils/auto_wrapper.py
@@ -8,6 +8,13 @@ from .control_flow_analyzer import ControlFlowAnalyzer
 from .name_analyzer import NameAnalyzer
 
 
+def _is_only_raise(node: ast.AST) -> bool:
+    """Return True when ``node`` consists solely of ``raise`` statements."""
+    if isinstance(node, ast.If):
+        return all(_is_only_raise(child) for child in node.body)
+    return isinstance(node, ast.Raise)
+
+
 class AutoWrapper(ast.NodeTransformer):
     """
     Automatically wraps untracable code according to the following patterns:
@@ -110,6 +117,23 @@ class AutoWrapper(ast.NodeTransformer):
             Otherwise, return a wrapper function call + assignment
         """
         try:
+            # force a wrap whenever the test combines multiple conditions, even
+            # if statically evaluable. FX cannot constant-fold across
+            # ``if`` branches whose test depends on values produced by the
+            # model itself, so replacing a mixed test with ``True``/``False``
+            # raises during tracing.
+            # Examples: ``(a is None) ^ (b is not None)`` or
+            # ``a is None and b is None``.
+            if isinstance(node.test, ast.BoolOp):
+                raise Exception("If statement combines multiple conditions")
+
+            # force a wrap whenever the body raises. ``if True/False: raise ...``
+            # would otherwise become a literal raise during tracing, killing
+            # the trace before any call_module nodes are emitted.
+            for stmt in node.body:
+                if isinstance(stmt, ast.Raise):
+                    raise Exception("If statement raises")
+
             value = bool(self._eval_expr(node.test))
 
             # force a wrap if any assignments occur within the if statement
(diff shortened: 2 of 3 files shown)


brian-dellabetta:
Hi @hammertoe , thanks for preparing this. I discussed with @kylesayrs , your 1st bullet point should be covered by something recently merged into main (hence the merge conflicts). The others shouldn't be needed, I forgot that qwen omni models just require a slight modification to be able to run within oneshot and the tracing logic. I will reply to your initial issue regarding that

hammertoe:
@brian-dellabetta fantastci, thanks!

kylesayrs:
Yep, first issue covered by https://github.com/vllm-project/llm-compressor/pull/3079

src/llmcompressor/pipelines/sequential/ast_helpers.py
```diff
@@ -95,18 +52,15 @@ def autowrap_forward(module: torch.nn.Module, ignore: list[str]):
             "`oneshot`. For example, `oneshot(model.thinker, ...)`"
         )
 
-    # get source code of module forward
-    forward_fn = get_unwrapped_forward(module)
-    source = inspect.getsource(forward_fn)
+    # get source code of module forward, unwrapping decorators (e.g.
+    # ``@can_return_tuple``) so the AST reflects the raw forward body
+    source = inspect.getsource(inspect.unwrap(module.forward))
```
kylesayrs, line 57: This is now landed on main

src/llmcompressor/pipelines/sequential/ast_helpers.py
```diff
@@ -129,9 +83,15 @@ def autowrap_forward(module: torch.nn.Module, ignore: list[str]):
...
-    # patch forward with autowrapped forward
+    # patch forward with autowrapped forward on both the instance and the
+    # class. FX's tracer resolves forward via ``type(module).forward``, so
+    # patching only the instance leaves the decorated class forward in place
+    # and FX cannot trace through decorators like ``@capture_outputs``.
     new_forward = namespace["forward"].__get__(module)
-    with patch_attr(module, "forward", new_forward):
+    with (
+        patch_attr(module, "forward", new_forward),
+        patch_attr(type(module), "forward", namespace["forward"]),
```
kylesayrs, line 93: We intentionally do not patch the class forward. Each module has the potential execute differently, and therefore must be autowrapped individually. FX uses the instance forward

src/llmcompressor/pipelines/sequential/ast_utils/auto_wrapper.py
```diff
@@ -111,6 +117,23 @@ def visit_If(
...
         try:
+            # force a wrap whenever the test combines multiple conditions, even
+            # if statically evaluable. FX cannot constant-fold across
+            # ``if`` branches whose test depends on values produced by the
+            # model itself, so replacing a mixed test with ``True``/``False``
+            # raises during tracing.
+            # Examples: ``(a is None) ^ (b is not None)`` or
+            # ``a is None and b is None``.
+            if isinstance(node.test, ast.BoolOp):
+                raise Exception("If statement combines multiple conditions")
```
kylesayrs, line 128: I don't see why static evaluation alone isn't sufficient?

src/llmcompressor/pipelines/sequential/ast_utils/auto_wrapper.py
```diff
@@ -111,6 +117,23 @@ def visit_If(
...
+
+            # force a wrap whenever the body raises. ``if True/False: raise ...``
+            # would otherwise become a literal raise during tracing, killing
+            # the trace before any call_module nodes are emitted.
+            for stmt in node.body:
+                if isinstance(stmt, ast.Raise):
```
kylesayrs, line 134: This should be covered by either
1. The `if` is not statically evaluatable, in which case the raise is wrapped
2. The `if` is statically evaluatable, in which case it will not trigger during tracing

kylesayrs (changes requested):
I believe that most of the issues that you are seeing will be fixed by just passing and quantizing the thinker part of the model

PR #3118, opened 2026-08-31
feat: preserve and optionally quantize MTP layers in oneshot

soyr-redhat:
## Summary

Current Transformers loading omits MTP layers in these supported checkpoint layouts. This PR processes them during `oneshot()` saving, both `output_dir` and deferred `model.save_pretrained()`, and writes the MTP shard, safetensors index, and runtime quantization metadata alongside the compressed backbone.

- `mtp_quant_scheme=None` (default) reproduces a compatible source MTP format through the compressed-tensors converter. A preset or `QuantizationScheme` applies data-free quantization.
- `mtp_dequantize=True` saves BF16 when no scheme is requested and permits BF16 fallback if conversion cannot be applied. It does not override an explicit target scheme; backbone and MTP dequantization are independent.
- Supported layouts: Qwen3.5/Qwen3.8-27B dense, GLM-5.3 Flash/DSA, and NemotronH. Qwen3.5 MoE supports dense MTP preservation and BF16 conversion only.
- Source layouts, shard headers, and native FP8 scales are validated; fused projections are quantized together. Missing shards, source dequantization errors, and checkpoint write failures remain fatal.

This path is data-free. Calibration-dependent MTP quantization (static FP8, NVFP4 W4A4, GPTQ/AWQ) and nonuniform MTP schemes are out of scope. See the saving guide for API examples and supported formats.

## Test plan

- Focused tests cover small Qwen load/save fixtures and a pinned GLM-5.3 Flash MTP checkpoint, including immediate/deferred save, source-format reproduction, BF16 conversion, data-free quantization, index/config integrity, and failure cases. `make quality` and `git diff --check` pass.
- Full-model vLLM speculative-decoding smoke tests passed for Qwen3.8-27B and GLM-5.3 Flash/DSA on B200s. See the [validation table](https://github.com/vllm-project/llm-compressor/pull/3118#issuecomment-5669821394) and [pinned checkpoints and runtime evidence](https://huggingface.co/datasets/soyrsoyr/pr3118-mtp-e2e-validation). These runs were on identified earlier PR commits, not a serving rerun of the current head or quality/performance benchmarks.

## Compatibility and known limits

- Native block-FP8 MTP dequantization requires compressed-tensors 0.18.1: `FP8BlockDequantizer.validate()` in 0.18.0 returns `None`. The release dependency still pins 0.18.0 and needs updating before merge.
- Twelve native-FP8 Qwen test cases fail during Transformers 5.16.1 model loading, before `oneshot()`. They pass locally with 5.17.0; the current release range allows 5.16.1 but excludes 5.17.0. Resolve the supported range or this load path before claiming it.
- Nemotron MXFP4 serving remains blocked by a vLLM MoE scale-loading issue.

Broader MTP representations can be revisited as upstream Transformers support is integrated.


soyr-redhat:
Validation results and tested checkpoints are listed in the [PR metrics table](https://github.com/vllm-project/llm-compressor/pull/3118#issuecomment-5669821394).

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -79,6 +79,255 @@ def _retie_embeddings(model: PreTrainedModel):
...
+        if quant_config_dict:
+            scheme = _extract_mtp_scheme(quant_config_dict)
+
+    if scheme is not None:
+        output_tensors = {}
+        for key, tensor in mtp_tensors.items():
+            if not key.endswith(".weight"):
+                output_tensors[key] = tensor
+                continue
+            module_name = key[: -len(".weight")]
```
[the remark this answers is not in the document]
soyr-redhat: added

dsikka:
Thank you! Will review tomorrow

soyr-redhat:
nvfp4 pending smoke test

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -79,6 +79,389 @@ def _retie_embeddings(model: PreTrainedModel):
     logger.info("Re-tied input/output embeddings; saving a single shared table.")
 
 
+def _extract_mtp_scheme(quant_config_dict: dict):
+    """
+    Derive a QuantizationScheme suitable for MTP layers from the saved
+    quantization_config. Returns None if the config cannot be parsed.
```
brian-dellabetta: why not just have the user explicitly define which scheme to use in `save_mtp_tensors_to_checkpoint`? It could default to bfloat16 for backwards compatibility

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -79,6 +79,389 @@ def _retie_embeddings(model: PreTrainedModel):
...
+            "be calibrated for MTP layers; MTP layers will be weight-only quantized."
+        )
+        input_acts = None
+
+    return QuantizationScheme(
+        targets=["re:.*\\.weight"], weights=weights_args, input_activations=input_acts
+    )
+
+
+def _compress_mtp_linears(
```
brian-dellabetta: could this be reduced to using the ModelFreePtqConverter directly? to reduce code redundancy?

soyr-redhat: totally,  mfptq refactor hasnt been merged yet so I was just trying to get some progress before to ideally combine after the fact!

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
soyr-redhat: I do like that idea, keeping it a bit more explicit sounds solid, would you prefer i do that?

brian-dellabetta: I know from a UX perspective we try to default to being explicit as much as possible, maybe worth discussing in standup

soyr-redhat: The user selects mtp_scheme explicitly: None preserves source MTP precision, "BF16" requests dequantization/casting, and a data-free preset or QuantizationScheme controls optional MTP quantization independently of the backbone.

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
soyr-redhat: good to go!

soyr-redhat:
@gemini-code-assist review

src/llmcompressor/transformers/compression/mtp.py
```diff
@@ -0,0 +1,633 @@
...
+    source_model = getattr(model, "name_or_path", None) or model.config._name_or_path
+    with suspend_distributed_timeout():
+        if is_source_process():
+            _quantize_and_save_mtp_tensors(
+                source_model,
+                save_directory,
+                model.config,
+                mtp_scheme=mtp_scheme,
+                revision=revision,
+            )
```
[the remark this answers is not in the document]
soyr-redhat: Added an explicit error when an MTP-configured model has no source checkpoint path or Hub ID. I intentionally kept MTP loading/preservation failures fatal: silently returning after the backbone save would produce an incomplete checkpoint and violate this PR’s preservation guarantee.

src/llmcompressor/entrypoints/model_free/microscale.py
```diff
@@ -52,11 +66,23 @@
...
         r"re:.*(attn|attention)\.wkv_a_with_mqa\.weight$",
     ],
-    [r"re:.*mlp\.gate_proj\.weight$", r"re:.*mlp\.up_proj\.weight$"],
+    [
+        r"re:.*(attn|attention)\.q_a_proj\.weight$",
+        r"re:.*(attn|attention)\.kv_a_proj_with_mqa\.weight$",
+    ],
+    [
+        r"re:.*mlp(?:\.(?:experts\.\d+|shared_experts))?\.gate_proj\.weight$",
+        r"re:.*mlp(?:\.(?:experts\.\d+|shared_experts))?\.up_proj\.weight$",
```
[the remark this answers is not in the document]
soyr-redhat: Fixed the in-memory fused matcher to recognize feed_forward consistently with the dependency mapping. Added a regression test for feed_forward gate/up fusion.

src/llmcompressor/transformers/compression/mtp.py
```diff
@@ -0,0 +1,633 @@
...
+    )
+    managed_ignores = set(layout.ignores + layout.full_precision_ignores())
+    managed_ignores.update(
+        f"re:^{re.escape(prefix)}\\." for prefix in layout.source_prefixes
+    )
+    ignores = [
+        value
+        for value in (quantization_config.ignore if quantization_config else [])
+        if value not in managed_ignores
+    ]
```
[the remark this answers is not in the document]
soyr-redhat: Fixed by defaulting a nullable quantization_config.ignore to an empty list. Added a regression test covering MTP preservation with ignore=null.

soyr-redhat:
@gemini-code-assist review

src/llmcompressor/transformers/compression/mtp.py
```diff
@@ -0,0 +1,647 @@
...
+        return tensors
+
+    quantization_config = getattr(config, "quantization_config", None)
+    if hasattr(quantization_config, "to_dict"):
+        quantization_config = quantization_config.to_dict()
+    block_size = (
+        tuple(quantization_config.get("weight_block_size", (128, 128)))
+        if isinstance(quantization_config, dict)
+        else (128, 128)
+    )
```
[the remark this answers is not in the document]
soyr-redhat: Null and missing weight_block_size use the existing (128, 128) default; explicit block sizes are preserved. The dequantization regression test covers all three cases and checks the output tensor values.

soyr-redhat:
Validation status: the full pretrained GLM-5.3 DSA result was tested at PR commit `572c6390fe64db5a9b3fd549174f9b150b1bb617`; GLM-5.3-Flash at `7029daba084fcdb251d4c87a43dc76b60d30ecd3`; the Qwen copy → FP8 → BF16 pipeline was tested at PR commit `3dd54c307dd68074c8af055cb5017093a4880e54`; other rows retain earlier results recorded at `87347881b46df8786b9fc463965f80504914a98f`.

| Family / tested path | Status | Evidence and limitations | Tested checkpoints |
| --- | --- | --- | --- |
| Qwen3.8-27B, BF16 copy → FP8 → BF16 | Smoke passed | PR `3dd54c307` MTP save helper; FP8 backbone reused unchanged. Exact BF16 copy and reference dequantization checks passed. One B200, TP=1, vLLM 0.28.0, one MTP token, one 64-token completion per checkpoint: **30/34 (88.24%), 29/34 (85.29%), 30/33 (90.91%)** accepted, respectively. [Evidence](https://huggingface.co/datasets/soyrsoyr/pr3118-mtp-e2e-validation/blob/81ea4d6ef34c7e78b84496c6bad2adc1dc02235d/reports/qwen38-mtp-roundtrip.json). | [BF16 copy (None)](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-Copy-PR3118/tree/b9049e2b6b8cd526111324a14a3215a4f0757d8e)<br>[FP8_DYNAMIC](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-FP8-Roundtrip-PR3118/tree/224e70420bdea25577f7863029106bc4d0605c20)<br>[BF16 from FP8](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-BF16-Roundtrip-PR3118/tree/c1300e73ba56443b9797fb9213c2b77b10caef79) |
| Qwen3.8-27B, seven variants | Passed | Reported runtime checks passed for MTP preservation, quantization, and FP8-to-FP4 conversion variants. | [BF16](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-BF16-pr3118-validation/tree/bbbfdb242bab29bca528efaf0f5295e0ff38e9d7)<br>[FP8-DYNAMIC](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-FP8-DYNAMIC-pr3118-validation/tree/f169d7e3590cd256ce58b4b55c93912f37c6d2f6)<br>[FP8-BLOCK](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-FP8-BLOCK-pr3118-validation/tree/be6f9226c3cdc5c715b3fb33e01ed3ba270e1fed)<br>[NVFP4A16](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-NVFP4A16-pr3118-validation/tree/8795a15e5e898bca41040e6716a6bc204608b0fe)<br>[MXFP4](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-MXFP4-pr3118-validation/tree/2cd57a5a7237aec81405f60fa8c469a9230ff3bc)<br>[NVFP4A16-FromFP8](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-NVFP4A16-FromFP8-pr3118-validation/tree/692591fa91ee9cd5389c9fa6f011359344070745)<br>[MXFP4-FromFP8](https://huggingface.co/soyrsoyr/Qwen3.8-27B-FP8Dyn-MTP-MXFP4-FromFP8-pr3118-validation/tree/e66b4df21d91c3870032dfb77ae17ac6ab0226f9) |
| Qwen3.5-35B-A3B, dense-MTP preservation | Passed | The preservation path passed; this does **not** establish quantized-MTP support for this MoE architecture. | [BF16](https://huggingface.co/soyrsoyr/Qwen3.5-35B-A3B-FP8Dyn-MTP-BF16-pr3118-validation/tree/246352bbe5f884187721fce4e1ae77ca6d5e932b) |
| Nemotron, BF16, FP8, NVFP4A16 MTP | Passed | Tested variants passed, including FP8-source MTP -> NVFP4A16. | [BF16](https://huggingface.co/soyrsoyr/Nemotron3.5-Lightning-30B-A3B-FP8Dyn-MTP-BF16-pr3118-validation/tree/d53357e1d5550a43372e7afb705e047018ebc725)<br>[FP8-DYNAMIC](https://huggingface.co/soyrsoyr/Nemotron3.5-Lightning-30B-A3B-FP8Dyn-MTP-FP8-DYNAMIC-pr3118-validation/tree/bb392cc00dcb4dfb36f7c6bf5176ed78941acc7f)<br>[NVFP4A16](https://huggingface.co/soyrsoyr/Nemotron3.5-Lightning-30B-A3B-FP8Dyn-MTP-NVFP4A16-pr3118-validation/tree/47de2c040bf1ffc84031156153350abdf20b5034)<br>[NVFP4A16-FromFP8MTP](https://huggingface.co/soyrsoyr/Nemotron3.5-Lightning-30B-A3B-FP8Dyn-MTP-NVFP4A16-FromFP8MTP-pr3118-validation/tree/c72995058172bdf996043dac3b753a05d595aaf4) |
| Nemotron, two MXFP4 variants | Runtime failure | Observed failures are in [vLLM's MXFP4 MoE scale preparation/reshape path](https://github.com/vllm-project/vllm/blob/767d1c4d473a3e81429563aac995f22f7151c325/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w4a4_mxfp4.py#L188-L193). The runtime fix is deferred to upstream vLLM. | [MXFP4](https://huggingface.co/soyrsoyr/Nemotron3.5-Lightning-30B-A3B-FP8Dyn-MTP-MXFP4-pr3118-validation/tree/bb17b8b511bb33cd65b160e7b91e3d85d6c90451)<br>[MXFP4-FromFP8MTP](https://huggingface.co/soyrsoyr/Nemotron3.5-Lightning-30B-A3B-FP8Dyn-MTP-MXFP4-FromFP8MTP-pr3118-validation/tree/729bcbb1b7d01acc3c93ca2be2b1addd6f49c399) |
| GLM-5.3 (DSA), full pretrained model | E2E smoke passed | Data-free NVFP4A16 backbone MLP + MTP; 777 quantized MTP weights. Four B200s, TP=4, vLLM 0.28.0, one MTP token: **1 × 128-token completion; 58/70 draft tokens accepted (82.86%)**. MFPTQ backbone + PR MTP save path; run-local MFPTQ adjustments documented in the linked model card. [Runtime evidence](https://huggingface.co/datasets/soyrsoyr/pr3118-mtp-e2e-validation/blob/c6f26241db46f9e4713e69ec0b546ea929b6d670/reports/glm53-full-b200-runtime.json). Single-request serving smoke, not a quality or performance benchmark. | [NVFP4A16 backbone MLP + MTP](https://huggingface.co/soyrsoyr/GLM-5.3-NVFP4A16-MTP-PR3118/tree/be5d8973114ef4425b90247f477bc70aaa60cc6d) |
| GLM-5.3-Flash, full pretrained model | E2E smoke passed | Data-free NVFP4A16 backbone MLP + MTP; 867 quantized MTP weights. Two B200s, TP=2, one MTP token: **4 × 64-token completions; 116/137 draft tokens accepted (84.7%)**. Tested HF revision `7466c614ce4d1cd9816d62ed5e479309efaa1af3`. [Model card](https://huggingface.co/soyrsoyr/GLM-5.3-Flash-NVFP4A16-MTP-PR3118) / [runtime evidence](https://huggingface.co/soyrsoyr/GLM-5.3-Flash-NVFP4A16-MTP-PR3118/blob/28906bd84916f2b253363e30a8d469757840b985/b200-runtime.json). Raw completion prompts without a chat template; serving/MTP execution only, not a quality or performance benchmark. | [NVFP4A16 backbone MLP + MTP](https://huggingface.co/soyrsoyr/GLM-5.3-Flash-NVFP4A16-MTP-PR3118/tree/7466c614ce4d1cd9816d62ed5e479309efaa1af3) |

Checkpoint links pin the published validation artifacts; link labels identify the MTP format or conversion path. Runtime status remains scoped to the checks described in each row; the linked index contains the current active checkpoints and runtime evidence, alongside labeled historical reports.

**Known upstream limitation:** Nemotron MXFP4 fails in vLLM's MoE scale-loading path; its runtime fix is deferred to upstream vLLM. Rows marked passed establish only their respective checks; this is not blanket validation of every architecture/format combination or a quality/performance benchmark.

Published artifacts are indexed at [pr3118-mtp-e2e-validation](https://huggingface.co/datasets/soyrsoyr/pr3118-mtp-e2e-validation). Backbone and MTP formats are recorded separately: NVFP4A16 denotes weight-only FP4 with 16-bit activations, not calibrated NVFP4 W4A4. Historical H100 reports remain available for provenance.

src/llmcompressor/transformers/compression/mtp.py
```diff
@@ -0,0 +1,647 @@
...
+        try:
+            quantization_input = _dequantize_fp8_blocks(tensors, config)
+            quantized, output = _partition_mtp_tensors(quantization_input, layout)
+            output.update(_compress_mtp_weights(quantized, scheme))
+            output = {
+                name: tensor if tensor.is_contiguous() else tensor.contiguous()
+                for name, tensor in output.items()
+            }
+            save_file(output, shard_path)
+        except (RuntimeError, ValueError) as error:
```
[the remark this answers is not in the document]
soyr-redhat: Optional partitioning and quantization catch Exception and fall back to preserving supported source MTP precision. Source loading/dequantization and checkpoint writes remain fatal. Regression tests cover RuntimeError, ValueError, TypeError, AttributeError, and KeyError across all four layouts, plus native-FP8 fallback and source immutability.

src/llmcompressor/transformers/compression/mtp.py
```diff
@@ -0,0 +1,647 @@
...
+        )
+    with suspend_distributed_timeout():
+        if is_source_process():
+            _quantize_and_save_mtp_tensors(
+                source_model,
+                save_directory,
+                model.config,
+                mtp_scheme=mtp_scheme,
+                revision=revision,
+            )
```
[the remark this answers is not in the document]
soyr-redhat: nah

docs/guides/saving_a_model.md
```diff
@@ -78,6 +78,72 @@ model.save_pretrained(
 tokenizer.save_pretrained(SAVE_DIR)
 ```
 
+### Preserving and Quantizing MTP Layers
+
+Some models ship Multi-Token Prediction (MTP) layers used as the draft model for speculative decoding in vLLM. Transformers omits these layers from the loaded backbone, so LLM Compressor reads them separately from the source checkpoint when saving. Both `oneshot(output_dir=...)` and `oneshot(...); model.save_pretrained(...)` write MTP alongside the backbone. Source layouts and shards are validated before calibration.
+
+Use `mtp_scheme=None` (the default) to preserve source MTP precision. Dense tensors are copied unchanged; supported quantized weights and scales are retained with compatible output metadata. Use `mtp_scheme="BF16"` to explicitly dequantize or cast MTP to BF16 and exclude its runtime modules from quantization. Compressed-tensors sources must include compatible `mtp_group` metadata.
```
HDCharles: This name is a little vague, imo mtp_quant_scheme would be better. When I see mtp_scheme I think speculator vs mtp heads, though I get the recipe arg for normal quant is called scheme.

Is uniform quantization a requirement for MTP layers. If not it feels like maybe we should just use the existing interface e.g. config groups or use that in combination with a field that enables dequant to bf16, since that's new functionality. If e.g. we decide quantizing token head 2/3 with fp8 and rest with nvfp4 is a good strategy, it seems like this interface wouldn't work well.

docs/guides/saving_a_model.md
```diff
@@ -78,6 +78,72 @@ model.save_pretrained(
...
+
+For a native FP8 checkpoint, first load the backbone with Transformers'
+`FineGrainedFP8Config(dequantize=True)`, then pass that model to `oneshot`.
+The backbone must be fully dequantized. MTP processing reads the original
+source separately: `None` preserves supported FP8 weights and scales, while
+`"BF16"` dequantizes them. Backbone and MTP precision are independent.
+A still-quantized FP8 backbone remains unsupported.
```
HDCharles, line 92: This dual interface is confusing I would expect

fp8 -> save fp8: all normal, scheme None
fp8 -> save bf16: dequantize=True, scheme None
fp8 -> save NVFP4: dequantize=True, scheme NVFP4

Which would separate the dequant and quant interfaces. I.e. dequant is set in this FP8Config object quant is specified in one shot.

docs/guides/saving_a_model.md
soyr-redhat: Yeah, makes sense to me, I think the rename is completely logical. In terms of mixed precision, I honestly hadn't taken that into consideration. afaict, I don't think uniform precision is a requirement, and could maybe down the line be an interesting candidate for HIGGS applications. I'll start off by that smaller api change and then do some digging on how the scope might change based on those efforts

docs/guides/saving_a_model.md
soyr-redhat: Okay sure, I think this is totally fair too. My only concern in the current moment is that since we read backbone/mtp separately, if we wanted to wire them in together, we would lose the ability to dequant the backbone whilst preserving mtp

docs/guides/saving_a_model.md
soyr-redhat: Okay, renamed. Also, I made sure that mtp goes through the converter in the same way as the rest of the pipeline. That should be conducive to mixed precision down the line (ideally just allowing mtp_scheme to accept a config_groups mapping), could be a good follow up. thoughts?

docs/guides/saving_a_model.md
soyr-redhat: Okay, tried to follow your interface advice, 


 So the cases map to:
- fp8 -> fp8: mtp_quant_scheme=None (reproduced through the converter)
- fp8 -> bf16: mtp_dequantize=True
- fp8 ->  NVFP4: mtp_quant_scheme="NVFP4A16" (intermediate dequant handled automatically)


From claude

"The reason dequant is a separate MTP arg rather than living in the load-time FineGrainedFP8Config: Transformers doesn't load the MTP layers, so we read them from the source checkpoint at save time. The backbone's FineGrainedFP8Config(dequantize=True) never reaches that path, so backbone and MTP precision have to be controlled independently — otherwise "dequantize the backbone but keep MTP in fp8" isn't expressible."

I'm not personally certain this is the perfect solution, but I think given the constraints we have, this isn't too bad

soyr-redhat:
@gemini-code-assist review

soyr-redhat:
to land, would need ct 0.18.1, we call into validate() which in 0.18.0 returns None

src/llmcompressor/entrypoints/oneshot.py
```diff
@@ -302,6 +323,30 @@ def validate_model(self, model: PreTrainedModel):
...
+                    and buffer.dtype not in (float16, bfloat16, float32)
+                )
+                for name, buffer in model.named_buffers()
+            )
+        ):
+            # Transformers removes its quantizer after an explicit dequantizing
+            # load. The on-disk FP8 config still describes unloaded MTP tensors,
+            # which must be converted separately by the save pathway.
+            _remove_fp8_save_roundtrip(model)
+            return
```
HDCharles, line 348: why is this needed? i am struggling to see a reason for specific fp8 handling, i'm also unsure if this is a branch for if we are quantizing to fp8 vs quantizing from fp8, but it seems like this should either be needed or not needed in all cases.

also this is validate, why are we modifying things?

src/llmcompressor/entrypoints/utils.py
```diff
@@ -109,6 +109,9 @@ def post_process(
     actions. The model is saved in a compressed format if specified in `model_args`.
     Additionally, the tokenizer or processor, if available, is also saved.
 
+    :param model_args: Model loading and saving arguments.
+    :param recipe_args: Recipe arguments used by the completed oneshot run.
+    :param output_dir: Directory where the compressed artifact is saved.
```
HDCharles, line 114: rebase failure?

src/llmcompressor/transformers/compression/compressed_tensors_utils.py
```diff
@@ -29,6 +29,79 @@
 __all__ = ["modify_save_pretrained"]
 
 
+def _remove_fp8_save_roundtrip(model: PreTrainedModel) -> None:
+    """Keep model renamings/reshapes, but do not undo an FP8 dequantizing load.
```
HDCharles, line 33: what does this mean?

PR #3123, opened 2026-09-01
Fix noisy GPT-OSS linearize test

PranjalAdhikari:
## Summary

Fix the noisy GPT-OSS linearize test by explicitly initializing
available projection biases to zero.

## Changes

- Initialize `up_proj` bias to zero when present.
- Initialize `down_proj` bias to zero when present.

This makes the test setup deterministic for models that contain
projection biases.

## Test Plan

Not run locally because the test requires a CUDA-enabled GPU and
the local environment is CPU-only.


kylesayrs (commented):
Nice catch! Can we initialize biases with `normal_` instead?

kylesayrs (approved):

brian-dellabetta (approved):

HDCharles (approved):

PR #3127, opened 2026-09-03
[Datatsets] Respect num_calibration_samples when loading calibration data

dsikka:
Previously num_calibration_samples was only applied in the DataLoader sampler, after the entire split had been loaded and tokenized. And with no split specified, all splits were loaded as a DatasetDict and a downstream fallback only knew how to pick a "train" split (breaking ultrachat_200k, which has train_sft/test_sft).

- Add TextGenerationDataset.DEFAULT_SPLIT so each dataset resolves its own default split (ultrachat_200k -> train_sft, flickr30k/peoples_speech -> test, custom -> None); base defaults to "train".
- Collapse a multi-split DatasetDict to a single split before processing.
- Trim to num_calibration_samples before the preprocess/tokenize maps so we don't tokenize samples the sampler would only discard.
- Simplify get_processed_dataset's split handling accordingly.

## Calibration data resolution

| Config | Calibration samples come from |
|---|---|
| `splits` + `num_calibration_samples` | `num` samples **from your split** (shuffled → seed-42 subset; unshuffled → the `num` longest in the split) |
| `splits` only | Same, with the implicit default 512 (split < 512 → whole split + warning) |
| `num_calibration_samples` only | 512 from the class `DEFAULT_SPLIT` (`train`, `train_sft`, `test`, …) |
| neither | `DEFAULT_SPLIT`, trimmed to the default 512 |
| `num_calibration_samples=None` | The whole split (no trimming) |
| `concatenate_data=True` | `num` *packed chunks* selected by the sampler (packing may yield fewer) |

The trim is only applied where it doesn't change the sampled set: it is skipped when `concatenate_data=True` (the sampler must count packed chunks, not rows) and when `shuffle_calibration_samples=False` (`LengthAwareSampler` must see the whole split to pick the longest sequences). The multi-split `DatasetDict` fallback picks the lexicographically first split so the choice is deterministic.


src/llmcompressor/datasets/utils.py
```diff
@@ -103,20 +106,13 @@ def get_processed_dataset(
...
-                dataset = dataset["train"]
-            else:
-                raise ValueError(
-                    "No split specified and 'train' split not found in dataset. "
-                    "Please specify `splits` explicitly."
-                )
+            raise ValueError(
+                f"Expected a single calibration split but loaded {type(dataset)}. "
+                "Please specify `splits` explicitly."
+            )
```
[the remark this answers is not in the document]
kylesayrs: Not worried about this atm

src/llmcompressor/transformers/data/base.py
```diff
@@ -81,13 +81,20 @@ class TextGenerationDataset(RegistryMixin):
...
     def __init__(
         self,
         dataset_args: DatasetArguments,
         split: str,
         processor: Processor,
     ):
         self.dataset_args = dataset_args
+        if split is None:
+            split = self.DEFAULT_SPLIT
         self.split = split
```
kylesayrs:
```suggestion
        self.split = split or self.DEFAULT_SPLIT
```

src/llmcompressor/transformers/data/base.py
```diff
@@ -134,8 +141,18 @@ def __call__(self, add_labels: bool = True) -> DatasetType:
...
+        # When no split is specified a DatasetDict of every split is loaded; collapse
+        # it to a single split before doing any (expensive) processing on it.
+        dataset = self._select_split(dataset)
         logger.debug(f"Raw dataset: {get_columns(dataset)}")
 
+        # Limit to `num_calibration_samples` *before* the expensive preprocessing and
+        # tokenization maps. Otherwise the entire split (potentially hundreds of
+        # thousands of samples) is processed just to have the calibration sampler
+        # discard all but `num_calibration_samples` of it.
+        dataset = self._limit_calibration_samples(dataset)
```
kylesayrs, line 152: Looks good

kylesayrs (approved):

kylesayrs (approved):
Looks good, thanks

PR #3109, opened 2026-08-27
Intergrate triton fuse gptq update kernels

Isotr0py:
SUMMARY:
"please provide a brief summary"
- This PR intergrate a triton kernel for fuse gptq update to speed up the slow gptq update

TEST PLAN:
"please outline how the changes were tested"
<details>

<summary>Kernel microbench script</summary>

```python3
"""Microbenchmark (lean): fused Triton kernel + batched GPTQ vs baselines.

Shapes mirror Qwen3.8-Flash-Next MoE experts: gate_up [1280, 2560],
down [2560, 640]. Schemes: NVFP4 (tensor_group g16) and W4A16 int4 g128.
Fast measurements print first; the slow eager baseline runs last, 1 rep only.
"""

import copy
import os
import time

import torch
from compressed_tensors.quantization import QuantizationArgs, QuantizationScheme

from llmcompressor.modifiers.gptq.gptq_quantize import (
    quantize_weight,
    quantize_weight_batched,
)
from llmcompressor.modifiers.quantization.calibration import (
    initialize_observer,
    observe,
)

DEV = "cuda"
SCHEMES = {
    "NVFP4-g16": QuantizationArgs(
        num_bits=4, type="float", symmetric=True, strategy="tensor_group",
        group_size=16,
    ),
    "INT4-g128": QuantizationArgs(
        num_bits=4, symmetric=True, strategy="group", group_size=128
    ),
}
SHAPES = {"gate_up": (2560, 1280), "down": (640, 2560)}  # (in, out)


def make_mod(in_f, out_f, qa, seed):
    torch.manual_seed(seed)
    m = torch.nn.Linear(in_f, out_f, bias=False).to(DEV)
    m.quantization_scheme = QuantizationScheme(targets=["Linear"], weights=qa)
    initialize_observer(m, "weight")
    observe(m, "weight")
    return m


def make_hessian(n, seed):
    g = torch.Generator("cpu").manual_seed(seed)
    A = torch.randn(n, n, generator=g)
    return (A @ A.T + torch.eye(n)).to(device=DEV, dtype=torch.float32)


def bench(fn, warmup=1, iters=3):
    for _ in range(warmup):
        fn()
    torch.cuda.synchronize()
    t0 = time.perf_counter()
    for _ in range(iters):
        fn()
    torch.cuda.synchronize()
    return (time.perf_counter() - t0) / iters


def report(scheme_name, shape_name, name, t):
[...]

PR #3128, opened 2026-09-03
Add batched GPTQ quantization with actorder support

### LINKED ISSUE #1496: Use torch.compile to speed up GPTQ algorithm

HDCharles:
## Summary
based on https://github.com/vllm-project/llm-compressor/pull/3109 
fixes https://github.com/vllm-project/llm-compressor/issues/1496

- adds gptq triton kernel (~15x faster)
- adds option to batch gptq together for layers with same shape. (~1.67x [~30x total] faster in limited tests)
 (good for e.g. MoE)
- removed hessian offloading
- improved eager path (this actually yields a 1.5-2x speedup even ignoring batching/triton)
  - not calculating error w-q twice: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L235-L237
  - not passing zp to fake_quantize if its 0 https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L224
  - improved intra block update err1 - [out_feat x 1] @ Hinv1[i, i:] -  [1 x remaining_block] matmul is less efficient than elementwise mul: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L238C46-L238C59
  - using out=H rather H=op(H,...) in hessian inversion: https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/modifiers/gptq/gptq_quantize.py#L144-L147
 
 
### results: 

Runtime (s)
| Implementation | Batching | Llama 3 8B | Qwen3-30B-A3B, one layer | Qwen3-30B-A3B, full model |
|---|---:|---:|---:|---:|
| Main - Hessian offload | N/A | 408.70 s | 146.28 s | N/A |
| Main - No Hessian offload | N/A | 266.39 s | 128.42 s | 6,653.18 s |
| PR eager | No | 157.61 s | 75.10 s | 3,779.61 s |
| PR eager | Yes | 115.17 s | 6.10 s | 278.94 s |
| PR Triton | No | 17.90 s | 7.14 s | 336.76 s |
| PR Triton | Yes | 17.19 s | 4.69 s | 212.79 s |
| PR Triton, attention + `ExpertMLPWithGate` 8 targets | Yes | N/A | 7.04 s | 281.99 s |

all told, for MoE which spends more time doing GPTQ and can make most use of batching its about a 30x speed up e2e, for the dense model, its a 15x speed up, primarily because batching has no effect. Batching yields a ~1.5x improvement e2e for the MoE model.

Memory (GiB)
| Implementation | Batching | Llama 3 8B | Qwen3-30B-A3B, one layer | Qwen3-30B-A3B, full model |
|---|---:|---:|---:|---:|
| Main - Hessian offload | N/A | 19.31 GiB | 3.45 GiB | 1.60 GiB  |
| Main - No Hessian offload | N/A | 19.31 GiB | 7.02 GiB | 5.96 GiB |
| PR eager | No | 18.43 GiB | 6.95 GiB | 5.90 GiB |
| PR eager | Yes | 18.43 GiB | 16.18 GiB | 15.11 GiB |
| PR Triton | No | 18.43 GiB | 6.95 GiB | 5.90 GiB |
| PR Triton | Yes | 18.43 GiB | 16.18 GiB | 15.11 GiB |
| PR Triton, attention + `ExpertMLPWithGate` 8 targets | Yes | N/A | 3.34 GiB | 1.25 GiB |

batching uses a lot of memory though we only adaptively batch based on available space. One thing to note is that since we removed the hessian offloading, we were worried that large models would run into memory issues. However i tested for small sequential targets which does even better than the offloading did.

Kernel Microbenchmarks
| Model shape | Triton B=1 | Eager B=1 | Triton speedup | Batch | Batched | Loop extrapolation | Batch speedup |
|---|---:|---:|---:|---:|---:|---:|---:|
| Qwen attention Q/O (`2048×2048`) | 22.4 ms | 402.6 ms | 17.96x | 2 | 42.4 ms | 44.8 ms | 1.06x |
| Qwen attention K/V (`256×2048`) | 10.8 ms | 244.9 ms | 22.60x | 2 | 17.8 ms | 21.7 ms | 1.22x |
| Qwen expert gate/up (`768×2048`) | 14.1 ms | 254.1 ms | 18.05x | 128 | 1,062.3 ms | 1,802.7 ms | 1.70x |
| Qwen expert down (`2048×768`) | 8.8 ms | 104.6 ms | 11.85x | 128 | 893.9 ms | 1,129.7 ms | 1.26x |
| Llama attention Q/O (`4096×4096`) | 74.2 ms | 860.2 ms | 11.60x | 2 | 144.5 ms | 148.3 ms | 1.03x |
| Llama attention K/V (`1024×4096`) | 33.9 ms | 520.6 ms | 15.36x | 2 | 62.1 ms | 67.8 ms | 1.09x |
| Llama MLP gate/up (`14336×4096`) | 235.4 ms | 766.7 ms | 3.26x | 2 | 464.8 ms | 470.8 ms | 1.01x |
| Llama MLP down (`4096×14336`) | 482.1 ms | 2,249.3 ms | 4.67x | 1 | 486.0 ms | 482.1 ms | 0.99x |

we hit a max of 15x * 1.09x -> 16.35 for llama and a max of 18x * 1.7x -> 30x for Qwen-30B-A3B, surprisingly those numbers are very close to our final e2e improvement over previous main due to other improvements beyond the kernel

repro: https://gist.github.com/HDCharles/bdc242021d4ba4c9e430f72221daedb7

### base.py
1) (before) 2 paths compress_module_batch & quantize_weight_batched + compress_module_list & quantize_weight (now) always use batching, fallback to batchsize=1, path is now compress_module_list (handles stacking/batching) -> compres_batch -> quantize_weight
2) actorder+stacking applied before quantize_weight (avoids keeping an extra additional copy of the stack in gpu memory)
3) made max_batch_size more accurate, takes into account remaining space, better estimate of the remaining allocations including multiplicity.
4) moved fallback behavior to be controlled here rather than in the triton kernel code

### gptq_quantize.py
1) use the triton kernel registration frameowkr for gptq_block_update/_gptq_block_update_triton
2) eager path now uses fake_quantize to align with previous main numerically (scale rounded to scale_dtype)
3) remove quantize_weight_single
4) allow fp8 and block quant to go through triton/eager paths

### gptq_triton.py
1) added DEQUANT_DTYPE to match existing numerics
2) added fp8 to triton kernel
3) added block quantization to triton kernel

## Validation
- `PYTHONPATH=. pytest -q tests/llmcompressor/modifiers/gptq/test_gptq_quantize.py --confcutdir=tests/llmcompressor/modifiers/gptq --disable-warnings --maxfail=1`
- 45 passed
- batched actorder and fallback smoke tests passed

compared main/PR numerics, got bitwise parity (https://gist.github.com/HDCharles/bdc242021d4ba4c9e430f72221daedb7)


HDCharles:
@Isotr0py can you give your thoughts, i reworked a lot of things to make it more maintainable, let me know what you think

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -65,46 +124,55 @@ def accumulate_hessian(
...
+    if actorder and perm is None:
+        raise ValueError("actorder requires pre-permuted weights, hessians, and perm")
+    final_dtype = weights.dtype
+    device = weights.device
+    # The caller provides a disposable stacked weight tensor, so use it as the
+    # working buffer when it is already FP32 instead of allocating another copy.
+    W = weights.to(device=device, dtype=GPTQ_PRECISION)
+    # The stacked Hessian is the disposable working buffer. The caller retains
+    # the original per-module Hessians separately for batch fallback.
+    H = hessians.to(device=device, dtype=GPTQ_PRECISION)
```
kylesayrs, line 164: Thanks for the comments

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,35 +310,207 @@ def compress_modules(self):
...
+                            batch,
+                            quant_args,
+                            weights,
+                            hessians,
+                            scales,
+                            zero_points,
+                            global_scales,
+                            perm,
+                        )
+                    except Exception as error:
```
kylesayrs: Should we keep this fallback? It seems like a lot of complexity for something that we should hopefully fix upstream

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,35 +310,207 @@ def compress_modules(self):
...
+
+                try:
+                    hessian_list = []
+                    for module in batch:
+                        with self._maybe_onload_hessian(module):
+                            hessian = self._hessians[module]
+                            num_samples = self._num_samples[module].to(
+                                device=hessian.device
+                            )
+                            hessian_list.append(hessian / num_samples)
```
kylesayrs: We can get better memory characteristics here if we remove the fallback and allow for inplace operation on the hessian

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,35 +310,207 @@ def compress_modules(self):
...
+                            if triton_failed:
+                                retry_stack.enter_context(
+                                    patch.dict(
+                                        os.environ,
+                                        {"LLMCOMPRESSOR_DISABLE_GPTQ_TRITON": "1"},
+                                    )
+                                )
+                            self.compress_module_list(
+                                batch,
+                                qparams={module: qparams[module] for module in batch},
```
kylesayrs: We can remove `qparams` if we get rid of the fallback

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -116,153 +184,334 @@ def quantize_weight(
...
+    i1: int,
+) -> None:
+    """Run one GPTQ block with the registered Triton backend."""
+    kernel_config = _get_triton_gptq_config(quant_args)
+    if kernel_config is None:
+        raise ValueError(f"Unsupported Triton GPTQ scheme: {quant_args}")
+
+    block_width = W1.shape[-1]
+    if block_width > 256 or block_width & (block_width - 1):
+        raise ValueError("Triton GPTQ block width must be a power of two <= 256")
```
kylesayrs: Should this be added to the requirement function?

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -116,153 +184,334 @@ def quantize_weight(
...
-            "of calibration samples, or shuffling the calibration dataset. "
-            "Falling back to round-to-nearest for this module."
-        )
-        used_rtn_fallback = True
-        Hinv = H = torch.eye(num_columns, dtype=H.dtype, device=H.device)
-
+    damp = percdamp * torch.diagonal(H, dim1=-2, dim2=-1).mean(dim=-1)
+    torch.diagonal(H, dim1=-2, dim2=-1).add_(damp.unsqueeze(-1))
+    info = torch.empty(batch_size, dtype=torch.int32, device=device)
+    torch.linalg.cholesky_ex(H, check_errors=False, out=(H, info))
```
kylesayrs, line 202: Nice

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -116,153 +184,334 @@ def quantize_weight(
...
+    bad = info.nonzero(as_tuple=False).flatten()
+    if bad.numel() and batch_size > 1:
+        raise torch.linalg.LinAlgError("batched GPTQ Hessian inversion failed")
```
kylesayrs: Can we not still proceed in the batched case? I feel like we shouldn't need to raise this error

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -116,153 +184,334 @@ def quantize_weight(
...
+    if bad.numel():
+        H[bad[0]].copy_(torch.eye(num_columns, dtype=H.dtype, device=device))
+        used_rtn_fallback[bad[0]] = True
+    else:
+        torch.cholesky_inverse(H, out=H)
+        torch.linalg.cholesky(H, upper=True, out=H)
+    Hinv = H
     # See section 3.4 of https://arxiv.org/abs/2203.07259
     for i1 in range(0, num_columns, blocksize):
```
kylesayrs, line 217: We should consider cudagraph/ torch compiling this at some point

src/llmcompressor/modifiers/gptq/gptq_quantize.py
kylesayrs: Imho make this an assert. It should be caught by the requirement

kylesayrs (commented):
Looks good. I'd strongly prefer to remove the "fallback" logic if possible. Since this is our core algorithm, we should be extremely aware if any programming errors occur

src/llmcompressor/modifiers/gptq/base.py
HDCharles: the main reason for the fallback is OOM situations with batching, its less of a worry now that i fixed all the batch creation logic to use remaining space so i can get rid of it if you think its not worth it

src/llmcompressor/modifiers/gptq/base.py
HDCharles: agreed i had this at one point but changed it for the fallback

src/llmcompressor/modifiers/gptq/base.py
HDCharles: yeah

src/llmcompressor/modifiers/gptq/gptq_quantize.py
HDCharles: fixed

src/llmcompressor/modifiers/gptq/gptq_quantize.py
HDCharles: agree

src/llmcompressor/modifiers/gptq/base.py
HDCharles: removed

dsikka:
_Note: for internal purposes only. External contributors are safe to ignore this message._

Tracked in JIRA: https://redhat.atlassian.net/browse/INFERENG-10585

**JIRA Details:**
- Issue Type: Task
- Priority: Normal
- Component: LLM Compressor
- Status: Backlog

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -116,153 +182,337 @@ def quantize_weight(
...
+    i1: int,
+) -> None:
+    """Run one GPTQ block with the registered Triton backend."""
+    kernel_config = _get_triton_gptq_config(quant_args)
+    if kernel_config is None:
+        raise ValueError(f"Unsupported Triton GPTQ scheme: {quant_args}")
+
+    block_width = W1.shape[-1]
+    assert (
+        block_width > 256 or block_width & block_width - 1
```
kylesayrs: Invert
```suggestion
        0 < block_width <= 256 and not block_width & (block_width - 1)
```

kylesayrs (changes requested):

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,35 +306,177 @@ def compress_modules(self):
...
+        zero_points: torch.Tensor,
+        global_scales: torch.Tensor | None,
+    ):
+        names = [self._module_names[module] for module in modules]
+        logger.info(f"Quantizing {len(modules)} module(s): {names}")
+
+        try:
+            with contextlib.ExitStack() as ctx_stack:
+                comp_loggers = [
+                    ctx_stack.enter_context(CompressionLogger(module))
```
kylesayrs, line 374: I suspect that polling the GPU for memory statistics might cut into runtime significantly

src/llmcompressor/modifiers/gptq/gptq_quantize.py
HDCharles: fixed

src/llmcompressor/modifiers/gptq/base.py
HDCharles: its like .1s per layer, not super significant, going to leave it for now.

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -126,7 +130,8 @@ class GPTQModifier(Modifier, QuantizationMixin):
     block_size: int = 128
     dampening_frac: float | None = 0.01
     actorder: ActivationOrdering | Sentinel | None = Sentinel("static")
-    offload_hessians: bool = False
+    batched_quantization: bool = True
+    batch_memory_fraction: float = 0.75
```
kylesayrs: Lmk what you think of this interface. I think these two fields are slightly redundant
```python
batch_size: str | int | None = "auto"
"""
"auto" := 75% of free memory
int := actual size
None := disable batching
"""
```

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,70 +299,207 @@ def compress_modules(self):
         # broadcast compressed modules to each rank
         broadcast_qparams_and_cleanup(module_list, module_to_rank, _GPTQ_Q_PARAMS)
 
-    def compress_module_list(self, module_list):
-        for module in module_list:
-            name = self._module_names[module]
-            num_samples = self._num_samples[module]
-            quant_args = getattr_chain(module, "quantization_scheme.weights")
+    def compress_module_list(self, module_list, qparams=None):
```
kylesayrs: Is this function ever passed with `qparams` not being `None`? If not, can we remove this argument?

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,70 +299,207 @@ def compress_modules(self):
...
                 torch.no_grad(),
-                align_module_device(module),
-                self._maybe_onload_hessian(module),
-                CompressionLogger(module) as comp_logger,
+                contextlib.ExitStack() as ctx_stack,
             ):
-                loss, q_param_dict, used_rtn_fallback = quantize_weight(
-                    module=module,
+                for module in batch:
+                    ctx_stack.enter_context(align_module_device(module))
```
kylesayrs: This context is equivalent to just using the `disable_offloading` context
```suggestion
                disable_offloading(),
):
```

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,70 +299,207 @@ def compress_modules(self):
...
+                for module in batch:
+                    hessian = self._hessians.pop(module)
+                    num_samples = self._num_samples.pop(module).to(
+                        device=hessian.device
+                    )
+                    hessian_list.append(hessian / num_samples)
+                    del hessian, num_samples
+
+                hessians = torch.stack(hessian_list)
+                del hessian_list
```
kylesayrs: Consider breaking out into a "batch_and_normalize_hessians" helper

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,70 +299,207 @@ def compress_modules(self):
...
+                    [qparam["zero_point"] for qparam in batch_qparams]
+                )
+                global_scales = None
+                if batch_qparams[0]["global_scale"] is not None:
+                    global_scales = torch.stack(
+                        [
+                            qparam["global_scale"].reshape(-1)[0]
+                            for qparam in batch_qparams
+                        ]
+                    )
```
kylesayrs: Similarly, this might be nice as a `batch_qparams` helper. Helps logically group with the `qparams` variable as well

```python
def batch_qparams(modules):
   qparams = ...  # only has to exist here locally, not at top level
```

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,70 +299,207 @@ def compress_modules(self):
...
+                    hessian = self._hessians.pop(module)
+                    num_samples = self._num_samples.pop(module).to(
+                        device=hessian.device
+                    )
+                    hessian_list.append(hessian / num_samples)
+                    del hessian, num_samples
+
+                hessians = torch.stack(hessian_list)
+                del hessian_list
+                weights = torch.empty(
```
kylesayrs: Is this meant to be empty?

src/llmcompressor/modifiers/gptq/base.py
kylesayrs: Yeah that's fine, but I think the argument around excessive logging still stands

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,70 +299,207 @@ def compress_modules(self):
...
+                q_param_dict["weight_global_scale"] = global_scales[index].to(
+                    dtype=module.weight.dtype
+                )
             for attr, val in q_param_dict.items():
                 update_offload_parameter(module, attr, val)
 
+    def _make_batches(
+        self, module_list: list[torch.nn.Module]
+    ) -> list[list[torch.nn.Module]]:
+        """Partition modules into execution batches."""
```
kylesayrs: Maybe add a more verbose docstring, mentioning moes

src/llmcompressor/modifiers/gptq/base.py
```diff
@@ -297,70 +299,207 @@ def compress_modules(self):
...
+            else:
+                pending.setdefault(key, []).append(module)
+
+        for members in pending.values():
+            max_batch = self._max_batch_size(members[0])
+            for start in range(0, len(members), max_batch):
+                batches.append(members[start : start + max_batch])
+        return batches
+
+    def _batch_key(self, module: torch.nn.Module) -> tuple | None:
```
kylesayrs: docstring

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -7,13 +8,73 @@
...
+    del permuted_hessians
 
-__all__ = ["make_empty_hessian", "accumulate_hessian", "quantize_weight"]
+    permuted_weights = torch.gather(
+        weights,
+        -1,
+        weight_perm.unsqueeze(-2).expand(-1, num_rows, -1),
+    )
+    weights.copy_(permuted_weights)
+    del permuted_weights
```
kylesayrs: variables are garbage collected on function exit

src/llmcompressor/modifiers/gptq/gptq_quantize.py
```diff
@@ -64,47 +125,108 @@ def accumulate_hessian(
     return H, num_samples
 
 
+def _factorize_hessian(
```
kylesayrs: Why do you use the term "factorize" rather than "invert"?

kylesayrs (approved):
Looks good, mostly code comments.
Consider adding a `helpers.py` to help clean up the `base.py` implementation, since we're adding 1K lines

src/llmcompressor/modifiers/gptq/gptq_quantize.py
HDCharles: because it takes the inverse and then factorizes that into the upper triangular part which is what GPTQ uses.

src/llmcompressor/modifiers/gptq/base.py
HDCharles: yes

PR #3130, opened 2026-09-03
fix(autoround): pre-seed fp_ref from IntermediatesCache to skip collect_reference

### LINKED ISSUE #3129: [Bug] AutoRound + SequentialPipeline: collect_reference redundant FP16 forward exhausts CPU RAM for large N
## Environment

- llm-compressor: `main` (post-#3055)
- auto-round: 0.14.2
- PyTorch: 2.x, CUDA
- Hardware tested: 8 × NVIDIA H20 (140 GB VRAM, ~1.8 TiB RAM)
- Model: `Qwen3_5MoeForConditionalGeneration` (122B MoE, BF16)

---

## Bug Description

When `SequentialPipeline` is used with `AutoRoundModifier` and `propagate_error=False`
(the AutoRound default), `quantize_block` begins every block by calling `collect_reference`
— a full FP16 forward pass over all N/world_size calibration inputs — to obtain reference
outputs for the 200-iter SignSGD loss computation. These outputs are stored in CPU RAM.

The `SequentialPipeline` already holds the **identical tensors** in `IntermediatesCache[k+1]`
(the next subgraph's inputs, which equal the current subgraph's unquantized FP16 outputs).
`collect_reference` re-derives them from scratch, adding ~39 GB CPU RAM per rank (N=8192,
8-GPU DDP). This duplication exhausts the node's physical memory, causing the OS to
SIGKILL one or more ranks; the remaining ranks then hang on the next NCCL collective
until the 600 s watchdog fires.

---

## Reproduction

Triggers with `propagate_error=False` (default), large N, and a large model:

```python
recipe = AutoRoundModifier(
    targets="Linear",
    scheme="W4A16",
    iters=200,
    # propagate_error defaults to False
)
oneshot(
    model="<122B MoE model>",
    recipe=recipe,
    dataset=calibration_data,
    num_calibration_samples=8192,  # N=8192 triggers OOM; N=4096 is marginal
)
```

**Observed failure pattern** (consistent across all affected runs):

```
Last successful log line before termination:
[Rank N] apply_autoround | INFO - Applying AutoRound on layer model.layers.X

Then silence — the rank is SIGKILL'd by the OS OOM killer before Python
can write an exception. The DDP launcher reports:
  exitcode: -9 (SIGKILL) / exitcode: 1
  error_file: <N/A>

Remaining ranks hang on NCCL ALLGATHER until watchdog fires:
[Rank 7] Watchdog caught collective operation timeout:
  WorkNCCL(OpType=ALLGATHER) ran for 600425 milliseconds before timing out.
```

CPU RAM spike observed in Memory Monitor logs right before termination:

```
After subgraph k calibration completes (IntermediatesCache[k+1] filled):
[Memory Monitor] peak_ram: 310.27 GB   ← IntermediatesCache, N=8192, 8 ranks

collect_reference would add another ~310 GB (same tensor, re-derived)
→ total ~620 GB + model weights → exceeds 1.8 TiB node RAM → SIGKILL
```

---

## Measured Impact

```
CPU RAM overhead per rank = (N / world_size) × seq × hidden × 2B
```

| N | N/rank | collect_reference CPU RAM overhead/rank | Without fix |
|---|-------:|----------------------------------------:|-------------|
| 4096 | 512 | ~19 GB (512 × 6144 × 3072 × 2) | marginal, passes |
| **8192** | **1024** | **~39 GB (1024 × 6144 × 3072 × 2)** | **CPU RAM OOM → SIGKILL → NCCL timeout** |

With N=8192 and 8 ranks, `IntermediatesCache` already consumes ~310 GB system RAM.
`collect_reference` adds another ~310 GB of identical data → total ~620 GB (plus
~244 GB model weights, Python heap, etc.) → exceeds 1.8 TiB node limit.

---

## Root Cause

With `propagate_error=False`, every subgraph sees the original unquantized activations
as input. Therefore:

```
IntermediatesCache[k+1]   =  FP16 output of block k  (input = cur_inputs[k], unquantized)
collect_reference output  =  FP16 output of block k  (input = cur_inputs[k], unquantized)
```

Same block + same input + same unquantized weights → **outputs are identical**.

`CompressionOrchestrator.quantize_block` (in `auto_round/compressors/orchestrator.py`) does
not currently accept `reference_output` — it always runs the collect_reference forward pass.
llm-compressor's `apply_autoround` therefore has no way to supply the cached tensors, even
though `IntermediatesCache` already holds the identical data.

The fix requires adding `reference_output=None` to both `CompressionOrchestrator.quantize_block`
and `AlgorithmComposer.compress_block` (in `auto_round/algorithms/composer.py`), so the
forward pass can be skipped when a pre-seeded result is provided by the caller.

---

## Proposed Fix

Two coordinated changes — one in llm-compressor (this PR), one companion PR to intel/auto-round:

### 1. intel/auto-round — add `reference_output` to `quantize_block` / `compress_block`

`CompressionOrchestrator.quantize_block` and `AlgorithmComposer.compress_block` need a
new optional parameter. When provided, Step 3 (`collect_reference` forward pass) is skipped:

```python
auto_round/compressors/orchestrator.py
def quantize_block(self, block, inputs, q_input=None, device="cpu",
                   auto_offload=True, reference_output=None, ...):
    ...
    new_q_input, reference_output = self.alg_composer.compress_block(
        ..., reference_output=reference_output,
    )

auto_round/algorithms/composer.py
def compress_block(self, block, fp_inputs, input_others, block_ctx,
                   q_inputs=None, input_ids=None, reference_output=None, **kwargs):
    ...
    quant_hooks = self._get_fp_act_hooks(block)
    if reference_output is None:          # ← skip forward pass when pre-seeded
        reference_output = block_forward_fn(block, fp_inputs, input_others)
    ...
```

### 2. llm-compressor (this PR) — preseed `reference_output` from `IntermediatesCache`

1. **`pipelines/sequential/pipeline.py`**: Before `sequential_epoch_end` fires, extract
   the next subgraph's hidden states from `IntermediatesCache` and deliver them via
   `modifier.set_fp_ref_outputs(hidden_states)` (duck-typed, no circular import).

2. **`modifiers/autoround/base.py`**: Add `_fp_ref_outputs: list | None = PrivateAttr(default=None)`
   and a `set_fp_ref_outputs()` method as the public intake. In `apply_autoround`, consume
   `_fp_ref_outputs` and pass it to `quantize_block(reference_output=...)`, setting the
   attribute to `None` immediately after to release the reference.

When `_fp_ref_outputs` is present, `quantize_block` uses it directly and skips
the collect_reference forward pass. When absent (e.g. `propagate_error=True`, or last subgraph),
behaviour is unchanged — `reference_output=None` falls back to the existing path.

**Prerequisite**: PR #3055 (merged) — offloads calibration inputs to CPU, which is a
necessary condition for the memory budget to hold here.

**Requires**: companion PR to intel/auto-round adding `reference_output` parameter to
`CompressionOrchestrator.quantize_block` and `AlgorithmComposer.compress_block`.
(Parameter does not exist in any released version; must land concurrently.)

---

### Completion runs

| Run | Model | N | elapsed | Result |
|-----|-------|---|---------|--------|
| Before fix | Qwen3.5-MoE 122B (`Qwen3_5MoeForConditionalGeneration`) | 8192 | — | CPU RAM OOM, rank SIGKILL'd, NCCL watchdog timeout |
| After fix | Qwen3.5-MoE 122B (`Qwen3_5MoeForConditionalGeneration`) | 8192 | 35834s (~10h) | ✅ SUCCESS, 49/49 blocks |

Hardware: 8 × NVIDIA H20 (140 GB VRAM, ~1.8 TiB RAM), 8-GPU DDP, seqlen=6144, iters=200, `enable_torch_compile=False`.

Memory profile post-fix (N=8192, 8-GPU DDP):
```
peak system RAM  ≈ 310 GB total  (~39 GB/rank × 8, IntermediatesCache only, no duplicate)
VRAM allocated   ≈ 9.5 GB/rank  (one block on GPU during quantize_block, weights offloaded to CPU)
```

### Quantization quality: N scaling (Qwen3.5-MoE 122B, CTK-OFF, seq=6144)

BF16 baseline: GPQA=0.4545, IFEval=0.7449, GSM8K=0.8362, MATH500=0.5260

| Config | GPQA | IFEval | GSM8K | MATH500 | GPQA recovery |
|--------|-----:|-------:|------:|--------:|--------------:|
| N=4096 (before fix) | 0.4444 | 0.7301 | 0.8393 | 0.5300 | 97.8% |
| N=6144 (before fix) | 0.4444 | 0.7024 ⚠️ | 0.8324 | 0.5200 | 97.8% |
| **N=8192 (this fix)** | **0.5101** | **0.7338** | **0.8347** | **0.5420** | **112.2%** ✅ |

N=4096 and N=6144 were both achievable before the fix; GPQA did not improve with
larger N until the fix enabled N=8192 (+14.6 pp absolute, surpasses BF16 baseline).
N=10240 excluded: uses seq_len=5120 (≠ 6144), not calibration-aligned with the above.

---

## Additional Finding: `enable_torch_compile` and DDP stability

During validation, a separate failure was observed unrelated to the CPU RAM fix:
`AutoRoundModifier` defaults to `enable_torch_compile=True`, which calls `torch.compile()`
on the block before the optimization loop. In an 8-GPU DDP run with a 122B model,
compilation time varies across ranks (can exceed 10 minutes); ranks that finish earlier
block on an NCCL ALLGATHER waiting for the slow rank, triggering the 600 s watchdog.

**Workaround**: set `enable_torch_compile=False` when running multi-GPU DDP with large
models. This is unrelated to the `collect_reference` fix but is recommended practice:

```python
recipe = AutoRoundModifier(
    targets="Linear",
    scheme="W4A16",
    iters=200,
    enable_torch_compile=False,  # required for DDP stability on large models
)
```

This will be noted in the `AutoRoundModifier` docstring in the accompanying PR.

---

## PR Ready

xesdiny:
## Summary

AutoRound's \`collect_reference\` runs a full FP16 forward pass over all
calibration inputs at the start of every \`quantize_block\` call to obtain
reference outputs for SignSGD optimization. With \`propagate_error=False\`
(the default), \`SequentialPipeline\` already holds the identical tensors
in \`IntermediatesCache[k+1]\` — \`collect_reference\` is a duplicate that
exhausts CPU RAM for large N on big models (~39 GB/rank at N=8192,
seq=6144, hidden=3072, 8-GPU DDP).

Fix: before \`sequential_epoch_end\` fires, extract the next subgraph's
hidden states from \`IntermediatesCache\` and deliver them to the modifier
via \`set_fp_ref_outputs()\`. \`apply_autoround\` passes them to
\`quantize_block(reference_output=...)\`, skipping \`collect_reference\`
entirely. Falls back to the existing path when absent (last subgraph, or
\`propagate_error=True\`).

Two files changed in llm-compressor, ~45 lines.

> **Companion PR to intel/auto-round merged** — intel/auto-round#2289 (merged 2026-09-04)
> added \`reference_output\` to \`CompressionOrchestrator.quantize_block\`.
> The parameter is guarded with \`inspect.signature\` so existing installs
> (auto-round <=0.15.0, the current PyPI max) silently fall back to \`collect_reference\`
> as before. The optimized path activates automatically once auto-round 0.16.0 is
> released on PyPI; a follow-up PR will bump the \`setup.py\` pin to \`>=0.16.0\` at
> that point.

Also adds a docstring note to \`AutoRoundModifier.enable_torch_compile\` recommending
\`False\` for multi-GPU DDP with large models (per-rank compilation time can exceed the
NCCL watchdog timeout).

Closes #3129
Depends on: #3055 (merged)

## Test Plan

- Ran N=8192, seq=6144, iters=200, \`enable_torch_compile=False\` on 122B MoE
  (8 × H20 140 GB, 8-GPU DDP):
  - v1-2 model: before fix → CPU RAM OOM + NCCL watchdog timeout at layer 44;
    after fix → ✅ 49/49 blocks completed, ~10 h, peak system RAM ~310 GB
  - v1-3 model: after fix → ✅ all 8 ranks SUCCESS, 31291 s (~8h42m)
- \`make quality\` passes
- See linked Issue for quantization quality evaluation results


xesdiny:
@gemini-code-assist review

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -157,6 +157,15 @@ def __call__(
...
+                    # Expose activations and subgraph context so modifiers can
+                    # read cached tensors during sequential_epoch_end callbacks.
+                    session.state.activations = activations
+                    session.state.next_subgraph_input_names = (
+                        subgraphs[subgraph_index + 1].input_names
+                        if subgraph_index < num_subgraphs - 1
+                        else None
+                    )
+
                     LifecycleCallbacks.sequential_epoch_end(subgraph.submodules(model))
```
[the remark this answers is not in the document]
xesdiny: Good catch. Applied in 3ae51ac0.

For context on the impact: we ran the full 122B-model run (49 MoE layers, 8×H20) and the per-layer transition spike stabilized at exactly 410.42 GB from layer 15 onwards with zero growth, so the existing GC was reclaiming the cache in practice. That said, leaving a strong reference on `session.state` after callers have finished is a correctness hazard if anything ever extends the state's lifetime, so the explicit nullification is the right call.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -380,12 +381,38 @@ def apply_autoround(self, state, modules):
...
+                    for _name in _sorted_names:
+                        if _name in _batch:
+                            _v = _batch[_name].value
+                            if isinstance(_v, torch.Tensor) and (
+                                _name in _HIDDEN_STATE_KEYS or _v.ndim == 3
+                            ):
+                                _refs.append(_v)
+                                break
+                if len(_refs) == len(cur_inputs):
+                    _fp_ref = _refs
```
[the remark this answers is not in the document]
xesdiny: Applied the `min(len(cur_inputs), len(batch_intermediates))` bound in 3ae51ac0 as a defensive guard.

Skipped the `hasattr(_activations, "batch_intermediates")` check: `IntermediatesCache` is a typed object that always exposes `batch_intermediates`, so the guard would mislead readers into thinking the attribute might be absent. The existing `if len(_refs) == len(cur_inputs)` check at the bottom already provides the fallback to `collect_reference` for any mismatch.

xesdiny:
@yiliu30 The PR is ready for review. All Sept 4 feedback (r3930367414, r3930367419) has been addressed and squashed into a single clean commit (33d9a7c4). Would appreciate your review when you get a chance!

yiliu30:
@codex review

yiliu30 (commented):
Thanks @xesdiny! This requires the change in https://github.com/intel/auto-round/pull/2289 as well, right?

We’re preparing a patch release that will include your changes. I’ll ping you once it’s released.

xesdiny:
@yiliu30 Yeah. This PR depends on [intel/auto-round#2289](https://github.com/intel/auto-round/pull/2289) .
I verified and executed it  double-check locally again this morning.
The latest commit (`4c4a961`) also adds a backward-compat guard (`inspect.signature`) so installs without `reference_output` support (< 0.16.0) fall back to `collect_reference` transparently. No action needed from your side on that.

xesdiny:
@yiliu30 To clarify my last message: this PR is ready to merge now — no need to wait for the auto-round patch [release](https://github.com/intel/auto-round/releases).

The `inspect.signature` makes it fully backward-compatible: with the current auto-round ≤0.15.0 the `reference_output` path is silently skipped and `collect_reference` runs as before. The optimized path activates automatically once the patch ships on PyPI. I'll send a follow-up PR to bump the `setup.py` pin to `>=0.16.0` at that point.

yiliu30:
> @yiliu30 To clarify my last message: this PR is ready to merge now — no need to wait for the auto-round patch [release](https://github.com/intel/auto-round/releases).
> 
> The `inspect.signature` makes it fully backward-compatible: with the current auto-round ≤0.15.0 the `reference_output` path is silently skipped and `collect_reference` runs as before. The optimized path activates automatically once the patch ships on PyPI. I'll send a follow-up PR to bump the `setup.py` pin to `>=0.16.0` at that point.

Hi @xesdiny, Thanks for the clarification, and the auto-round 0.15.1 was shipped today; I will prepare a PR soon.

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -380,12 +382,46 @@ def apply_autoround(self, state, modules):
             # per-batch CPU->GPU transfer automatically.
             ar_inputs = [((args, kwargs),) for args, kwargs in cur_inputs]
 
+            # Read FP16 reference outputs from IntermediatesCache if pipeline
+            # exposed them in state. Skips collect_reference (~39 GB CPU RAM/rank).
+            _state = active_session().state
+            _activations = getattr(_state, "activations", None)
+            _next_names = getattr(_state, "next_subgraph_input_names", None)
+            _fp_ref = None
+            if _activations is not None and _next_names is not None:
```
yiliu30, line 390: The `_fp_ref` extraction is a bit of a footgun. How about following a best-effort strategy: *try* to capture the next subgraph's hidden states from the cache (name match first, shape match for models whose inputs don't use the standard names), and if capture doesn't fully succeed, pass `reference_output=None` and let auto-round run its own `collect_reference`.

yiliu30 (approved):
Thanks @xesdiny , LGTM!

yiliu30:
Hi @kylesayrs @dsikka  could you take a look as well, thanks!

yiliu30:
Hi @xesdiny could you resolve the conflicts, thanks!

yiliu30 (approved):

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -157,8 +157,20 @@ def __call__(
...
                                 activations.delete(batch_idx, subgraph.consumed_names)
 
+                    # Expose activations and subgraph context so modifiers can
+                    # read cached tensors during sequential_epoch_end callbacks.
+                    session.state.activations = activations
+                    session.state.next_subgraph_input_names = (
+                        subgraphs[subgraph_index + 1].input_names
+                        if subgraph_index < num_subgraphs - 1
+                        else None
+                    )
```
kylesayrs: Please move this computation to within the AutoRoundModifier, rather than add clutter to the `sequential/pipline.py` file.

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -157,8 +157,20 @@ def __call__(
                                 activations.update(batch_idx, outputs)
                                 activations.delete(batch_idx, subgraph.consumed_names)
 
+                    # Expose activations and subgraph context so modifiers can
+                    # read cached tensors during sequential_epoch_end callbacks.
+                    session.state.activations = activations
```
kylesayrs: Let's expose sequential activations as `session.state.sequential_activations` to make it clear that this only applies to the sequential pipeline

Please add a field to https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/core/state.py#L96

src/llmcompressor/pipelines/sequential/pipeline.py
kylesayrs:
```suggestion
            session.state.sequential_activations = IntermediatesCache.from_dataloader(
```

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -157,8 +157,20 @@ def __call__(
...
+                    session.state.next_subgraph_input_names = (
+                        subgraphs[subgraph_index + 1].input_names
+                        if subgraph_index < num_subgraphs - 1
+                        else None
+                    )
+
                     LifecycleCallbacks.sequential_epoch_end(subgraph.submodules(model))
 
+                    session.state.activations = None
+                    session.state.next_subgraph_input_names = None
```
kylesayrs: If you agree to move to `session.state.sequential_activations`, then let's move this to the end of the function call, rather than per-layer

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -383,12 +384,55 @@ def apply_autoround(self, state, modules):
             # per-batch CPU->GPU transfer automatically.
             ar_inputs = [((args, kwargs),) for args, kwargs in cur_inputs]
 
+            # Read FP16 reference outputs from IntermediatesCache if pipeline
+            # exposed them in state. Skips collect_reference (~39 GB CPU RAM/rank).
+            _state = active_session().state
+            _activations = getattr(_state, "activations", None)
```
kylesayrs: You'll want to guard against accidentally grabbing activations when propgate_error=True. Just to be safe, consider something like this:

```
if _state.sequential_activations is not None and not data_args.propgate_error:
    _activations = _state.sequential_activations
```

src/llmcompressor/modifiers/autoround/base.py
```diff
@@ -383,12 +384,55 @@ def apply_autoround(self, state, modules):
             # per-batch CPU->GPU transfer automatically.
             ar_inputs = [((args, kwargs),) for args, kwargs in cur_inputs]
```
kylesayrs, line 392: The LLM Compressor default is `propgate_error=True`. Consider writing an info/warning log in `initialize()` which recommends to the user that they use `propgate_error=False` for better runtime and memory

kylesayrs (changes requested):

kylesayrs:
This feature is a good idea. I have some suggestions on how to better integrate without polluting sequential pipeline too much

xesdiny:
@kylesayrs  Thx.I am eager to hear your insights and learn from your experience.

src/llmcompressor/pipelines/sequential/pipeline.py
xesdiny: Done. Removed `next_subgraph_input_names` from `pipeline.py` entirely. AutoRound now searches the cache directly with `for n in _batch`, filtered by `_HIDDEN_STATE_KEYS = {"hidden_states", "inputs_embeds"}` plus a 3-D shape fallback. Since `activations.update()` overwrites same-named keys, the cache holds at most one `hidden_states` at any time, so iteration is equivalent and correct.

src/llmcompressor/modifiers/autoround/base.py
xesdiny: Good catch. When `propagate_error=True`, the cache holds the current subgraph's inputs (not outputs) at `epoch_end` time, so reading it as fp_ref would be wrong. Fixed with a guard: AutoRound only reads the cache when `sequential_propagate_error=False`. In practice the pipeline already force-sets `propagate_error=False` for AutoRound (pipeline.py L89–94), so this guard is defensive but correct.

src/llmcompressor/pipelines/sequential/pipeline.py
xesdiny: Done. Renamed to `state.sequential_activations` (and `added state.sequential_propagate_error`), following the `sequential_prefetch` precedent. Lifecycle is now: set once at cache creation (line 119), cleared once at function end — no per-subgraph set/clear.

src/llmcompressor/modifiers/autoround/base.py
xesdiny: Done. Added an `INFO` log in `on_initialize()` explaining that setting `propagate_error=False` in `dataset_args` activates the optimisation and skips `collect_reference`.

xesdiny:
@mergify update

PR #3140, opened 2026-09-04
fix(args): escape '%' in DatasetArguments.splits help so argparse builds on Python 3.14

### LINKED ISSUE #3139: [Bug]: Python 3.14 compatibility: unescaped '%' in DatasetArguments.splits help text raises ValueError

<details>
### Environment

- Python: 3.14.6
- OS: Ubuntu 26.04
- `llmcompressor`: `main`, commit `b0c8df1a110afe4f1ec493a3c29230d3ad894a3e`
- `transformers`: 5.16.1
- `compressed-tensors`: 0.18.1a20260903

The issue is reproducible independently of Ubuntu-specific behavior. Ubuntu 26.04 is included here because it was the downstream CI environment where the failure was observed.

</details>

### Summary

On Python 3.14, calling LLMC's public `oneshot()` entry point fails while LLMC is constructing its argument parser. The failure happens before model loading, calibration, or quantization.

The affected help text is the `DatasetArguments.splits` metadata, which contains examples such as:

```text
train[:50%]
```

`argparse` treats help strings as percent-format strings. Because the percent sign is followed by `]`, the sequence `%]` is interpreted as an invalid format directive.

### Minimal reproducer

Save the following as `reproduce.py`:

```python

# The placeholder model name is never resolved because parser construction fails first.
oneshot(model="unused")
```

Run it with Python 3.14:

```bash
python reproduce.py
```

The call raises an exception similar to:

```text
ValueError: unsupported format character ']' (0x5d) at index 77
...
ValueError: badly formed help string
```

Anai-Guo:
## Summary

On Python 3.14, calling the public `oneshot()` entry point fails while `HfArgumentParser` builds its argument parser — before any model loading or calibration:

```
ValueError: unsupported format character ']' (0x5d) at index ...
ValueError: badly formed help string
```

`argparse` treats every `help` string as a percent-format template and expands it with `help % params`. The `splits` help text in `DatasetArguments` contains `train[:50%]`, and the sequence `%]` is an invalid format directive. Python 3.14 validates help strings eagerly during parser construction, so this now raises up front (earlier versions only tripped when help was actually rendered).

Fixes #3139.

## Fix

Escape the literal percent signs as `%%` in the two `train[:50%]` examples in the `splits` help metadata. `argparse` un-escapes `%%` back to `%` when rendering, so the displayed help is unchanged (`train[:50%]`), while parser construction no longer raises.

## Verification

```python
import argparse
p = argparse.ArgumentParser()
p.add_argument("--x", help="'train[:50%]'")    # ValueError: unsupported format character ']'
p.add_argument("--y", help="'train[:50%%]'")   # OK; format_help() renders 'train[:50%]'
```

Reproduced the exact `ValueError` on the unescaped string and confirmed the escaped form both builds the parser and renders `train[:50%]` in `--help`.

### DIFF
--- a/src/llmcompressor/args/dataset_arguments.py
+++ b/src/llmcompressor/args/dataset_arguments.py
@@ -143,9 +143,9 @@ class DatasetArguments(CustomDatasetArguments):
         metadata={
             "help": (
                 "Optional dataset split selector. Passing a string like 'train' or "
-                "'train[:50%]' is strongly recommended. Legacy dict input is "
+                "'train[:50%%]' is strongly recommended. Legacy dict input is "
                 "deprecated and only supported for calibration compatibility "
-                "(for example: {'calibration': 'train[:50%]'})."
+                "(for example: {'calibration': 'train[:50%%]'})."
             )
         },
     )


dsikka (commented):
Thanks! How does this render for 3.12/3.13?

kylesayrs (approved):
Thanks! @dsikka I checked that the escaping works properly with older versions of `argparse.HelpFormatter._expand_help`

PR #3141, opened 2026-09-04
[Examples] [Tests] Add download limit for perfect blend

kylesayrs:
## Background ##
https://github.com/vllm-project/llm-compressor/pull/3062 and https://github.com/vllm-project/llm-compressor/pull/3063 both moved examples and tests to using the "perfectblend" dataset. However, they also inadvertently removed the limit on downloading the dataset split. This causes the entire train split to be downloaded and tokenized, dramatically increasing runtime.

Some of the tests never had a download limit to begin with, but this was not an issue for "small" datasets like ultrachat

## Changes ##
* Add split limit for perfectblend dataset

## Testing ##
* Tested examples and tests locally

### DIFF
--- a/examples/.claude/skills/shared_quantization.md
+++ b/examples/.claude/skills/shared_quantization.md
@@ -289,6 +289,7 @@ Use the shared template at `.claude/skills/templates/oneshot_with_data.py` as th
   oneshot(
       model=model,
       dataset="perfectblend",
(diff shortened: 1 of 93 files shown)


brian-dellabetta (approved):
nice. i hit this too running locally

HDCharles (approved):

PR #3144, opened 2026-09-05
Fix activation caching for dataclasses that validate their fields

avifenesh:
Caching activations containing a strict dataclass currently reconstructs the object with `type(value)(**wrapped_fields)`. Its constructor validates each `IntermediateValue` wrapper against the original field type and rejects it. The concrete Mistral object is `transformers.models.mistral3.configuration_mistral3.Mistral3Config` used by Mistral-Small-3.2; its nested text config is another configuration object.

The fix shallow-copies the dataclass and writes the transient wrapped fields with `object.__setattr__`, then uses the same mechanism to restore fields on load. This bypasses constructor and assignment validation without mutating the original object.

Reproduced on 2026-09-14 with Transformers 5.14.1 and huggingface_hub 1.31.0, without model weights or PyTorch:

```python
from dataclasses import dataclass, fields
from transformers import Mistral3Config

@dataclass
class IntermediateValue:
    value: object
    device: object = None

config = Mistral3Config()
wrapped = {f.name: IntermediateValue(getattr(config, f.name))
           for f in fields(config)}
type(config)(**wrapped)
```

This raises `StrictDataclassFieldValidationError` for `transformers_version`: `IntermediateValue(value=None, device=None)` is neither `str` nor `None`. The exact `_replace_fields` helper from this PR successfully wrapped and restored that same real config, with `restored.to_dict() == config.to_dict()` and the original left unchanged.

The existing regression test exercises a validating dataclass through the actual cache. Its original PR receipt reports the full cache file at 17 passed. The 2026-09-14 revision adds the missing DCO sign-off only; its code tree is byte-identical to cf2ead8127ddbc7ef8855f51e30d156e86c2b769. Fresh hosted checks and maintainer approval remain required.


kylesayrs (commented):
@avifenesh Can you point to exactly which mistral object is problematic? These changes look fine but I want to get a better sense of when this is useful

avifenesh:
The missing DCO sign-off is fixed and DCO is green. I added a concrete `Mistral3Config` reproducer to the description: Transformers 5.14.1 plus huggingface_hub 1.31.0 rejects the wrapped transformers_version field, while the exact PR helper wraps/restores the real config without mutating it. Could a maintainer re-review?

PR #3147, opened 2026-09-08
Add resave_config to preserve original model config on save

kylesayrs:
## Purpose

There are often differences between the original model config and the config regenerated by transformers on `save_pretrained`. vLLM always supports loading the original model config, but only *sometimes* supports loading the transformers-serialized one. This PR adds a helper to write out the original config (patched with the fields llmcompressor actually modifies) so saved models stay loadable in vLLM.

## Changes

Adds `resave_config(config: PretrainedConfig, save_dir: str)` to `src/llmcompressor/transformers/compression/compressed_tensors_utils.py`. It:

1. Uses `config._name_or_path` to locate the original config file (local dir/file, or downloaded from the hub).
2. Loads the original `config.json`.
3. Overwrites only the fields llmcompressor has modified — currently the expert-sparsity fields from `NUM_EXPERTS_CONFIG_KEYS` (`src/llmcompressor/modifiers/pruning/reap/utils.py`), which change under REAP expert pruning. Fields are patched whether they live at the top level or nested under `text_config`.
4. Writes the result to `save_dir/config.json`.

The set of patched fields is intentionally scoped to `NUM_EXPERTS_CONFIG_KEYS` for now and can be extended as other modifiers need it.

## Testing

* Verified locally that `resave_config` restores fields dropped by the transformers serialization while patching the expert count to the pruned value.
* Added unit tests

### DIFF
--- a/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
+++ b/src/llmcompressor/transformers/compression/compressed_tensors_utils.py
@@ -20,6 +20,7 @@ from transformers.utils import http_user_agent
 from llmcompressor.core import active_session
 from llmcompressor.modifiers.pruning.reap.utils import NUM_EXPERTS_CONFIG_KEYS
 from llmcompressor.pytorch.model_load.helpers import copy_python_files_from_model_cache
+from llmcompressor.sentinel import Sentinel
 from llmcompressor.transformers.utils import RECIPE_FILE_NAME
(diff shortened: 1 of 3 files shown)


HDCharles (approved):
seems like a good idea

dsikka (commented):
somewhat related, we see this issue with tokenizers as well (load with transformers + resave has variations from the original model provided tokenizer)

dsikka:
failing tests

PR #3151, opened 2026-09-09
[Bugfix] Restore model offload after save failures

### LINKED ISSUE #3149: [Bug]: Failed save leaves model in temporary Accelerate offload state
### Environment

CPU-only WSL Linux 6.6.87.2, Python 3.12.3, torch 2.11.0+cpu, Transformers 5.15.0.

Source-built llm-compressor `8f96fe61feb501b98c2f1c61de2053b50f8d8534` and compressed-tensors `099fa98fea7f3533a8e304a081795a1136ef67c0`. Shallow local builds report `0.1.dev1+g8f96fe6.d20260908` and `0.1.dev1+g099fa98.d20260908`; these are not published release versions. Packages were installed in isolated targets. The first diagnostic used Transformers 5.14.1; I subsequently repeated it with the declared minimum 5.15.0 and observed the same result.

### Bug

`modify_save_pretrained` converts CT offloading to Accelerate for saving, but calls `from_accelerate(model)` only on the success path. A save-side exception after conversion leaves the live model in that temporary representation.

Below, an invalid output directory triggers a real `NotADirectoryError` while writing `recipe.yaml`. Transformers logs that the path is a file and returns before that wrapper metadata operation. The bug reported here is missing offload restoration, not the expected rejection of an invalid output path.

Observed:

| State | CT parameter caches | Accelerate hooks |
| --- | ---: | ---: |
| Before save | 20 | 0 |
| After failed save | 0 | 20 |
| After explicit `from_accelerate` | 13 | 0 |

The initial fixture also offloads empty modules, so the restored cache count need not be exactly 20. After explicit recovery, forward logits matched the original exactly. I am not claiming weight corruption or that forward inference necessarily fails before recovery.

### Reproduction (no model download or GPU)

```python
import tempfile
from pathlib import Path
import torch
from transformers import LlamaConfig, LlamaForCausalLM
from compressed_tensors.offload import OffloadCache, offload_module, from_accelerate
from llmcompressor.transformers.compression.compressed_tensors_utils import modify_save_pretrained

def state(model):
    return (
        sum(isinstance(m._parameters, OffloadCache) for m in model.modules()),
        sum(hasattr(m, '_hf_hook') for m in model.modules()),
    )

torch.manual_seed(42)
model = LlamaForCausalLM(LlamaConfig(
    vocab_size=32, hidden_size=16, intermediate_size=32,
    num_hidden_layers=1, num_attention_heads=2, num_key_value_heads=2,
    max_position_embeddings=32, tie_word_embeddings=False,
)).eval()
    offload_module(module, 'cpu', 'cpu')
modify_save_pretrained(model)
print('before:', state(model))
with tempfile.TemporaryDirectory() as directory:
    target = Path(directory) / 'not-a-directory'
    target.write_text('sentinel')
    try:
        model.save_pretrained(target, save_compressed=False)
    except OSError as error:
        print(type(error).__name__, error)
    print('after failure:', state(model))
    from_accelerate(model)
    print('after explicit recovery:', state(model))
```

### Scope / proposed follow-up

The successful-save restoration was introduced in #2403. #2788 concerns a different failure inside restoration itself (multiple offload directories), rather than the restoration call being skipped. I searched related save/restoration issues and PRs but may have missed ongoing work.

I'd be interested in adding exception-safe restoration and focused regressions if this is available. I have not started an implementation. Please confirm the expected contract and assignment first: after successful conversion, should all save/metadata failures attempt restoration while preserving the original error?

Distributed handling needs explicit care: `from_accelerate` broadcasts the device map, and save uses source-rank-only operations plus the timeout context's barriers. CPU single-process evidence does not establish a distributed fix; all-rank collective ordering and recovery failure semantics must be tested before claiming support.

Investigation and reproduction were AI-assisted. No full suite, quantized GPU, real disk-full, or multi-rank failure test is claimed. This is a focused existing-save-path issue, not an implementation of #3131.

### Update: real two-rank CPU/Gloo reproduction

Subsequently ran the same tiny-model setup under `torch.distributed.run --standalone --nproc-per-node=2`, initializing Gloo with a 5-second default-group timeout after constructing/offloading each model. An independent 20-second Gloo control group gathered results after the save exception, keeping rank 0 alive while rank 1 completed its failing collective. No production functions were patched. An outer 50-second timeout was not reached; the diagnostic completed normally after capturing both expected exceptions.

- Rank 0: `NotADirectoryError` writing `recipe.yaml`, leaving 0 CT caches / 20 Accelerate hooks.
- Rank 1: `save_pretrained_wrapper -> from_accelerate -> dist.broadcast_object_list -> broadcast -> work.wait` raises `RuntimeError: Timed out waiting 5000ms for recv operation to complete`. It has already removed its Accelerate hooks but not redispatched CT caches (0 / 0).

This supersedes only the earlier lack of multi-rank evidence: a CPU/Gloo failure is now reproduced. It does not establish NCCL/GPU behavior or prove a fix. Both ranks were kept alive through the control-group result gathering, so this was not a connection-closed error caused by early source-process termination.

LOGO127:
## SUMMARY

Fixes #3149.

Following the small `try/finally` approach suggested by @Roderick-Wu in the issue, restore the live model's compressed-tensors offload representation even if saving raises. Previously, an exception after `to_accelerate(model)` skipped `from_accelerate(model)`, leaving Accelerate hooks installed when the caller caught the error and continued using the model.

The `finally` covers the save and metadata operations after successful conversion, on every rank. It preserves the recently added `resave_config` operation and its ordering. This does not change the successful-save flow or claim weight corruption or broad training impact.

Adds six CPU regression cases using a locally constructed tiny Llama and real offload conversion: failures in weight saving, config resaving, recipe saving and Python-file copying; an invalid destination followed by a retry through the same model; and a successful-save/reload control. Injected failures check the original exception object, hook cleanup and unchanged logits.

## TEST PLAN

From a development installation with the repository's compressed-tensors dependency available:

```bash
OMP_NUM_THREADS=1 HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 \
  python -m pytest tests/llmcompressor/transformers/compression/test_save_recovery.py -q
```

- Final integrated tree: **6 passed**. No downloaded model, tokenizer or dataset is needed.
- Before the upstream config-resave integration, the original five-case test file produced **4 failed / 1 passed** on the unfixed base and **5 passed** with the fix. The added config case was tested on the integrated candidate; no six-case baseline result is claimed.
- Full-repository Ruff 0.4.10 lint and format checks passed (487 files already formatted); CUDA static lint and `git diff --check` passed.
- Additional local probes passed: invalid destination, a POSIX file-size-limit serialization failure, two-rank CPU/Gloo failure restoration, and two-rank CPU/Gloo successful save/reload. These supplementary probes are not part of the six repository tests. The file-size-limit failure is EFBIG, not a disk-full/ENOSPC simulation.

Local environment: WSL Linux, Python 3.12, torch 2.11.0+cpu, Transformers 5.15.0, compressed-tensors source `099fa98fea7f3533a8e304a081795a1136ef67c0`. Validation reused local dependencies with this checkout's `src` first on PYTHONPATH and locally generated package metadata; it was not a clean-install validation. No GPU/NCCL, quantized-model or full-suite validation is claimed.

Scope limits: this does not recover partial `to_accelerate` conversion, roll back partially written files, or introduce distributed error broadcasting. If restoration itself raises, the original save error remains exception context rather than necessarily the top-level exception.

AI assistance: Codex assisted with implementation and validation. LOGO127 reviewed the changes and personally added the DCO sign-offs, including the upstream integration update.


LOGO127:
The signed head `f80a16a` is code-complete for the scoped save-failure recovery fix. The six local CPU regression cases, supplementary CPU/Gloo probes, and repository quality checks described in the PR body have passed; the current DCO and quality checks are also green. The environment and untested GPU/quantized-model boundaries remain as documented.

The reminder asks for the `ready` label to complete the testing workflow. I do not have triage/write permission to add repository labels. Could a maintainer please add `ready` when appropriate, or let me know if there is another contributor-accessible step I should take? This is a request to enter the normal review/testing workflow, not to bypass checks or approval. Thank you! This follow-up was prepared with AI assistance.

LOGO127:
Compatibility note for #3152: I tested a temporary combined tree of this PR's `f80a16a` and #3152's `6291d3d`. They merge without a textual conflict, and the following two modules passed together (**20 passed**):

```bash
python -m pytest \
  tests/llmcompressor/transformers/compression/test_save_recovery.py \
  tests/llmcompressor/transformers/compression/test_resave_config.py -q
```

This covers the save-recovery controls, including injected `resave_config` failure, alongside the updated config/dtype tests. The combined Git tree was `1f39774168741ebd0ec9bd6ac18b46366cb9fdf5`; neither PR branch was modified. Environment: WSL Linux, Python 3.12, CPU Torch 2.11.0, Transformers 5.14.1 and compressed-tensors source `099fa98`. Fourteen warnings concerned deprecated `torch.jit.script_method`.

This is scoped CPU evidence, not a fresh-install, GPU/NCCL or full-model quantization validation. An initial attempt failed during collection because old temporary dependency directories had disappeared; the successful run used the pinned compressed-tensors source explicitly. Thanks for the dtype fix! This compatibility check was performed with Codex assistance.

kylesayrs (changes requested):
See comment: https://github.com/vllm-project/llm-compressor/issues/3149#issuecomment-5602987992

LOGO127:
Thanks for taking a look, @kylesayrs. That's a fair question. I agree that restoring the live model has little value when the process exits after the save error, and sample generation alone is not a strong justification.

The narrower case I reproduced is catching a save error and retrying with the same model after correcting the destination. `test_invalid_save_directory_restores_offload` covers that with a locally constructed CPU model, including a successful retry/reload. This is a regression scenario, not evidence of an affected production workflow; I haven't identified an in-tree caller that catches the save failure and continues.

My reason for putting recovery here was that this wrapper introduces the temporary `to_accelerate` conversion and already reverses it on success. Pairing those operations also keeps the existing restoration on every rank when source-rank saving raises. This is only offload-representation cleanup, not rollback of compression or partially written files, and recovery itself can fail.

Your caller-side recovery suggestion is a reasonable alternative if that is the intended responsibility boundary. I'd be happy to drop the wrapper change if you prefer recovery to remain opt-in; I won't expand the API or add another recovery layer without agreement. Would you prefer that direction, or is the narrow retry case sufficient motivation for keeping the cleanup here?

This analysis and response were prepared with Codex assistance.

kylesayrs:
@LOGO127 Thanks for the response! Yes, I think this kind of recovery falls under the responsibility of the caller and should not be an upstream responsibility

PR #3158, opened 2026-09-10
[MFPTQ] Implement profiler-based memory estimates

kylesayrs:
## Background ##
Fixes `model_free_ptq/gemma4_fp8_block.py` example test failure. The test was failing because the current naive estimate (input_file_size * 3) was far too pessimistic for the Gemma shard size of [49.8 GB](https://huggingface.co/google/gemma-4-31B-it/resolve/main/model-00001-of-00002.safetensors).

This PR ports abbreviated profiler code from [PTTP](https://github.com/kylesayrs/PTTP) and uses it to profile the validate function, capturing the peak memory usage on the meta device.

## Prerequisites ##
* https://github.com/vllm-project/compressed-tensors/pull/870 for meta tensor calculations

## Changes ##
* Add `TensorProfiler` for profiling meta tensor movement
* Estimate memory usage during validation, profile meta device memory using `TensorProfiler`
  * Implementation has sufficient fallbacks if the tensor profiler fails for any reason
* For MFPTQ converter, do **not** move tensors to cpu first before saving. Keep on GPU. This seems to be a unnecessary device movement and breaks meta tensor processing, pending feedback from @soyr-redhat
* For MFPTQ converter, actually run `process` with meta tensors

## Testing ##
* Added unit tests for tensor profiler
* Ran Gemma4 example E2E. Previously it was estimated that ~`150` Gb of memory was needed for the `49.8` Gb shard. Now it only estimates `73.49` Gb, which is much more sane.
```
2026-09-10T05:29:29.8565 | model_free_ptq | INFO - Distributing 2 shard(s), estimated memory: 19.88-73.49 GB per shard, 93.37 GB total
```
* Ran Llama 8B NVFP4 E2E and observed sane estimations
```
2026-09-10T05:25:05.0658 | model_free_ptq | INFO - Distributing 4 shard(s), estimated memory: 2.48-10.42 GB per shard, 31.96 GB total
```

* Nightly: https://github.com/neuralmagic/llm-compressor-testing/actions/runs/34500092613


src/llmcompressor/entrypoints/model_free/memory.py
```diff
@@ -0,0 +1,157 @@
...
+class TensorProfiler(TorchDispatchMode):
+    _tracked: set[int]
+    _memory: MemoryProfile
+    _events: list[tuple[int, str]]
+    _exception: BaseException | None
+
+    def __init__(self, catch_exception: bool = True):
+        self._tracked = set()
+        self._memory = MemoryProfile()
+        self._events = list()
```
HDCharles: looks unused? are you just matching implementation here and leaving unused things that exist in PTTP?

src/llmcompressor/entrypoints/model_free/memory.py
```diff
@@ -0,0 +1,157 @@
...
+    def exception(self) -> BaseException | None:
+        return self._exception
+
+    # ::::::::::::::::::::::::::::::::::::::::::::::::
+    # ⚙️ Tracking - Dispatch overload and finalizers
+    # ::::::::::::::::::::::::::::::::::::::::::::::::
+
+    def __torch_dispatch__(self, func, types, args, kwargs=None):
+        ret = func(*args, **(kwargs or {}))
+        if isinstance(ret, torch.Tensor):
```
HDCharles: as mentioned in the gemini comment, this will only catch tensors, not nested structures with tensors, i think we need to decide if this is a specific validation helper where thats not needed, or if this is intended to be more general. If the former it should probably be named that way, if the latter then handling nested structures with pytree would probably make sense.

src/llmcompressor/entrypoints/model_free/converter.py
```diff
@@ -143,7 +133,7 @@ def _process_standard(
             del tensors[name]
             prefix = module_name + "."
             for key, value in module.state_dict(prefix=prefix).items():
-                tensors[key] = value.to("cpu")
+                tensors[key] = value
```
soyr-redhat, line 136: No objections here! with better scheduling we shouldnt need any offloading. Thanks for making that catch :D

soyr-redhat (commented):
Looks great, Addressed your mention in converter.py, all looks good to me, better scheduling should remove the need for offloading

soyr-redhat:
Addressed your mention in `converter.py` — all looks good to me. Better scheduling should remove the need for offloading to CPU.

src/llmcompressor/entrypoints/model_free/memory.py
kylesayrs: I'm keeping some things and minimizing others. This stuff I should remove

src/llmcompressor/entrypoints/model_free/memory.py
kylesayrs: Torch dispatch triggers whenever a tensor performs an operation (normally used to determine which kernel to dispatch).

From this perspective, I don't see  how nested structures are relevant.

src/llmcompressor/entrypoints/model_free/__init__.py
```diff
@@ -212,10 +208,27 @@ def _process_shard(
...
+            tensor.nbytes
+            for tensor in tensors.values()
+            if isinstance(tensor, torch.Tensor)
+        )
+        fallback_estimate = int(fallback_estimate * 2.5)
+        logger.warning(
+            f"Failed to estimate memory usage for {save_path}. Falling back to "
+            f"2.5x size of tensor inputs ({fallback_estimate / 1e9:.2f} GB)."
+        )
+        return fallback_estimate
```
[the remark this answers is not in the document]
kylesayrs: This is valid, but I also think that we can assume that `load_tensors_from_inverse_weight_map` will not fail (if it does fail for some reason, then the real job will fail and we should error early.

The error is not intuitive and this situation is not ideal, but I think this is the best tradeoff solution for a case which I don't expect to ever happen regularly.

src/llmcompressor/entrypoints/model_free/memory.py
```diff
@@ -0,0 +1,157 @@
...
+    def memory_timeline(self) -> dict[torch.device, list[int]]:
+        return self._memory.timeline
+
+    @property
+    def memory_timeline_mib(self) -> dict[torch.device, list[float]]:
+        return {
+            device: [value / (1024 * 1024) for value in timeline]
+            for device, timeline in self._memory.timeline
+        }
+
```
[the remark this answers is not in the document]
kylesayrs: I've removed this logic

src/llmcompressor/entrypoints/model_free/memory.py
kylesayrs: Yeah I see the issue now, resolving

kylesayrs: Added pytree flattening and tested with `test_structured_object_operations`

src/llmcompressor/entrypoints/model_free/memory.py
```diff
@@ -0,0 +1,157 @@
...
+    # ⚙️ Tracking - Dispatch overload and finalizers
+    # ::::::::::::::::::::::::::::::::::::::::::::::::
+
+    def __torch_dispatch__(self, func, types, args, kwargs=None):
+        ret = func(*args, **(kwargs or {}))
+        if isinstance(ret, torch.Tensor):
+            storage = ret.untyped_storage()
+            self._track(storage)
+
+        return ret
```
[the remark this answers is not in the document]
kylesayrs: Addressed

src/llmcompressor/entrypoints/model_free/memory.py
```diff
@@ -0,0 +1,157 @@
...
+        if hash in self._tracked:
+            return
+
+        # track
+        self._memory.add(device, size)
+        self._tracked.add(hash)
+
+        # register finalizer to subtract memory
+        finalizer = partial(self._untrack, hash, size, device)
+        weakref.finalize(storage, finalizer)  # triggers regardless of gc
```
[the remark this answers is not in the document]
kylesayrs: Good point, but this is not a big deal since
1. `TensorProfiler` memory usage is very small
2. We know that the tensors being tracked will die

HDCharles (approved):
Lookin slick

soyr-redhat (approved):
lgtm! awesome stuff :)

PR #3163, opened 2026-09-10
Add GLM 5.3 MXFp4 example

dsikka:
SUMMARY:
- The example first dequantizes from FP8 and then quantizes to MXFP4
 

Produces: https://huggingface.co/RedHatAI/GLM-5.3-MXFP4

### DIFF
--- a/examples/model_free_ptq/glm_5_3_mxfp4.py
+++ b/examples/model_free_ptq/glm_5_3_mxfp4.py
@@ -1,0 +1,28 @@
+from compressed_tensors.entrypoints.convert import FP8BlockDequantizer
+
+from llmcompressor import model_free_ptq
+
+
+MODEL_ID = "zai-org/GLM-5.3"
+SAVE_DIR = "/mnt/nvme_stripe/playground/dsikka/" + MODEL_ID.rstrip("/").split("/")[-1] + "-MXFP4"
+
+ignore = [
+    "re:.*mlp.gate$",
+    "re:.*lm_head",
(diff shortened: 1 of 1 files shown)


examples/model_free_ptq/glm_5_3_mxfp4.py
```diff
@@ -0,0 +1,28 @@
...
+
+model_free_ptq(
+    model_stub=MODEL_ID,
+    save_directory=SAVE_DIR,
+    scheme="MXFP4",
+    # wk IS fp8 in the source (dequantizer dequantizes it), but vLLM fuses
+    # wk + weights_proj into a dense (quant_config=None) layer whose fp8-only
+    # load path can't unpack mxfp4 weights — keep wk in bf16 instead
+    ignore=ignore + ["re:.*self_attn.indexer.wk$"],
+    converter=FP8BlockDequantizer(ignore=ignore),
```
kylesayrs, line 25: Don't you need to pass targets?

Also see https://github.com/vllm-project/compressed-tensors/issues/872

kylesayrs: I read a bit more, apparently empty targets has the behavior of targeting everything. I find this behavior very unintuitive, but I'm glad it worked out this time.

https://github.com/vllm-project/compressed-tensors/blob/a2a210e9001d6ac45b6f294f8a66a87384c7da55/src/compressed_tensors/utils/match.py#L484-L487

kylesayrs (approved):

Issue #3154, opened 2026-09-09
[Feature] Layerwise KL Divergence Reporting


kylesayrs:
## Background ##
When quantizing models, it can sometimes be difficult to measure the effect of quantization and to validate that a quantization was successful. While algorithms such as GPTQ will report the apparent local error from applying the algorithm, other algorithms do not, and local error reporting is not reflective of the effect on output activations.

Related to/ fixes https://github.com/vllm-project/llm-compressor/issues/2031

## Request ##
Add a `log_sequential_error` option to `oneshot`. This option, when enabled, reports the KL divergence between layer outputs after each sequential subgraph is compressed.

This will involve
* Capturing [unquantized outputs](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/pipeline.py#L153) when enabled
* Comparing with [quantized outputs](https://github.com/vllm-project/llm-compressor/blob/main/src/llmcompressor/pipelines/sequential/pipeline.py#L173) when enabled

Remaining questions
* How do we filter the outputs to only include the main activation (maybe check the largest tensor?)
* How do we handle when the sequential target is a Linear layer rather than a Decoder layer?

dichn:
@kylesayrs I'd like to work on this issue.

kylesayrs:
@dichn FYI I've updated this ticket to use SQNR because, as @HDCharles points out, KL-divergence is for comparing probability distributions, whereas here we're calculating differences in values

PR #3165, opened 2026-09-11
Add log_sequential_error option for layerwise SQNR-based error reporting

dichn:
## Purpose

Reports the Signal-to-Quantization-Noise Ratio (SQNR) between each sequential subgraph's pre- and post compression outputs during oneshot calibration, using the largest output tensor as a proxy for the subgraph's main activation.

SQNR (not KL divergence — the issue was updated after discussion, since KL divergence is for comparing probability distributions, whereas here we're comparing differences in raw activation values) is computed directly from accumulated signal/noise power sums:

    SQNR = 10 * log10(signal_power_sum / noise_power_sum)

Enabling this option forces `propagate_error=True` since both unquantized and quantized outputs are needed for comparison.

## Changes

* Add `log_sequential_error` option to `DatasetArguments` and the `oneshot()` entrypoint (default `False`)
* New module `pipelines/sequential/error_logging.py` holds all SQNR logic, keeping `pipeline.py`'s original loop structure intact:
  * `cache_pre_compression_output()` — selects a subgraph's main activation (largest tensor in its output dict) and stashes it in the shared `IntermediatesCache` under a reserved key, so it inherits the pipeline's existing CPU/disk offloading instead of accumulating in a raw Python list/dict (avoids OOM on large calibration datasets)
  * `accumulate_batch_power()` — compares the cached pre-compression tensor against the post-compression output for a batch, returning `(signal_power_sum, noise_power_sum)`, then frees the cached tensor
  * `compute_sqnr()` — computes one SQNR value from accumulated power sums across all batches in a subgraph, rather than averaging a per-batch SQNR (dB is a logarithmic ratio, so arithmetic-averaging it is not meaningful, and would let a single zero-noise batch skew the whole subgraph's result to `inf`); mirrors the accumulate-then-normalize pattern already used by `AWQModifier._compute_loss`
* In `SequentialPipeline`, force `propagate_error=True` when `log_sequential_error` is enabled, taking precedence over the existing `AutoRoundModifier` override, and log the per-subgraph SQNR via a `SequentialPipeline`-named logger patch (matching `IndependentPipeline`'s existing convention)
* Add unit tests (`test_log_sequential_error.py`) covering both the enabled and disabled paths using the `nm-testing/tinysmokellama-3.2` smoke model

## Test Plan

```python

oneshot(
    model="Qwen/Qwen3-0.6B",
    dataset="open_platypus",
    splits="train[:64]",
    max_seq_length=256,
    num_calibration_samples=64,
    pipeline="sequential",
    log_sequential_error=True,            # For #3154
    recipe=recipe,
)

### DIFF
--- a/src/llmcompressor/pipelines/sequential/error_logging.py
+++ b/src/llmcompressor/pipelines/sequential/error_logging.py
@@ -1,0 +1,101 @@
+import math
+from typing import Any
+
+import torch
+
+from llmcompressor.pipelines.cache import IntermediatesCache
+
+__all__ = ["compute_sqnr", "cache_pre_compression_output", "accumulate_batch_power"]
+
+_PRE_ERROR_KEY = "__pre_error__"
+
+
+def _get_largest_tensor(outputs: dict[str, Any]) -> torch.Tensor | None:
+    """
+    Select the largest tensor (by number of elements) from a subgraph output dict.
+    This is used as a heuristic to identify the "main activation" (e.g. hidden
+    states) among a subgraph's outputs, which may also include pass-through values
+    such as attention masks or position ids.
+
+    :param outputs: dictionary of subgraph output values
+    :return: the largest tensor value, or None if no tensor values are present
+    """
+    largest = None
+    largest_numel = 0
+    for value in outputs.values():
+        if isinstance(value, torch.Tensor) and value.numel() > largest_numel:
+            largest = value
+            largest_numel = value.numel()
+    return largest
+
+
+def compute_sqnr(signal_power_sum: float, noise_power_sum: float) -> float:
+    """
+    Compute the Signal-to-Quantization-Noise Ratio (SQNR), in decibels, from
+    accumulated signal and noise power sums; higher values indicate less
+    distortion. Callers should sum power across every batch in a subgraph
+    (see :func:`accumulate_batch_power`) and call this once on the totals,
+    since SQNR is a logarithmic ratio and cannot be meaningfully averaged
+    batch-by-batch.
+
+    :param signal_power_sum: sum of squared pre-compression activation values
+    :param noise_power_sum: sum of squared compression noise values
+    :return: SQNR in dB, or ``inf`` if there is no noise
+    """
+    if noise_power_sum == 0:
+        return float("inf")
+    return 10 * math.log10(signal_power_sum / noise_power_sum)
+
+
+def cache_pre_compression_output(
+    activations: IntermediatesCache,
+    batch_idx: int,
+    outputs: dict[str, Any],
+) -> None:
+    """
+    Save the main activation tensor from a subgraph's pre-compression output
+    into the intermediates cache for later error comparison.
+
+    :param activations: intermediates cache shared across the pipeline
+    :param batch_idx: index of the current calibration batch
+    :param outputs: subgraph output dict from the calibration (pre-compression) pass
+    """
+    main_tensor = _get_largest_tensor(outputs)
+    if main_tensor is not None:
+        activations.update(batch_idx, {_PRE_ERROR_KEY: main_tensor})
+
+
+def accumulate_batch_power(
+    activations: IntermediatesCache,
+    batch_idx: int,
+    output: dict[str, Any],
+) -> tuple[float, float] | None:
+    """
+    Compute the signal and noise power sums for a single batch by comparing
+    the saved pre-compression activation against the post-compression output,
+    then remove the cached pre-compression tensor to free memory. Power is
+    summed (not averaged) so callers can add it across batches and compute a
+    single SQNR from the totals via :func:`compute_sqnr`.
+
+    :param activations: intermediates cache holding the pre-compression tensor
+    :param batch_idx: index of the current propagation batch
+    :param output: subgraph output dict from the propagation (post-compression) pass
+    :return: (signal power sum, noise power sum) for this batch, or None if
+        either tensor is unavailable
+    """
+    pre_data = activations.fetch(batch_idx, [_PRE_ERROR_KEY])
+    pre_tensor = pre_data.get(_PRE_ERROR_KEY)
+    if pre_tensor is None:
+        return None
+    # clean up cached tensor; delete only after confirming the key exists,
+    # since IntermediatesCache.delete raises KeyError for missing keys
+    activations.delete(batch_idx, [_PRE_ERROR_KEY])
+
+    post_tensor = _get_largest_tensor(output)
+    if post_tensor is None:
+        return None
+
+    noise = post_tensor.float() - pre_tensor.float()
+    signal_power_sum = (pre_tensor.float() ** 2).sum().item()
+    noise_power_sum = (noise**2).sum().item()
+    return signal_power_sum, noise_power_sum
--- a/src/llmcompressor/pipelines/sequential/pipeline.py
+++ b/src/llmcompressor/pipelines/sequential/pipeline.py
@@ -119,6 +119,8 @@ class SequentialPipeline(CalibrationPipeline):
         :param dataloader: loads data for calibration
         :param dataset_args: dataset arguments relevant to pipelines
         """
+        _logger = logger.patch(lambda r: r.update(function="SequentialPipeline"))
+
         session = active_session()
 
         # prepare model for sequential onloading
@@ -260,7 +262,7 @@ class SequentialPipeline(CalibrationPipeline):
 
                         if dataset_args.log_sequential_error and batch_kls:
                             avg_kl = sum(batch_kls) / len(batch_kls)
-                            logger.log(
+                            _logger.log(
                                 "METRIC",
                                 f"subgraph {subgraph_index + 1}/{num_subgraphs} | "
                                 f"sequential error (KL): {avg_kl:.6f}",
--- a/src/llmcompressor/pipelines/sequential/pipeline.py
+++ b/src/llmcompressor/pipelines/sequential/pipeline.py
@@ -185,8 +185,11 @@ class SequentialPipeline(CalibrationPipeline):
                 prop_desc = f"({subgraph_index + 1}/{num_subgraphs}): Propagating"
 
                 # per-batch largest tensor captured before compression, used by
-                # log_sequential_error to compare against post-compression outputs
-                pre_compression_outputs: list[torch.Tensor] = []
+                # log_sequential_error to compare against post-compression outputs.
+                # keyed by batch_idx (rather than appended to a list) so that a
+                # batch skipped in pass 1 (e.g. _get_largest_tensor returns None)
+                # cannot desynchronize the alignment with pass 2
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by dichn** [comment]
A Live Run for validating `log_sequential_error=True`:
```
oneshot(
    model='Qwen/Qwen3-0.6B',
    dataset='open_platypus',
    splits='train[:64]',
    max_seq_length=256,
    num_calibration_samples=64,
    pipeline='sequential',
    log_sequential_error=True,
    recipe=recipe,
)


2026-09-11T14:10:50.3101 | __init__ | WARNING - Disabling tokenizer parallelism due to threading conflict between FastTokenizer and Datasets. Set TOKENIZERS_PARALLELISM=false to suppress this warning.
Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
Loading weights: 100%|██████████████████████████████████████████| 311/311 [00:05<00:00, 54.16it/s]
Tokenizing: 100%|████████████████████████████████████████| 64/64 [00:00<00:00, 1186.21 examples/s]
Adding labels: 100%|█████████████████████████████████████| 64/64 [00:00<00:00, 5757.45 examples/s]
2026-09-11T14:11:10.2933 | reset | INFO - Compression lifecycle reset
2026-09-11T14:11:10.3125 | from_modifiers | INFO - Creating recipe from modifiers
Applying quantization config: 100%|██████████████████████████| 196/196 [00:00<00:00, 11632.87it/s]
2026-09-11T14:11:10.3608 | initialize | INFO - Compression lifecycle initialized for 1 modifiers
W0911 14:11:10.521000 3223 torch/fx/_symbolic_trace.py:57] is_fx_tracing will return true for both fx.symbolic_trace and torch.export. Please use is_fx_tracing_symbolic_tracing() for specifically fx.symbolic_trace or torch.compiler.is_compiling() for specifically torch.export/compile.
Preparing cache: 100%|██████████████████████████████████████████| 64/64 [00:00<00:00, 3540.20it/s]
(1/29): Calibrating: 100%|████████████████████████████████████████| 64/64 [00:02<00:00, 23.86it/s]
(1/29): Propagating: 100%|███████████████████████████████████████| 64/64 [00:00<00:00, 365.13it/s]


2026-09-11T14:38:55.6590 | SequentialPipeline | METRIC - subgraph 1/29 | sequential error (KL): 0.000000
...
2026-09-11T14:39:03.6402 | SequentialPipeline | METRIC - subgraph 3/29 | sequential error (KL): 0.031213
...
2026-09-11T14:40:40.6636 | SequentialPipeline | METRIC - subgraph 28/29 | sequential error (KL): 14.582560
```

File: src/llmcompressor/pipelines/sequential/pipeline.py [hunk 1]
```diff
@@ -138,6 +186,13 @@ def __call__(
                 calib_desc = f"({subgraph_index + 1}/{num_subgraphs}): Calibrating"
                 prop_desc = f"({subgraph_index + 1}/{num_subgraphs}): Propagating"
 
+                # per-batch largest tensor captured before compression, used by
+                # log_sequential_error to compare against post-compression outputs.
+                # keyed by batch_idx (rather than appended to a list) so that a
+                # batch skipped in pass 1 (e.g. _get_largest_tensor returns None)
+                # cannot desynchronize the alignment with pass 2
+                pre_compression_outputs: dict[int, torch.Tensor] = {}
```
kylesayrs, line 194: Doing this will create OOMs if the dataset is very large. We should save output activations to `activations` if the kl divergence logging is enabled

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -177,5 +243,32 @@ def __call__(
...
+                                    post_tensor = _get_largest_tensor(output)
+                                    if post_tensor is not None:
+                                        pre_tensor = pre_compression_outputs[
+                                            batch_idx
+                                        ].to(post_tensor.device)
+                                        batch_kls.append(
+                                            _compute_kl_divergence(
+                                                pre_tensor, post_tensor
+                                            )
+                                        )
```
kylesayrs, line 261: FYI I think we should try to hide as much logic behind functions as possible to keep the sequential pipeline as clean as possible

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -48,6 +49,45 @@ def _get_batches(
         yield batch_idx, inputs
 
 
+def _get_largest_tensor(outputs: dict[str, Any]) -> torch.Tensor | None:
```
kylesayrs, line 52: Nice

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -48,6 +49,45 @@ def _get_batches(
...
+    largest = None
+    largest_numel = 0
+    for value in outputs.values():
+        if isinstance(value, torch.Tensor) and value.numel() > largest_numel:
+            largest = value
+            largest_numel = value.numel()
+    return largest
+
+
+def _compute_kl_divergence(
```
kylesayrs: https://github.com/vllm-project/llm-compressor/issues/3154#issuecomment-5637289536

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -157,11 +171,25 @@ def __call__(
...
+                            dataset_args.log_sequential_error
+                            and subgraph_index < num_subgraphs - 1
+                        ):
+                            cache_pre_compression_output(
+                                activations, batch_idx, outputs
+                            )
+
                     LifecycleCallbacks.sequential_epoch_end(subgraph.submodules(model))
 
                     if dataset_args.propagate_error:
```
kylesayrs, line 184: Maybe you can decouple `propagate_error` and `log_sequential_error` using something like this?

```python
if log_sequential_error:
    seq_error_cache = ActivationsCache(...)

for subgraph in subgraphs:
    outputs = subgraph(inputs)
    
    if log_sequential_error:
        seq_error_cache.update(outputs)

    seq_epoch_end()

    if propagate_error or log_sequential_error:
        outputs = subgraph(inputs)
        if propagate_error:
            activations.update(outputs)
    
        if log_sequential_error:
            compute_sqnr(seq_error_cache, outputs)
```

kylesayrs (commented):
Looking good!

src/llmcompressor/pipelines/sequential/pipeline.py
dichn: @kylesayrs are you expecting something like this, after decoupling `propagate_error` and `log_sequential_error`?
```
oneshot(
    model="Qwen/Qwen3-0.6B",
    dataset="open_platypus",
    recipe=recipe,
    output_dir="/tmp/qwen3_decoupled_propTrue",
    max_seq_length=512,
    num_calibration_samples=32,
    log_sequential_error=True,     # new introduced
    propagate_error=True,          # new introduced
)
```

kylesayrs: @dichn `propagate_error` already exists as an argument

If we do something like what I've outlined above, we can support flows like this:
```python
oneshot(
    model="Qwen/Qwen3-0.6B",
    dataset="open_platypus",
    recipe=recipe,
    output_dir="/tmp/qwen3_decoupled_propTrue",
    max_seq_length=512,
    num_calibration_samples=32,
    log_sequential_error=True,     # new introduced
    propagate_error=False,
)
```

dichn:
force-pushed for
- decouple `propagate_error` and `log_sequential_error`
- squash all the previous changes to one commit
- simplify the unit test and eliminate wildcards   
- commit message upgrade

CC: @kylesayrs

src/llmcompressor/pipelines/sequential/error_logging.py
```diff
@@ -0,0 +1,166 @@
...
+    :param activations: intermediates cache shared across the pipeline
+    :param batch_idx: index of the current batch
+    :param consumed_names: names no longer needed by any subsequent subgraph
+    """
+    pre_outputs = cache.fetch(batch_idx)
+    activations.update(batch_idx, pre_outputs)
+    activations.delete(batch_idx, consumed_names)
+
+
+def record_batch_error(
```
kylesayrs: I think it's slightly misleading to have the function name `record` but also modify `activations`

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -152,16 +171,32 @@ def __call__(
...
+                                not dataset_args.log_sequential_error
+                                and has_next_subgraph
+                            ):
                                 activations.update(batch_idx, outputs)
                                 activations.delete(batch_idx, subgraph.consumed_names)
 
+                        if seq_error_cache is not None and has_next_subgraph:
+                            cache_pre_compression_output(
+                                seq_error_cache, batch_idx, outputs
+                            )
```
kylesayrs: Can we inline this function, since it's only one line?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -171,11 +206,31 @@ def __call__(
...
                                 output = subgraph.forward(model, **inputs)
-                                if subgraph_index < num_subgraphs - 1:
+                                if dataset_args.propagate_error and has_next_subgraph:
                                     activations.update(batch_idx, output)
                                     activations.delete(
                                         batch_idx, subgraph.consumed_names
                                     )
 
+                                if seq_error_cache is not None and has_next_subgraph:
+                                    batch_power = record_batch_error(
```
kylesayrs: It's misleading that this function says "record", but also modifies `activations` in place

src/llmcompressor/pipelines/sequential/error_logging.py
```diff
@@ -0,0 +1,166 @@
...
+    when propagate_error is disabled.
+
+    :param cache: intermediates cache holding the pre-compression outputs
+    :param activations: intermediates cache shared across the pipeline
+    :param batch_idx: index of the current batch
+    :param consumed_names: names no longer needed by any subsequent subgraph
+    """
+    pre_outputs = cache.fetch(batch_idx)
+    activations.update(batch_idx, pre_outputs)
+    activations.delete(batch_idx, consumed_names)
```
kylesayrs: This will onload, then re-offload the values, inducing unnecessary runtime cost. Consider just replacing `activations` (inputs_acts) with `cache` (outputs_acts) directly (see other comment)

kylesayrs (commented):
Looking good! Just a few nits. The only missing functionality left is being able to update offloads directly between caches, rather than onloading + offloading.

Please also make sure we have the following:
* Simple docs/diagram describing how the activation cache is updated depending on different configurations of (`propagate_error`, `log_sequential_error`)
* A test that demonstrates that oneshot outputs are the same across all of the different configurations listed above
* (extra) support the basic pipeline and rename `log_sequential_error` to `log_error`)

PR #3184, opened 2026-09-15
[Bugfix] Support upfront model decompression, Kimi-K3

kylesayrs:
## Background ##
Previously, kimi-k3 was quantized using [layerwise decompression](#2995). However, when adapting this test, the adapted script used upfront decompression. Unfront decompression works fine when using gpu/cpu offloading with small models, but fails for large models and disk offloading.

These changes fix issues with the example script and provide a canonical way of decompressing a model with disk offloading

## Changes ##
* Only target mlp for quantization (some attention layers will not receive quantization data)
* Modify script to use DDP
* Modify script to call decompression directly, which is supported with disk offloading as of https://github.com/vllm-project/compressed-tensors/pull/878
* Fix kimi-k3 model definition to send all tokens to all experts based on `get_calibrate_all_experts_flag()`, rather than `self.training` (which is false for our calibration flows)

## Prerequisites ##
* https://github.com/vllm-project/compressed-tensors/pull/890

## Testing ##
* Able to quantize `inference-optimization/Kimi-K3-0.40B-MXFP4` without warnings or errors


docs/key-models/kimi-k3/nvfp4-example.md
```diff
@@ -5,70 +5,83 @@
...
+
 recipe = QuantizationModifier(
-    targets="Linear",
+    targets="re:.*mlp.*",
     scheme="NVFP4",
     ignore=[
         "lm_head",
         r"re:.*block_sparse_moe\.gate",
         "re:.*vision_tower.*",
+        "re:.*mlp_res_proj$",
```
kylesayrs, line 59: TODO: ignore non-experts mlp

src/llmcompressor/modeling/kimi_k3/modeling_kimi_linear.py
```diff
@@ -959,10 +959,7 @@ def moe_train(self, x, topk_ids, topk_weight):
         for expert_idx, expert in enumerate(self.experts):
             top_k_pos, token_indices = torch.where(expert_mask[expert_idx])
 
-            if get_calibrate_all_experts_flag():
-                expert_out = expert(x)[token_indices]
-            else:
-                expert_out = expert(x[token_indices])
```
HDCharles, line 965: Why?

HDCharles (approved):

src/llmcompressor/modeling/kimi_k3/modeling_kimi_linear.py
kylesayrs: It's now controlled in the frame above this one.

This will soon be removed in favor of using a proper `LinearExperts2D` class

kylesayrs:
Fixed the qconfig patch, good to go

PR #3185, opened 2026-09-15
[Bugfix] Fix distributed hang from mis-shaped weight_global_scale writeback

kylesayrs:
## Purpose

Fix a hang in distributed GPTQ caused by writing back an improperly formatted `weight_global_scale`.

## Background

In distributed GPTQ, each module is compressed on a single owning rank, which writes back its quantized weight and qparams via `update_offload_parameter`. That function updates an offload parameter **in place** (asynchronously) only when the new value matches the registered parameter's shape and dtype. When the shape/dtype does **not** match, it instead synchronously re-offloads a brand new parameter.

Previously the global scale was written as `global_scales[index].to(dtype=module.weight.dtype)` — a 0-dim tensor in the weight dtype — which does not match the registered `[1]` float32 `weight_global_scale`. Because only the owning rank performs this write, the ranks end up disagreeing on the parameter's shape/dtype, which stalls the collectives that later broadcast/save these params and hangs the run.

## Changes

- **`src/llmcompressor/modifiers/gptq/helpers.py`**: format the `weight_global_scale` writeback as a `[1]` float32 tensor (`global_scales[index].unsqueeze(0).to(dtype=torch.float32)`) so it matches the registered parameter and is updated in place on every rank.
- **`tests/.../gptq/test_gptq_ddp.py`**: new multi-GPU regression test that drives `GPTQModifier` directly (no `oneshot`) over the distributed compression path. It:
  - offloads a toy model to CPU with GPU onload to exercise the offload-cache writeback path,
  - uses a `tensor_group` float weight scheme (the case that produces `weight_global_scale`),
  - initializes the process group with a low NCCL timeout + blocking wait so any residual deadlock fails fast instead of hanging the suite,
  - asserts cross-rank `weight_global_scale` shape/dtype consistency and that broadcast quantized weights are identical across ranks.

## Testing

- `test_gptq_distributed_global_scale_writeback` passes with the fix (~14s).
- Reverting the source fix makes it fail deterministically (~3s) on the cross-rank shape assertion (`torch.Size([]) != torch.Size([1])`).


tests/llmcompressor/modifiers/gptq/test_gptq_ddp.py
```diff
@@ -0,0 +1,92 @@
+"""
```
HDCharles, line 1: Don't the smoke tests catch this?

HDCharles (approved):
Looks good

tests/llmcompressor/modifiers/gptq/test_gptq_ddp.py
HDCharles: Looks like no, but we could change w4a16 to nvfp4 to catch this

kylesayrs: It's probably good to have a smaller dist gptq test anyways. We should def have a nvfp4 gptq test

PR #3187, opened 2026-09-16
Skip CUDA compute-capability tests on XPU

HDCharles:
## Summary

- skip CUDA compute-capability requirements on non-CUDA accelerators such as XPU
- add a regression test for XPU-style capability dictionaries

## Validation

- `python -m pytest -q tests/test_testing_utils.py`
- `python -m ruff check tests/testing_utils.py tests/test_testing_utils.py`


kylesayrs (approved):
This is fine, but maybe in the future we can generalize this to allow such tests to run on xpu

PR #3190, opened 2026-09-16
[AutoRound] Pad captured inputs to a common sequence length

kylesayrs:
## Purpose

Fix `test_ddp_smoke_autoround`, which regressed after `pad_to_max_length`
was changed to default `False` (#3180).

With that change, calibration samples are no longer padded to a fixed
`max_seq_length`. AutoRound batches calibration samples together and
concatenates the captured decoding-layer inputs along the batch dimension,
which requires every sample to share the same sequence length. With
variable-length samples this fails:

```
RuntimeError: Sizes of tensors must match except in dimension 0.
Expected size 512 but got size 436 for tensor number 3 in the list.
```

## Changes

Before handing captured inputs to AutoRound, right-pad them to the longest
captured sequence:

- `hidden_states`, `position_ids`, `position_embeddings` (`cos`/`sin`) and
  `cache_position` are padded along their sequence dimension.
- Attention masks are padded correctly by kind: integer padding masks with
  `0`, and float causal masks with the dtype minimum along the key
  dimension (so valid queries can't attend to pad keys) and `0` elsewhere.
- Per-sample validity masks (`[1, max_seq]`, zero on pad positions) are
  passed to AutoRound so pad positions are excluded from the tuning loss.

Right-padding is safe because causal attention prevents valid positions
from attending to trailing pad positions, so the reconstruction targets for
valid tokens are unchanged. When all captured samples already share a
length (e.g. `pad_to_max_length=True`), inputs are returned untouched and
the previous behavior is preserved.

## Testing

- Reproduced the failure with `oneshot(..., pad_to_max_length=False)` on
  `nm-testing/tinysmokeqwen3moe`; passes with the fix.
- `tests/llmcompressor/modifiers/autoround/test_base.py` and
  `tests/llmcompressor/transformers/autoround/test_autoround_oneshot.py`
  pass.


yiliu30 (approved):
Thanks @kylesayrs , LGTM!

PR #3194, opened 2026-09-17
[Bugfix] Load preprocessing functions from file paths in import_from_path

### LINKED ISSUE #3193: [Bug]: import_from_path cannot load a preprocessing function from a file path
### Describe the bug

`import_from_path` documents three accepted forms, and
`TextGenerationDataset.preprocessing_func` repeats the first one in its own comment
(`# load func_name from "/path/to/file.py:func_name"`):

```
"/path/to/file.py:func_or_class_name"
"/path/to/file:focn"
"path.to.file:focn"
```

Only the third one works. Pointing `preprocessing_func` at a file — the documented way
to supply a custom calibration preprocessing function — always fails.

### Expected behavior

All three documented forms resolve the object.

### Steps to reproduce

```python
from pathlib import Path
from llmcompressor.utils import import_from_path

Path("myprep.py").write_text("def preprocess(example):\n    return example\n")

import_from_path(f"{Path('myprep.py').resolve()}:preprocess")   # absolute path
import_from_path("llmcompressor.pytorch.utils.helpers:get_quantized_layers")
```

Actual:

```
TypeError: the 'package' argument is required to perform a relative import for '.tmp.lcrepro.myprep'
AttributeError: Cannot find get_quantized_layers in llmcompressor.pytorch.utils.helpers
```

(the symbol does exist: `importlib.import_module("llmcompressor.pytorch.utils.helpers")`
resolves it fine)

### Cause

`importlib.import_module` takes dotted module names, not file system paths, so the
function first rewrites the path:

```python
path = original_path.split(".py")[0]
path = re.sub(r"/+", ".", path)
```

- A leading `/` becomes a leading `.`, so an absolute path becomes a relative import and
  raises `TypeError`. That is not an `ImportError`, so the surrounding handler does not
  catch it and the intended `Cannot find module with path ...` message never appears.
- `split(".py")` cuts at the first `.py` substring anywhere, not at the extension, so a
  dotted path is truncated at any package starting with `py`.
  `llmcompressor.pytorch.utils.helpers` becomes `llmcompressor`, which imports fine — so
  the wrong module is searched and the failure surfaces as a misleading `AttributeError`.
- Separately, `path.split(":")` raises `too many values to unpack` on a Windows drive
  letter such as `C:/models/prep.py:preprocess`.

### Proposed fix

Load the module from the file when the path points at one
(`importlib.util.spec_from_file_location`), falling back to `importlib.import_module`
for dotted names, and split the object name with `rsplit(":", 1)`.

`import_from_path` currently has no test coverage; I would add tests for the three
documented forms and both error paths.

### Environment

`main` at 6693441, Python 3.12.

---

I have this ready and would like to open the PR if the approach looks right.

Evihut:
## Purpose

Fixes #3193

`import_from_path` documents three accepted forms, and
`TextGenerationDataset.preprocessing_func` repeats the first one in its own comment:

```
"/path/to/file.py:func_or_class_name"
"/path/to/file:focn"
"path.to.file:focn"
```

Only the third worked, so pointing `preprocessing_func` at a file — the documented way to
supply a custom calibration preprocessing function — always failed.

`importlib.import_module` takes dotted module names, not file system paths, so the
function rewrote the path first:

```python
path = original_path.split(".py")[0]
path = re.sub(r"/+", ".", path)
```

- A leading `/` becomes a leading `.`, so an absolute path becomes a relative import and
  raises `TypeError`, which the surrounding `except ImportError` does not catch — the
  intended `Cannot find module with path ...` message never appears.
- `split(".py")` cuts at the first `.py` substring anywhere rather than at the extension,
  truncating a dotted path at any package starting with `py`.
  `llmcompressor.pytorch.utils.helpers` becomes `llmcompressor`, which imports
  successfully, so the wrong module is searched and the user sees a misleading
  `AttributeError`.
- `path.split(":")` also raises `too many values to unpack` on `C:/models/prep.py:preprocess`.

## Changes

- load the module from the file when the path points at one, falling back to
  `importlib.import_module` for dotted names
- split the object name with `rsplit(":", 1)`
- `import_from_path` had no test coverage; add tests for the three documented forms and
  both error paths

Behaviour for the form that already worked (a plain dotted module name) is unchanged, as
are the `ImportError` / `AttributeError` messages.

## Test Plan

```
pytest tests/llmcompressor/utils/test_helpers.py -k import_from_path
pytest tests/llmcompressor/utils/
make quality
```

## Test Result

- `-k import_from_path`: 5 passed. 4 of the 5 fail on `main` and pass with this change
  (the fifth covers the missing-module path, which already worked).
- `tests/llmcompressor/utils/`: 5 failed, 15 passed — identical 5 failures on `main`
  (5 failed, 10 passed there); they are `test_disable_lm_head` and
  `test_untie_word_embeddings_gpu`, which need a model download and CUDA respectively and
  cannot run on my machine. No new failures.
- `make quality`: clean.


Evihut:
Thanks for the review — all three points were worth acting on and I pushed a follow-up
commit. Two of the specific failure modes named along the way did not reproduce though,
one of them my own extrapolation, so I want to be precise about what actually changed.

**Re-execution — real, and this is what the commit fixes.** Without the `sys.modules`
entry the file is executed on every `import_from_path` call, so anything it does at
import time repeats. A module incrementing a counter at import ran 3 times for 3 calls,
and once after the change.

**Pickling — real for the standard library, but not the failure mode I first assumed.**
`pickle.dumps` on the loaded function did fail:

```
PicklingError: Can't pickle <function preprocess at 0x...>: import of module 'myprep' failed
```

I expected that to break `preprocessing_func` under
`dataset.map(num_proc=...)`, and it does not: `datasets` serializes through `multiprocess`
/ `dill`, which handles the function by value. I ran the real thing both ways and it
worked before the change as well, under `fork` and under `spawn`:

```
map(num_proc=2, start=fork):  OK
map(num_proc=2, start=spawn): OK
```

So registering is the documented recipe and fixes stdlib pickling, but it is not
repairing a broken multiprocessing path.

**`inspect` — did not reproduce.** `inspect.getsource` on the loaded function worked
without the registration too.

**Name collisions** only matter once the module is registered, so that came along with
the fix: the module name is `f"{__name__}._loaded_{stem}_{sha256(abs_path)[:16]}"`
rather than the full path with separators substituted — same collision-freedom, bounded
length, and namespaced under this module. Cleanup catches `BaseException`, so an
interrupt during `exec_module` also unregisters the half-built module.

Tests cover single execution, stdlib pickling, two same-named files staying distinct, and
no leftover entry when the file raises at import. `make quality` is clean;
`pytest tests/llmcompressor/utils/` adds no new failures (the remaining ones need a model
download and CUDA and fail the same way on `main` here).

Could a maintainer add the `ready` label? The bot asks for it and I cannot set labels.

Evihut:
Closing this file-path fix as superseded by #3227, which begins deprecating `preprocessing_func` following the direction agreed in #3193. Thank you for the review and feedback.

PR #3201, opened 2026-09-18
use load_context in llm-compressor

HDCharles:
SUMMARY:
use load_context instead of load_offloaded_model


kylesayrs (approved):
You have a couple lines which say
```
from compressed_tensors.offload import load_context
```

otherwise all good

kylesayrs (approved):
Rereviewed, looks good

PR #3208, opened 2026-09-18
Sequential Offloading, Linearization + Repacking

Roderick-Wu:
Changes to linearization, repacking, sequential pipeline offloading. 

### Sequential Pipeline Offloading
Removed the use of the offload context in the sequential pipeline. Onload/offload is manually managed now. 

Onloading is done as
```python
offload_kwargs = onload(subgraph_modules)
```
`subgraph_modules` is a hash as dict[str, module] where the key is the module's name. 
`offload kwargs` are in the form dict[str, dict]. Values are in the form returned by `get_cache_init_kwargs`

Offloading done as:
```python
offload(offload_modules, dict(offload_kwargs))
```

We do this so that we can swap out modules from the `subgraph_modules` when doing eager linearization/repacking easily. These operations create new module objects that need to be bound to the model. In addition, the old ones need to have all references deleted so that they can be freed, regulating our memory usage. 

Onload/Offload functions under `src/llmcompressor/pipelines/sequential/offloading.py`. 

### Linearization and Repacking

We add flags `moe_eager_linearize_repack` (Linearize and/or repack each subgraph during calibration onload if `True`, linearize and/or repack before/after the whole model is calibrated for `False) and `repack_moe` (Whether or not to pack experts to 3d). 

Before calibration, we trace the whole model and build a hash for fast lookup for all moe modules -- packed or not packed. This is stored on the model as `model._moe_lookup`. Generated by `get_moe_linear_status`. 

This lookup is important because using model.get_submodules(name) traverses the whole model which is a taxing operation for large models. 

From here, we have functions as follows:
`linearize_moe_model`
`linearize_moe_subgraph`
`linearize_moe_layer`

and their corresponding repacking/reverse counterparts
`repack_moe_model`
`repack_moe_subgraph`
`repack_moe_layer`

The `*_model` functions are used with `sequential_linearize_repack=False`. They iterate over moe layers, onloading/offloading each layer individually (If you can't fit an entire moe layer on gpu at once, you're out of luck). 

Some things to note:

Weird hard-to-read code such as:
```python
    linearized = [
        (moe_lookup[module], module)
        for module in subgraph_set
        if module in moe_lookup and isinstance(module, LinearExperts2D)
    ]

    for i in tqdm.tqdm(range(len(linearized)), desc="Repacking experts in subgraph"):
        repack_moe_layer(model, linearized[i][0], linearized[i][1])
```

In the case that we are linearizing/repacking and the resulting new module is a new object, we want to free the old object. This can only be done when the python object is garbage collected. So here we make sure that the old module's reference is deleted by the end of its iteration. This is also why we need:

```python
        gc.collect()
        if torch.accelerator.is_available():
            torch.accelerator.empty_cache()
```

At the end of a sequential pipeline subgraph iteration, we don't generate enough objects to force automatic garbage collection (at least for the tests I've run). The cache release might have a solution but I haven't been able to dig it up yet. Without that part it seems that we constantly have allocated but unused gpu memory.


src/llmcompressor/args/dataset_arguments.py
```diff
@@ -298,6 +298,26 @@ class DatasetArguments(CustomDatasetArguments):
             "for faster calibration when GPU memory allows (two batches on device)."
         },
     )
+    moe_lazy_linearization_and_repack: bool = field(
```
HDCharles, line 301: nice

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -140,7 +153,15 @@ def __call__(
...
                 # reduce memory movement by keeping modules onloaded
                 num_batches = len(dataloader)
-                with disable_offloading():
+
+                subgraph_modules = subgraph.submodules(model)
+
+                # Everything onloaded in this context, offloaded outside
+                with disable_offloading_controlled(model, subgraph_modules):
+                    # linearize moe layers just before calibration,
+                    if dataset_args.moe_lazy_linearization_and_repack:
```
HDCharles: this conditional is not necessary right? 

linearize_moe_subgraph is idempotent so it only linearizes un-linearized things right? so we can just always to the linearize at this point?

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -115,6 +123,11 @@ def __call__(
         with contextlib.ExitStack() as stack:
             stack.enter_context(calibration_forward_context(model))
             stack.enter_context(DisableQuantization(model))
+
+            # linearize MoE layers upfront if not using layer-wise linearization
```
HDCharles, line 127: lazy* linearization

src/llmcompressor/pipelines/sequential/offloading.py
```diff
@@ -0,0 +1,162 @@
...
+
+@contextlib.contextmanager
+def disable_offloading_controlled(
+    model: torch.nn.Module,
+    subgraph: Iterable[torch.nn.Module] | None = None,
+):
+    """
+    Context manager to disable offloading for a specific subgraph of the model.
+
+    Intended for sequential pipeline. Onload the entire subgraph, then offload it.
```
HDCharles: what does this mean? why onload then offload it?

this should explain what it is doing differently from disable_offloading

src/llmcompressor/pipelines/sequential/offloading.py
```diff
@@ -0,0 +1,162 @@
...
+    reserved = torch.accelerator.memory_reserved(device) / 1024**3
+    peak_allocated = torch.accelerator.max_memory_allocated(device) / 1024**3
+    logger.debug(
+        f"{prefix} | cuda:{device} allocated={allocated:.2f} GiB "
+        f"reserved={reserved:.2f} GiB peak_allocated={peak_allocated:.2f} GiB"
+    )
+
+
+@contextlib.contextmanager
+def disable_offloading_controlled(
```
HDCharles: not sure about this name, it seems to be targetting not controlled

disable_offloading_for_subgraph?

src/llmcompressor/pipelines/sequential/offloading.py
```diff
@@ -0,0 +1,162 @@
+import contextlib
```
HDCharles: this will need to be in CT

src/llmcompressor/pipelines/sequential/pipeline.py
```diff
@@ -140,7 +153,15 @@ def __call__(
 
                 # reduce memory movement by keeping modules onloaded
                 num_batches = len(dataloader)
-                with disable_offloading():
+
+                subgraph_modules = subgraph.submodules(model)
+
+                # Everything onloaded in this context, offloaded outside
+                with disable_offloading_controlled(model, subgraph_modules):
```
HDCharles: why is this change needed? it looks like offloading was generally disabled previously, why is offloading the subgraph only beneficial here?

src/llmcompressor/pipelines/sequential/pipeline.py
Roderick-Wu: Yes that is correct, I can actually leave that in

src/llmcompressor/pipelines/sequential/offloading.py
Roderick-Wu: That's probbaly a better name, but I thought in this branch I had it removed altogether so I might need to update this again

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,66 +217,51 @@ def linearize_moe(model: PreTrainedModel):
...
+) -> None:
     """
-    return [
-        (name, module)
-        for name, module in model.named_modules()
-        if isinstance(module, LinearExperts2D)
-    ]
-
+    Linearize MoE layers within a subgraph during sequential calibration.
+    Offloading is deferred so calibration can run on the newly created modules before
```
HDCharles: isn't this a requirement, not a function?

"offloading must be disabled" right?

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,66 +217,51 @@ def linearize_moe(model: PreTrainedModel):
...
+        if module in moe_lookup and not isinstance(module, LinearExperts2D)
+    ]
 
-    for name, module in tqdm.tqdm(linearized, desc="Repacking experts"):
-        fused = module.to_experts_module()
-        model.set_submodule(name, fused)
+    for i in tqdm.tqdm(
+        range(len(non_linearized)), desc="Linearizing experts in subgraph"
+    ):
+        linearize_moe_layer(model, non_linearized[i][0], non_linearized[i][1])
```
HDCharles:
```suggestion
        linearize_moe_layer(model, *non_linearized[i])
```

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,66 +217,51 @@ def linearize_moe(model: PreTrainedModel):
...
 
-    for name, module in tqdm.tqdm(non_linearized_moes, desc="Linearizing experts"):
-        config = getattr(module, "config", model.config)
-        linear_experts_cls = LinearExperts2D.get_linear_experts_cls(module.__class__)
-        linear_moe = linear_experts_cls.from_experts_module(module, config)
-        model.set_submodule(name, linear_moe)
+    # This is also sequential
+    for i in tqdm.tqdm(range(len(non_linearized)), desc="Linearizing experts"):
+        with disable_offloading_controlled(non_linearized[i][1]):
+            linearize_moe_layer(model, non_linearized[i][0], non_linearized[i][1])
```
HDCharles:
```suggestion
            linearize_moe_layer(model, *non_linearized[i])
```

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -93,22 +97,117 @@ def patched(cls, *args, **kwargs):
...
-    Linearize a mixture-of-experts model after it has been loaded. For more
-    runtime-efficient loading, please see `load_quantizable_moe`.
+    Repack a single linearized :class:`LinearExperts2D` module back into its native
+    fused 3D expert module.
+    """
+    fused = module.to_experts_module()
+    replace_module(model, name, module, fused)
+
 
+def linearize_moe_model(model: PreTrainedModel) -> None:
```
HDCharles, line 204: I would probably structure this whole thing differently

you have 

linearize_moe_model - calls linearize_moe_layer recursively on whole model
linearize_moe_subgraph - calls linearize_moe_layer recursively on subgraph
linearize_moe_layer - does linearization, then does assignment into the model

1) linearize_moe_model and linearize_moe_subgraph seem redundant, can make linearize_moe_subgraph do the same thing as linearize_moe_model if subgraph is e.g. None or something
2) all 3 of these have to handle full model scope, it makes more sense to have linearize_moe_subgraph have full scope and then linearize_moe_layer have layer scope i.e. then linearize_moe_layer shouldn't do the assignment into the model, the higher scope function should do this. linearize_moe_layer should just take a layer in and send a layer out.

actually i see there's a difference between the offloading handling and that feels like something that should be handled by the caller rather than the function itself. 

edit: actually is the offloading handling different, it looks like linearize_moe_model has offloading inside its function but whenever linearize_moe_subgraph is called its in a disable offloading context

src/llmcompressor/modeling/moe/linearize.py
```diff
@@ -118,66 +217,51 @@ def linearize_moe(model: PreTrainedModel):
...
     )
 
-    for name, module in tqdm.tqdm(non_linearized_moes, desc="Linearizing experts"):
-        config = getattr(module, "config", model.config)
-        linear_experts_cls = LinearExperts2D.get_linear_experts_cls(module.__class__)
-        linear_moe = linear_experts_cls.from_experts_module(module, config)
-        model.set_submodule(name, linear_moe)
+    # This is also sequential
+    for i in tqdm.tqdm(range(len(non_linearized)), desc="Linearizing experts"):
+        with disable_offloading_controlled(non_linearized[i][1]):
```
HDCharles: what exactly is this doing, i'm not following?

src/llmcompressor/pipelines/sequential/pipeline.py
Roderick-Wu: Offloading was actually still possible in the previous context, we want to explicit outline here how we want to control the offloading. Which for now we think the simplest is to onload the entire subgraph, perform whatever operations are necessary, then offload everything. 

Also we want this changes so that later on we can add streaming and overlap onload/offload with calibration so we aren't blocked by disk io

HDCharles:
i'd really prefer for this to be 2 PRs, it looks like this is doing 2 things

1) changing the way we do offloading/onloading
2) adding a lazy linearization method

these should be separated out

